PR #164, opened 2014-09-26
CXX-351 Add a basic os x buildvariant

samantharitter:


acmorrow:
@samantharitter I think this is unblocked, see https://jira.mongodb.org/browse/BUILD-383

PR #223, opened 2015-01-07
CXX-454 test enumeration of collections and indexes

TylerBrock:


src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -394,6 +394,67 @@ namespace {
...
+        auto_ptr<DBClientCursor> cursor = c.enumerateCollections(TEST_DB, bob.obj(), batch_size);
+
+        if (serverGTE(&c, 2, 8))
+            ASSERT_EQUALS(cursor->getns(), TEST_DB + ".$cmd.listCollections");
+        else
+            ASSERT_EQUALS(cursor->getns(), TEST_DB + ".system.namespaces");
+
+        int coll_count = 0;
+        while (cursor->more()) {
+            // The shims don't support the objsLeftInBatch cursor interface for initial batch.
```
acmorrow: I'm not sure that that is OK. We are implementing the interface for DBClientCursor, so... we need to implement the interface?

src/mongo/client/dbclientinterface.h
```diff
@@ -1145,7 +1145,8 @@ namespace mongo {
          *  }
          */
         std::auto_ptr<DBClientCursor> enumerateCollections( const std::string& db,
-                                                            const BSONObj& filter = BSONObj() );
+                                                            const BSONObj& filter = BSONObj(),
+                                                            int batchSize = 0 );
```
acmorrow: Kind of a shame to have this wart in the interface just for testing. No user should ever care. However, I can live with it I guess.

acmorrow:
@TylerBrock Can't say I love adding this new parameter which is of no value to callers, but I don't see any other way to verify the behavior.

I'm also a little worried about the comments indicating that shim cursors don't fully honor the cursor interface. Can we repair that?

src/mongo/integration/standalone/dbclient_test.cpp
TylerBrock: The way it was designed is that more() and next() are the only virtual parts of the DBClientCursor interface and the shims don't provide any cursor like behavior besides what is provided by those two methods. This has been the case since aggregation was first implemented, why do we want to change it now?

src/mongo/client/dbclientinterface.h
TylerBrock: Yeah... it's not great but its another knob so its a plus and a minus.

TylerBrock:
I repaired that part of the cursor interface. Seems to be well liked by our CI but it's kind of ugly.

src/mongo/client/dbclientcursorshimcursorid.cpp
```diff
@@ -36,7 +36,9 @@ namespace mongo  {
             cursor.ns = ele["ns"].String();
 
             if (!ele["firstBatch"].eoo()) {
-                iter = BSONObjIterator(ele["firstBatch"].Obj());
+                BSONObj firstBatch = ele["firstBatch"].Obj();
+                objects_left = firstBatch.nFields();
```
TylerBrock: That is correct, we'd have to walk the fields. Better than copying them out via getOwned right? The first batch can be 4mb so you'd count 100k in the worst case (all of the collections are single characters and have no options)

```js
> 4 * 1024 * 1024 / Object.bsonsize(db.test.runCommand({
    "listCollections": "test", cursor: {}}
)["cursor"]["firstBatch"][0])
```

Result: 135300 with the only collection being named "a"

acmorrow: ok. scanning 4MB of memory for elements is going to take a while, but if you have >100k collections, well, some things are going to be slower.

acmorrow:
So, I guess our options are:

 - Always return 0 for objsLeftInBatch
- Potentially do a linear BSON walk of up to 4MB of memory when returning things.

I'm honestly not sure which is worse. What do you think?

TylerBrock:
Well for listCollections/Indexes I think taking the hit for the walk is acceptable. However, for aggregation results I do not think it is. That being said... if your aggregation results always fit in one batch you could run the non-cursor returning version. That version may not be long for this world though.

TylerBrock:
Ok, we decided that we are going to leave objsLeftInBatch() broken for the first batch rather than take the performance hit. Updated code is available for your review.

acmorrow:
And... back to where we started! OK, this LGTM now, after we explored the alternatives.

PR #225, opened 2015-01-07
CXX-355 remove thread name support

TylerBrock:
This is kind of a hack-job because it modifies the LogstreamBuilder class so it makes it harder for us to take back ports but... it does remove the thread name support. What do we think? :+1: or :-1:


acmorrow:
@TylerBrock Right... now I remember why we didn't do this before.

What if we left the contextName parameter as is, and had the logging macros and functions just pass std::string() has the value of the contextName parameter?

TylerBrock:
New and improved. Let me know what you think.

acmorrow:
LGTM if CI system likes it.

PR #226, opened 2015-01-07
CXX-452 don't pass -KPIC to gcc when building on solaris

amidvidy:


amidvidy:
with this patch I was successfully able to build with `--sharedclient` on gelatin.

src/SConscript.client
```diff
@@ -409,6 +409,16 @@ if buildShared:
             ))
     mongoClientPrefixInstalls.append(mongoClientSharedLibPrefixInstall)
 
+    if solaris:
+        # On solaris SCONS adds -KPIC via SHCCFLAGS and SHCXXFLAGS, which are unsupported on gcc.
```
amidvidy: which *is* unsupported on GCC.

acmorrow:
@amidvidy As mentioned when we discussed this, I think I'd prefer to tackle this by removing the 'default' tool from the Environment construction, and only pulling in the tools we really need. The 'gcc' tool, for instance, buys us nothing. The suncc, sunlink, etc, tools are just wrong for us. There are also many tools getting initialized for things we definitely don't need (fortran, latex, etc.)

Lets see if getting to the minimal set of tools solves this (I'm pretty sure it will).

amidvidy:
:+1:  going to keep this open in case paring down the tools turns out to be infeasible for some unseen reason.

amidvidy:
closing now that we have a better solution in review.

PR #227, opened 2015-01-08
CXX-451 polyfill strnlen for platforms that don't have it

amidvidy:
fixes compile on solaris.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1651,10 +1651,15 @@ def doConfigure(myenv):
         if not conf.CheckLib("execinfo"):
             Exit(1)
 
+    # check for presence of timegm(3) and polyfill if needed
     conf.env['MONGO_HAVE_TIMEGM'] = conf.CheckDeclaration(
         'timegm', includes="#include <time.h>", language='C')
     if conf.env['MONGO_HAVE_TIMEGM']:
         conf.env.Append(CPPDEFINES=['MONGO_HAVE_TIMEGM'])
+
+    # check for presence of strnlen(3) and polyfill if needed
+    if conf.CheckDeclaration('strnlen', includes="#include <string.h>", language='C'):
+        conf.env['MONGO_HAVE_STRNLEN'] = True
 
     return conf.Finish()
 
--- a/src/SConscript.client
+++ b/src/SConscript.client
@@ -58,6 +58,7 @@ configSubstitutions = [
     libEnv.makeConfigHDefine('@mongoclient_have_gcc_atomic_builtins@', 'MONGO_HAVE_GCC_ATOMIC_BUILTINS'),
     libEnv.makeConfigHDefine('@mongoclient_have_gcc_sync_builtins@', 'MONGO_HAVE_GCC_SYNC_BUILTINS'),
     libEnv.makeConfigHDefine('@mongoclient_have_std_is_trivially_copyable@', 'MONGO_HAVE_STD_IS_TRIVIALLY_COPYABLE'),
+    libEnv.makeConfigHDefine('@mongoclient_have_strnlen@', 'MONGO_HAVE_STRNLEN'),
     libEnv.makeConfigHDefineWithValue('@mongoclient_byte_order@', 'MONGO_BYTE_ORDER'),
 ]
 libEnv.Substfile('mongo/config.h.in', SUBST_DICT=configSubstitutions)
@@ -130,6 +131,7 @@ clientSourceBasic = [
     'mongo/logger/message_event_utf8_encoder.cpp',
     'mongo/logger/message_log_domain.cpp',
     'mongo/platform/random.cpp',
+    'mongo/platform/strnlen.cpp',
     'mongo/util/assert_util.cpp',
     'mongo/util/background.cpp',
     'mongo/util/base64.cpp',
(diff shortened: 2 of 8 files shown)


src/mongo/platform/strnlen.cpp
```diff
@@ -0,0 +1,14 @@
+#include "mongo/platform/strnlen.h"
```
amidvidy: need to add copyright header.

src/mongo/bson/util/builder.h
```diff
@@ -22,7 +22,6 @@
 #include <sstream>
 #include <stdio.h>
 #include <string>
-#include <string.h>
```
amidvidy: removed this unnecessary include while looking for other uses of strnlen

SConstruct
```diff
@@ -1644,11 +1644,18 @@ def doConfigure(myenv):
...
             Exit(1)
 
+    # check for presence of timegm(3) and polyfill if needed
     conf.env['MONGO_HAVE_TIMEGM'] = conf.CheckDeclaration(
         'timegm', includes="#include <time.h>", language='C')
     if conf.env['MONGO_HAVE_TIMEGM']:
         conf.env.Append(CPPDEFINES=['MONGO_HAVE_TIMEGM'])
 
+    # check for presence of strnlen(3) and polyfill if needed
+    conf.env['MONGO_HAVE_STRNLEN'] = conf.CheckDeclaration(
```
acmorrow: I think this needs to be done as a config.h entry, rather than a CPPDEFINE fs we ever call strnlen from a header, and it appears that we do in bsonelement.h.

Have a look at how MONGO_HAVE_HEADER_UNISTD_H is handled.

src/SConscript.client
```diff
@@ -155,7 +155,9 @@ clientSourceBasic = [
     'third_party/murmurhash3/MurmurHash3.cpp',
     ]
 
+# polyfills for platforms that don't have certain system provided functions
 clientSourceTz = [] if (clientEnv['MONGO_HAVE_TIMEGM'] or windows) else ['third_party/tz/timegm.c']
+clientSourceStrnlen = [] if (clientEnv['MONGO_HAVE_STRNLEN']) else ['mongo/platform/strnlen.cpp']
```
acmorrow: Rather than making the list of files more complicated to reason about, maybe just always compile strnlen.cpp, but wrap the entire body in #ifndef MONGO_HAVE_STRNLEN?

src/mongo/platform/strnlen.h
```diff
@@ -0,0 +1,23 @@
...
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#ifdef MONGO_HAVE_STRNLEN
+#include <string.h>
+#else
+#include <stddef.h>
+extern "C" size_t strnlen(const char *s, size_t maxlen);
```
acmorrow: Rather than making this extern and doing this globally, should we just import the name into the mongo namespace:

#include <cstring>
#include <cstddef>

namespace mongo {
#ifdef MONGO_HAVE_STRNLEN
    using ::strnlen;
#else
    size_t strnlen(const char* s, size_t maxlen);
#endif
} // namespace mongo

src/mongo/platform/strnlen.cpp
```diff
@@ -0,0 +1,29 @@
+/*    Copyright 2014 MongoDB Inc.
```
acmorrow: 2015

src/mongo/platform/strnlen.h
```diff
@@ -0,0 +1,23 @@
+/*    Copyright 2014 MongoDB Inc.
```
acmorrow: 2015

src/mongo/platform/strnlen.h
```diff
@@ -0,0 +1,23 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#ifdef MONGO_HAVE_STRNLEN
```
acmorrow: When you move this to a config header thing, you will need to incldue config.h here

amidvidy:
@acmorrow could you give this another look when you get a chance?

src/SConscript.client
```diff
@@ -154,6 +156,7 @@ clientSourceBasic = [
     'third_party/murmurhash3/MurmurHash3.cpp',
     ]
 
+# polyfills for platforms that don't have certain system provided functions
```
acmorrow: Is this line out of date, now?

src/mongo/platform/strnlen.h
```diff
@@ -0,0 +1,29 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include <stddef.h>
```
acmorrow: cstddef and cstring

acmorrow:
LGTM mod nits.

src/SConscript.client
amidvidy: it still applies to clientSourceTz, so i think it might make sense to keep it

src/mongo/platform/strnlen.h
amidvidy: done

src/SConscript.client
acmorrow: Fair enough, fine to leave it.

PR #228, opened 2015-01-08
CXX-452 only enable required Tools on OSX, Linux, Windows, and Solaris

amidvidy:
this required some trial and error :)

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -362,6 +362,27 @@ if len(mongoclientVersionComponents) not in (3,4):
     print("Error: client version most be of the form w.x.y[-rcz][-string]")
     Exit(1)
 
+# The Scons 'default' tool enables a lot of tools that we don't actually need to enable.
+# On platforms like Solaris, it actually does the wrong thing by enabling the sunstudio
+# toolchain first. As such it is simpler and more efficient to manually load the precise
+# set of tools we need for each platform.
+# If we aren't on a platform where we know the minimal set of tools, we fallback to loading
+# the 'default' tool.
+def decide_platform_tools():
+    if windows:
+        # we only support MS toolchain on windows
+        return ['msvc', 'mslink', 'mslib']
+    elif linux:
+        return ['gcc', 'g++', 'gnulink', 'ar']
+    elif solaris:
+        return ['gcc', 'g++', 'gnulink', 'ar']
+    elif darwin:
+        return ['gcc', 'g++', 'applelink', 'ar']
+    else:
+        return ["default"]
+
+tools = decide_platform_tools() + ["unittest", "integration_test", "textfile"]
+
 # We defer building the env until we have determined whether we want certain values. Some values
 # in the env actually have semantics for 'None' that differ from being absent, so it is better
 # to build it up via a dict, and then construct the Environment in one shot with kwargs.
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -415,6 +415,27 @@ if windows:
...
+# The Scons 'default' tool enables a lot of tools that we don't actually need to enable.
+# On platforms like Solaris, it actually does the wrong thing by enabling the sunstudio
+# toolchain first. As such it is simpler and more efficient to manually load the precise
+# set of tools we need for each platform.
+# If we aren't on a platform where we know the minimal set of tools, we fallback to loading
+# the 'default' tool.
+def decide_platform_toolchain():
+    if windows:
+        # we only support MS toolchain on windows
+        return ['msvc', 'mslink', 'mslib']
```
acmorrow: Do we want mssdk here too?

SConstruct
```diff
@@ -415,6 +415,27 @@ if windows:
...
+    elif linux:
+        return ['gcc', 'g++', 'gnulink', 'ar']
+    elif solaris:
+        return ['gcc', 'g++', 'gnulink', 'sunar']
```
acmorrow: I think we just want 'ar' here for sun, as sunar affects SHLINKFLAGS and adds -G

SConstruct
```diff
@@ -415,6 +415,27 @@ if windows:
 env = Environment(**envDict)
```
acmorrow: Rather than loading some of the tools here, when the envDict is fed to Environment, and then some later, maybe just adorn the envDict's 'tools' field based on decide_platform_toolchain, and then load all of the tools at once?

SConstruct
```diff
@@ -425,20 +446,6 @@ if has_option("cache"):
         Exit(1)
     env.CacheDir(str(env.Dir(cacheDir)))
 
-# This could be 'if solaris', but unfortuantely that variable hasn't been set yet.
-if "sunos5" == os.sys.platform:
```
acmorrow: Happy to see this go away

SConstruct
amidvidy: we don't need it currently, but as we discussed in person, consumers of the library may need it.

amidvidy:
@acmorrow I made the changes you requested.

acmorrow:
I like it. For sanity checking, we need to:
- Put it through MCI
- Do a manual build on the SPARC box
- Do a build on an OS X box (we don't currently have that on the CI system).

If it is green across the board for those, LGTM.

SConstruct
```diff
@@ -362,6 +362,30 @@ if len(mongoclientVersionComponents) not in (3,4):
...
+    elif linux:
+        return ['gcc', 'g++', 'gnulink', 'ar']
+    elif solaris:
+        return ['gcc', 'g++', 'gnulink', 'ar']
+    elif darwin:
+        return ['gcc', 'g++', 'applelink', 'ar']
+    else:
+        return ["default"]
+
+tools += decide_platform_tools()
```
acmorrow: Small subtlety:

I think SCons runs these tools in order. So, before, the 'default' tools would get loaded, then ours. This reverses that order. If we ever had a tool that wanted to manipulate flags *after* the toolchain tools were loaded, it might not work well. 

Can we reverse the order so that we load the 'minimized defaults' first, and then our custom tools?

amidvidy:
patch build: https://mci.10gen.com/version/54aef0323ff122386f00000e_0

This works on gelatin with my (WIP patch) for CXX-451 cherry-picked on top.

Also have built locally with this patch on my MBP, so I think that takes care of it.

SConstruct
amidvidy: yep, good catch.

amidvidy:
as discussed we are going to remove mssdk. will commit after recompiling on MCI, local box & gelatin.

acmorrow:
Is this PR dead?

amidvidy:
yeah, i didn't push to the branch before pushing to legacy. it should get automagically marked as merged the next commit lands on legacy

PR #229, opened 2015-01-09
CXX-465 remove questionable typedefs and cleanup bsondemo

amidvidy:
also got rid of `using namespace std;` in bsondemo since that is pretty gross.


amidvidy:
will update wiki after this gets in.

acmorrow:
LGTM. Please add this as a breaking change to the wiki.

amidvidy:
wiki is updated. going to close out the ticket.

PR #231, opened 2015-01-15
CXX-389 Add auth to copydb helper (MONGODB-CR and SCRAM-SHA-1)

samantharitter:
This code borrows heavily from the [V8 implementation in the server](https://github.com/mongodb/mongo/blob/master/src/mongo/scripting/v8_db.cpp#L392-L478).


src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -1,4 +1,4 @@
-/*    Copyright 2014 MongoDB Inc.
+/*    Copyight 2014 MongoDB Inc.
```
samantharitter: sorry will fix

samantharitter:
@agralius would you mind taking a look?  Thanks!

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
         return runCommand(db.c_str(), b.done(), *info);
     }
 
-    bool DBClientWithCommands::copyDatabase(const string &fromdb, const string &todb, const string &fromhost, BSONObj *info) {
+    string DBClientWithCommands::_authKeyCopyDB(const string &username,
```
amidvidy: i know style was already weird in this file, but for new code we should stick to the usual `const string& foo`

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                                                const string &password,
+                                                const string &nonce)
+    {
+        md5digest d;
+        string passwordDigest = createPasswordDigest( username, password );
+        {
+            md5_state_t st;
+            md5_init(&st);
+            md5_append(&st, (const md5_byte_t *) nonce.c_str(), nonce.size() );
```
amidvidy: Can we avoid the C-style cast? I think a static_cast<md5_byte_t*> should work since md5_byte_t is just unsigned char, if not we can use reinterpret_cast.

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+
+        // If we don't have a username, or if we're copying locally,
+        // just run the command without authenticating
+        if ( username == "" || fromhost == "" ) {
+            return runCommand("admin", copydbCmd.done(), *info);
+        }
+
+        // Otherwise, take or guess the auth mechanism
+        string authMech;
+        if ( strcmp( mechanism.c_str(), "DEFAULT" ) != 0 ) {
```
amidvidy: Why use `strcmp` here? Can't we just use the [std::string operator==](http://en.cppreference.com/w/cpp/string/basic_string/operator_cmp)

Can we define "DEFAULT", "MONGODB-CR", and "SCRAM-SHA-1" as constants kAuthMechanismDefault, kAuthMechanismMongoCR, kAuthMechanismScram, etc?

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+            if ( strcmp( mechanism.c_str(), "MONGODB-CR" ) != 0 &&
+                 strcmp( mechanism.c_str(), "SCRAM-SHA-1" ) != 0 ) {
+                throw UserException( 13422,
```
amidvidy: we don't want to add new uassert codes here. Also, can you use the uassert macro to write this as

uassert(0, "auth mechanism must be...",
            (authMech == kAuthMechanismDefault) || 
            (authMech == kAuthMechanism ....)

or put them in an array kAuthMechanisms and use std::count

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                                     "auth mechanism must be MONGODB-CR or SCRAM-SHA-1" );
+            }
+            authMech = mechanism;
+        }
+#ifdef MONGO_SSL
+        else if ( (static_cast<DBClientBase*>(this))->getMaxWireVersion() >= 3 ) {
+            authMech = "SCRAM-SHA-1";
```
amidvidy: replace string literal with constant

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+        }
+#endif
+        else {
+            authMech = "MONGODB-CR";
+        }
+
+        if ( strcmp( authMech.c_str(), "MONGODB-CR" ) == 0 ) {
```
amidvidy: use operator==

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+
+            copydbCmd.append("username", username);
+            copydbCmd.append("nonce", nonce);
+            copydbCmd.append("key", _authKeyCopyDB(username, password, nonce));
+            return runCommand("admin", copydbCmd.done(), *info);
+        }
+        else {
+            // run SCRAM-SHA-1 copydb, but only with SSL
+#ifndef MONGO_SSL
+            throw UserException( 13423, "SCRAM-SHA-1 requires driver to be built with SSL" );
```
amidvidy: use uasserted macro & no new uassert codes

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+#endif
+#ifdef MONGO_SSL
+            std::string hashedPwd = createPasswordDigest(username, password);
```
amidvidy: we should be consistent between namespace qualifying i.e. pick either `std::string` or `string` or not

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+            bool isServerDone = false;
+
+            // send copydbsaslstart, then continue to send copydb until we are done.
+            while (!session->isDone()) {
+                std::string payload;
+                BSONType type;
+
+                Status status = saslExtractPayload(inputObj, &payload, &type);
+                if (!status.isOK()) {
+                    throw DBException( str::stream() << "sasl session failure: " << status.reason(), 14355 );
```
amidvidy: no new error codes in the driver, use 0 or something from ErrorCodes

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                status = session->step(payload, &responsePayload);
+                if (!status.isOK()) {
+                    throw DBException( str::stream() << "sasl session failure: " << status.reason(), 14356 );
+                }
+
+                // build command to send to server
+                BSONObjBuilder commandBuilder;
+                commandBuilder.appendElements(saslCommandPrefix);
+                commandBuilder.appendBinData(saslCommandPayloadFieldName,
+                                             int(responsePayload.size()),
```
amidvidy: use static_cast<int>, this is equivalent to a c-style cast

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                BSONObj command = commandBuilder.obj();
+                bool ok = runCommand("admin", command, inputObj);
+
+                ErrorCodes::Error code =
+                    ErrorCodes::fromInt(inputObj[saslCommandCodeFieldName].numberInt());
+
+                if (!ok || code != ErrorCodes::OK) {
+                    if (code == ErrorCodes::OK)
+                        code = ErrorCodes::UnknownError;
+                    // TODO handle
```
amidvidy: is this TODO still relevant?

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                    // attempt to give a sane error message
+                    if ( inputObj.hasField("errmsg") ) {
+                        BSONElement e = inputObj.getField("errmsg");
+                        if (e.type() == String &&
+                            strstr( e.valuestr(), "no such cmd: saslStart")) {
+                            throw UserException( 13423,
```
amidvidy: no new error codes in the driver

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                                                 "fromhost doesn't support SCRAM-SHA-1, use MONGODB-CR");
+                        }
```
amidvidy: what if errmsg is a string, but a different error?

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                    }
+                    throw OperationException( inputObj );
+                }
+
+                isServerDone = inputObj[saslCommandDoneFieldName].trueValue();
+                saslCommandPrefix = saslFollowupCommandPrefix;
+            }
+            if (!isServerDone) {
+                assert(false); // TODO throw something better
```
amidvidy: perhaps uassert with a sensible ErrorCode

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+            }
+            return true;
+#endif /* MONGO_SSL */
+        }
+        assert(false);
```
amidvidy: use invariant(false) instead of assert for this

src/mongo/client/dbclientinterface.h
```diff
@@ -1248,6 +1254,13 @@ namespace mongo {
         virtual void _auth(const BSONObj& params);
 
         /**
+         * Create and return an MD5 digest for authentication for copydb.
+         */
+        std::string _authKeyCopyDB(const std::string &username,
+                                   const std::string &password,
```
amidvidy: if this function is not used elsewhere (or unit tested), can we just put it in an anonymous namespace and get rid of this declaration?

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+
+                Status status = saslExtractPayload(inputObj, &payload, &type);
+                if (!status.isOK()) {
+                    throw DBException( str::stream() << "sasl session failure: " << status.reason(), 14355 );
+                }
+
+                std::string responsePayload;
+                status = session->step(payload, &responsePayload);
+                if (!status.isOK()) {
+                    throw DBException( str::stream() << "sasl session failure: " << status.reason(), 14356 );
```
amidvidy: no new error codes in the driver use 0 or something from ErrorCodes

amidvidy:
mostly style stuff, still need a proper review of the sasl/scram stuff from @agralius

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+        if ( strcmp( mechanism.c_str(), "DEFAULT" ) != 0 ) {
+            if ( strcmp( mechanism.c_str(), "MONGODB-CR" ) != 0 &&
+                 strcmp( mechanism.c_str(), "SCRAM-SHA-1" ) != 0 ) {
+                throw UserException( 13422,
+                                     "auth mechanism must be MONGODB-CR or SCRAM-SHA-1" );
+            }
+            authMech = mechanism;
+        }
+#ifdef MONGO_SSL
+        else if ( (static_cast<DBClientBase*>(this))->getMaxWireVersion() >= 3 ) {
```
amidvidy: I think this is probably undefined behavior since DBClientBase is not a parent of DBClientWithCommands. I think we should use dynamic_cast here, but I will defer to @acmorrow

amidvidy: actually I think this is safe, nevermind.

src/mongo/client/dbclient.cpp
samantharitter: then we can't give a nice explanation and we'll have to just throw an operation exception instead

src/mongo/client/dbclient.cpp
samantharitter: well, it's not a user error at this point, it's a failure of some internal mechanism that would have caused the sasl session object and the server's sasl conversation to fall out-of-sync

amidvidy: how about OperationException ?

samantharitter: spose so

samantharitter: Actually I'm curious what @agralius thinks about this, it seems to me like a case that shouldn't really happen, perhaps it should be an invariant.  I ported this bit from the server's code.

src/mongo/client/dbclient.cpp
agralius: Can we make the name more descriptive. At least indicate that it's MONGODB-CR specific.

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+        string authMech;
+        if ( strcmp( mechanism.c_str(), "DEFAULT" ) != 0 ) {
+            if ( strcmp( mechanism.c_str(), "MONGODB-CR" ) != 0 &&
+                 strcmp( mechanism.c_str(), "SCRAM-SHA-1" ) != 0 ) {
+                throw UserException( 13422,
+                                     "auth mechanism must be MONGODB-CR or SCRAM-SHA-1" );
+            }
+            authMech = mechanism;
+        }
+#ifdef MONGO_SSL
```
agralius: I'm a little on the fence about this logic. Do we really want different behavior if SSL is enabled or not? It seems a little strange to me to default to MONGODB-CR if MONGO_SSL is not defined. At least we should verify that it's consistent with how the "normal" auth code chooses default mech.

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1070,175 @@ namespace mongo {
...
+                nonce = e.valuestr();
+            }
+
+            copydbCmd.append("username", username);
+            copydbCmd.append("nonce", nonce);
+            copydbCmd.append("key", _authKeyCopyDB(username, password, nonce));
+            return runCommand("admin", copydbCmd.done(), *info);
+        }
+        else {
+            // run SCRAM-SHA-1 copydb, but only with SSL
```
agralius: See my note above, we are already catching incorrect use of SCRAM-SHA-1 without SSL here.

src/mongo/client/dbclient.cpp
agralius: Well it is an invariant on the SASL mechanism used (SCRAM-SHA-1) in this case and the client and server implementations. The client "should" never finish before the server but it could theoretically happen if some version of the server considers SCRAM-SHA-1 taking more iterations then client.

src/mongo/client/dbclientinterface.h
```diff
@@ -897,7 +897,13 @@ namespace mongo {
 
            returns true if successful
         */
-        bool copyDatabase(const std::string &fromdb, const std::string &todb, const std::string &fromhost = "", BSONObj *info = 0);
+        bool copyDatabase(const std::string &fromdb,
+                          const std::string &todb,
+                          const std::string &fromhost = "",
+                          const std::string &mechanism = "DEFAULT",
```
agralius: You might consider providing SCRAM-SHA-1 or MONGODB-CR as the default to simplify the implementation.

src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -925,6 +953,57 @@ namespace {
         ASSERT_TRUE(doc["test"].boolean());
     }
 
+    TEST_F(DBClientTest, CopyDatabaseMONGODBCR) {
+        // only run this test if we have auth
+        if (serverGTE(&c, 2, 4) && !serverGTE(&c, 2, 7)) {
```
agralius: This test should work for new servers too. Not sure if I misunderstood the logic.

src/mongo/client/dbclient.cpp
samantharitter: The issue is that if the driver is build without --ssl, native_sasl_client_session.h seems not to be included in the driver, and this code won't compile without that conditional.  I don't think it's within the scope of this ticket to change that behavior, that would be a discussion for @acmorrow 

In the case that no mechanism is given and the max wire version is >= 3, would it be preferable to default to SCRAM-SHA, then fall through to the error that you pointed out at L1153 if the driver is not built with SSL?

src/mongo/client/dbclient.cpp
samantharitter: We'll only hit this case if the user provides "SCRAM-SHA-1" as a mechanism (so we don't have to infer it) and the driver is built without SSL.

src/mongo/client/dbclientinterface.h
samantharitter: The [DRIVERS ticket](https://jira.mongodb.org/browse/DRIVERS-188) for this was pretty explicit that the default value be either "DEFAULT" or another special value, so though I agree it would be nice to simplify it, I don't think it's an option.

src/mongo/integration/standalone/dbclient_test.cpp
samantharitter: The logic is that if the server is a new server, then we'll run the SCRAM-SHA-1 test instead of the MONGODB-CR test.

src/mongo/client/dbclient.cpp
samantharitter: Hm, so do you think it is worth making this an invariant at the risk of having to change it, should a future server version take more iterations?

samantharitter:
Thanks @amidvidy and @agralius I believe I've addressed all of your comments in this patch

src/mongo/client/dbclient.cpp
amidvidy: bump

src/mongo/client/dbclient.cpp
amidvidy: bump

src/mongo/client/dbclient.cpp
```diff
@@ -848,7 +874,7 @@ namespace mongo {
                 !(params.hasField(saslCommandUserDBFieldName)
                   && params.hasField(saslCommandUserSourceFieldName)));
 
-        if (mechanism == StringData("MONGODB-CR", StringData::LiteralTag())) {
+        if (mechanism == StringData(kAuthMechMongoCR, StringData::LiteralTag())) {
```
amidvidy: I think this line can just be

` if (mechanism == kAuthMechMongoCR) {`

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1092,153 @@ namespace mongo {
...
+
+        // If we don't have a username, or if we're copying locally,
+        // just run the command without authenticating
+        if ( username == "" || fromhost == "" ) {
+            return runCommand("admin", copydbCmd.done(), *info);
+        }
+
+        // Otherwise, take or guess the auth mechanism
+        string authMech;
+        if ( mechanism.c_str() != kAuthMechDefault ) {
```
amidvidy: you shouldn't need .c_str() here, `(mechanism != kAuthMechDefault)` should work

src/mongo/client/dbclient.cpp
```diff
@@ -1066,15 +1092,153 @@ namespace mongo {
...
+            uassert( 0, "auth mechanism must be MONGODB-CR or SCRAM-SHA-1",
+                     ( mechanism.c_str() != kAuthMechMongoCR &&
+                       mechanism.c_str() != kAuthMechScramSha1));
```
amidvidy: two things, first i think the logic here is inverted, second, as above you don't need c_str()

shouldn't this be
`uassert( 0, ..., (mechanism == kAuthMongoCR) || (mechanism == kAuthMechScramSha1)`)

src/mongo/client/dbclient.cpp
```diff
@@ -180,6 +188,24 @@ namespace mongo {
             }
             return bob.obj();
         }
+
+        string _authKeyCopyDBMongoCR(const string& username,
```
amidvidy: you don't need the preceding underscore in the declaration anymore since this isn't a private method.

amidvidy:
once you resolve the last few comments can you run a patch build on MCI for all platforms we target?

samantharitter:
done! https://mci.10gen.com/version/54bfcaf23ff1226829000128_0

src/mongo/client/dbclient.cpp
amidvidy: bump

src/mongo/client/dbclient.cpp
amidvidy: bump2

src/mongo/client/dbclient.cpp
amidvidy: bump2

amidvidy:
i just bumped a few nit comments, LGTM after those get resolved

Neustradamus:
Dear @mongodb team,

I wish you a Happy New Year 2022!

It is possible to add SCRAM-SHA-256 support?

Thanks in advance.

kevinAlbs:
> Dear @mongodb team,
> 
> I wish you a Happy New Year 2022!
> 
> It is possible to add SCRAM-SHA-256 support?
> 
> Thanks in advance.

SCRAM-SHA-256 was added to the C driver 1.11.0 in CDRIVER-2491. Please upgrade the the C++ driver dependency of the C driver to 1.11.0 or higher.

Please file bug tickets and feature requests in the MongoDB Jira CXX project: http://jira.mongodb.org/browse/CXX
For other assistance, please post in the community forums: https://www.mongodb.com/community/forums/tag/cxx

Neustradamus:
@kevinAlbs: Thanks for your reply and your confirmation! :)

PR #236, opened 2015-01-22
CXX-388 CXX-440 test against MongoDB 2.2

amidvidy:
@samantharitter  can you review?

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -26,9 +26,11 @@ cxx_driver_variables:
   version_test_list: &version_tests
     - name: "compile"
     - name: "unit test"
+    - name: "integration test 2.2"
     - name: "integration test 2.4"
     - name: "integration test 2.6"
     - name: "integration test latest"
+    - name: "client test 2.2"
     - name: "client test 2.4"
     - name: "client test 2.6"
     - name: "client test latest"
@@ -145,10 +147,10 @@ cxx_driver_variables:
           git clone https://github.com/10gen/mongo-orchestration.git
           cd mongo-orchestration
           python setup.py install
+          cd ..
           echo "Starting Mongo Orchestration..."
           echo "{ \"releases\": { \"default\": \"c:\\\\mongodb\\\\bin\" }, \"last_updated\": \"2014-08-29 20:57:00.000000\" }" > orchestration.config
           mongo-orchestration -f orchestration.config -e default -s wsgiref start
-          cd ..
       linux: &mongo_orchestration_linux
         ## mongo orchestration is included on linux, no installation needed
         start_mongo_orchestration: |
@@ -224,14 +226,21 @@ functions:
     params:
       updates:
         - key: "mongo_url"
-          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|"enterprise-"}${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|enterprise}-${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
 
   "set version 2.4" :
     command: expansions.update
     params:
       updates:
         - key: "mongo_url"
-          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|"subscription-"}${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|subscription}-${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
+
+  "set version 2.2" :
+    command: expansions.update
+    params:
+      updates:
+        - key: "mongo_url"
+          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|subscription}-${mongo_url_platform|}2.2.7.${mongo_url_extension|tgz}
 
   "run integration tests" :
     command: shell.exec
@@ -332,6 +341,15 @@ tasks:
                 set -o verbose
                 ${scons|scons} ${scons_flags} ${warning_flags} ${compile_flags} ${toolchain_flags} ${extrapath} ${dllpath} ${cpppath} ${libpath} unit
 
+    - name: "integration test 2.2"
+      depends_on:
+        - name: "compile"
+      commands:
+        - func: "set version 2.2"
+        - func: "fetch artifacts"
+        - func: "fetch mongodb"
+        - func: "run integration tests"
+
     - name: "integration test 2.4"
       depends_on:
         - name: "compile"
@@ -378,6 +396,15 @@ tasks:
         - func: "fetch artifacts"
         - func: "fetch mongodb"
         - func: "run integration tests"
+
+    - name: "client test 2.2"
+      depends_on:
+        - name: "compile"
+      commands:
+        - func: "set version 2.2"
+        - func: "fetch artifacts"
+        - func: "fetch mongodb"
+        - func: "run client tests"
 
     - name: "client test 2.4"
       depends_on:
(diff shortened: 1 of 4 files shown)


amidvidy:
whoops there were some issues with how i reordered the commits, going to rebase and force push

amidvidy:
ok good to go.

src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -1414,7 +1414,11 @@ namespace {
         if (serverGTE(&c, 2, 4) && (!serverGTE(&c, 2, 7) || kCompiledWithSSL)) {
             createUser(c, TEST_DB, "user3", "password3");
             std::string errmsg;
-            ASSERT_FALSE(c.auth("test", "user3", "notPassword3", errmsg));
+            try {
+                ASSERT_FALSE(c.auth("test", "user3", "notPassword3", errmsg));
+            } catch (const DBException&) {
+                //Expected on 2.2
```
acmorrow: Should we do the same song and dance about checking serverLTE here?

acmorrow: And if so, should we hoist serverLTE and serverGTE somewhere common?

amidvidy: yep, this is cherry-picked from spencer's earlier PR but i'll just add it on top

.mci.yml
```diff
@@ -224,14 +230,21 @@ functions:
     params:
       updates:
         - key: "mongo_url"
-          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|"enterprise-"}${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|enterprise}-${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
```
samantharitter: Some of the buildvariants don't have an enterprise keyword, I'd expect this to construct invalid urls with "--" in them.

I'd agree that it's a bit backwards for the default to be "enterprise-" and override with "" but that's the way it's currently set up for some reason.

.mci.yml
```diff
@@ -145,15 +147,19 @@ cxx_driver_variables:
...
           echo "{ \"releases\": { \"default\": \"c:\\\\mongodb\\\\bin\" }, \"last_updated\": \"2014-08-29 20:57:00.000000\" }" > orchestration.config
           mongo-orchestration -f orchestration.config -e default -s wsgiref start
-          cd ..
       linux: &mongo_orchestration_linux
-        ## mongo orchestration is included on linux, no installation needed
+        ## we are going to manually install mongo-orchestration for now on Linux until it is stable enough that we don't always to install from HEAD
         start_mongo_orchestration: |
           echo "Starting Mongo Orchestration..."
           trap 'set +o errexit; mongo-orchestration stop;' EXIT
+          git clone https://github.com/10gen/mongo-orchestration.git
```
samantharitter: Echo an additional line here for the installation of mongo orchestration, as above

.mci.yml
```diff
@@ -224,14 +230,21 @@ functions:
...
         - key: "mongo_url"
-          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|"subscription-"}${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|subscription}-${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
+
+  "set version 2.2" :
+    command: expansions.update
+    params:
+      updates:
+        - key: "mongo_url"
+          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|subscription}-${mongo_url_platform|}2.2.7.${mongo_url_extension|tgz}
```
samantharitter: same comment as above with the hyphens

src/mongo/client/examples/authTest.cpp
```diff
@@ -31,6 +31,18 @@
 
 using namespace mongo;
 
+bool serverLTE(DBClientBase* c, int major, int minor) {
```
acmorrow: Lets move this and serverGTE somewhere common.

.mci.yml
```diff
@@ -224,14 +230,21 @@ functions:
...
-          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|"enterprise-"}${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_enterprise_keyword|enterprise}-${mongo_url_platform|}2.6.5.${mongo_url_extension|tgz}
 
   "set version 2.4" :
     command: expansions.update
     params:
       updates:
         - key: "mongo_url"
-          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|"subscription-"}${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
+          value: ${mongo_url_prefix}${mongo_url_subscription_keyword|subscription}-${mongo_url_platform|}2.4.12.${mongo_url_extension|tgz}
```
samantharitter: ditto hyphens

src/mongo/integration/standalone/dbclient_test.cpp
amidvidy: i don't think there's a good place to put serverLTE and serverGTE unless we want some kind of mixin or utility file for both test fixtures and client examples

acmorrow: OK. Not worth it then. Go ahead.

src/mongo/client/examples/authTest.cpp
acmorrow: Nevermind...

src/mongo/integration/standalone/dbclient_test.cpp
samantharitter: Too bad, I feel like very similar fns are already defined elsewhere in our tests

acmorrow: Yeah, I forgot that this was an example. I'd prefer that the examples be standalone so that people can copy and paste them into things and have them run.

I'm as in favor of DRY as possible, but, this duplication is far from the worst thing in this codebase.

amidvidy: I think it might even be worth providing a std::pair<int, int> serverVersion() method on DBClientBase itself (with a properly bikeshedded return value), but at this point in the release cycle it seems better to leave as is.

.mci.yml
```diff
@@ -145,15 +147,19 @@ cxx_driver_variables:
           git clone https://github.com/10gen/mongo-orchestration.git
           cd mongo-orchestration
           python setup.py install
+          cd ..
```
amidvidy: note that i am moving this up so that mongo-orchestrations server.pid file will be in the main test directory so that mongo-orchestration stop works cleanly

src/mongo/integration/standalone/dbclient_test.cpp
amidvidy: also regarding the original comment, this test won't actually ever run on 2.2 so I think we can just get rid of the entire try/catch block and comment

.mci.yml
samantharitter: good idea, I was wondering why you'd done that

amidvidy:
addressed first round of comments

acmorrow:
Code changes LGTM. I defer to Samantha on MCI changes.

samantharitter:
Minus the picky hyphen stuff LGTM!

amidvidy:
cool, I just made a change to run the mongo orchestration install with sudo if we aren't root (fixes ubuntu build), i will push once that is green

amidvidy:
ubuntu is green https://mci.10gen.com/task/mongodb_cpp_driver_ubuntu1404_e240e0604678b1028aaee63e8de98e18047f7f31_15_01_22_22_05_09_integration_test_latest_with_WiredTiger_ubuntu1404

PR #238, opened 2015-01-23
create /data/db if it doesn't already exist on MCI

amidvidy:


.mci.yml
```diff
@@ -282,6 +282,9 @@ pre:
     params:
       script: |
         rm -rf "mongo-cxx-driver"
+        if [ ! -d /data/db ]; then
+            mkdir /data/db
```
samantharitter: Create /data directory first if there is none?

.mci.yml
```diff
@@ -282,6 +282,9 @@ pre:
...
+        fi
```
acmorrow: Maybe be a bit more defensive?
`
if [ ! -e /data/db ]; then
    mkdir /data/db
elif [ ! -d /data/db ]; then
    # raise an error that it exists but isn't a dir
fi
`

.mci.yml
acmorrow: Agree, just use -p?

amidvidy: /data directory is automatically created by the hostinit script when mci provisions a new builder. we wouldn't want to create it ourselves anyway, because we need it to be mounted on ephemeral storage.

acmorrow: Ah, right. OK then.

.mci.yml
amidvidy: done

acmorrow:
LGTM

PR #242, opened 2015-01-27
Prepare for 1.0.0 release

acmorrow:
There are two separate commits here, the BUMP and the post, along with more stubstantial than usual updates to the README. I figured, since we got this wrong for rc4 we should codereview.

The two commits need to be reviewed independently, not as cumulative edits.


acmorrow:
FYI you can view the readme here: https://github.com/acmorrow/mongo-cxx-driver/tree/release

samantharitter:
LGTM

amidvidy:
LGTM, A++ would bump again.

PR #243, opened 2015-01-27
Add a 32-bit windows buildvariant

samantharitter:
Note: the new buildvariant will only compile and run the unit tests.  The ssl and sasl libraries are not available for 32-bit windows, so we'll build without them and skip the client and integration tests.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -43,9 +43,11 @@ cxx_driver_variables:
       mongo_url_prefix: "http://downloads.mongodb.com/linux/mongodb-linux-x86_64-"
       mongo_url_platform: "ubuntu1404-"
     windows64-bitcomm: &mongo_url_windows64
-      mongo_url_extension: "zip"
       mongo_url_prefix: "http://downloads.mongodb.com/win32/mongodb-win32-x86_64-"
       mongo_url_platform: "windows-64-"
+    windows32: &mongo_url_windows32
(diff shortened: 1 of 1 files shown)


samantharitter:
https://mci.10gen.com/version/54c801ff3ff12237340000d2_0

amidvidy:
the refactor looks good to me. I think we're going to need to wait until we have a patch with both compile and unit green, though.

acmorrow:
The failed test is... odd.

samantharitter:
It is odd, it just kind of... stopped.  I really don't have any ideas about it, as the unit tests worked for me manually on a remote host.  I'm tempted to just run it again.

acmorrow:
It would be interesting if it crashed, but maybe just see if re-run makes it happy. New builder, we will keep an eye on it.

TylerBrock:
You can try adding `-s wsgiref` which will make it less likely to 404 when asking for a host.

acmorrow:
That doesn't look like the kind of problem here, this was a unit test that terminated.

samantharitter:
Right, this buildvariant won't even have mongo orchestration installed

TylerBrock:
Must have been looking at a different failure, nvm.

samantharitter:
We had a green patch on mci!  New version of the script is up, here's a new patch build:
https://mci.10gen.com/version/54c949313ff1220f390000b3_0

amidvidy:
changes LGTM, lets wait on the patch tho

PR #244, opened 2015-01-28
CXX-498 make OperationException inherit from DBException

amidvidy:
also a bit of cleanup in test exception handling.


amidvidy:
@TylerBrock, @acmorrow can you review?

src/mongo/client/exceptions.cpp
```diff
@@ -22,7 +22,8 @@ namespace mongo {
     }
 
     OperationException::OperationException(const BSONObj& errorObj)
-        : _lastError(errorObj)
+        : DBException(errorObj.toString(), 0)
+        , _lastError(errorObj)
```
acmorrow: So.... how many error strings do we have in this exception, now?

acmorrow:
Seems OK to me mod the thing about the strings, feel like we can consolidate?

acmorrow:
LGTM.

PR #247, opened 2015-02-03
fix mongocxx installation

amidvidy:
- fixes install target to actually install headers (currently nothing is installed)
- fixes install target to copy generated version.hpp from the build directory (without polluting the source tree)


amidvidy:
@TylerBrock this is pretty minor, but let me know if i can merge

TylerBrock:
:ship: it

PR #251, opened 2015-02-13
CXX-524 document the rest of bsoncxx

amidvidy:
I mostly punted on documenting builder::stream as it wouldn't really help users much. Going to write separate bson examples for that.


TylerBrock:
Seems fine.

PR #252, opened 2015-02-13
CXX-530 remove string_or_literal

amidvidy:


src/bsoncxx/builder/core.hpp
```diff
@@ -23,6 +23,7 @@
 #include <bsoncxx/array/view.hpp>
 #include <bsoncxx/document/value.hpp>
 #include <bsoncxx/document/view.hpp>
+#include <bsoncxx/stdx/string_view.hpp>
```
TylerBrock: Could possibly forward declare here?

src/bsoncxx/document/element.cpp
```diff
@@ -65,7 +65,7 @@ string_or_literal element::key() const {
 
     const char* key = bson_iter_key(&iter);
 
-    return string_or_literal{key, std::strlen(key)};
+    return stdx::string_view{key, std::strlen(key)};
```
acmorrow: Do you need the strlen here?

src/bsoncxx/document/element.cpp
```diff
@@ -144,8 +144,8 @@ types::b_regex element::get_regex() const {
     const char* options;
     const char* regex = bson_iter_regex(&iter, &options);
 
-    return types::b_regex{string_or_literal{regex, std::strlen(regex)},
-                          string_or_literal{options, std::strlen(options)}};
+    return types::b_regex{stdx::string_view{regex, std::strlen(regex)},
```
acmorrow: Ditto re strlen, everywhere

src/bsoncxx/builder/core.hpp
amidvidy: actually this is completely unnecessary

amidvidy:
@hanumantmk, how does this look

src/bsoncxx/test/bson_builder.cpp
```diff
@@ -583,10 +583,29 @@ TEST_CASE("array builder finalizes", "[bsoncxx::builder::stream]") {
...
 
+TEST_CASE("document core builder ownership", "[bsoncxx::builder::core]") {
+    builder::core b(false);
+
+    SECTION("when passing a std::string key, ownership transfers to the instance") {
+        {
+            std::string key{"falafel"};
+            b.key_owned(key);
+        }
+        b.append(1);
```
amidvidy: TODO, actually assert that the builder's view contains the correct keys.

src/bsoncxx/builder/core.cpp
```diff
@@ -178,7 +185,9 @@ class core::impl {
     bson_t _root;
 
     itoa _itoa_key;
-    string_or_literal _user_key;
+
+    stdx::string_view _user_key_view;
+    std::string _user_key_owning;
```
amidvidy: note to self, rename to _user_key_owned for consistency

src/bsoncxx/builder/stream/key_context.hpp
```diff
@@ -36,12 +37,12 @@ class key_context {
 
     template <std::size_t n>
     value_context<key_context> operator<<(const char (&v)[n]) {
-        _core->key_literal(v, n - 1);
+        _core->key_view(stdx::string_view{v, n - 1});
```
hanumantmk: I suspect we'll want a stdx::string_view overload as well as the literal and std::string options

hanumantmk:
My one critique would be that we probably want stdx::string_view overloads everywhere we have std::string overloads for the basic and stream builder apis.  As it is, we'll only pick up literals and std::strings, so we're probably doing some unnecessary copying if the user actually has string_views from some other source.

Other than that, LGTM

PR #254, opened 2015-02-18
Update SConscript.client

const-volatile:
If python is installed in a directory containing spaces (i.e. C:\Program Files(x86)\Python27) the scons script didn't worked (output was something like "the command C:\Program could not be found").

Adding the double quotes there solves the issue.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -456,7 +456,7 @@ tools = decide_platform_tools() + ["unittest", "integration_test", "textfile"]
 envDict = dict(BUILD_DIR=buildDir,
                VARIANT_DIR=get_variant_dir(),
                EXTRAPATH=get_option("extrapath"),
-               PYTHON=buildscripts.utils.find_python(),
+               PYTHON=File(buildscripts.utils.find_python()),
                tools=tools,
                PYSYSPLATFORM=os.sys.platform,
                CONFIGUREDIR=sconsDataDir.Dir('sconf_temp'),
--- a/src/SConscript.client
+++ b/src/SConscript.client
@@ -34,7 +34,7 @@ del env
 
 libEnv.Command(['mongo/base/error_codes.h', 'mongo/base/error_codes.cpp',],
                ['mongo/base/generate_error_codes.py', 'mongo/base/error_codes.err'],
-               '$PYTHON $SOURCES $TARGETS')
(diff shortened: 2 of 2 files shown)


acmorrow:
Hi @const-volatile could you please file a ticket in the CXX JIRA Project (https://jira.mongodb.org/browse/CXX) for this? We will need that in order to merge this change.

const-volatile:
I've added a JIRA ticket here: https://jira.mongodb.org/browse/CXX-538
Please let me know, if there is anything missing. Thank you

acmorrow:
Hi @const-volatile I agree that this is a problem but I think it would be better to fix this by wrapping the result of find_python in a File object here

https://github.com/mongodb/mongo-cxx-driver/blob/legacy/SConstruct#L459.

SCons will then know to quote it when invoking.

Can you see if that works for you, and, if so, update the PR? Also, please reference the ticket number at the beginning of the commit message.

Finally, have you signed the contributors agreement referenced in CONTRIBUTING.txt?

Thanks,
Andrew

const-volatile:
Changing this line in the SConstruct file to
`PYTHON="\""+buildscripts.utils.find_python()+"\"",`
solves the problem for me too. So should I rather close this pull request and create a new pull request with this change instead?

I've signed the contributors agreement today - thank you for pointing that out

acmorrow:
@const-volatile You can either close this and open a new PR, or if you are not opposed to doing so, just re-write the commit and force push to this branch. The PR will update to reflect that change.

RE using File vs. quoting: I do think that saying File is preferable to explicit quoting, as it delegates to SCons the decision about whether or not to quote, and in what way, and it doesn't require you to explicitly escape and concatenate.

Thanks for signing the agreement too.

const-volatile:
@acmorrow Thank you for your continuous help. Using "File()" works also fine on my machine - so I made this changes to this branch & reverted the change before. It seems the Travis CI build failed, but that doesn't seem related to me

acmorrow:
@const-volatile Agreed that the travis failure isn't you. If you could squash these into one commit and update the commit message to start with CXX-538 (have a look at the structure of most commits on the legacy branch) I'll be happy to merge this for you.

amidvidy:
merged in 245afd54

PR #255, opened 2015-02-18
CXX-536: Query::getMaxTimeMs() does not use the correct field name

ioancea:
Follow up of https://groups.google.com/forum/#!topic/mongodb-dev/oKnOaF56gto for the getMaxTimeMs() part:
- added static variable to hold the string "$maxTimeMS" in order to avoid mistakes
- adapted the tests in order to check if the *{m|M}axTimeMs() functions behave correctly


TylerBrock:
Thanks for reporting this and for the pull request.

PR #256, opened 2015-02-18
CXX-537: Add support for $maxTimeMS

ioancea:
Followup of https://groups.google.com/forum/#!topic/mongodb-dev/oKnOaF56gto for the maxTimeMS and 26compat branch part:
- merged the $maxTimeMS implementation from the legacy branch


TylerBrock:
We decided not to take this on the 26compat branch as it is meant to be a snapshot of upstream as of 2.6 and not have additional features. The feature can be found in the legacy driver however.

PR #257, opened 2015-02-18
CXX-533 safely expose implementation to users

TylerBrock:
I think it is OK to expose the interface as suggested by @acmorrow using stdx::optional so that we reserve the right to optionally return a pointer but actually always do so in the implementation. Questions, comments, and tips are all welcome.

### DIFF
--- a/src/mongocxx/bulk_write.cpp
+++ b/src/mongocxx/bulk_write.cpp
@@ -84,6 +84,10 @@ void bulk_write::append(const model::write& operation) {
     }
 }
 
+void* bulk_write::implementation() const {
+    return _impl->operation_t;
+}
+
 void bulk_write::write_concern(class write_concern wc) {
     libmongoc::bulk_operation_set_write_concern(_impl->operation_t, wc._impl->write_concern_t);
 }
--- a/src/mongocxx/bulk_write.hpp
+++ b/src/mongocxx/bulk_write.hpp
@@ -89,6 +89,15 @@ class MONGOCXX_API bulk_write {
     void append(const model::write& operation);
 
     ///
+    /// Gets a handle to the underlying implementation.
+    ///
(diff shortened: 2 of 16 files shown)


src/mongocxx/bulk_write.cpp
```diff
@@ -84,6 +84,10 @@ void bulk_write::append(const model::write& operation) {
     }
 }
 
+bsoncxx::stdx::optional<void*> bulk_write::implementation() const {
```
acmorrow: FYI, if you include mongocxx/stdx.h the bsoncxx polyfills are automatically imported into the mongocxx namespace.

src/mongocxx/bulk_write.cpp
```diff
@@ -84,6 +84,10 @@ void bulk_write::append(const model::write& operation) {
...
+    return bsoncxx::stdx::optional<void*>(_impl.get());
```
acmorrow: Do you actually need to name the return type here? Or can you just say return _impl.get()?

src/mongocxx/bulk_write.hpp
```diff
@@ -89,6 +89,13 @@ class MONGOCXX_API bulk_write {
     void append(const model::write& operation);
 
     ///
+    /// Gets a handle to the underlying implementation.
+    ///
+    /// @return Pointer to implementation for this object.
```
acmorrow: Maybe add ', or nullptr if no implementation object is available', here and elsewhere.

acmorrow: Also, add a note mentioning that the returned pointer is only valid for as long as this object is?

acmorrow:
LGTM with some suggestions for somewhat more comprehensive documentation, that I think you can just paste everywhere.

PR #258, opened 2015-02-23
CXX-546 test should pass 'flags' option key instead of 'usePowerOf2Sizes'

amidvidy:
see: http://docs.mongodb.org/manual/reference/command/create/#dbcmd.create

still unsure why it was passing in the first place.


TylerBrock:
LGTM

PR #260, opened 2015-02-26
CXX-489 add inline macros

amidvidy:
redux due to git snafu


amidvidy:
depends on #261

acmorrow:
@amidvidy This also looks odd as there are several commits from Tyler here?

amidvidy:
this PR got screwed up, and rebasing it seems to have confused github. I am going to open a new one after the json stuff gets in.

PR #261, opened 2015-02-26
CXX-547 de-inline json.hpp

amidvidy:
need this for CXX-489.

### DIFF
--- a/src/bsoncxx/json.cpp
+++ b/src/bsoncxx/json.cpp
@@ -1,0 +1,268 @@
+// Copyright 2015 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/json.hpp>
+
+#include <iomanip>
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <vector>
+
+#include <bsoncxx/document/view.hpp>
+#include <bsoncxx/stdx/make_unique.hpp>
+#include <bsoncxx/stdx/string_view.hpp>
+#include <bsoncxx/types.hpp>
+#include <bsoncxx/types/value.hpp>
+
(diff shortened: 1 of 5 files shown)


acmorrow:
@amidvidy something looks wrong about this PR.

amidvidy:
fixed pr and added MONGOCXX_API.

src/bsoncxx/json.cpp
```diff
@@ -0,0 +1,267 @@
+// Copyright 2014 MongoDB Inc.
```
acmorrow: 2015 now

src/bsoncxx/json.cpp
```diff
@@ -0,0 +1,267 @@
...
+            pad();
+            out << "]";
+        }
+
+        void visit_value(const types::b_binary& value) {
+
+            std::size_t b64_len;
+
+            b64_len = (value.size / 3 + 1) * 4 + 1;
+            std::unique_ptr<char> b64(reinterpret_cast<char*>(operator new(b64_len)));
```
acmorrow: Is there any reason that this isn't:

auto b64 = stdx::make_unique<char>(b64_len);

src/bsoncxx/json.cpp
```diff
@@ -0,0 +1,267 @@
...
+            visit_children(value.value);
+            pad();
+            out << "]";
+        }
+
+        void visit_value(const types::b_binary& value) {
+
+            std::size_t b64_len;
+
+            b64_len = (value.size / 3 + 1) * 4 + 1;
```
acmorrow: Why?

src/bsoncxx/json.cpp
amidvidy: I'm going to defer to @hanumantmk, he wrote this code, I'm just shuffling it around.

src/bsoncxx/json.cpp
amidvidy: @hanumantmk?

src/bsoncxx/json.cpp
amidvidy: i think that would make sense though.

hanumantmk: We can certainly change it over.  I think I had a bit of a psychological hangup over using new[] for things I associate with malloc when I was doing bsoncxx.

There's no real rational reason to use operator new() here though, definitely change it

src/bsoncxx/json.cpp
hanumantmk: the size computation?

It's how large things can get after you run them through ntop.  You started with binary and you're going to a string encoding, so you get a growth rate associated with going from 8bits/byte to 6bits/byte.  I think I cribbed that particular arithmetic from somewhere reputable, but I can't quite remember where now.

acmorrow:
LGTM

PR #262, opened 2015-02-27
CXX-516 support nolock in eval

amidvidy:
i know we hate bool parameters, but a tag or enum seemed like overkill here.

### DIFF
--- a/src/mongo/client/dbclientinterface.h
+++ b/src/mongo/client/dbclientinterface.h
@@ -1075,12 +1075,13 @@ namespace mongo {
            retValue  return value from the jscode function.
            args      args to pass to the jscode function.  when invoked, the 'args' variable will be defined
                      for use by the jscode.
+           nolock    if true, the server will not take a global write lock when executing the jscode.
 
            returns true if runs ok.
 
            See testDbEval() in dbclient.cpp for an example of usage.
         */
-        bool eval(const std::string &dbname, const std::string &jscode, BSONObj& info, BSONElement& retValue, BSONObj *args = 0);
+        bool eval(const std::string &dbname, const std::string &jscode, BSONObj& info, BSONElement& retValue, BSONObj *args = 0, bool nolock = false);
 
         /** validate a collection, checking for errors and reporting back statistics.
             this operation is slow and blocking.
--- a/src/mongo/client/dbclient.cpp
+++ b/src/mongo/client/dbclient.cpp
@@ -1458,11 +1458,15 @@ namespace mongo {
         return result.obj();
     }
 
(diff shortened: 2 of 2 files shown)


src/mongo/client/dbclientinterface.h
```diff
@@ -1080,7 +1081,7 @@ namespace mongo {
 
            See testDbEval() in dbclient.cpp for an example of usage.
         */
-        bool eval(const std::string &dbname, const std::string &jscode, BSONObj& info, BSONElement& retValue, BSONObj *args = 0);
+        bool eval(const std::string &dbname, const std::string &jscode, bool nolock, BSONObj& info, BSONElement& retValue, BSONObj *args = 0);
```
acmorrow: So, I realize we don't offer ABI compatibility, but I'd like to try if we can, and I'd even be a little sad to issue a patch release to 1.0.0 that broke callers at the source level.

We could make it a defaulted parameter which would at least retain source compatibility.

Or, we could introduce a new function.

Thoughts?

amidvidy: I think default parameter is the most sane of those options.
To be honest, few users will likely even call this version of eval rather than one of the more sane overloads that I left unchanged.
I think introducing an `evalNolock` function would be overkill and would (further) clutter the API.

acmorrow: Lets go with defaulted parameter.

amidvidy: done

src/mongo/client/dbclientinterface.h
```diff
@@ -1080,7 +1081,7 @@ namespace mongo {
 
            See testDbEval() in dbclient.cpp for an example of usage.
         */
-        bool eval(const std::string &dbname, const std::string &jscode, BSONObj& info, BSONElement& retValue, BSONObj *args = 0);
+        bool eval(const std::string &dbname, const std::string &jscode, BSONObj& info, BSONElement& retValue, bool nolock=false, BSONObj *args = 0);
```
acmorrow: I think that if you are adding a new default parameter, you must add it last. Otherwise callers passing a BSONObj for 'args' will suddenly get that coerced to bool and interpreted as a 'nolock' value, and still pass a NULL args, which probably isn't what they wanted.

amidvidy: whoops, fixed.

acmorrow:
Code LGTM. Any thoughts on testing this?

amidvidy:
i'm not sure what I would test other than ensuring that nolock is actually sent to the server when nolock=true, which seems fairly trivial.

acmorrow:
"trivial" as in trivial to test, or trivial in that it would provide little information/insight?

amidvidy:
the latter.

acmorrow:
OK, I think I agree, lets ship it.

PR #263, opened 2015-02-27
CXX-519 no null bytes allowed in fieldnames or regexes

amidvidy:


src/mongo/bson/bsonobjbuilder.h
```diff
@@ -752,6 +785,7 @@ namespace mongo {
         BufBuilder &subarrayStart() { return _b.subarrayStart( num() ); }
 
         BSONArrayBuilder& appendRegex(const StringData& regex, const StringData& options = "") {
+            uassert(0, "regex cannot contain null bytes", regex.find('\0') == std::string::npos);
```
TylerBrock: Can we use checkFieldName here too?

src/mongo/bson/bsonobjbuilder_test.cpp
```diff
@@ -244,4 +244,26 @@ namespace {
...
+        BSONObjBuilder b;
+        std::string middleNull("foo\0bar", 7);
+        std::string beginNull("\0foobar", 7);
+        std::string endNull("foobar\0", 7);
+        ASSERT_THROWS(b.append(middleNull, "baz"), UserException);
+        ASSERT_THROWS(b.append(beginNull, "baz"), UserException);
+        ASSERT_THROWS(b.append(endNull, "baz"), UserException);
+    }
+
+    TEST(BSONObjBuilderTest, AppendFieldNameContainingNullIsExceptionSafe) {
```
TylerBrock: Nice test.

src/mongo/bson/bsonobjbuilder.h
amidvidy: I wanted the error message to be clear that it's a regex as opposed to a fieldName.

samantharitter: Why not have a checkRegex method?

amidvidy: as we only use it twice, it's less code to just do the uassert inline.

acmorrow:
I think the code LGTM. However, I'm still not 100% that we should do this, based on our in-person discussion. Can you push back on the motivation for this in the underlying drivers ticket?

amidvidy:
merget in ae293a2fb

PR #264, opened 2015-03-03
CXX-518 don't allow sending _id with nested $ key

amidvidy:


src/mongo/SConscript
```diff
@@ -53,14 +53,15 @@ unittests = [
     'base/parse_number_test',
     'bson/bson_field_test',
     'bson/bson_obj_test',
```
amidvidy: extra changes are from sorting these lines

src/mongo/client/insert_write_operation.cpp
```diff
@@ -66,8 +66,10 @@ namespace mongo {
     }
 
     BSONObj InsertWriteOperation::_ensureId(const BSONObj& doc) {
-        if (doc.hasField("_id"))
+        if (doc.hasField("_id")) {
+            uassert(0, "value of _id element cannot contain any fields starting with $", doc.okForStorage());
```
acmorrow: Validating the whole object seems excessive. The point of this is to protect old servers from a certain unchecked failure mode, but only w.r.t. to the _id element.

I think you should re-write this, more like:

```
BSONElement id = doc.getField("_id");
if (!id.eoo()) {
    if (id has substructure) {
        BSONObj idObj = // get element as object
        uassert(0, "...", idObj.okForStorage();
    }
}
```

amidvidy: done.

src/mongo/client/insert_write_operation_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+        bob.append("foo", "bar");
+        const BSONObj doc = bob.done();
+        ASSERT_THROWS(InsertWriteOperation w(doc), UserException);
+    }
+
+    TEST(InsertWriteOperation, IdFieldCannotContainDollarSignNested) {
+        BSONObjBuilder bob;
+        BSONObjBuilder subbob;
+        BSONObjBuilder subsubbob;
+        BSONObjBuilder subsubsubbob;
```
acmorrow: This is a somewhat odd way of using BSONObjBuilder. Using 'subobjStart' might work better.

src/mongo/client/insert_write_operation.cpp
```diff
@@ -66,8 +67,11 @@ namespace mongo {
     }
 
     BSONObj InsertWriteOperation::_ensureId(const BSONObj& doc) {
-        if (doc.hasField("_id"))
+        BSONElement id = doc.getField("_id");
+        if (!id.eoo()) {
+            uassert(0, "value of _id element cannot contain any fields starting with $", !id.isABSONObj() || id.Obj().okForStorage());
```
acmorrow: Does isABSONObj also handle the case where _id is an array (I don't know that is possible, btw).

amidvidy: yes, it does.

src/mongo/client/insert_write_operation_test.cpp
amidvidy: my bad, I've been spoiled by `bsoncxx`. fixed

acmorrow:
LGTM

PR #265, opened 2015-03-03
CXX-517 nest $maxDistance inside $near and $nearSphere

amidvidy:


acmorrow:
Code LGTM.

Can we write a test?

Also do you agree that this will work against 2.4 servers? I think, from my reading of the document referenced in the DRIVERS ticket, is that it will, but I would like to double check.

PR #266, opened 2015-03-03
CXX-489 add inline macros

amidvidy:


amidvidy:
note that we can't include the inline keyword because we also need this when functions are defined at their declarations. libc++ does the same thing.

acmorrow:
LGTM

src/bsoncxx/builder/basic/impl.hpp
```diff
@@ -57,6 +60,7 @@ generic_append(core* core, T&& t) {
 }  // namespace
 
 template <typename T>
+BSONCXX_INLINE
```
TylerBrock: Might be nicer on the same line as template keyword?

amidvidy: i kind of like having it on its own line for templates so it doesn't distract from the template types.

PR #267, opened 2015-03-11
CXX-527 provide an API for BSON validation

amidvidy:
this is for a review of interface and implementation. Docs and tests will be added to this PR in a forthcoming commit.

### DIFF
--- a/src/bsoncxx/stdx/optional.hpp
+++ b/src/bsoncxx/stdx/optional.hpp
@@ -29,6 +29,7 @@ namespace stdx {
 
 using ::core::optional;
 using ::core::nullopt;
+using ::core::make_optional;
 
 }  // namespace stdx
 }  // namespace bsoncxx
@@ -41,6 +42,7 @@ namespace stdx {
 
 using ::std::experimental::optional;
 using ::std::experimental::nullopt;
+using ::std::experimental::make_optional;
 
 }  // namespace stdx
 }  // namespace bsoncxx
--- a/src/bsoncxx/test/bson_validate.cpp
+++ b/src/bsoncxx/test/bson_validate.cpp
@@ -1,0 +1,137 @@
+#include "catch.hpp"
+
+#include <bsoncxx/validate.hpp>
+
+#include <array>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/builder/stream/helpers.hpp>
+
+using namespace bsoncxx;
+
+// These are basic sanity tests to make sure the API works.
+// Libbson has a much more complete suite of tests that covers various corrupt BSON
+// inputs.
+
+namespace {
+    // polymorphic lambdas would be nice here.
+    template<typename T>
+    bool is_engaged(const stdx::optional<T>& opt) {
+        return opt != stdx::nullopt;
+    }
+
+    template<typename T>
+    bool is_disengaged(const stdx::optional<T>& opt) {
+        return opt == stdx::nullopt;
+    }
+}  // namespace
+
+
+TEST_CASE("validate accepts bson we produce", "[bsoncxx::validate]") {
+    builder::stream::document doc{};
+    doc << "hello" << "world";
+    auto view = doc.view();
+    // option has explicit bool conversion operator
+    REQUIRE(is_engaged(validate(view.data(), view.length())));
+}
+
+TEST_CASE("validate doesn't accept random bytes", "[bsoncxx::validate]") {
+    std::array<uint8_t, 12> arr{ 0xDE, 0xAD, 0xBE, 0xEF, 0xF0, 0x0B, 0x45 };
+    REQUIRE(is_disengaged(validate(arr.data(), arr.size())));
+}
+
+TEST_CASE("configuring optional validations", "[bsoncxx::validate]") {
+    builder::stream::document doc{};
+    validator vtor{};
+
+    SECTION("we can check for valid utf8") {
+        vtor.check_utf8(true);
+
+        // an invalid 3 octet sequence - thanks stackoverflow!
+        std::string invalid_utf8("\0xF0\0x28\0x8C\0x28", 3);
+        doc << "bar" << invalid_utf8;
+        auto view = doc.view();
+        REQUIRE(is_disengaged(validate(view.data(), view.length(), vtor)));
+    }
+
+    SECTION("we can check for valid utf8, but allow null bytes") {
+        vtor.check_utf8_allow_null(true);
+        std::string valid_utf8_with_null("foo\0\0bar", 8);
+        doc << "bar" << valid_utf8_with_null;
+        auto view = doc.view();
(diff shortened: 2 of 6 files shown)


amidvidy:
another question is if its worth PIMPLing the validator.

amidvidy:
added docs.

src/bsoncxx/stdx/optional.hpp
```diff
@@ -29,6 +29,7 @@ namespace stdx {
 
 using ::core::optional;
 using ::core::nullopt;
+using ::core::make_optional;
```
acmorrow: Neat

src/bsoncxx/validate.hpp
```diff
@@ -0,0 +1,130 @@
...
+/// @param validator
+///   A validator used to configure what checks are done. If the document if found
+///   to be invalid, it will contain the offset at which the document was found to
+///   be invalid.
+///
+/// @returns
+///   An engaged optional containing a view if the document is valid, or
+///   an unengaged optional if the document is invalid.
+///
+BSONCXX_API stdx::optional<document::view> validate(const std::uint8_t* data,
```
acmorrow: I see why you wanted variant. So you could return either a document view, or an object describing the failed validation.

src/bsoncxx/validate.hpp
```diff
@@ -0,0 +1,130 @@
...
+    ///   If true, UTF-8 validation (with null bytes allowed) is performed.
+    ///
+    void check_utf8_allow_null(bool check_utf8_allow_null);
+
+    ///
+    /// Verifies that document keys are not preceeded with '$'. Such keys
+    /// are reserved for MongoDB internal use.
+    ///
+    /// @param check_dollar_keys
+    ///   If true, UTF-8 validation (with null bytes allowed) is performed.
```
amidvidy: bad copy pasta

src/bsoncxx/validate.hpp
```diff
@@ -0,0 +1,130 @@
...
+    ///
+    void check_dollar_keys(bool check_dollar_keys);
+
+    ///
+    /// Verifies that document keys do not contain any '.' characters. The dot
+    /// character is illegal in MongoDB keys.
+    ///
+    /// @param check_dollar_keys
+    ///   If true, UTF-8 validation (with null bytes allowed) is performed.
```
amidvidy: ditto

src/bsoncxx/validate.hpp
```diff
@@ -0,0 +1,130 @@
...
+/// precise validation that is performed, and will give the caller access to the
+/// offset at which the document was found to be invalid.
+///
+/// @param data
+///   A buffer containing a BSON document to validate.
+/// @param length
+///   The size of the buffer.
+/// @param validator
+///   A validator used to configure what checks are done. If the document if found
+///   to be invalid, it will contain the offset at which the document was found to
```
amidvidy: less repetitive wording

src/bsoncxx/validate.hpp
amidvidy: yeah. expected<T, E> would be better, but the [mnmlstc version](http://mnmlstc.github.io/core/optional.html#expected:T:) differs from the version [proposed for the standard](https://isocpp.org/files/papers/n4015.pdf).

amidvidy:
added tests. @acmorrow care to give this a quick look?

src/bsoncxx/test/bson_validate.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+// These are basic sanity tests to make sure the API works.
+// Libbson has a much more complete suite of tests that covers various corrupt BSON
+// inputs.
+
+namespace {
+    // polymorphic lambdas would be nice here.
+    template<typename T> 
+    bool is_engaged(const stdx::optional<T>& opt) {
+        return opt != stdx::nullopt;
```
acmorrow: Can you just return static_cast<bool>(opt) to explictly invoke the explicit conversion to bool?

src/bsoncxx/test/bson_validate.cpp
```diff
@@ -0,0 +1,135 @@
...
+    }
+    
+    template<typename T>
+    bool is_disengaged(const stdx::optional<T>& opt) {
+        return opt == stdx::nullopt;
+    }
+}
```
acmorrow: // namespace

acmorrow:
Something of the design of this seems like it could be improved, to me. In particular, the fact that we need to friend in the validate function to the validator, and that the validator is stateful both seem not quite right. I think part of it is that you don't have the Either<T, U> return type.

I think there is a simpler design available, where the validator doesn't own the offset state, and is constant w.r.t. performing validation.

amidvidy:
we should probably discuss this one IRL tomorrow.

src/bsoncxx/validate.hpp
```diff
@@ -0,0 +1,136 @@
...
+///   A validator used to configure what checks are done. If validation fails, it
+///   will contain the offset at which the document was found to be invalid.
+///
+/// @returns
+///   An engaged optional containing a view if the document is valid, or
+///   an unengaged optional if the document is invalid.
+///
+BSONCXX_API stdx::optional<document::view> validate(const std::uint8_t* data,
+                                                    std::size_t length,
+                                                    validator& validator);
```
acmorrow: Lets make the invalid offset an out parameter:

BSONCXX_API stdx::optional<document::view> validate(
    const std::uint8_t* data,
    std::size_t length,
    const validator& validator = validator(),
    std::size_t** invalid_offset = NULL)

And then we can remove the friendship, and the member state.

src/bsoncxx/validate.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+stdx::optional<document::view> validate(const std::uint8_t* data, std::size_t length) {
+    validator vtor{};
+    return validate(data, length, vtor);
+}
+
+stdx::optional<document::view> validate(const std::uint8_t* data, std::size_t length, validator& validator) {
+    ::bson_validate_flags_t flags = BSON_VALIDATE_NONE;
+
+    auto flip_if = [&flags](bool cond, ::bson_validate_flags_t flag) {
```
acmorrow: can be const

src/bsoncxx/validate.cpp
```diff
@@ -0,0 +1,97 @@
...
+        validator._impl->_invalid_offset = stdx::make_optional(offset);
+        return {};
+    }
+
+    if (!::bson_validate(&bson, flags, &offset)) {
+        validator._impl->_invalid_offset = stdx::make_optional(offset);
+        return {};
+    }
+
+    return stdx::make_optional(document::view{data, length});
```
acmorrow: Can this be make_optional<Document>(data, length)?

amidvidy:
@acmorrow PTAL?

acmorrow:
LGTM. Next ABI rev maybe we will have a standards conformant Expected and we can get rid of the output parameter.

PR #268, opened 2015-03-15
CXX-560 Mark mongo::OID::asTimeT() and mongo::OID::asDateT() const

mapx:
As `mongo::OID::asDateT()` calls `mongo::OID::asTimeT()`, which calls `OID::getTimestamp()`, which is a const member function, `mongo::OID::asDateT()` and `mongo::OID::asTimeT()` should also be marked as const.


amidvidy:
Hey Yuchen,

This has been merged into the legacy driver in 624146f. Thanks for contributing!

Adam

PR #269, opened 2015-03-19
CXX-556 prevent converting rvalue document::values to document::views

amidvidy:
During skunkworks I found that this was a source of bugs. It is currently very easy to create a dangling document::view by passing a temporary value to an options struct. This change should make it a bit harder to do so.

### DIFF
--- a/src/bsoncxx/array/value.hpp
+++ b/src/bsoncxx/array/value.hpp
@@ -81,8 +81,11 @@ class BSONCXX_API value {
     ///
     /// Get a view over the document owned by this value.
     ///
-    inline BSONCXX_INLINE array::view view() const noexcept;
-    inline BSONCXX_INLINE operator array::view() const noexcept;
+    inline BSONCXX_INLINE array::view view() const & noexcept;
+    inline BSONCXX_INLINE array::view view() const && = delete;
+
+    inline BSONCXX_INLINE operator array::view() const & noexcept;
+    inline BSONCXX_INLINE operator array::view() const && = delete;
 
     ///
     /// Transfer ownership of the underlying buffer to the caller.
@@ -100,11 +103,11 @@ class BSONCXX_API value {
     std::size_t _length;
 };
 
-inline BSONCXX_INLINE array::view value::view() const noexcept {
+inline BSONCXX_INLINE array::view value::view() const & noexcept {
     return array::view{static_cast<uint8_t*>(_data.get()), _length};
 }
 
-inline BSONCXX_INLINE value::operator array::view() const noexcept {
+inline BSONCXX_INLINE value::operator array::view() const & noexcept {
     return view();
 }
 
--- a/src/bsoncxx/array/value.hpp
+++ b/src/bsoncxx/array/value.hpp
@@ -82,10 +82,10 @@ class BSONCXX_API value {
     /// Get a view over the document owned by this value.
     ///
     inline BSONCXX_INLINE array::view view() const & noexcept;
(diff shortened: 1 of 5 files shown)


src/bsoncxx/array/value.hpp
```diff
@@ -81,8 +81,11 @@ class BSONCXX_API value {
     ///
     /// Get a view over the document owned by this value.
     ///
-    inline BSONCXX_INLINE array::view view() const noexcept;
-    inline BSONCXX_INLINE operator array::view() const noexcept;
+    inline BSONCXX_INLINE array::view view() const & noexcept;
+    inline BSONCXX_INLINE array::view view() const && = delete;
+
+    inline BSONCXX_INLINE operator array::view() const & noexcept;
+    inline BSONCXX_INLINE operator array::view() const && = delete;
```
acmorrow: I think you can drop inline BSONCXX_INLINE on the deleted ones

amidvidy: done.

src/bsoncxx/array/value.hpp
```diff
@@ -81,8 +81,11 @@ class BSONCXX_API value {
     ///
     /// Get a view over the document owned by this value.
     ///
-    inline BSONCXX_INLINE array::view view() const noexcept;
-    inline BSONCXX_INLINE operator array::view() const noexcept;
+    inline BSONCXX_INLINE array::view view() const & noexcept;
+    inline array::view view() const && = delete;
```
acmorrow: You don't really need the 'inline' keyword on the deleted functions either.

acmorrow:
LGTM - two thoughts:

- Have you tested with both const and non-const rvalue refs to make sure that the 'const' && = delete actually does what we want? I remember having some debate about that over the fall.
- Is it worth (non-doxygen) commenting one of the deleted functions with an example of how it goes wrong if you don't delete those functions?

acmorrow:
@amidvidy can we get this pushed?

amidvidy:
will wrap this up tomorrow morning, have been focusing on server work.

acmorrow:
Looking at this again, something is puzzling me. I thought the whole point of the finalize token was to make explicitly the pattern that you are forbidding possible.

acmorrow:
I wonder if we would do better by deleting value&& overloads for things that store a copy of a view:

    void sort(bsoncxx::document::value&& ordering) = delete;

etc.

amidvidy:
Deleting value&& works as well - but is still unsatisfying as we now need a lot of boilerplate declarations. 
Additionally the compiler errors from this are a bit unfriendly. 

I think what we really want is the `view_or_value` that @hanumantmk proposed. Then if the user passes a view, we store a copy of it, and if the user passes a value, we take ownership. This gives the most usable interface, without sacrificing performance.

I think we could use the mnmlstc variant as it almost the same as Boost.Variant and will likely be quite similar to what eventually makes it in to the standard.

Then we can just do `using view_or_value = stdx::variant<document::view, document::value>`.

hanumantmk:
@amidvidy, I'm in pretty strong agreement.  Adding in a view/value variant to the options makes them massively more usable at a marginal implementation cost.  And if we're at all interested in giving up some performance for a nicer interface, we should just bite the bullet and go for it.

On the other hand, I think we probably want to avoid stdx'ing variant.  I doubt user's will want/need to use their own polyfill.  Maybe we're happy just dragging in the hard mnmlstc dep?

acmorrow:
@amidvidy - Is this review dead in light of more recent work? If so we should close it.

PR #270, opened 2015-03-26
CXX-479 Backport server r2.6.7..r2.6.9 changes

acmorrow:
Note that this is the backport of both 2.6.8 and 2.6.9. Server 2.6.8 had a client affecting bug, so we aren't going to be issuing a 26compat 2.6.8 release; opting instead to move on to 2.6.9 which contains the fix for the regression.

Please give it a look. There is more here than I expected. I'm particularly interested in someone evaluating the listCollections changes.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -96,7 +96,7 @@ def add_option( name, help, nargs, contributesToVariantDir,
     if type == 'choice' and not metavar:
         metavar = '[' + '|'.join(choices) + ']'
 
-    AddOption( "--" + name , 
+    AddOption( "--" + name ,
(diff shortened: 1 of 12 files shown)


amidvidy:
I stared at listCollections for a while and it looks sensible.

Question - are the SSL changes relevant? I'm not sure you would ever want to change the supported ciphers in the client, that would be done in the server.

Other than that LGTM.

amidvidy:
alright, given that we basically merge anything into 26compat that would have been in the driver had we not forked it, this LGTM as well.

PR #271, opened 2015-03-27
CXX-559 Backport server r3.0.0-rc6..r3.0.1 changes

acmorrow:
I skipped the following two SHAs from that range:
- 197fbbd - it wasn't clear that we wanted it or needed it
- afc23f0 - Please give this a close look. My read is that we don't need it, but the listX stuff has been a potent source of confusion.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -246,14 +246,7 @@ add_option('variables-help',
            "Print the help text for SCons variables", 0, False)
 
 if darwin:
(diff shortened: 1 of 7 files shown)


amidvidy:
I think we do want afc23f0, otherwise we could segfault if the system.indexes/namespaces query fails.

amidvidy:
after discussing IRL afc23f0 is unnecessary in the driver. LGTM!

PR #272, opened 2015-03-31
CXX-500 Only start mongo-orchestration from Scons when running examples

acmorrow:
This isn't pretty, but it gets rid of the annoying startup warnings and and delay when running things that aren't the examples.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1837,7 +1837,6 @@ Export("darwin windows solaris linux freebsd nix")
 Export("debugBuild optBuild")
 Export("use_clang")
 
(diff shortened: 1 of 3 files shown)


amidvidy:
can you put up a patch build?

acmorrow:
@amidvidy https://evergreen.mongodb.com/version/551ae9b73ff122135600089f_0

acmorrow:
Updated with a fix for the windows build variant failures. Still not clear on what is causing problems with the integration test suites: this shouldn't have affected it.

acmorrow:
@amidvidy given the M-O issues, should we go ahead with this? The client tests seem to be working fine.

amidvidy:
LGTM.

PR #273, opened 2015-04-09
CXX-572 Backport server r3.0.1..r3.0.2 changes

acmorrow:
Very small delta on this for the driver relevant changes between 3.0.1 and 3.0.2. These will flow into legacy-1.0.2 whenever that happens. No rush on this one, we haven't even shipped legacy-1.0.1 yet.


amidvidy:
LGTM

PR #277, opened 2015-04-13
CXX-599 call std::abort() if SetUpTestCase fails in read_preference_test

amidvidy:
I am going to limit this to read_preference_test as that is the only place we have the problem of test setup failing.


acmorrow:
LGTM

PR #278, opened 2015-04-14
CXX-555 add test coverage for collection::distinct

amidvidy:
I'm not using the mocking framework here because there isn't a clean way to interpose distinct without
using private APIs of the C driver.

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -1,7 +1,12 @@
 #include "catch.hpp"
 
+#include <vector>
+
 #include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <bsoncxx/stdx/string_view.hpp>
 #include <bsoncxx/types.hpp>
+
 #include <mongocxx/collection.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/pipeline.hpp>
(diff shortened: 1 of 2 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -388,4 +393,57 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
...
+        document doc1{};
+        doc1 << "foo" << "baz" << "garply" << 1;
+        document doc2{};
+        doc2 << "foo" << "bar" << "garply" << 2;
+        document doc3{};
+        doc3 << "foo" << "baz" << "garply" << 2;
+        document doc4{};
+        doc4 << "foo" << "quux" << "garply" << 9;
+
+        bulk_write bulk{false /* unordered */};
```
acmorrow: Off topic: Do we really just use raw bool's for the ordered-ness of a bulk_write? That makes me sad. Does CRUD spec allow us leeway here?

acmorrow:
So this test will require a live mongod somewhere to run, unlike all of our other tests?

amidvidy:
we already have a number of tests that require a live mongod (i.e. all the other tests in that file)

acmorrow:
OK then. LGTM

src/mongocxx/test/collection.cpp
amidvidy: yep. It's pretty gross. An enum class would definitely be better

acmorrow: Worth fixing?

PR #279, opened 2015-04-22
CXX-477 include error information in exceptions

amidvidy:


src/mongocxx/exception/write.hpp
```diff
@@ -16,13 +16,19 @@
...
 #include <mongocxx/exception/operation.hpp>
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace exception {
 
-class MONGOCXX_API write : public operation {};
+class MONGOCXX_API write : public operation {
+ public:
+   write(bsoncxx::document::value raw_server_error);
```
acmorrow: Can you use inherited constructors here?

src/mongocxx/exception/base.hpp
```diff
@@ -16,13 +16,41 @@
...
+#include <mongocxx/stdx.hpp>
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace exception {
 
-class MONGOCXX_API base : public std::exception {};
+using error_and_code_t = std::tuple<std::string, std::int32_t>;
+
+class MONGOCXX_API base : public virtual std::exception {
```
acmorrow: I think 'virtual public' is more idiomatic

src/mongocxx/exception/base.hpp
```diff
@@ -16,13 +16,41 @@
...
+
+    ///
+    /// @returns The raw server error, if it is available.
+    ///
+    const stdx::optional<bsoncxx::document::value>& raw_server_error();
+
+    ///
+    /// @returns The error message and code, if it is available.
+    ///
+    const stdx::optional<error_and_code_t>& error_and_code();
```
acmorrow: Blank line before 'private'

src/mongocxx/exception/base.hpp
```diff
@@ -16,13 +16,41 @@
...
+class MONGOCXX_API base : public virtual std::exception {
+   public:
+    base(bsoncxx::document::value raw_server_error);
+    base(error_and_code_t error_and_code);
+    base(bsoncxx::document::value raw_server_error, error_and_code_t error_and_code);
+
+    ///
+    /// @returns The raw server error, if it is available.
+    ///
+    const stdx::optional<bsoncxx::document::value>& raw_server_error();
```
acmorrow: Method can be const

src/mongocxx/exception/base.hpp
acmorrow: Method can be const?

src/mongocxx/database.cpp
```diff
@@ -14,6 +14,10 @@
 
 #include <mongocxx/database.hpp>
 
+#include <utility>
+
+#include <bsoncxx/stdx/make_unique.hpp>
```
acmorrow: used?

acmorrow: nm, just moved

src/mongocxx/exception/authentication.hpp
```diff
@@ -22,7 +22,10 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace exception {
 
-class MONGOCXX_API authentication : public operation {};
+class MONGOCXX_API authentication : public base {
+   public:
+    authentication(bsoncxx::document::value raw_server_error);
```
acmorrow: Having looked at several of these exception declarations, I'm pretty convinced that you can capitalize on constructor inheritance fully. Just using in the base constructors you want, in each derived exception, via the immediate super classes name.

src/mongocxx/exception/base.hpp
```diff
@@ -16,13 +16,41 @@
...
+#include <bsoncxx/stdx/optional.hpp>
+
+#include <mongocxx/stdx.hpp>
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace exception {
 
-class MONGOCXX_API base : public std::exception {};
+using error_and_code_t = std::tuple<std::string, std::int32_t>;
```
acmorrow: Is there a reason to favor two element tuple over two pair?

Also, does this type have meaning outside of exception::base? Should it be nested within?

Also also, no _t please.

src/mongocxx/exception/base.hpp
acmorrow: Should there be non-const overloads, so that you could move from the optionals if you wanted?

src/mongocxx/exception/base.hpp
amidvidy: since it has a meaning for all exceptions, I'd like to leave it in the exception namespace.

src/mongocxx/exception/base.hpp
amidvidy: yeah, I think that's a fair idea.

amidvidy:
@acmorrow I think I've addressed everything you brought up.

acmorrow:
LGTM

PR #280, opened 2015-04-22
CXX-553 include inline keyword in inline macros

amidvidy:
fixes warnings on GCC.


acmorrow:
LGTM

acmorrow:
FYI please adjust the fixVersion on this ticket when you resolve it since it is currently slated for 0.2.0

PR #281, opened 2015-04-29
CXX-588 Cannot use kvp() with a std::string param

hanumantmk:
sub_document.append() specializations for std::string and
stdx::string_view keyed tuples fail to forward for the various cv and
ref qualified variants.  To keep up the perfect forwarding, these
functions need to take the key type as a template type parameter and
SFINAE themselves to disambiguate.

### DIFF
--- a/src/bsoncxx/builder/basic/sub_document.hpp
+++ b/src/bsoncxx/builder/basic/sub_document.hpp
@@ -34,6 +34,11 @@ void value_append(core* core, T&& t);
 /// Users should almost always construct a builder::basic::document instead.
 ///
 class BSONCXX_API sub_document {
+    template <typename T, typename U>
+    struct is_same_stripped_type
+        : public std::is_same<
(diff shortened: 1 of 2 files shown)


src/bsoncxx/builder/basic/sub_document.hpp
```diff
@@ -34,6 +34,11 @@ void value_append(core* core, T&& t);
 /// Users should almost always construct a builder::basic::document instead.
 ///
 class BSONCXX_API sub_document {
+    template <typename T, typename U>
+    struct is_same_stripped_type
+        : public std::is_same<
+              typename std::remove_reference<typename std::remove_cv<T>::type>::type, U> {};
```
amidvidy: can you just make this std::is_same<typename std::decay<T>::type, U> {} ?

acmorrow:
@hanumantmk Can we try to close this one out?

PR #282, opened 2015-04-30
CXX-589 Add a view() method on pipeline

tracyJsonar:
Added a to_string function in mongocxx::pipeline.

Not sure if we want to check there whether the pipeline is empty, or to leave that to the caller?

### DIFF
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -151,9 +151,9 @@ class MONGOCXX_API pipeline {
     pipeline& unwind(std::string field_name);
 
     ///
-    /// Return a string representation of the pipeline.
+    /// @return A view of the BSON pipeline.
     ///
-    std::string to_string();
+    bsoncxx::document::view view() const;
 
    private:
     friend class collection;
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -150,6 +150,11 @@ class MONGOCXX_API pipeline {
     ///
     pipeline& unwind(std::string field_name);
 
+    ///
+    /// Return a string representation of the pipeline.
+    ///
+    std::string to_string();
+
    private:
     friend class collection;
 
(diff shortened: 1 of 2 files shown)


amidvidy:
Hi @tracyJsonar, thanks for the PR!

I think I would prefer to have this as a free function rather than a member method, so it can be picked up by ADL.

i.e. it should be called like
```
mongocxx::to_string(pipeline);
```
You will need to make a new member function of `pipeline` called view() that returns a `document::view` of the underlying BSON. I think an empty pipeline should just print an empty array - so no need to check. Currently [printing an empty array throws an exception](https://jira.mongodb.org/browse/CXX-587), but we are planning to fix that in the next release.

Can you also file a ticket on the [issue tracker](https://jira.mongodb.org/browse/CXX) and link this pull request in the comments?

Thanks,
Adam

tracyJsonar:
Thanks for the speedy response!

All of that sounds good - will get on that.
-Tracy

tracyJsonar:
Hi @amidvidy,

I wasn't sure whether to put the free function into pipeline.hpp, but now that there's `pipeline::view()`, it seems redundant to have a to_string function that simply calls `bsoncxx::to_json()`, which I'm already using in my code.

So for what I'm trying to do, having view() is sufficient.  However, if you think `to_string()` would still be handy, where should I put it, and would it do anything more/different than calling to_json()?

(Glad to know about the empty array fix!)

Thanks,
Tracy

src/mongocxx/pipeline.hpp
```diff
@@ -150,6 +150,11 @@ class MONGOCXX_API pipeline {
     ///
     pipeline& unwind(std::string field_name);
 
+    ///
+    /// @return A view of the BSON pipeline.
```
amidvidy: I would change this to `@return A view of the underlying BSON document this pipeline represents`

amidvidy:
Hey Tracy,

I agree that pipeline::view() is all that is really needed. I think this looks good to merge after you update the comment. Thanks so much for contributing to the C++ driver!

Adam

tracyJsonar:
That makes way more sense.  Thanks, Adam!

amidvidy:
This was merged in 7e69a7c0

PR #283, opened 2015-05-01
CXX-591: Added list_databases and list collections functionality

yhager:


src/mongocxx/client.cpp
```diff
@@ -66,5 +66,11 @@ class database client::database(bsoncxx::stdx::string_view name) const & {
     return mongocxx::database(*this, name);
 }
 
+cursor client::list_databases() const {
+    bson_error_t error;
+    auto result = libmongoc::client_find_databases(_impl->client_t, &error);
```
amidvidy: if result is nullptr, we should throw an exception::operation containing the message and code. See collection::create_index for an example.

src/mongocxx/client.hpp
```diff
@@ -167,6 +167,13 @@ class MONGOCXX_API client {
     MONGOCXX_INLINE class database operator[](bsoncxx::stdx::string_view name) const &;
     MONGOCXX_INLINE class database operator[](bsoncxx::stdx::string_view name) const && = delete;
 
+    ///
+    /// Enumerates the databases in the client
+    ///
+    /// @return mongocxx::cursor containing the databases information.
+    ///
+    cursor list_databases() const;
```
amidvidy: add ```@throws exception::operation if the underlying 'listIndexes' command fails```

src/mongocxx/client.hpp
```diff
@@ -167,6 +167,13 @@ class MONGOCXX_API client {
     MONGOCXX_INLINE class database operator[](bsoncxx::stdx::string_view name) const &;
     MONGOCXX_INLINE class database operator[](bsoncxx::stdx::string_view name) const && = delete;
 
+    ///
+    /// Enumerates the databases in the client
```
amidvidy: period at the end of the sentence.

src/mongocxx/client.hpp
```diff
@@ -167,6 +167,13 @@ class MONGOCXX_API client {
...
+    ///
+    /// @return mongocxx::cursor containing the databases information.
```
amidvidy: `A mongocxx::cursor containing a BSON document for each database. Each document contains Each document contains a name field with the database name, a sizeOnDisk field with the total size of the database file on disk in bytes, and an empty field specifying whether the database has any data.

`@see http://docs.mongodb.org/manual/reference/command/listDatabases/`

src/mongocxx/database.cpp
```diff
@@ -44,6 +44,14 @@ void* database::implementation() const {
     return _impl->database_t;
 }
 
+cursor database::list_collections(bsoncxx::document::view filter) {
+    libbson::scoped_bson_t filter_bson{filter};
+    bson_error_t error;
+
+    auto result = libmongoc::database_find_collections(_impl->database_t,
```
amidvidy: throw exception::operation if result is nullptr, as above.

src/mongocxx/database.hpp
```diff
@@ -113,9 +113,12 @@ class MONGOCXX_API database {
     ///
     /// Enumerates the collections in this database.
     ///
+    /// @see http://docs.mongodb.org/manual/reference/command/listCollections/
+    ///
+    /// @param filter an optional query expression to filter the returned list
```
amidvidy: Should be formatted like (also with minor wording change)

```
@param filter
   An optional query expression to filter the returned collections. 
```

src/mongocxx/database.hpp
```diff
@@ -113,9 +113,12 @@ class MONGOCXX_API database {
...
     /// @return mongocxx::cursor containing the collection information.
```
amidvidy: Also add `@throws` similar to `list_databases`.

amidvidy:
Hey @yhager, this is a good start. I have some comments regarding error handling and documentation formatting.

acmorrow:
@amidvidy @yhager Please ensure that the new APIs are in conformance with the published specification. See https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst

Should probably also do
https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst
?

amidvidy:
Yeah, mongocxx::collection::list_indexes is also unimplemented, I will create a new ticket for that.

amidvidy:
our list_collections interface is in conformance with the spec, as is our list_indexes method (though as I mentioned, it is unimplemented)

yhager:
Updated based on the comments above.

amidvidy:
merged! Great work @yhager. Thanks for your contributions.

PR #285, opened 2015-05-08
CXX-593: added bsoncxx::document::view::empty() helper

yhager:


src/bsoncxx/document/view.hpp
```diff
@@ -121,6 +121,13 @@ class BSONCXX_API view {
     ///
     std::size_t length() const;
 
+    ///
+    /// Checks if the underlying document is empty
```
amidvidy: I would elaborate a bit: 

```Checks if the underlying document is empty, i.e. it is equivalent to the trivial document '{}'.```

Also please put periods at the end of sentences.

amidvidy:
Hi @yhager,

Thanks for the PR and apologies for the delayed reply as I was out of the office last week. This looks good to me, with a minor nit about the comment.

acmorrow:
@amidvidy @yhager A question: What should happen in the view constructor if you pass NULL,0, or a pointer to valid bytes but length of 4?

amidvidy:
The result of that would be undefined as the contract of the view constructor that takes a buffer/len is that the buffer contains *valid BSON*. Of course, the caller could use `bsoncxx::validate` to ensure this.

acmorrow:
@amidvidy makes sense, thanks.

yhager:
modified the comment as you requested. 
@acmorrow, AFAICT, a valid bson document cannot have 4 bytes.

amidvidy:
merged in 05c1ca89. Thanks again @yhager!

PR #287, opened 2015-05-11
CXX-599 Backport server r3.0.2..r3.0.3 changes

acmorrow:
This is a little ahead of the game because we haven't even released legacy-1.0.2, but these are the server backports that would flow into a 1.0.3. There is some actual work in here, since I had to refactor how the SSL changes worked in some places, since we don't have SSLGlobalParams like the server does.

FYI @spencerjackson would appreciate you giving this a look as much of this is backports of your work.


acmorrow:
Closing. Wrong branch. Will re-submit

PR #288, opened 2015-05-12
CXX-571 Backport server 2.6.9..2.6.10 changes

acmorrow:
I'm being an optimist and assuming that server 2.6.10-rc0 will become server 2.6.10. If so, then these are the changes we would ship in the 26compat driver. Note that we would fix up the version before we did our label.

### DIFF
--- a/src/mongo/util/version.cpp
+++ b/src/mongo/util/version.cpp
@@ -31,7 +31,7 @@ namespace mongo {
      *      1.2.3-rc4-pre-
      * If you really need to do something else you'll need to fix _versionArray()
      */
-    const char versionString[] = "2.6.9";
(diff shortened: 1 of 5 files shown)


src/mongo/util/version.cpp
```diff
@@ -31,7 +31,7 @@ namespace mongo {
      *      1.2.3-rc4-pre-
      * If you really need to do something else you'll need to fix _versionArray()
      */
-    const char versionString[] = "2.6.9";
+    const char versionString[] = "2.6.10-rc0";
```
acmorrow: When 2.6.10 GA goes, I'll cherry-pick in the commit to bump this.

PR #290, opened 2015-05-12
CXX-595 Implement a basic from_json

acmorrow:


acmorrow:
Probably needs a run through clang-format

src/bsoncxx/test/json.cpp
```diff
@@ -0,0 +1,32 @@
+#include "catch.hpp"
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+
+namespace {
+    const auto k_invalid_json = R"({])";
```
amidvidy: constexpr? :)

acmorrow: Done

acmorrow:
Other thought is whether we want a version with more complex error reporting (like where we found the first wrong byte, etc.). I'm not sure it is worth the effort right now.

amidvidy:
LGTM

amidvidy:
merged in c9f1ad13

PR #291, opened 2015-05-13
CXX-588 Cannot use kvp() with a std::string param

yhager:
sub_document.append() specializations for std::string and
stdx::string_view keyed tuples fail to forward for the various cv and
ref qualified variants.  To keep up the perfect forwarding, these
functions need to take the key type as a template type parameter and
SFINAE themselves to disambiguate.

Credit: @hanumantmk, see PR #281


amidvidy:
This LGTM. @hanumantmk  PTAL

hanumantmk:
LGTM for me as well

PR #292, opened 2015-05-13
CXX-596 implement collection::list_indexes

amidvidy:


acmorrow:
Well, hard to argue with something that simple. LGTM. Please double check that we are meeting the requirements in https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst

PR #293, opened 2015-05-19
Fix for corruption of data appended to a gridfs  with appendChunk

craftit:
This is a fix for bug CXX-603, which I reported. 

The class mongo::GridFileBuilder, drops parts of appended data blocks from the stream, corrupting it.

Using mongo::GridFileBuilder:appendChunk to append data with arbitrary block sizes can drop part of the data block. The function call _appendChunk( data + size, length - size,false) is used without checking if there is any further data to add to the pending buffer.   The patch checks the returned value and adds it to the append buffer as done in other parts of the code.

### DIFF
--- a/src/mongo/client/gridfs.cpp
+++ b/src/mongo/client/gridfs.cpp
@@ -321,7 +321,14 @@ namespace mongo {
             invariant( _pendingDataSize <= _chunkSize );
             if (_pendingDataSize == _chunkSize) {
                 _appendPendingData();
-                _appendChunk( data + size, length - size, false );
+                const char* const end = data + length;
+                data = _appendChunk( data + size, length - size, false );
+                if (data != end) {
+                    invariant(data < end);
+                    size_t nsize = static_cast<size_t>(end - data);
+                    memcpy( _pendingData.get() + _pendingDataSize, data, nsize );
+                    _pendingDataSize += nsize;
+                }
             }
         }
         else {
--- a/src/mongo/integration/standalone/gridfs_test.cpp
+++ b/src/mongo/integration/standalone/gridfs_test.cpp
@@ -226,21 +226,45 @@ namespace {
 
     TEST_F(GridFSTest, GridFileBuilder) {
         GridFileBuilder gfb(_gfs.get());
-        for (int i=0; i<DATA_LEN; i+=2)
-            gfb.appendChunk(DATA + i, min(2, DATA_LEN - i));
+        size_t totalSize = 0;
+        for (int i=0; i<DATA_LEN; i+=2) {
+            size_t chunkSize = min(2, DATA_LEN - i);
+            gfb.appendChunk(DATA + i, chunkSize);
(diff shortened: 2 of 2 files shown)


src/mongo/client/gridfs.cpp
```diff
@@ -321,7 +321,13 @@ namespace mongo {
             invariant( _pendingDataSize <= _chunkSize );
             if (_pendingDataSize == _chunkSize) {
                 _appendPendingData();
-                _appendChunk( data + size, length - size, false );
+                const char* const end = data + length;
+                data = _appendChunk( data + size, length - size,false);
+                if (data != end) {
+                    size_t nsize = static_cast<size_t>(end-data);
```
acmorrow: spaces around arithmetic op. Also, consider invariant(end > data), maybe before entering the if block.

src/mongo/client/gridfs.cpp
```diff
@@ -321,7 +321,13 @@ namespace mongo {
             invariant( _pendingDataSize <= _chunkSize );
             if (_pendingDataSize == _chunkSize) {
                 _appendPendingData();
-                _appendChunk( data + size, length - size, false );
+                const char* const end = data + length;
+                data = _appendChunk( data + size, length - size,false);
```
acmorrow: space after comma before 'false'

acmorrow:
Looks good. Do you think you could add a test to src/mongo/integration/standalone/gridfs_test.cpp that would fail without this, and pass with it?

craftit:
I am working on a improved unit test now.

craftit:
Unfortunately I haven't been able to get the driver integration tests to run, but I have modified the tests to pick up one of the discrepancies I noticed during testing.

acmorrow:
@craftit Thanks for the fixups.

Yes, the integration tests are tricky to run. I've verified that your new tests fail when your fix is not applied and pass when it is.

src/mongo/client/gridfs.cpp
```diff
@@ -321,14 +321,21 @@ namespace mongo {
...
+                    _pendingDataSize += nsize;
+                }
             }
         }
         else {
             const char* const end = data + length;
             // split data in _chunkSize blocks, and store as pending the last block if
             // necessary
-            data = _appendChunk( data, length, false );
+            data = _appendChunk( data, length, false);
```
acmorrow: Unrelated? I know the formatting is inconsistent in this file, but I think you should restore the space you removed here.

acmorrow:
LGTM. Could you please squash to one commit and start the commit message with "CXX-603" (see the commit log for other examples), and I will get this merged.

Thanks for the bug report and for contributing to the C++ driver! This fix will be in the upcoming legacy-1.0.3-rc0 release.

craftit:
Done. Thanks for the all great work!

acmorrow:
Closed by 9980a0bfcb36b485c540a0fa372081eb610aa17d

PR #296, opened 2015-06-24
CXX-612 attach read preference properly in count/distinct commands

amidvidy:


src/mongo/integration/replica_set/read_preference.cpp
```diff
@@ -103,40 +103,43 @@ namespace {
...
-        test_conn->count(TEST_NS, q, QueryOption_SlaveOk);
+        BSONObj ignoredResult;
+        test_conn->runCommand(TEST_DB,  makeTestCommand("count", rp), ignoredResult);
     }
 
     void distinct(const auto_ptr<DBClientReplicaSet>& test_conn, ReadPreference rp) {
-        Query q = Query().readPref(rp, BSONArray());
-        test_conn->distinct(TEST_NS, "a", q);
+        BSONObj ignoredResult;
+        test_conn->runCommand(TEST_DB, makeTestCommand("count", rp), ignoredResult);
```
acmorrow: distinct?

acmorrow:
This looks good for commit so that the read_preference tests are actually testing what we think they are. However, as discussed yesterday, this doesn't really address the underlying problem. However, we also know that we don't need to fix the problem before server 3.2 ships.

So, LGTM, and please commit, but do not resolve the CXX-612 ticket. Instead, update its fix version to be 'planning', and kick it out of the sprint.

src/mongo/integration/replica_set/read_preference.cpp
amidvidy: whoops, yeah that should be fixed.

PR #297, opened 2015-06-25
CXX-611 Account for 3.1.4 storage engine swap

acmorrow:


acmorrow:
I can't patch queue or commit this yet, because the images need to be updated to mongo-orchestration 0.4 (see the ticket linked in CXX-611).

.mci.yml
```diff
@@ -394,12 +419,12 @@ tasks:
         - func: "fetch mongodb"
```
amidvidy: add "use WiredTiger storage" to be explicit.

.mci.yml
```diff
@@ -385,6 +400,16 @@ tasks:
...
+      depends_on:
+        - name: "compile"
+      commands:
+        - func: "set version 3.0"
+        - func: "use WiredTiger storage"
+        - func: "fetch artifacts"
+        - func: "fetch mongodb"
+        - func: "run integration tests"
+
     - name: "integration test latest"
```
amidvidy: change this to "integration test latest with WiredTiger"

amidvidy:
LGTM % nits

.mci.yml
acmorrow: I sort of left it off intentionally, the idea being that we only explicitly name the storage engine when it differs from the default on that version.

.mci.yml
acmorrow: Same idea: I don't want to specify what latest uses by default: I just want whatever the default is.

PR #298, opened 2015-06-30
CXX-616 simplify is_functor

amidvidy:


src/bsoncxx/util/functor.hpp
```diff
@@ -21,79 +21,12 @@
...
-
-template <typename F, typename S>
-struct is_functor_impl<F, S, false> : is_function_with_signature<F, S> {};
-
-}  // namespace functor
-
-template <typename C, typename S>
-struct is_functor : functor::is_functor_impl<C, S, std::is_class<C>::value> {};
+template <typename FunctionLike, typename Signature>
+using is_functor = std::is_convertible<FunctionLike, std::function<Signature>>;
```
acmorrow: If this works, I like it a lot. @hanumantmk?

hanumantmk: Seems plausible to me.  If it works I'm happy to have all that complexity out of the way

PR #300, opened 2015-07-20
CXX-549 add pool abstraction 

amidvidy:


src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+   public:
+    ///
+    /// Creates a pool associated with a connection string.
+    ///
+    /// @param mongodb_uri
+    ///  A MongoDB URI representing the connection parameters
+    /// @param ssl_options
+    ///  Optional SSL options to use when connecting to the MongoDB deployment.
+    pool(const uri& mongodb_uri = mongocxx::uri(),
+         stdx::optional<options::ssl> ssl_options = stdx::nullopt);
```
acmorrow: Are SSL options not expressible in the URI?

amidvidy: the c driver provides a separate function to set them: http://api.mongodb.org/c/current/mongoc_client_pool_set_ssl_opts.html

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    ~pool();
+
+    ///
+    /// A unique_client is a handle on a @c client object acquired via the pool.
+    ///
+    /// @note The lifetime of any unique_client object must be a subset of the pool object
+    ///  from which it was acquired.
+    ///
+    using unique_client = std::unique_ptr<client, pool::release>;
```
acmorrow: I think you can make release private by using a type here.

```
#include <iostream>
#include <memory>

class Things {

public:

    class Thing {
    public:
        Thing() { std::cout << "Creating new Thing\n"; }
        ~Thing() { std::cout << "Destroying Thing\n"; }
    };

private:
    class Deleter {
    public:
        Deleter(Things* things) :
            _things(things) {}

        void operator()(Things::Thing* thing) {
            _things->_release(thing);
        }

    private:
        Things* _things;
    };
    friend Deleter;

public:
    using ThingPtr = std::unique_ptr<Thing, Deleter>;

    ThingPtr get() {
        return ThingPtr(new Thing, Deleter(this));
    }

private:
    void _release(Thing* thing) {
        delete thing;
    }
};

int main(int argc, char** argv) {

    Things things;

    {
        auto x = things.get();
    }

    return 0;
}
```

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+class MONGOCXX_API pool {
+   public:
+    ///
+    /// Creates a pool associated with a connection string.
+    ///
+    /// @param mongodb_uri
+    ///  A MongoDB URI representing the connection parameters
+    /// @param ssl_options
+    ///  Optional SSL options to use when connecting to the MongoDB deployment.
+    pool(const uri& mongodb_uri = mongocxx::uri(),
```
acmorrow: How does the C driver configure the pool? Do we need to add options to the instance?

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    pool(pool&&) noexcept;
+
+    ///
+    /// Move assigns a pool.
+    ///
+    pool& operator=(pool&&) noexcept;
+
+    ///
+    /// Destroys a pool.
```
acmorrow: What happens to connections that I have acquired from the pool?

amidvidy: See the lifetime comment on unique client - I will add a note here as well.

src/mongocxx/pool.hpp
amidvidy: the c driver uses the connection string to configure the pool, aside from SSL options.

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+    ///  A MongoDB URI representing the connection parameters
+    /// @param ssl_options
+    ///  Optional SSL options to use when connecting to the MongoDB deployment.
+    pool(const uri& mongodb_uri = mongocxx::uri(),
+         stdx::optional<options::ssl> ssl_options = stdx::nullopt);
+
+    ///
+    /// Move constructs a pool.
+    ///
+    pool(pool&&) noexcept;
```
acmorrow: Moving a pool sounds dangerous to me. What happens if I try to return a connection to a moved from pool?

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,128 @@
...
+    /// @deprecated Future versions of the driver reserve the right to change the implementation
+    ///   and remove this interface entirely.
+    ///
+    /// @return Pointer to implementation of this object, or nullptr if not available.
+    ///
+    MONGOCXX_DEPRECATED void* implementation() const;
+
+   private:
+    class MONGOCXX_PRIVATE impl;
+    std::unique_ptr<impl> _impl;
```
acmorrow: If you decide against move for this, just make this a const std::unique_ptr<impl> and that should auto inhibit move.

src/mongocxx/pool.hpp
amidvidy: Would it be better to make the deleter completely opaque?
```c++
#include <iostream>                                                                                                                                                                                                
#include <memory>                                                                                                                                                                                                  
                                                                                                                                                                                                                   
//================ in header                                                                                                                                                                                       
                                                                                                                                                                                                                   
class Things {                                                                                                                                                                                                     
                                                                                                                                                                                                                   
public:                                                                                                                                                                                                            
  class Thing {                                                                                                                                                                                                    
  public:                                                                                                                                                                                                          
    Thing() { std::cout << "Creating new Thing\n"; }                                                                                                                                                               
    ~Thing() { std::cout << "Destroying Thing\n"; }                                                                                                                                                                
  };                                                                                                                                                                                                               
  using ThingPtr = std::unique_ptr<Thing, void (*)(Thing *)>;                                                                                                                                                      
                                                                                                                                                                                                                   
  ThingPtr get();                                                                                                                                                                                                  
};                                                                                                                                                                                                                 
                                                                                                                                                                                                                   
//============ in CPP file                                                                                                                                                                                         
                                                                                                                                                                                                                   
namespace {                                                                                                                                                                                                        
void deleter(Things::Thing *thing) { delete thing; }                                                                                                                                                               
}                                                                                                                                                                                                                  
                                                                                                                                                                                                                   
Things::ThingPtr Things::get() { return ThingPtr(new Thing, deleter); }                                                                                                                                            
                                                                                                                                                                                                                   
int main(int argc, char **argv) {                                                                                                                                                                                  
                                                                                                                                                                                                                   
  Things things;                                                                                                                                                                                                   
                                                                                                                                                                                                                   
  { auto x = things.get(); }                                                                                                                                                                                       
                                                                                                                                                                                                                   
  return 0;                                                                                                                                                                                                        
} 

```c++

Comment (acmorrow) [inline src/mongocxx/pool.hpp]: Well, the point of this example was that in the _release function you wanted to be back in the context of 'Things', since in the pool you want to add it back, not destroy it. I don't see how you are going to do that with a bare function.

Comment (acmorrow) [inline src/mongocxx/pool.hpp]: However, I don't care *that* much. I just agree that it is important that the release function be private, so users can't mess it up.

Comment (amidvidy) [inline src/mongocxx/pool.hpp]: The issue with my solution is that the deleter needs to have access to the inner state of Things to release the Thing - I'll think I'm going to play around with the code when I implement it so that we get something that is opaque as possible to a reader of the header file.

File: src/mongocxx/pool.hpp (hunk 5 above)
Comment (amidvidy) [inline src/mongocxx/pool.hpp]: I agree that this class should be non-moveable

File: src/mongocxx/pool.hpp (hunk 3 above)
Comment (acmorrow) [inline src/mongocxx/pool.hpp]: I wonder if, overall, we should try to standardize the C++11 driver interface on always using the URL to configure connections, and then internally do whatever song and dance is required to make the C driver SSL options by parsing and rebuilding the URI? It seems unfortunate that we don't follow the connection string standard

**Comment by amidvidy** [comment]
Drew, I am going to make the changes we discussed and write the implementation and tests in the next commit, I will ping again when it's ready for you to take a look at

File: src/mongocxx/pool.hpp (hunk 3 above)
Comment (amidvidy) [inline src/mongocxx/pool.hpp]: I don't think there is a standard for how drivers configure SSL options - lets discuss this in person.

Comment (acmorrow) [inline src/mongocxx/pool.hpp]: I think one is coming though. We can probably do it later though. If we are going to fix here, we should fix everywhere.

**Comment by amidvidy** [comment]
added implementation and tests.

**Comment by amidvidy** [comment]
@acmorrow  PTAL?

File: src/third_party/catch/include/helpers.hpp [hunk 7]
```diff
@@ -15,108 +15,124 @@
 #pragma once
```
amidvidy: changes in this file due to clang format.

src/mongocxx/client.cpp
```diff
@@ -23,10 +23,15 @@
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 
+// TODO: actually set client options
```
acmorrow: File a ticket for this?

src/mongocxx/client.hpp
```diff
@@ -183,8 +183,12 @@ class MONGOCXX_API client {
     cursor list_databases() const;
 
    private:
-    friend class database;
     friend class collection;
+    friend class database;
+    friend class pool;
+    friend class client_deleter;
```
acmorrow: Do you think this is better than having the pool have a private non-exported "_release" method?

src/mongocxx/pool.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <mongocxx/private/pool.hpp>
+#include <mongocxx/private/uri.hpp>
+#include <mongocxx/private/ssl.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+class client_deleter {
+   public:
+    client_deleter(pool* pool) : _pool(pool) {
```
acmorrow: explicit?

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,111 @@
...
+    ///
+    ~pool();
+
+    ///
+    /// A unique_client is a handle on a @c client object acquired via the pool.
+    ///
+    /// @note The lifetime of any unique_client object must be a subset of the pool object
+    ///  from which it was acquired.
+    ///
+    using unique_client = std::unique_ptr<client, std::function<void(client*)>>;
```
acmorrow: Should we call this 'entry'?

So:

pool::entry e = pool->acquire();

I don't care too much, but unique_client seems a little misleading.

src/mongocxx/pool.hpp
```diff
@@ -0,0 +1,111 @@
...
+    ///   and remove this interface entirely.
+    ///
+    /// @return Pointer to implementation of this object, or nullptr if not available.
+    ///
+    MONGOCXX_DEPRECATED void* implementation() const;
+
+   private:
+    friend class client_deleter;
+
+    stdx::optional<options::ssl> _ssl_options;
```
acmorrow: Why doesn't this live in the impl?

src/mongocxx/private/pool.hpp
```diff
@@ -0,0 +1,43 @@
...
+        libmongoc::client_pool_destroy(client_pool_t);
+    }
+
+    mongoc_client_pool_t* client_pool_t;
+};  // class impl
+
+MONGOCXX_INLINE_NAMESPACE_END
+};  // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
```
acmorrow: I'm not sure we need the postlude in private headers. Do we have it in the others?

src/mongocxx/client.cpp
amidvidy: filed CXX-640

src/mongocxx/private/pool.hpp
amidvidy: we do have it in the others.

src/mongocxx/pool.hpp
amidvidy: done

src/mongocxx/pool.hpp
amidvidy: done

src/mongocxx/pool.cpp
amidvidy: obviated by _release

src/mongocxx/client.hpp
amidvidy: agreed

amidvidy:
responded to comments.

src/mongocxx/private/pool.hpp
acmorrow: ok, lets keep it then

acmorrow:
Looks great to me. Ship it!

PR #301, opened 2015-07-21
CXX-639 Don't make assert behavior build type dependent

acmorrow:


acmorrow:
@amidvidy @hanumantmk Any thoughts on this one? It seems reasonable to me, and only affects debug builds. I'm inclined to ship it in a legacy-1.0.4 today. PTAL.

amidvidy:
LGTM

PR #302, opened 2015-07-22
CXX-578 expose ability to set log handler

amidvidy:


amidvidy:
going to add a test.

amidvidy:
added test - @acmorrow  PTAL?

src/mongocxx/private/libmongoc_symbols.hpp
```diff
@@ -12,20 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-MONGOCXX_LIBMONGOC_SYMBOL(bulk_operation_new)
```
amidvidy: I sorted these.

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,63 @@
 
 #include <mongocxx/instance.hpp>
 
+#include <mongoc.h>
```
acmorrow: Do we need this, if we are including libmongoc.hpp

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,63 @@
...
+#include <utility>
+
 #include <bsoncxx/stdx/make_unique.hpp>
 
-#include <mongoc.h>
+#include <mongocxx/private/libmongoc.hpp>
+#include <mongocxx/logger.hpp>
```
acmorrow: Sort

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,63 @@
...
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
 
 namespace {
-void log_handler(mongoc_log_level_t, const char *, const char *, void *) {
+extern "C" void null_log_handler(::mongoc_log_level_t, const char *, const char *, void *) {
```
acmorrow: Maybe move this down next to user_log_handler so it is easier to see the relationship

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,63 @@
...
+            // TODO: MONGOCXX_UNREACHABLE (CXX-628)
+            std::abort();
+    }
 }
 
-namespace mongocxx {
-MONGOCXX_INLINE_NAMESPACE_BEGIN
+extern "C" void user_log_handler(::mongoc_log_level_t mongoc_log_level, const char* log_domain,
+                                 const char* message, void* user_data) {
+    (*static_cast<logger*>(user_data))(convert_log_level(mongoc_log_level),
```
acmorrow: I think that you need to either:
- wrap this in try catch and do something on an exception, so that you don't throw through C context, OR
- require that the handler operator() be noexcept

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,63 @@
...
+}
+
+}  // namespace
 
 class instance::impl {
    public:
     impl() {
-        mongoc_init();
-        mongoc_log_set_handler(log_handler, nullptr);
+        libmongoc::init();
```
acmorrow: Nice catch

src/mongocxx/instance.hpp
```diff
@@ -49,10 +50,17 @@ class MONGOCXX_API instance {
     ///
     ~instance();
 
+    ///
+    /// Sets a user-provided @c logger for the driver to use.
+    ///
+    /// @param
+    ///  logger The logger that the driver will direct log messages to.
+    ///
+    void set_logger(std::unique_ptr<logger> logger);
```
acmorrow: What is the thread-safety level for this function? What happens if someone calls set_logger(a), then calls set_logger(b) from a's operator()?

src/mongocxx/logger.hpp
```diff
@@ -0,0 +1,66 @@
...
+    k_debug,
+    k_trace,
+};
+
+///
+/// The interface that all user-defined loggers must implement.
+///
+class MONGOCXX_API logger {
+   public:
+    virtual ~logger() = default;
```
acmorrow: I feel like we want the =default in the .cpp file for ABI reasons? For the ctor as well?

src/mongocxx/logger.hpp
```diff
@@ -0,0 +1,66 @@
...
+    /// this is called, such as log the output to a file or send it to a remote server for analysis.
+    ///
+    /// @param level
+    ///   The log level of the current log message
+    /// @param domain
+    ///   The domain of the current log message, such as 'client'
+    /// @param message
+    ///   The text of the current log message.
+    virtual void operator()(log_level level, stdx::string_view domain,
+                            stdx::string_view message) = 0;
```
acmorrow: See my other comment, but it might make sense to make this noexcept

src/mongocxx/logger.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+   protected:
+    logger() = default;
+};
+
+MONGOCXX_INLINE_NAMESPACE_END
+}  // namespace mongocxx
```
acmorrow: Needs a postlude header

src/mongocxx/instance.hpp
amidvidy: that would cause a deadlock - I will update the docs.

acmorrow: Perhaps there should also be a mutex around the logger object in the impl

amidvidy: added mutex.

amidvidy:
responded to comments - PTAL?

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,65 @@
...
+        default:
+            // TODO: MONGOCXX_UNREACHABLE (CXX-628)
+            std::abort();
+    }
+}
+
+extern "C" void null_log_handler(::mongoc_log_level_t, const char *, const char *, void *) {
 }
+
+extern "C" void user_log_handler(::mongoc_log_level_t mongoc_log_level, const char* log_domain,
```
acmorrow: I don't think you need extern "C" here

src/mongocxx/instance.cpp
```diff
@@ -14,28 +14,65 @@
...
+        libmongoc::log_set_handler(null_log_handler, nullptr);
     }
 
     ~impl() {
-        mongoc_cleanup();
+        libmongoc::cleanup();
     }
+
+    std::unique_ptr<logger> user_logger;
+    std::mutex logger_mtx;
```
acmorrow: Nit: put the lock before the things it guards

src/mongocxx/instance.cpp
```diff
@@ -45,5 +82,11 @@ instance::instance(instance &&) noexcept = default;
 instance &instance::operator=(instance &&) noexcept = default;
 instance::~instance() = default;
 
+void instance::set_logger(std::unique_ptr<logger> logger) {
+    std::lock_guard<std::mutex> lk(_impl->logger_mtx);
+    _impl->user_logger = std::move(logger);
```
acmorrow: This is unfortunately still racy. Thread A could be the currently configured logger while thread B is calling into set_logger.

Per discussion, lets make the logger a ctor arg to instance, and remove set_logger (and the mutex).

src/mongocxx/instance.cpp
```diff
@@ -14,31 +14,73 @@
...
+            return log_level::k_debug;
+        case MONGOC_LOG_LEVEL_TRACE:
+            return log_level::k_trace;
+        default:
+            // TODO: MONGOCXX_UNREACHABLE (CXX-628)
+            std::abort();
+    }
+}
+
+void null_log_handler(::mongoc_log_level_t, const char *, const char *, void *) {
```
acmorrow: This appears to be unused now. I think you should set it as the log handler in the instance::impl ctor for the case where the logger unique_ptr is null.

src/mongocxx/instance.cpp
```diff
@@ -14,31 +14,73 @@
...
+            libmongoc::log_set_handler(user_log_handler, user_logger.get());
+        }
     }
 
     ~impl() {
-        mongoc_cleanup();
+        libmongoc::cleanup();
     }
+
+    std::unique_ptr<logger> user_logger;
```
acmorrow: Pedantic, but, this is a public mutable class member. I'd suggest making this private, renaming to _user_logger, and making it const.

src/mongocxx/instance.cpp
```diff
@@ -14,31 +14,73 @@
...
+    impl(std::unique_ptr<logger> logger) : user_logger(std::move(logger)) {
+        libmongoc::init();
+        if (user_logger) {
+            libmongoc::log_set_handler(user_log_handler, user_logger.get());
+        }
     }
 
     ~impl() {
-        mongoc_cleanup();
+        libmongoc::cleanup();
```
acmorrow: Presumably, cleanup can log. Are we assured that after calling cleanup, the C driver will make no more calls to the logger? Or should we explicitly set it back to the null log handler here, rather than leaving it with a dangling pointer to a defunct user logger when we leave this scope?

src/mongocxx/instance.cpp
```diff
@@ -14,31 +14,73 @@
 
 #include <mongocxx/instance.hpp>
 
+#include <mutex>
```
acmorrow: This header is no longer needed

src/mongocxx/logger.hpp
```diff
@@ -0,0 +1,68 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/config/prelude.hpp>
+
+#include <memory>
+
+#include <bsoncxx/stdx/string_view.hpp>
+
```
acmorrow: Kill the blank line. Or, why do you need both stdx headers.

src/mongocxx/test/instance.cpp
```diff
@@ -0,0 +1,69 @@
...
+using namespace mongocxx;
+
+class test_log_handler : public logger {
+   public:
+    using event = std::tuple<log_level, std::string, std::string>;
+
+    test_log_handler(std::vector<event>* events) : _events(events) {
+    }
+
+    std::vector<event>* _events;
```
acmorrow: Make _events private

src/mongocxx/test/instance.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    void operator()(log_level level, stdx::string_view domain,
+                    stdx::string_view message) noexcept final {
+        try {
+            _events->emplace_back(level, std::string(domain), std::string(message));
+        } catch (...) {
```
acmorrow: The function is noexcept, so if it throws it will automatically call terminate. I think you can drop the try/catch

acmorrow:
I like this much better as a constructor argument. A bunch of nitpicky comments but I think this is basically ready to ship

src/mongocxx/logger.hpp
amidvidy: mongocxx/stdx.hpp doesn't actually include the bsoncxx/stdx headers, it just pulls stdx into the mongocxx namespace.

src/mongocxx/instance.cpp
amidvidy: done

src/mongocxx/instance.cpp
```diff
@@ -14,31 +14,79 @@
...
-        mongoc_cleanup();
+        // If we had a user logger, remove it so that it can't be used by the driver after it goes
+        // out of scope
+        if (_user_logger) {
+            libmongoc::log_set_handler(null_log_handler, nullptr);
+        }
+        libmongoc::cleanup();
     }
+
+    const std::unique_ptr<logger> _user_logger;
```
acmorrow: Make private

acmorrow:
LGTM

PR #304, opened 2015-08-05
CXX-646 force usage of SCRAM-SHA-1 when connecting to replica set members that support it

amidvidy:
The problem this fix solves is that we use the maxWireVersion of the `DBClientWithCommands` to pick an authentication mechanism in `DBClientWithCommands::auth`.

DBClientReplicaSet (unfortunately) inherits from DBClientWithCommands, so it uses its own _maxWireVersion - which doesn't make sense for a connection to a replica set as there may be a mixture of maxWireVersions among the sets members (during upgrade).

The quick solution is to just rewrite the auth parameters 'blob' with the correct mechanism when have obtained a connection from the monitor.

### DIFF
--- a/src/mongo/client/dbclient_rs.cpp
+++ b/src/mongo/client/dbclient_rs.cpp
@@ -22,6 +22,7 @@
 #include <memory>
 
 #include "mongo/bson/util/builder.h"
+#include "mongo/bson/bsonobjbuilder.h"
 #include "mongo/client/dbclientcursor.h"
 #include "mongo/client/replica_set_monitor.h"
 #include "mongo/client/sasl_client_authenticate.h"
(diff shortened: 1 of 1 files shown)


src/mongo/client/dbclient_rs.cpp
```diff
@@ -426,10 +427,25 @@ namespace {
                     break;
                 }
 
-                conn->auth( params );
+                BSONObj interposedParams;
```
acmorrow: Maybe start with BSONObj interposedParams = params. That way, you can eliminate the else block below.

src/mongo/client/dbclient_rs.cpp
```diff
@@ -426,10 +427,25 @@ namespace {
...
+                }
+                else {
+                    interposedParams = params;
+                }
+
+                conn->auth( interposedParams );
 
                 // Cache the new auth information since we now validated it's good
-                _auths[params[saslCommandUserDBFieldName].str()] = params.getOwned();
+                _auths[interposedParams[saslCommandUserDBFieldName].str()] =
```
acmorrow: I know you just changed params to interposedParams here, correctly. However, this code looks a little strange. If params/interposedParams here doesn't have a saslCommandUserDBFieldName, then this will store an entry in _auths with a key of ""?

acmorrow:
It is definitely a hack, but I agree that it is probably the most reasonable way to get the desired behavior. Any thoughts on testing?

Overall though, LGTM

acmorrow:
@amidvidy With server 3.0.5 out, it is time that we release legacy-1.0.5-rc0. Do you think we should try to land this change for legacy-1.0.5, or hold it for legacy-1.0.6? It is currently slated for the latter release in JIRA.

src/mongo/client/dbclient_rs.cpp
amidvidy: I think the logic, sadly enough, is that if the line above doesn't throw, then the params are valid.

amidvidy:
I think it may be worth waiting until legacy-1.0.6 so we can get this configuration tested with mongo orchestration.

src/mongo/client/dbclient_rs.cpp
acmorrow: Couldn't we be here but using an auth mechanism other than SASL? Or is the saslCommandUserDBFieldName field misnamed?

acmorrow:
OK, agree, lets hold this until after legacy-1.0.5

acmorrow:
@amidvidy FYI window for legacy-1.0.6 is now open. Should we move this forward?

amidvidy:
Superseded by #319.

PR #310, opened 2015-08-24
CXX-622 Backport server r3.0.5..r3.0.6 changes

acmorrow:
This is all of the code changes, and none of the clang format changes. The format work will go in under another ticket.

### DIFF
--- a/src/mongo/util/net/message_port.cpp
+++ b/src/mongo/util/net/message_port.cpp
@@ -167,7 +167,9 @@ namespace mongo {
     
     bool MessagingPort::recv(Message& m) {
         try {
+#ifdef MONGO_SSL
 again:
+#endif
             //mmm( log() << "*  recv() sock:" << this->sock << endl; )
             MSGHEADER::Value header;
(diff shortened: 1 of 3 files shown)


acmorrow:
Ping?

src/mongo/util/net/message_port.cpp
```diff
@@ -184,13 +186,6 @@ namespace mongo {
                 send( s.c_str(), s.size(), "http" );
                 return false;
             }
-            else if ( len == -1 ) {
-                // Endian check from the client, after connecting, to see what mode server is running in.
-                unsigned foo = 0x10203040;
```
amidvidy: is this change relevant? Not sure we can ever hit this codepath in the client

acmorrow: Most likely no, but it is generally easier to keep the code as close as possible, so that someday down the road we don't end up with some sort of merge conflict here and then need to do a bunch of archaeology.

amidvidy: ack

amidvidy:
LGTM

PR #313, opened 2015-09-09
CXX-664 Option to enable IPv6

v-nikulichev:


amidvidy:
Hi @v-nikulichev,  this looks good to me. I'm going to double check that this builds on our internal test infrastructure, but then it should be good to merge.

amidvidy:
Hey v-nikullchev - one issue - since you made this change we have reformatted our codebase. Could you run clang format on your changes, then re-upload this patch?

PR #314, opened 2015-09-21
CXX-669 Support Cmake 3.0.0

koraa:
Ubuntu vivid only provides binaries <= 3.1.0 and I am still on trusty for work :D

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(MONGO_CXX_DRIVER LANGUAGES CXX)
 
-cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.0.0 FATAL_ERROR)
 
 # Add in our modules
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/")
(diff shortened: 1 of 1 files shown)


acmorrow:
Hi @koraa thank you for the pull request. In order to simplify merging this, could you please:

- Open an associated issue in our JIRA tracker at https://jira.mongodb.org/browse/CXX
- Amend the commit message to start with the ticket ID (see the git history for examples)
- Sign the MongoDB contributors agreement at https://www.mongodb.com/legal/contributor-agreement

Thanks!

acmorrow:
@koraa - I'm sort of on the fence about this one.

I can see why this is helpful given that 3.1 isn't everywhere yet, but we did pick 3.1 as an explicit minimum precisely because of its support for CMAKE_CXX_STANDARD. The hack we added to work around the fact that it is broken for clang (at least in 3.1.0, possibly fixed in newer 3.1?) is something we would hope to remove, rather than go in the other direction and extend it to GCC.

Additionally, I wouldn't want to preclude using other novel CMake 3.1 features in the future.

However, since we aren't (other than CMAKE_CXX_STANDARD) using any of those features, I see why it would be frustrating when CMake 3.1 isn't widely available via package managers (yet).

acmorrow:
@koraa - In fact, given that it looks like CMake 3.2 fixed the issue, I'm somewhat inclined to actually move the cmake minimum forward, rather than back.

I realize that this is inconvenient but we are taking a hard line with this project that it will be released as 1.0.0 using the most modern tools available and with the absolute minimum of backward compatibility changes. There will be plenty of time for those to accrete in the future!

koraa:
There isn't really much to add to what you said, except that I choose this version of the driver for future production use because I thought I'd rather patch this one than deal with the old one :)
That sort of beta testing might be something beneficial to facilitate, then again, anyone who'd be patching the driver could be compiling cmake too…

acmorrow:
Yeah, the state of the C++11 driver is definitely in 'experts only' territory right now, but I'm happy to hear that you are using it and finding it to be an improvement over the old driver. We think it is much better too. I hope that my decision on this PR doesn't discourage you from submitting future patches: we are *absolutely* interested in beta testing as we work towards 1.0.

acmorrow:
Per discussion, closing this.

PR #315, opened 2015-09-24
CXX-675 Static bsoncxx library wrongly called libbson.a

koraa:
...


acmorrow:
This LGTM. Thanks for the fix!

PR #316, opened 2015-09-24
CXX-679 basic::builder should support concatenate

hanumantmk:
This adds support for concatenate in the basic builder api in a way that mirrors the stream api (with a tagged type)


acmorrow:
LGTM. Took me a minute to sort out what was going on, but I see it now.

acmorrow:
Merged in 8225f830dbb3fdd05fcdf8602730ef4a27e96802

Machyne:
Looks like this was merged. @acmorrow can we close this?

PR #319, opened 2015-10-06
CXX-646-redux - test with auth and fix connecting to a replica set with auth enabled

amidvidy:
This supersedes #304.

quick summary of commits:

- add evergreen configuration for running tests with authentication enabled
- authenticate with SCRAM against 3.0+ replicaSets
- use mongodb_auth_uri when available so we can authenticate
- rewrite integration test fixtures to use ConnectionString for all connections

This needs mongo-orchestration v0.4.2 for it to work on evergreen.

@acmorrow PTAL?

### DIFF
--- a/src/mongo/client/dbclient_rs.cpp
+++ b/src/mongo/client/dbclient_rs.cpp
@@ -21,8 +21,8 @@
 
 #include <memory>
 
-#include "mongo/bson/util/builder.h"
 #include "mongo/bson/bsonobjbuilder.h"
+#include "mongo/bson/util/builder.h"
 #include "mongo/client/dbclientcursor.h"
 #include "mongo/client/replica_set_monitor.h"
 #include "mongo/client/sasl_client_authenticate.h"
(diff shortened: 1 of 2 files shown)


src/mongo/client/dbclient_rs.cpp
```diff
@@ -22,6 +22,7 @@
 #include <memory>
 
 #include "mongo/bson/util/builder.h"
+#include "mongo/bson/bsonobjbuilder.h"
```
markbenvenuto: header sort

markbenvenuto:
:+1: except for nit

src/mongo/client/dbclient_rs.cpp
```diff
@@ -417,10 +418,23 @@ void DBClientReplicaSet::_auth(const BSONObj& params) {
                 break;
             }
 
-            conn->auth(params);
+            BSONObj interposedParams;
```
acmorrow: Why not just start with

BSONObj interposedParams = params;

? Then you woudn't need the else block below on line 430

src/mongo/client/dbclient_rs.cpp
```diff
@@ -417,10 +418,23 @@ void DBClientReplicaSet::_auth(const BSONObj& params) {
...
+                BSONObjBuilder interposedParamsBob;
+                interposedParamsBob.append(saslCommandMechanismFieldName, "SCRAM-SHA-1");
+                interposedParamsBob.appendElementsUnique(params);
+                interposedParams = interposedParamsBob.obj();
+            } else {
+                interposedParams = params;
+            }
+
+            conn->auth(interposedParams);
             // Cache the new auth information since we now validated it's good
```
acmorrow: blank line before comment line please

src/mongo/integration/standalone/dbclient_writer_test.cpp
```diff
@@ -52,9 +52,11 @@ template <typename T>
 class DBClientWriterTest : public StandaloneTest {
 public:
     DBClientWriterTest() {
-        c.connect(server().uri());
```
acmorrow: How did .uri work here at all before?

src/mongo/client/dbclient_rs.cpp
```diff
@@ -417,10 +418,23 @@ void DBClientReplicaSet::_auth(const BSONObj& params) {
                 break;
             }
 
-            conn->auth(params);
+            BSONObj interposedParams;
 
+            // Force usage of SCRAM-SHA-1 if host has maxWireVersion > 3.
```
acmorrow: Given that this code doesn't in some sense really belong here but is here for pragmatic reasons, I think a more elaborated comment is called for.

src/mongo/client/dbclient_rs.cpp
amidvidy: done

src/mongo/integration/standalone/dbclient_writer_test.cpp
amidvidy: MongoResource::uri() has some parsing logic that basically gives you hostname:port

acmorrow:
LGTM

PR #321, opened 2015-10-07
CXX-699 compute nextBatchSize() correctly

amidvidy:
need to write a test.

### DIFF
--- a/src/mongo/client/dbclientcursor.cpp
+++ b/src/mongo/client/dbclientcursor.cpp
@@ -92,16 +92,11 @@ void DBClientCursor::_finishConsInit() {
 }
 
 int DBClientCursor::nextBatchSize() {
-    if (nToReturn) {
-        int remaining = nToReturn - nReturned;
-
-        if (batchSize && batchSize < remaining)
-            return batchSize;
-
-        return -remaining;
-    }
-
-    return batchSize;
+    if (nToReturn == 0)
+        return batchSize;
+    if (batchSize == 0)
+        return nToReturn;
+    return std::min(batchSize, nToReturn);
 }
 
 void DBClientCursor::_assembleInit(Message& toSend) {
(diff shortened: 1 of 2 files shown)


amidvidy:
added test.

markbenvenuto:
:+1:

src/mongo/client/dbclientcursor.cpp
```diff
@@ -92,16 +92,11 @@ void DBClientCursor::_finishConsInit() {
 }
 
 int DBClientCursor::nextBatchSize() {
```
acmorrow: Is there some documentation somewhere that specifies that this is the correct interpretation? This has been a potent source of errors in the past

src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -1413,5 +1413,60 @@ TEST_F(DBClientTest, ConnectionStringWithTestDB) {
...
+                                 << "c"
+                                 << "dddddddddddddddddddddddddddd"));
+        }
+        bulk.execute(&WriteConcern::acknowledged, &result);
+        ASSERT_EQUALS(static_cast<std::size_t>(result.nInserted()), bulkSize);
+    }
+
+    {
+        // A sanity check.
+        auto_ptr<DBClientCursor> limitQuery1 = c.query(TEST_NS, mongo::Query("{}"), 1);
```
acmorrow: I see that this test changes the limit, but I think it is sticking with the default batch size? Is there a way we can explore extremal values of batchSize and ensure that things still work?

acmorrow:
LGTM mod two questions

src/mongo/client/dbclientcursor.cpp
amidvidy: this is the same logic in the server

amidvidy:
added additional tests w/ non-default batchSize

acmorrow:
LGTM

PR #322, opened 2015-10-08
CXX-626: Added conversion between b_date and std::chrono::system_cloc…

oferei:
solved CXX-626. requires review.

### DIFF
--- a/src/bsoncxx/test/bson_b_date.cpp
+++ b/src/bsoncxx/test/bson_b_date.cpp
@@ -1,0 +1,32 @@
+#include "catch.hpp"
+
+#include <bsoncxx/types.hpp>
+#include <chrono>
+
+TEST_CASE("time_point is converted to b_date and back", "[bsoncxx::types::b_date]") {
+    using bsoncxx::types::b_date;
+    using std::chrono::system_clock;
+    using std::chrono::time_point_cast;
+    using std::chrono::milliseconds;
+
+    system_clock::time_point now1, now2;
+
+    now1 = system_clock::now();
+    b_date d{now1};
+    now2 = d;
+
+    REQUIRE(time_point_cast<milliseconds>(now1) == time_point_cast<milliseconds>(now2));
+}
+
+TEST_CASE("time_point is converted to b_date consistently", "[bsoncxx::types::b_date]") {
+    using bsoncxx::types::b_date;
+    using std::chrono::system_clock;
+
+    system_clock::time_point now;
+
+    now = system_clock::now();
+    b_date d1{now};
+    b_date d2{now};
+
+    REQUIRE(d1.value == d2.value);
+}
--- a/src/bsoncxx/types.hpp
+++ b/src/bsoncxx/types.hpp
@@ -16,6 +16,7 @@
 
 #include <bsoncxx/config/prelude.hpp>
 
+#include <chrono>
 #include <cstring>
 
 #include <bsoncxx/array/view.hpp>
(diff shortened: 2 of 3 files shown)


src/bsoncxx/types.hpp
```diff
@@ -201,9 +201,21 @@ BSONCXX_INLINE bool operator==(const b_bool& lhs, const b_bool& rhs) {
 struct BSONCXX_API b_date {
     static constexpr auto type_id = type::k_date;
 
+    BSONCXX_INLINE
+    explicit b_date(int64_t value)
+        : value(value) {}
+
+    BSONCXX_INLINE
+    explicit b_date(const std::chrono::system_clock::time_point& tp)
```
amidvidy: I think you need to add a `<chrono>` include.

amidvidy:
I think travis is failing due to the missing include. Could you add it so we can check if your test passes CI?

Otherwise this looks pretty good! Thanks for contributing.

src/bsoncxx/test/bson_b_date.cpp
```diff
@@ -0,0 +1,22 @@
...
+    system_clock::time_point now1, now2;
+
+    now1 = system_clock::now();
+    b_date d{now1};
+    now2 = d;
+
+    //std::cout << "now1: " << time_point_cast<milliseconds>(now1).time_since_epoch().count() << std::endl;
+    //std::cout << "now2: " << time_point_cast<milliseconds>(now2).time_since_epoch().count() << std::endl;
+
+    REQUIRE(time_point_cast<milliseconds>(now1) == time_point_cast<milliseconds>(now2));
```
amidvidy: can we also add a test that creates two b_date objects from the same time_point and then checks that their `value` members are equal?

amidvidy:
Also, I think it is fine to only accept system_clock::time_point. steady_clock and high_resolution_clock are both pretty unsuitable for this purpose - if someone needs that we can add it later.

src/bsoncxx/test/bson_b_date.cpp
oferei: yeah, sure.

src/bsoncxx/test/bson_b_date.cpp
```diff
@@ -0,0 +1,22 @@
...
+    using std::chrono::milliseconds;
+
+    system_clock::time_point now1, now2;
+
+    now1 = system_clock::now();
+    b_date d{now1};
+    now2 = d;
+
+    //std::cout << "now1: " << time_point_cast<milliseconds>(now1).time_since_epoch().count() << std::endl;
+    //std::cout << "now2: " << time_point_cast<milliseconds>(now2).time_since_epoch().count() << std::endl;
```
amidvidy: also, please remove these print statements :)

src/bsoncxx/types.hpp
```diff
@@ -17,6 +17,7 @@
 #include <bsoncxx/config/prelude.hpp>
 
 #include <cstring>
+#include <chrono>
```
amidvidy: This is a bit of a nitpick, but for each header subsection we prefer to sort them - so <chrono> goes before <cstring>

amidvidy:
Great, now can you just squash this in to one commit with the original `CXX-626 ....` commit message, and then I'll merge it. Thanks again!

src/bsoncxx/types.hpp
oferei: yup. funny that I could build it locally.

oferei:
um, how do I squash?

amidvidy:
The easiest way is to do an interactive rebase. 

e.g. `git rebase -i origin/master`

Whatever you configured $EDITOR to be in your environment should pop up with a list of commits. Then you leave the 'c' in front of the first commit, and change the other lines to start with 's' so they get "squashed" in to the first commit. More info: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

amidvidy:
note that after you squash you will need to force-push to your feature branch.

oferei:
squashed.

amidvidy:
Great! going to merge now.

oferei:
awesome. glad I could help.

PR #323, opened 2015-10-08
CXX-656 deprecate eval helpers

amidvidy:


markbenvenuto:
:+1:

src/mongo/client/dbclientinterface.h
```diff
@@ -1205,9 +1213,22 @@ class MONGO_CLIENT_API DBClientWithCommands : public DBClientInterface {
...
     template <class T>
-    bool eval(const std::string& dbname, const std::string& jscode, T parm1) {
+    MONGO_CLIENT_DEPRECATED("deprecated in MongoDB 3.0") bool eval(const std::string& dbname,
+                                                                   const std::string& jscode,
+                                                                   T parm1) {
+        return evalDeprecated(dbname, jscode, parm1);
+    }
+
+    template <class T>
+    MONGO_CLIENT_DEPRECATED("deprecated in MongoDB 3.0") bool evalDeprecated(
```
acmorrow: Why does this one with Deprecated in the name have the CLIENT_DEPRECATED tag?

src/mongo/client/dbclient.cpp
```diff
@@ -1416,6 +1424,10 @@ bool DBClientWithCommands::eval(const string& dbname,
 }
 
 bool DBClientWithCommands::eval(const string& dbname, const string& jscode) {
```
acmorrow: Any reason not to just define this in the header?

src/mongo/client/dbclient.cpp
```diff
@@ -1401,6 +1401,14 @@ bool DBClientWithCommands::eval(const string& dbname,
                                 BSONElement& retValue,
                                 BSONObj* args,
                                 bool nolock) {
+    return evalDeprecated(dbname, jscode, info, retValue, args, nolock);
```
acmorrow: Any reason not to just define this in the header?

src/mongo/client/dbclientinterface.h
amidvidy: that was a mistake, removed.

src/mongo/client/dbclient.cpp
amidvidy: meh, I prefer to de-inline unless necessary.

src/mongo/client/dbclient.cpp
amidvidy: same

amidvidy:
removed unneeded deprecated macro

src/mongo/client/dbclient.cpp
acmorrow: In general I agree with you.

In this case though you are just creating a new symbol when there doesn't really need to be one. If inlined, the compiler will just throw away the tagged one and replace with calls to the xDeprecated.

So, you are making the library one symbol bigger for no real reason.

acmorrow:
LGTM

src/mongo/client/dbclient.cpp
acmorrow: I should clarify that I'm not saying you need to change it. I don't think we care about an extra symbol.

amidvidy: i think you have a good point, I'm going to change it since this seems like actually a pretty reasonable place to inline

PR #325, opened 2015-10-14
CXX-687: add option class for index creation

Machyne:
Create a new options class to be used on index creation.
This class supports the partial filter expression, so this also fixes CXX-686.


src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,418 @@
...
+        ///
+        /// @return The current config_str.
+        ///
+        /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+        ///
+        const bsoncxx::stdx::optional<std::string>& config_str() const;
+
+       private:
+        friend collection;
+        const int type() const override { return 1; }
```
amidvidy: can you de-inline type ?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,418 @@
...
+        void config_str(std::string config_str);
+
+        ///
+        /// The current config_str setting.
+        ///
+        /// @return The current config_str.
+        ///
+        /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+        ///
+        const bsoncxx::stdx::optional<std::string>& config_str() const;
```
amidvidy: just stdx::optional, you can #include mongocxx/stdx.hpp so it pulls all the stdx stuff in to the mongocxx namespace

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+        opt.name = options.name().value().c_str();
+    }
+    if (options.sparse()) {
+        opt.sparse = options.sparse().value();
+    }
+    if (options.storage_options()) {
+        const options::index::wiredtiger_storage_options* wt_options;
+        libmongoc::index_opt_wt_init(&wt_opt);
+        switch (options.storage_options()->type()) {
+            case 1:
```
amidvidy: case ::MONGOC_INDEX_STORAGE_OPT_WIREDTIGER ?

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    libmongoc::index_opt_init(&opt);
+
+    if (options.background()) {
+        opt.background = options.background().value();
+    }
+    if (options.unique()) {
+        opt.unique = options.unique().value();
+    }
+    if (options.name()) {
+        opt.name = options.name().value().c_str();
```
amidvidy: when deref'ing an optional, instead of options.name().value().c_str() can you write options().name()->c_str() ?

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    }
+    if (options.sparse()) {
+        opt.sparse = options.sparse().value();
+    }
+    if (options.storage_options()) {
+        const options::index::wiredtiger_storage_options* wt_options;
+        libmongoc::index_opt_wt_init(&wt_opt);
+        switch (options.storage_options()->type()) {
+            case 1:
+                wt_options = dynamic_cast<const options::index::wiredtiger_storage_options*>(options.storage_options());
```
amidvidy: if you are already sure that it is a wiredtiger_storage_opt_t, you can just reinterpret_cast

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+                if (wt_options->config_str()) {
+                    wt_opt.config_str = wt_options->config_str().value().c_str();
+                }
+                opt.storage_options = reinterpret_cast<mongoc_index_opt_storage_t *>(&wt_opt);
```
amidvidy: no space between type and *

amidvidy:
added a few initial comments.
Can you also run clang-format ?

.travis.yml
```diff
@@ -38,7 +38,8 @@ before_install:
     - sudo apt-get update -qq
 
     # Mongo C Driver
-    - git clone -b r1.2 https://github.com/mongodb/mongo-c-driver.git
+    # TODO switch master (branch for current 1.3 dev) to r1.3 when released
```
acmorrow: Can you do this in a separate commit? Should definitely do it, but it isn't part of this ticket. Feel free to file a new ticket against 0.3.0 and send out a new PR. This had been on my TODO list.

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+                }
+                opt.storage_options = reinterpret_cast<mongoc_index_opt_storage_t*>(&wt_opt);
+                break;
+            default:
+                // do nothing
+                break;
+        }
+    }
+    if (options.expire_after_seconds()) {
+        opt.expire_after_seconds = options.expire_after_seconds().value();
```
amidvidy: use * instead of value().

e.g.
`opt.expire_after_seconds = *options.expire_after_seconds()`

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
 }
 
 bsoncxx::document::value collection::create_index(bsoncxx::document::view keys,
-                                                  bsoncxx::document::view options) {
+                                                  const options::index& options) {
     scoped_bson_t bson_keys{keys};
+    scoped_bson_t weights{options.weights()};
+    scoped_bson_t partial_filter_expression{options.partial_filter_expression()};
     bson_error_t error;
+    mongoc_index_opt_geo_t geo_opt{};
```
amidvidy: just to be paranoid, can you prefix all mongoc symbols with `::` (global namespace resolution operator)

e.g.

`::mongoc_index_opt_geo_t geo_opt{}`

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,419 @@
...
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+    ///
+    const stdx::optional<bool>& sparse() const;
+
+    ///
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage
+    /// engine options for the index
+    ///
+    /// @param storage_options
+    ///   The storage engine options for the index in.
```
amidvidy: Can you add a section to the comment explaining that the base_storage_options* instance must stay in scope until after the call to create_index?

@acmorrow curious to hear your thoughts on this interface - I suggested it due to how the C driver implements WT-specific options internally

amidvidy:
can you verify that the options::index header is installed when the driver is installed?

Machyne:
Verified – options::index header is installed with the driver!

src/mongocxx/options/index.hpp
Machyne: (warning added)

examples/mongocxx/index.cpp
```diff
@@ -32,4 +32,28 @@ int main(int, char**) {
         db["restaurants"].create_index(index_spec, {});
         // @end: cpp-create-compound-index
     }
+
+    // Create a unique index.
```
samantharitter: +1 for examples!

Machyne:
Updated, PTAL

Machyne:
ping

examples/mongocxx/index.cpp
```diff
@@ -32,4 +32,28 @@ int main(int, char**) {
...
+    {
+        // @begin: cpp-create-unique-index
+        document index_spec;
+        mongocxx::options::index index_options{};
+        index_spec << "cuisine" << 1;
```
samantharitter: Maybe pick a different field for your unique index?  It's not a great example because many restaurants should be able to have the same type of cuisine.

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    if (options.name()) {
+        opt.name = options.name()->c_str();
+    }
+    if (options.sparse()) {
+        opt.sparse = *options.sparse();
+    }
+    if (options.storage_options()) {
+        const options::index::wiredtiger_storage_options* wt_options;
+        libmongoc::index_opt_wt_init(&wt_opt);
+        switch (options.storage_options()->type()) {
```
samantharitter: Is it worth having a switch statement here for just the one case?  An if statement would be clearer.

src/mongocxx/collection.cpp
```diff
@@ -426,12 +426,89 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
     scoped_bson_t bson_keys{keys};
+    scoped_bson_t weights{options.weights()};
+    scoped_bson_t partial_filter_expression{options.partial_filter_expression()};
     bson_error_t error;
+    ::mongoc_index_opt_geo_t geo_opt{};
+    ::mongoc_index_opt_t opt{};
+    ::mongoc_index_opt_wt_t wt_opt{};
+    libmongoc::index_opt_init(&opt);
+
+    if (options.background()) {
```
samantharitter: Wondering now, since in other PRs we are moving towards options classes that transform themselves into useful forms, whether this logic should all be here or whether options::index should have a convenience method for this.  Thoughts?

src/mongocxx/options/index.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void index::twod_location_max(double twod_location_max) {
+    _twod_location_max = twod_location_max;
+}
+
+void index::haystack_bucket_size(double haystack_bucket_size) {
+    _haystack_bucket_size = haystack_bucket_size;
+}
+
+const stdx::optional<bool>& index::background() const {
```
samantharitter: If we go down the convenience method path, then these getters can all go away.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+class collection;
+
+namespace options {
+
+///
+/// Class representing the optional arguments to a MongoDB create index operation
```
samantharitter: nit, period at the end of the sentence

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,423 @@
...
+
+    ///
+    /// Whether or not to create a unique index so that the collection will not
+    /// accept insertion of documents where the index key or keys match an
+    /// existing value in the index.
+    ///
+    /// @param unique
+    ///   Whether or not to create a unique index.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
```
samantharitter: Here, I would link to the docs page about unique indexes instead of the createIndex options page: https://docs.mongodb.org/manual/core/index-unique/

src/mongocxx/options/index.hpp
samantharitter: Link here to the createIndex options page

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,423 @@
...
+    index();
+
+    ///
+    /// Whether or not to build the index in the background so that building
+    /// the index does not block other database activities.
+    ///
+    /// @param background
+    ///   Whether or not to build the index in the background.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
```
samantharitter: Link to page about background index builds instead: https://docs.mongodb.org/manual/tutorial/build-indexes-in-the-background/

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,423 @@
...
+    ///
+    void unique(bool unique);
+
+    ///
+    /// The current unique setting.
+    ///
+    /// @return The current unique.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
```
samantharitter: I don't think you need these links for all the methods.  I'd put this link on the options class, then link to relevant docs pages for specific features.  IMO getters don't need a link.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,423 @@
...
+    ///
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage
+    /// engine options for the index
+    ///
+    /// @param storage_options
+    ///   The storage engine options for the index in.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+    ///
+    /// @warning
```
samantharitter: I'd move this warning up higher, to right after the description, before the @param part.

src/mongocxx/collection.cpp
Machyne: I think it would be nice to have, but this one can't just return the bson view, it needs to return the mongoc options struct. We could just return a void*, but that seems a bit sketchy to me. Could we put the helper in a private header?

samantharitter: It could be a method that took a void * and populated it instead:
```
void export_options(void* opts) {
   // cast to mongoc_index_opt_t *
   // carry on
}
```

This is also sketchy but is marginally less sketchy, I think

samantharitter: It's also not a requirement.  If it turns out there are other commands that take index options (I can't think of any right now) then this is probably worth doing, otherwise it's up to you.

Machyne: It will only be this command.

I think leaving it for now will be fine. Maybe we will want to do a sweep later of adding these helpers to all options classes, and we could get this one then.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+        ///
+        /// The current config_str setting.
+        ///
+        /// @return The current config_str.
+        ///
+        const stdx::optional<std::string>& config_str() const;
+
+       private:
+        friend collection;
+        const int type() const override;
```
amidvidy: MONGOCXX_PRIVATE type()

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+///
+class MONGOCXX_API index {
+   public:
+    ///
+    /// Base class representing the optional storage engine options for indexes.
+    ///
+    class base_storage_options {
+       private:
+        friend collection;
+        virtual const int type() const = 0;
```
amidvidy: make type() MONGOCXX_PRIVATE

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -196,6 +210,14 @@ TEST_CASE("Collection", "[collection]") {
             REQUIRE_THROWS_AS(mongo_coll.create_index(index_spec), exception::operation);
         }
 
+        SECTION("Succeeds With Options") {
+            success = true;
+            options.unique(expected_unique);
+            options.expire_after_seconds(expected_expire_after_seconds);
+            options.name(expected_name);
+            REQUIRE_NOTHROW(mongo_coll.create_index(index_spec, options));
+        }
```
amidvidy: can you add a test specific for the wiredtiger_storage_options?

Machyne: Added!

amidvidy:
LGTM

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+    ///   Whether or not to create a unique index.
+    ///
+    void unique(bool unique);
+
+    ///
+    /// The current unique setting.
+    ///
+    /// @return The current unique.
+    ///
+    /// @see https://docs.mongodb.org/manual/core/index-unique/
```
samantharitter: I'd move this link (or copy it) to the getter

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// The current sparse setting.
+    ///
+    /// @return The current sparse.
+    ///
+    const stdx::optional<bool>& sparse() const;
+
+    ///
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage
+    /// engine options for the index
```
samantharitter: nit, period at end of sentence

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+    /// @return The current name.
+    ///
+    const stdx::optional<std::string>& name() const;
+
+    ///
+    /// Whether or not to create a sparse index. Sparse indexes only reference
+    /// documents with the indexed fields.
+    ///
+    /// @param sparse
+    ///   Whether or not to create a sparse index.
```
samantharitter: Probably good to link the sparse index page, too: https://docs.mongodb.org/manual/core/index-sparse/

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,361 @@
...
+    /// @return The current storage_options.
+    ///
+    const base_storage_options* storage_options() const;
+
+    ///
+    /// Set a value, in seconds, as a TTL to control how long MongoDB retains
+    /// documents in this collection.
+    ///
+    /// @param expire_after_seconds
+    ///   The amount of time, in seconds, to retain documents.
```
samantharitter: Ditto for TTL indexes: https://docs.mongodb.org/manual/core/index-ttl/

Machyne: Thanks for the doc's comments :) Should have been more careful when putting links...

Fixed!

Machyne:
@samantharitter ping

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+/// Class representing the optional arguments to a MongoDB create index operation.
+///
+/// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+///
+class MONGOCXX_API index {
+   public:
+    ///
+    /// Base class representing the optional storage engine options for indexes.
+    ///
+    class base_storage_options {
```
samantharitter: This base class needs a virtual destructor
```
virtual ~base_storage_options();
```
so that cleanup of inheriting classes can always happen properly.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+        ///
+        const stdx::optional<std::string>& config_str() const;
+
+       private:
+        friend collection;
+        MONGOCXX_PRIVATE const int type() const override;
+        stdx::optional<std::string> _config_str;
+    };
+
+    index();
```
samantharitter: You can leave this out, the compiler will generate a default constructor for you.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+
+    ///
+    /// Whether or not to build the index in the background so that building
+    /// the index does not block other database activities.
```
samantharitter: Add that the default is to build indexes in the foreground

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    /// The current background setting.
+    ///
+    /// @return The current background.
+    ///
+    const stdx::optional<bool>& background() const;
+
+    ///
+    /// Whether or not to create a unique index so that the collection will not
+    /// accept insertion of documents where the index key or keys match an
+    /// existing value in the index.
```
samantharitter: an existing document in the collection?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///   Whether or not to create a sparse index.
+    ///
+    /// @see https://docs.mongodb.org/manual/core/index-sparse/
+    ///
+    void sparse(bool sparse);
+
+    ///
+    /// The current sparse setting.
+    ///
+    /// @return The current sparse.
```
samantharitter: sparse setting

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    const stdx::optional<bool>& sparse() const;
+
+    ///
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage
+    /// engine options for the index.
+    ///
+    /// @warning
+    ///  The base_storage_options* instance must stay in scope until after
```
samantharitter: the base_storage_options instance, without *

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///  the call to create_index.
```
samantharitter: "the call to create_index" isn't quite right, because the lifetime of this options object isn't strictly tied to a single create_index() call.  Better to say that the storage_options object must outlive this index::options object.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// @param storage_options
+    ///   The storage engine options for the index in.
```
samantharitter: for the index.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    /// @return The current weights.
+    ///
+    const stdx::optional<bsoncxx::document::view>& weights() const;
+
+    ///
+    /// For text indexes, the language that determines the list of stop words
+    /// and the rules for the stemmer and tokenizer.
+    ///
+    /// @param default_language
+    ///   The default language for text indexes.
```
samantharitter: which text indexes?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// @param language_override
+    ///   The name of the field that contains the override language for text indexes.
+    ///
+    void language_override(std::string language_override);
+
+    ///
+    /// The current language_override setting.
+    ///
+    /// @return The current language_override.
```
samantharitter: The name of the field that contains the override language for text indexes.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// The current partial_filter_expression setting.
+    ///
+    /// @return The current partial_filter_expression.
+    ///
+    const stdx::optional<bsoncxx::document::view>& partial_filter_expression() const;
+
+    ///
+    /// For 2dsphere indexes, the 2dsphere index version number. Version can be
+    /// either 1 or 2.
```
samantharitter: What happens if you try to set it to 3?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+
+    ///
+    /// The current twod_sphere_version setting.
+    ///
+    /// @return The current twod_sphere_version.
+    ///
+    const stdx::optional<std::uint8_t>& twod_sphere_version() const;
+
+    ///
+    /// For 2d indexes, the number of precision of the stored geohash value of
```
samantharitter: just "the precision of"

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    /// the location data.
+    ///
+    /// @param twod_bits_precision
+    ///   The number of precision of the stored geohash value.
```
samantharitter: ditto above

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// @param twod_location_min
+    ///   The lower inclusive boundary.
+    ///
+    void twod_location_min(double twod_location_min);
+
+    ///
+    /// The current twod_location_min setting.
+    ///
+    /// @return The current twod_location_min.
```
samantharitter: take your more descriptive comment from above, the current lower inclusive boundary for longitude and latitude values.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    /// @param twod_location_max
+    ///   The upper inclusive boundary.
+    ///
+    void twod_location_max(double twod_location_max);
+
+    ///
+    /// The current twod_location_max setting.
+    ///
+    /// @return The current twod_location_max.
```
samantharitter: ditto

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    ///
+    const stdx::optional<double>& twod_location_max() const;
+
+    ///
+    /// For geoHaystack indexes, specify the number of units within which to
+    /// group the location values; i.e. group in the same bucket those location
+    /// values that are within the specified number of units to each other.
+    ///
+    /// @param haystack_bucket_size
+    ///   The geoHaystack bucket size.
```
samantharitter: docs link?

samantharitter:
Lots of nit comments but overall looking pretty good!

src/mongocxx/options/index.hpp
Machyne: I did mean "value in the index" here. I thought that for unique indexes, the indexes would not insert values already in the index. Do you think the other way is more clear?

(also, sorry, this line will be / was changed by clang format)

src/mongocxx/options/index.hpp
Machyne: Good catch! Thanks!

src/mongocxx/options/index.hpp
Machyne: Just tried it in the mongo shell and it worked... I have no idea what it does O.O
```
> db.collection.createIndex( { tmp : "2dsphere" }, { "2dsphereIndexVersion": 3 } )
{
	"createdCollectionAutomatically" : true,
	"numIndexesBefore" : 1,
	"numIndexesAfter" : 2,
	"ok" : 1
}
> db.collection.getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "asdf.collection"
	},
	{
		"v" : 1,
		"key" : {
			"tmp" : "2dsphere"
		},
		"name" : "tmp_2dsphere",
		"ns" : "asdf.collection",
		"2dsphereIndexVersion" : 3
	}
]
> db.colleciton.insert({tmp: [90,90]})
WriteResult({ "nInserted" : 1 })
```

The docs say that we only have two versions:
>MongoDB 2.6 introduces a version 2 of 2dsphere indexes. Version 2 is the default version of 2dsphere indexes created in MongoDB 2.6 and later series. To override the default version 2 and create a version 1 index, include the option { "2dsphereIndexVersion": 1 } when creating the index.

Thoughts??

src/mongocxx/options/index.hpp
Machyne: I need this to initialize `_storage_options` to `nullptr`.

Machyne:
Nits addressed! pinging again.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+
+       private:
+        friend collection;
+        MONGOCXX_PRIVATE virtual const int type() const = 0;
+    };
+
+    ///
+    /// Class representing the optional WiredTiger storage engine options for indexes.
+    ///
+    class wiredtiger_storage_options : public base_storage_options {
```
acmorrow: class can be final?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+/// @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#options
+///
+class MONGOCXX_API index {
+   public:
+    ///
+    /// Base class representing the optional storage engine options for indexes.
+    ///
+    class base_storage_options {
+       public:
+        virtual ~base_storage_options() = default;
```
acmorrow: You should declare this here, but define it in the .cpp file. You can still define it with = default.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage engine options for
+    /// the index.
+    ///
+    /// @warning
+    ///  The base_storage_options instance must outlive this options::index object.
+    ///
+    /// @param storage_options
+    ///   The storage engine options for the index.
+    ///
+    void storage_options(base_storage_options* storage_options);
```
acmorrow: As an alternative to these lifetime rules, this could take ownership if you passed in a unique_ptr<base_storage_options>.

src/mongocxx/options/index.hpp
acmorrow: You probably want to keep it even if you don't need to initialize _storage_options in the future, because you want the constructor definition to be part of the ABI, so it should be defined at .cpp scope. You could declare it here, and then define it in the .cpp file with = default.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+        void config_str(std::string config_str);
+
+        ///
+        /// The current config_str setting.
+        ///
+        /// @return The current config_str.
+        ///
+        const stdx::optional<std::string>& config_str() const;
+
+        ~wiredtiger_storage_options() override;
```
acmorrow: Lifecycle methods should come first.

src/mongocxx/options/index.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void index::wiredtiger_storage_options::config_str(std::string config_str) {
+    _config_str = config_str;
+}
+
+const stdx::optional<std::string>& index::wiredtiger_storage_options::config_str() const {
+    return _config_str;
+}
+
+index::wiredtiger_storage_options::~wiredtiger_storage_options() = default;
```
acmorrow: Try to keep definitions in .cpp files in the same order as declarations in the header

src/mongocxx/options/index.cpp
```diff
@@ -0,0 +1,172 @@
...
+void index::background(bool background) {
+    _background = background;
+}
+
+void index::unique(bool unique) {
+    _unique = unique;
+}
+
+void index::name(std::string name) {
+    _name = name;
```
acmorrow: std::move(name)

src/mongocxx/options/index.cpp
```diff
@@ -0,0 +1,172 @@
...
+void index::v(std::int32_t v) {
+    _v = v;
+}
+
+void index::weights(bsoncxx::document::view weights) {
+    _weights = weights;
+}
+
+void index::default_language(std::string default_language) {
+    _default_language = default_language;
```
acmorrow: std::move, and elsewhere below as needed.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,365 @@
...
+    const stdx::optional<bsoncxx::document::view>& weights() const;
+
+    ///
+    /// For text indexes, the language that determines the list of stop words and the rules for the
+    /// stemmer and tokenizer.
+    ///
+    /// @param default_language
+    ///   The default language used when creating text indexes.
+    ///
+    void default_language(std::string default_language);
```
acmorrow: For other parts of this API, we take things as views. But here we make an owning copy. Then there is the question of the storage options.

I'd prefer that the API be more consistent about what is viewed and what is owned. If we opt for un-owned everywhere, then I'd suggest that these std::strings should be string_views, both as parameters and as members.

Even if we opt for an owning model for this object (what do we do in other helpers)? I'd suggest making these parameters string_view, as that is our vocabulary type for passing in stringy things.

src/mongocxx/collection.cpp
```diff
@@ -431,12 +431,85 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
 }
 
 bsoncxx::document::value collection::create_index(bsoncxx::document::view keys,
-                                                  bsoncxx::document::view options) {
+                                                  const options::index& options) {
```
acmorrow: So this change touches on two fairly fundamental issues that we still need to discuss. The nice thing about taking a view for the document options is that it is 100% flexible. If new index options come into being, the API doesn't need to change. That gives us great ABI stability, but a poor interface.

On the other hand, the new options::index struct gives us a very nice interface, but locks us into a strict ABI. 

And, this is not the only place we face this. Have a look at https://jira.mongodb.org/browse/CXX-435

I think there are a few ways forward that we could consider:

- For every place that we take an option type, also provide an overload that takes a document::view. Then, make every option type have a to_bson method. That way, you can use the helper to build up as much as you can, then make any customizations, and call the ABI safe method.
- Alternatively, make every option type have an "extensions" member, that takes a document_view and which it will fold in to the options set via the option type.

This has been in the back of my mind for a while, but it seems like a good time to start talking about it.

src/mongocxx/options/index.hpp
acmorrow: Note that I have the same ABI concerns about this "sub option" as I do about the overall options class. What is its extensibility story? How can we allow it to describe new options without an ABI break. Obviously, we can extend the class to handle new options when we next do change ABI, but it would be unfortunate if we didn't have a workaround for cases where ABI can't change.

src/mongocxx/collection.cpp
Machyne: With other options moving towards a `to_document` method, I think I'm more in favor of the first option. Having an "extensions" member seems a little messy, and like things could go wrong. I like the idea of "use our nice options class or send arbitrary bson and suffer possible consequences".

src/mongocxx/options/index.hpp
Machyne: As far as I can tell, all other options classes take these as `std::string`'s and not views. I think the members for options classes should all be owning. It's nice to be able do something like `opts.name("name")` and not have to do `auto name = "name"; opts.name(name);`.

src/mongocxx/options/index.hpp
acmorrow: +1. Needs to be declared here, but should be defined out of line as = default

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,363 @@
...
+       public:
+        ~wiredtiger_storage_options() override;
+
+        ///
+        /// Set the WiredTiger configuration string.
+        ///
+        /// @param config_str
+        ///   The WiredTiger configuration string.
+        ///
+        void config_str(std::string config_str);
```
acmorrow: avoid abbreviations. is the 'str' even necessary? Or name it config_string

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,363 @@
...
+    const stdx::optional<bool>& sparse() const;
+
+    ///
+    /// Optionally used only in MongoDB 3.0.0 and higher. Specifies the storage engine options for
+    /// the index.
+    ///
+    /// @param storage_options
+    ///   The storage engine options for the index.
+    ///
+    void storage_options(std::unique_ptr<base_storage_options>& storage_options);
```
acmorrow: These signatures look strange to me. A mutable reference to a unique_ptr? If ownership is being transferred, take the unique_ptr by value. If not, pass it by const reference.

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,363 @@
...
+    void storage_options(std::unique_ptr<wiredtiger_storage_options>& storage_options);
+
+    ///
+    /// The current storage_options setting.
+    ///
+    /// @return The current storage_options.
+    ///
+    const base_storage_options* storage_options() const;
```
acmorrow: What is the utility of this method? Since I've lost the type information about the concrete type, I can't safely act on it without dynamic cast. Perhaps we should not expose this?

src/mongocxx/options/index.hpp
```diff
@@ -0,0 +1,363 @@
...
+    ///
+    const stdx::optional<std::int32_t>& expire_after_seconds() const;
+
+    ///
+    /// Sets the index version.
+    ///
+    /// @param v
+    ///   The index version.
+    ///
+    void v(std::int32_t v);
```
acmorrow: I know this is idiomatically called 'v', but can we do better? This is a pretty terrible method name! Also, signed/unsigned for the parameter?

src/mongocxx/collection.cpp
```diff
@@ -431,12 +431,85 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
     bson_error_t error;
+    ::mongoc_index_opt_geo_t geo_opt{};
+    ::mongoc_index_opt_t opt{};
+    ::mongoc_index_opt_wt_t wt_opt{};
+    libmongoc::index_opt_init(&opt);
+
+    if (options.background()) {
+        opt.background = *options.background();
+    }
+    if (options.unique()) {
```
acmorrow: blank line separating the close of each if statement } from the start of the next if statemement, please.

src/mongocxx/collection.cpp
```diff
@@ -431,12 +431,85 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    if (options.partial_filter_expression()) {
+        opt.partial_filter_expression = partial_filter_expression.bson();
+    }
+
+    if (options.twod_sphere_version() || options.twod_bits_precision() ||
+        options.twod_location_min() || options.twod_location_max() ||
+        options.haystack_bucket_size()) {
+        libmongoc::index_opt_geo_init(&geo_opt);
+
+        if (options.twod_sphere_version()) {
```
acmorrow: ditto

src/mongocxx/collection.cpp
```diff
@@ -431,12 +431,85 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
 }
 
 bsoncxx::document::value collection::create_index(bsoncxx::document::view keys,
-                                                  bsoncxx::document::view options) {
+                                                  const options::index& options) {
     scoped_bson_t bson_keys{keys};
+    scoped_bson_t weights{options.weights()};
+    scoped_bson_t partial_filter_expression{options.partial_filter_expression()};
```
acmorrow: can we move these declarations to point of use?

src/mongocxx/collection.cpp
```diff
@@ -431,12 +431,85 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    }
+    if (options.name()) {
+        opt.name = options.name()->c_str();
+    }
+    if (options.sparse()) {
+        opt.sparse = *options.sparse();
+    }
+    if (options.storage_options()) {
+        const options::index::wiredtiger_storage_options* wt_options;
+        libmongoc::index_opt_wt_init(&wt_opt);
```
acmorrow: is there a cleanup function associated with opt_wt_init, or is it simply an initializer, not an allocation?

Machyne: Just an initializer. It's a memcpy from the static default struct.

acmorrow:
LGTM

PR #326, opened 2015-10-15
minor: update evergreen / travis to use mongoc master

Machyne:


acmorrow:
LGTM. Please commit (note that we don't do merge commits in this repo, so you will need to merge to master and then push, rather than use the big green button).

amidvidy:
LGTM

PR #329, opened 2015-10-16
CXX-707 don't create v0 indexes against 3.1+ servers

amidvidy:


amidvidy:
@markbenvenuto PTAL

markbenvenuto:
:+1:

PR #331, opened 2015-10-16
CXX-704: add enum for special write concern w values

Machyne:


src/mongocxx/write_concern.hpp
```diff
@@ -54,6 +54,15 @@ class bulk_write;
 class MONGOCXX_API write_concern {
 
    public:
+    ///
+    /// A class to represent the special case values for write_concern::nodes.
+    /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
+    ///
+    enum w : std::int32_t {
```
amidvidy: `enum class w`

http://en.cppreference.com/w/cpp/language/enum#Scoped_enumerations

src/mongocxx/write_concern.hpp
```diff
@@ -54,6 +54,15 @@ class bulk_write;
...
+        UNACKNOWLEDGED = 0,
```
amidvidy: our naming scheme for enum values is to prefix them with `k` and use camelCase.

e.g. kUnacknowledged, kDefault, kMajority...

src/mongocxx/write_concern.hpp
Machyne: I didn't do this initially because then to compare `write_concern::w` to the return of `write_concern::nodes()` requires a cast. Is that okay?

amidvidy: yes that is a fair cast

src/mongocxx/test/write_concern.cpp
```diff
@@ -35,8 +35,7 @@ TEST_CASE("a default write_concern", "[write_concern]") {
     }
 
     SECTION("will require confirmation from just the primary or standalone mongod") {
-        // TODO: fix me
-        REQUIRE(-2 == wc.nodes());
+        REQUIRE(static_cast<std::int32_t>(write_concern::w::kDefault) == wc.nodes());
```
amidvidy: if you want to be fancy here... you can use std::underlying_type instead of explicitly saying std::int32_t

src/mongocxx/write_concern.hpp
```diff
@@ -54,6 +54,15 @@ class bulk_write;
 class MONGOCXX_API write_concern {
 
    public:
+    ///
+    /// A class to represent the special case values for write_concern::nodes.
+    /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
+    ///
+    enum class w : std::int32_t {
+        kUnacknowledged = 0,
```
amidvidy: where do these values come from?

amidvidy:
it seems like there should be a setter/getter on write_concern that takes a type of this enum

src/mongocxx/write_concern.hpp
Machyne: https://github.com/mongodb/mongo-c-driver/blob/master/src/mongoc/mongoc-write-concern.h#L30

Machyne:
(setter / getter added)

src/mongocxx/write_concern.hpp
```diff
@@ -183,7 +204,14 @@ class MONGOCXX_API write_concern {
     ///
     /// @return The number of required nodes.
     ///
-    std::int32_t nodes() const;
+    stdx::optional<std::int32_t> nodes() const;
```
samantharitter: For this method, would an unset optional mean that no write concern level had been set, or that no nodes are required?

Machyne: This is optional because the special kDefault, kMajority, and kUnacknowledged live in the same C int32 as the nodes, so only one of `nodes()` or `acknowledge_level()` will ever be set.

src/mongocxx/write_concern.hpp
```diff
@@ -180,10 +201,25 @@ class MONGOCXX_API write_concern {
 
     ///
     /// Gets the current number of nodes that this write_concern requires operations to reach.
+    /// This value will be unset iff the acknowled_level is set instead.
```
samantharitter: acknowledge_level

src/mongocxx/write_concern.hpp
samantharitter: Ok, got it.  Is it necessarily true that either nodes or acknowledge_level will be set?  Might they both be unset?

Machyne: Yes, one of them always has to be set. The default is ack level of `kDefault`.

samantharitter:
LGTM!

src/mongocxx/write_concern.cpp
```diff
@@ -84,8 +90,15 @@ bool write_concern::journal() const {
...
-std::int32_t write_concern::nodes() const {
-    return libmongoc::write_concern_get_w(_impl->write_concern_t);
+stdx::optional<std::int32_t> write_concern::nodes() const {
+    std::int32_t w = libmongoc::write_concern_get_w(_impl->write_concern_t);
+    return w >= 1 ? stdx::optional<std::int32_t>{w} : stdx::nullopt;
+}
+
+stdx::optional<write_concern::level> write_concern::acknowledge_level() const {
+    std::int32_t w = libmongoc::write_concern_get_w(_impl->write_concern_t);
+    return w < 1 ? stdx::optional<write_concern::level>{static_cast<write_concern::level>(w)}
```
acmorrow: There is a small problem here. What if the value that comes back from the C driver is something that isn't in range for the enumeration? e.g. you upgrade your libmongoc.so to something newer, and we have some new special write concern value, but the libmongocxx doesn't know that value. Right now you would get a 'write_concern::level' enum value back that wasn't actually in range.

src/mongocxx/write_concern.hpp
```diff
@@ -55,6 +55,15 @@ class bulk_write;
 class MONGOCXX_API write_concern {
 
    public:
+    ///
+    /// A class to represent the special case values for write_concern::nodes.
+    /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
+    ///
+    enum class level : std::int32_t {
```
acmorrow: Do we actually want to expose the numeric values here? This is sort of leaking an implementation detail. It might be better to make these symbolic and map. Also, per my other comment, you might need to represent kUnknown.

src/mongocxx/write_concern.hpp
```diff
@@ -125,7 +134,7 @@ class MONGOCXX_API write_concern {
     /// @param confirm_from
     ///   The number of replica set nodes that must acknowledge the write.
     ///
-    /// @warning Setting the number of nodes to 0 disables write acknowledgement and all other
+    /// @warning Setting the number of nodes to 0 disables write acknowledgment and all other
```
acmorrow: Thanks :)

src/mongocxx/test/write_concern.cpp
```diff
@@ -116,7 +121,7 @@ TEST_CASE("confirmation from tags, a repl-member count, and majority are mutuall
         write_concern wc{};
         wc.nodes(10);
         wc.majority(std::chrono::milliseconds(100));
-        REQUIRE(-3 == wc.nodes());
+        REQUIRE(write_concern::level::kMajority == *wc.acknowledge_level());
     }
 
     SECTION("setting majority unsets the tag") {
```
acmorrow: Consider adding a test that nodes xor ack_level

src/mongocxx/write_concern.hpp
```diff
@@ -55,6 +55,15 @@ class bulk_write;
 class MONGOCXX_API write_concern {
 
    public:
+    ///
+    /// A class to represent the special case values for write_concern::nodes.
+    /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
+    ///
+    enum class level : std::int32_t {
+        kUnacknowledged = 0,
+        kDefault = -2,
```
amidvidy: i think I was wrong about our enum style before. While the server uses e.g. 'kDefault', the driver should use 'k_default',

acmorrow: Yes, that is true.

PR #333, opened 2015-10-16
CXX-709: Fix reinterpret_cast for VS 2015

markbenvenuto:
I have verified this on all support platforms & compilers in Evergreen including the 32-bit builders.


amidvidy:
LGTM

PR #334, opened 2015-10-16
Minor padding fix json $regex output

yhagerjs:

PR #336, opened 2015-10-19
CXX-693: allow findAndModifys to take write concern

Machyne:


Machyne:
_the comment from 6030ce6 should have refferenced #334, not this_

src/mongocxx/collection.hpp
```diff
@@ -350,6 +350,10 @@ class MONGOCXX_API collection {
     /// @return The original or replaced document.
     /// @throws exception::write if the operation fails.
     ///
+    /// @note
```
acmorrow: Err... that seems sort of contrary to the spirit of the ticket, which I think was to actually support it?

samantharitter:
LGTM

src/mongocxx/collection.hpp
Machyne: Right now, this is the only thing possible with the C driver.
We had a bit of discussion [on CDRIVER-945](https://jira.mongodb.org/browse/CDRIVER-945) about this.

acmorrow: I see. I don't think that C driver can reasonably decline both this request and the "copy a collection/database" request. Might want to point that out to them.

acmorrow:
I'm fine with committing the comments, but I don't want to resolve the CXX ticket, since we actually still can't do that. Lets add the comments, but leave the ticket open, and move it back to 'open' state, and mark it as waiting on dependency. Leave 0.3.0 as the fix version, and we can address it later, either way.

acmorrow:
LGTM for comment changes.

PR #338, opened 2015-10-19
CXX-692: findAndModify take a write concern

markbenvenuto:
Add a `writeConcern` argument to `findAndModify`.

I needed to move `findAndModify` to `DBClientBase` since this is the class that had `writeConcern` has the connection-level default `writeConcern`.

### DIFF
--- a/src/mongo/client/dbclient.cpp
+++ b/src/mongo/client/dbclient.cpp
@@ -1336,14 +1336,15 @@ BSONObj DBClientWithCommands::distinct(const StringData& ns,
     return result.getField("values").Obj().getOwned();
 }
 
-void DBClientWithCommands::_findAndModify(const StringData& ns,
-                                          const BSONObj& query,
-                                          const BSONObj& update,
-                                          const BSONObj& sort,
-                                          bool returnNew,
-                                          bool upsert,
-                                          const BSONObj& fields,
-                                          BSONObjBuilder* out) {
+void DBClientBase::_findAndModify(const StringData& ns,
+                                  const BSONObj& query,
+                                  const BSONObj& update,
+                                  const BSONObj& sort,
+                                  bool returnNew,
+                                  bool upsert,
+                                  const BSONObj& fields,
+                                  const WriteConcern* writeConcern,
+                                  BSONObjBuilder* out) {
     BSONObjBuilder commandBuilder;
 
     commandBuilder.append("findAndModify", nsGetCollection(ns.toString()));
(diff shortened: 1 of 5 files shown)


src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -240,6 +240,23 @@ TEST_F(DBClientTest, FindAndModifyDuplicateKeyError) {
...
+                                      BSON("$inc" << BSON("i" << 1)),
+                                      false,
+                                      false,
+                                      BSONObj(),
+                                      BSONObj(),
+                                      &WriteConcern::journaled);
+
+    ASSERT_EQUALS(result.getIntField("_id"), 1);
+    ASSERT_EQUALS(result.getIntField("i"), 1);
+    ASSERT_EQUALS(c->count(TEST_NS), 1U);
```
amidvidy: is it worth doing the whole song & dance with the system profiler to check that the system actually registered the proper write concern?

src/mongo/client/dbclient.cpp
```diff
@@ -1365,6 +1366,13 @@ void DBClientWithCommands::_findAndModify(const StringData& ns,
     commandBuilder.append("new", returnNew);
     commandBuilder.append("upsert", upsert);
 
+    if (getMaxWireVersion() >= 4) {
+        const WriteConcern* operationWriteConcern =
+            writeConcern ? writeConcern : &getWriteConcern();
```
amidvidy: will &getWriteConcern() always be non-null?

src/mongo/client/dbclient.cpp
```diff
@@ -1365,6 +1366,13 @@ void DBClientWithCommands::_findAndModify(const StringData& ns,
     commandBuilder.append("new", returnNew);
     commandBuilder.append("upsert", upsert);
 
+    if (getMaxWireVersion() >= 4) {
```
amidvidy: if the maxWireVersion is too low, should we uassert that the operation is not supported on the remote server?

src/mongo/client/dbclient.cpp
markbenvenuto: Yes. This code pattern is also used throughout the driver.

src/mongo/client/dbclient.cpp
markbenvenuto: It is not required as part of this change, and I am not sure if there is a relevant wire version to check.

src/mongo/integration/standalone/dbclient_test.cpp
markbenvenuto: I do not think it is worth the effort. We do not do in any other tests, but I have checked the `system.profile` collection by hand.

amidvidy:
LGTM

src/mongo/client/dbclient.cpp
acmorrow: I feel like ignoring a user requested writeConcern is dangerous. I think we should take the suggestion from @amidvidy

src/mongo/integration/standalone/dbclient_test.cpp
acmorrow: I agree, not worth the effort.

acmorrow:
LGTM mod uassert per suggestion. No need to re-post.

markbenvenuto:
Merged with https://github.com/mongodb/mongo-cxx-driver/commit/76be500a206bf0d964ae6db81378b22476d9bce0

PR #339, opened 2015-10-19
CXX-712 Add ordered option to insert_many

Machyne:


src/mongocxx/options/insert.hpp
```diff
@@ -42,6 +43,20 @@ class MONGOCXX_API insert {
     void write_concern(class write_concern wc);
 
     ///
+    /// @note: This applies only to insert_many and is ignored for insert_one.
+    ///
+    /// If true, when an insert fails, return without performing the remaining
+    /// writes. If false, when a write fails, continue with the remaining
+    /// writes, if any. Defaults to true.
```
samantharitter: I'd add that in addition to the failure semantics, inserts could be performed in any order if "ordered" is false.

src/mongocxx/options/insert.hpp
```diff
@@ -50,8 +65,18 @@ class MONGOCXX_API insert {
...
+    ///
+    /// @return The current ordered value.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/method/db.collection.insert/
+    ///
+    const bool ordered() const;
+
    private:
     bsoncxx::stdx::optional<class write_concern> _write_concern;
+    bool _ordered;
```
samantharitter: I think we should make this an optional and put the defaulting logic at a higher level, for example on line 627 of collection.hpp.

src/mongocxx/collection.hpp
```diff
@@ -624,7 +624,7 @@ MONGOCXX_INLINE bsoncxx::stdx::optional<result::insert_many> collection::insert_
     document_view_iterator_type end,
     const options::insert& options
 ) {
-    class bulk_write writes(false);
+    class bulk_write writes(options.ordered().value_or(true));
```
acmorrow: Is the change in default behavior here intentional? If the user doesn't specify, we want to treat their updates as ordered? That seems reasonable. Is there a spec that mandates one choice or the other?

acmorrow:
@Machyne LGTM mod one small question.

src/mongocxx/collection.hpp
Machyne: That was intentional! The crud api specifies bulk ops defaulting to ordered
[(specified in BulkWriteOptions)](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst)

acmorrow: Thanks!

samantharitter:
LGTM

PR #340, opened 2015-10-19
CXX-666 Backport server r3.0.6..r3.0.7 changes to the legacy driver

acmorrow:
Server SHAs

447aa61
946c96a

See CXX-666 for details on why these commits.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -562,7 +562,8 @@ if has_option('runtime-library-search-path'):
     for lib in libs:
         env.AppendENVPath(envVar, env.Dir(lib), delete_existing=0)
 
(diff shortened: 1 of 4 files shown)


markbenvenuto:
There are three commits in this pull request. The change to the SSL manager is not listed in your pull request description or in the bug as non-strikethrough.

acmorrow:
@markbenvenuto Thanks - good catch. That came from a commit that I picked in, but then got reverted. Apparently I committed the pick but not the revert and didn't squash it out.

acmorrow:
@markbenvenuto updated, PTAL

markbenvenuto:
:+1:

PR #341, opened 2015-10-19
CXX-608 Allow default construction of client, collection, database

acmorrow:
Also a drive by fixup of some missing MONGOCXX_PRIVATE tags on private constructors of these types.

### DIFF
--- a/src/mongocxx/client.hpp
+++ b/src/mongocxx/client.hpp
@@ -53,6 +53,13 @@ class MONGOCXX_API client {
    public:
 
     ///
+    /// Default constructs a new client. The client is not connected and is equivalent to the
+    /// state of a moved-from client. The only valid actions to take with a default constructed
+    /// 'client' are to assign to it, or destroy it.
+    ///
+    client() noexcept;
+
+    ///
     /// Creates a new client connection to MongoDB.
     ///
     /// @param mongodb_uri
@@ -61,7 +68,7 @@ class MONGOCXX_API client {
     ///   Additional options that cannot be specified via the mongodb_uri
     ///
     client(
-        const class uri& mongodb_uri = mongocxx::uri(),
+        const class uri& mongodb_uri,
         const options::client& options = options::client()
     );
 
@@ -79,6 +86,12 @@ class MONGOCXX_API client {
     /// Destroys a client.
     ///
     ~client();
+
+    ///
+    /// Returns true if the client is valid, meaning it was not default constructed
+    /// or moved from.
+    ///
+    explicit operator bool() const noexcept;
 
     ///
     /// Gets a handle to the underlying implementation.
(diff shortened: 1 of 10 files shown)


acmorrow:
I should probably add some tests for these...

src/mongocxx/client.hpp
```diff
@@ -61,7 +68,7 @@ class MONGOCXX_API client {
     ///   Additional options that cannot be specified via the mongodb_uri
     ///
     client(
-        const class uri& mongodb_uri = mongocxx::uri(),
+        const class uri& mongodb_uri,
```
Machyne: I understand the logic behind Databases and Collections having default constructors, but could you explain why this is better for client? It seems weird to make people say `client mongo_client{uri{}};` instead of `client mongo_client{};`.

acmorrow: Yeah, this one could go either way. I think that outside of examples and tests though, almost all real uses of the driver will pass a non-default constructed uri argument, since they won't actually want to connect to localhost. So, maybe it won't be that strange in practice? Another option would be to eliminate the default constructor for URI and require providing a string? That way it wouldn't look quite so odd. Open to suggestions on this one.

Machyne: My preference would be to keep the default constructor for URI. I think you have a good point about non-default uri's being used almost everywhere, so I think this makes sense.

Machyne:
LGTM! :+1:

src/mongocxx/client.hpp
samantharitter: I don't feel strongly either way, but I think it's all right to force the caller to be deliberate about creating an empty URI.  As you mentioned it isn't normal behavior to do this, so it's fine to discourage doing it with the extra typing.

src/mongocxx/collection.hpp
```diff
@@ -75,6 +75,8 @@ class MONGOCXX_API collection {
 
    public:
 
+    collection() noexcept;
```
samantharitter: I would add a comment here like the one for the default client constructor above.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -30,6 +30,11 @@
 using namespace mongocxx;
 using namespace bsoncxx;
 
+TEST_CASE("A default constructed collection is false-ish", "[collection]") {
```
samantharitter: Add a test for the truth-y case?

src/mongocxx/test/database.cpp
```diff
@@ -22,11 +22,16 @@
 
 using namespace mongocxx;
 
+TEST_CASE("A default constructed database is false-ish", "[database]") {
```
samantharitter: ditto test for the truth-y case

samantharitter: Oh actually I think the REQUIRE on line 45 covers this.

src/mongocxx/test/collection_mocked.cpp
samantharitter: Also covered by the REQUIRE below, never mind!

samantharitter:
LGTM mod adding a comment for the default collection constructor.

PR #342, opened 2015-10-20
CXX-640: actually set options in client constructor

Machyne:


acmorrow:
@Machyne Do you think it is worth adding a test that these round trip, or does that seem like more work than is warranted? The amount of mockery involved might be silly for something so simple.

Otherwise LGTM.

Actually, one more question: Can you please verify that the set of SSL options described in our SSL options header covers all C driver SSL options as of C driver 1.2?

Also, can you look into whether there are other C driver client options that we should be exposing? Keep in mind that we don't want to let the presence of the C driver leak out of our abstractions, but if they have new/important client options we probably need to at least consider how to present that.

You don't need to change this PR, but if you find missing options groups, please file tickets for them so we can evaluate.

Machyne:
I think the mock may be to complex to be worth this small test. If we do more with client in the future and add tests for that, then this would be a good addition, but as it is, I think we should just wait until we can figure out SSL testing on Evergreen.

The one option missing was the CRL file. I've added a ticket for it.

PR #343, opened 2015-10-20
CXX-634 new home for return_document enum

Machyne:


src/mongocxx/options/find_one_and_update.hpp
```diff
@@ -27,6 +26,17 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
 ///
+/// Argument for find_one_and_update and find_one_and_replace to determine
+/// whether to return the document before or after modification.
+///
+/// @see https://docs.mongodb.org/manual/reference/method/db.collection.findAndModify/#return-data
+///
+enum class return_document {
```
acmorrow: So, why is the right home for it?

samantharitter: Thinking this should be in a separate place.  I don't know if the current file scheme allows for this, common things between options.

The Ruby driver had this concept of composable classes through mix-ins, which map loosely to multiple inheritance.  In that model we'd have some sort of "modify_return_order" option that both find_one_and_replace and find_one_and_update would include.  This might be a helpful model to handle other commonalities between options (for example implementing a "hintable" option class, and having all commands that take hints have "hintable" option classes)

Machyne: @samantharitter - is your suggestion to leave as?

I agreed with the original comment (thinking that it needed a new home). My opinion was that it belonged either in this file or in `find_one_and_replace.hpp`. As long as it's `mongocxx::options::return_document` and not `mongocxx::options::find_one_and_update::return_document`, this isn't something that a user would need to directly include anyway (you get all of this with collection.hpp), so it seemed okay to me.

samantharitter: I think it would be better if this lived in its own file, logically it's weird that either find_one_and_* class would have to rely on the other's header.

acmorrow: I think what @samantharitter is suggesting is closer to the original intent of the (admittedly not very illuminating) "where is my home" comment.

Machyne: Hmm... so should we just delete the comment, and call it good? I'm fine with that! It does seem weird for either of these classes to rely on the other's header.

Machyne: ping

samantharitter: No, if we leave it then we're still having one class rely on the other's header.  I opt for moving this to a new header for common find_one_* options.  It will admittedly be a very short header file but it seems like a cleaner solution.

Machyne: Sorry, I meant leave it how it is pre-changes (i.e. on master currently) and delete the comment.

samantharitter: Got it.  If we leave it in the fine_one_and_modify header without the comment though, we'll probably end up not fixing it ever, I think we should just do something else with it now.

samantharitter: I misunderstood, I didn't realize that find_one_and_modify had no matching operation, or .cpp file.  It is actually a common header for the other find_one_and* operation options.  I think we should do as you said and just remove the comment.  I also think we should rename this file to something clearer, maybe `find_one_common_options.hpp`

Machyne:
Undo moving enum into one of the headers.
Confusing file name is gone, along with the comment.
Has `return_document` finally found its home?(!?)

samantharitter:
LGTM

acmorrow:
LGTM. The little lost enum has found a forever home!

PR #344, opened 2015-10-20
CXX-681: fix bulk_write results to return int_32t values

Machyne:
The server and C driver both return int_32's for these commands, the bsoncxx `get_int64()` was causing and error.

### DIFF
--- a/src/mongocxx/result/update.hpp
+++ b/src/mongocxx/result/update.hpp
@@ -45,14 +45,13 @@ class MONGOCXX_API update {
     ///
     /// @return The number of documents that were matched.
     ///
-    std::int64_t matched_count() const;
+    std::int32_t matched_count() const;
 
     ///
     /// Gets the number of documents that were modified during this operation.
     ///
     /// @return The number of documents that were modified.
-    ///    std::int64_t matched_count() const;
-    std::int64_t modified_count() const;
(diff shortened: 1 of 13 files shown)


src/mongocxx/result/update.hpp
```diff
@@ -45,14 +45,14 @@ class MONGOCXX_API update {
...
-    std::int64_t matched_count() const;
+    std::int32_t matched_count() const;
 
     ///
     /// Gets the number of documents that were modified during this operation.
     ///
     /// @return The number of documents that were modified.
-    ///    std::int64_t matched_count() const;
-    std::int64_t modified_count() const;
+    ///    std::int32_t matched_count() const;
```
samantharitter: I think just remove this line from the comment

samantharitter:
LGTM, though this brings up a good question about our tests.  In general we prefer mocked tests to integration tests, but with those we can't catch bugs like these, where the result from the server is something different from what we expect.  It's a matter of testing our mocks I suppose.

src/mongocxx/result/update.hpp
acmorrow: It just looks like a copy-paste error. I agree

acmorrow:
LGTM as well. @samantharitter agree with you about the mocks.

PR #345, opened 2015-10-20
CXX-635: fix visibility of insert_many constructor

Machyne:
(contrary to the comment, this isn't / wasn't used in our tests...)

### DIFF
--- a/src/mongocxx/result/insert_many.hpp
+++ b/src/mongocxx/result/insert_many.hpp
@@ -24,6 +24,9 @@
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+class collection;
+
 namespace result {
 
 ///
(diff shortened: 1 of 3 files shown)


samantharitter:
LGTM!

src/mongocxx/result/insert_many.hpp
```diff
@@ -59,11 +57,15 @@ class MONGOCXX_API insert_many {
     ///
     /// @return The values of the _id field for inserted documents.
     ///
-    std::map<std::size_t, bsoncxx::document::element> inserted_ids();
+    id_map inserted_ids();
 
    private:
+    friend collection;
+
+    insert_many(result::bulk_write result, id_map inserted_ids);
```
acmorrow: I believe this is addressed in your other review, but this method when made C++ private should also be made ABI private with a MONGOCXX_PRIVATE annotation

Machyne: Yes! Thanks for pointing that out :)
I made sure to fix this in my other pull request

PR #346, opened 2015-10-21
CXX-633 Implement hints

samantharitter:
Ideas on how to test this for count?  The count method on collection just returns a number.

Edit: discovered the collection mocking framework!  Will update with tests, and fix Travis.


src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+    ///
+    explicit hint(bsoncxx::document::view index);
+
+    ///
+    /// Constructs a new hint.
+    ///
+    /// @param index
+    ///   String representing the name of the index to be used.
+    ///
+    explicit hint(std::string index);
```
amidvidy: suggest taking stdx::string_view instead

src/mongocxx/collection.cpp
```diff
@@ -414,6 +414,16 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
         rp_ptr = options.read_preference()->_impl->read_preference_t;
     }
 
+    if (options.hint()) {
+        auto filter_with_hint = bsoncxx::builder::stream::document{}
+                                << options.hint().value()
```
Machyne: I've been doing `*options.hint()` instead of `options.hint().value()`. Not sure if there's a difference, but if there isn't, I think the `*options.hint()` option is more common.

amidvidy:
you can write an integration test that uses the query profiler to test that the hint was part of the query. This is usually how other drivers test this

src/bsoncxx/document/view.cpp
```diff
@@ -162,7 +162,9 @@ view::iterator view::find(stdx::string_view key) const {
     bson_t b;
     bson_iter_t iter;
 
-    bson_init_static(&b, _data, _length);
+    if (!bson_init_static(&b, _data, _length)) {
```
acmorrow: When might this happen?

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/builder/stream/helpers.hpp>
+#include <bsoncxx/builder/stream/key_context.hpp>
+#include <bsoncxx/stdx/optional.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+using namespace bsoncxx::builder::stream;
```
acmorrow: Can't do this in a header

src/mongocxx/hint.hpp
acmorrow: +1

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+    ///   String representing the name of the index to be used.
+    ///
+    explicit hint(std::string index);
+
+    template <typename T>
+    friend key_context<T> operator<<(key_context<T> builder, const hint& obj);
+
+   private:
+    bsoncxx::stdx::optional<bsoncxx::document::view> index_doc;
+    bsoncxx::stdx::optional<std::string> index_string;
```
acmorrow: The lack of a variant type in the C++ standard library is a serious frustration.

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+    /// @param index
+    ///   String representing the name of the index to be used.
+    ///
+    explicit hint(std::string index);
+
+    template <typename T>
+    friend key_context<T> operator<<(key_context<T> builder, const hint& obj);
+
+   private:
+    bsoncxx::stdx::optional<bsoncxx::document::view> index_doc;
```
acmorrow: Have we been standardizing on the unqualified stdx::optional achievable via mongocxx/stdx.h? I think @Machyne has been making that change systemically. If so, please do so here as well.

acmorrow: Actually @Machyne, would you mind filing a CXX enhancement ticket to consistently use stdx names in the mongocxx directory via the mongocxx/stdx header and namespace, rather than bsoncxx?

acmorrow: Should index_doc be a view of the BSON, or a copied value?

Machyne: (ticket filed: CXX-716)

acmorrow:
What other operations should we be allowing hinting on, other than count? Do we need tickets for those?

Machyne:
The crud api only specifies hinting for count, but it seems to me like we might want it for find options as well.

acmorrow:
@Machyne That seems like an oversight? Can you check in with the CRUD spec authors about that?

acmorrow:
@samantharitter Could you move the relevant ticket (I think the hint/count) one to 'in code review' and add a link to this PR?

Machyne:
Checking up on that now. ([SPEC-367](https://jira.mongodb.org/browse/SPEC-367))

samantharitter:
There is an open ticket to add this for find, just wanted to get things clear with count before doing that: https://jira.mongodb.org/browse/CXX-691

Machyne:
Ah, the answer is that hinting can be done in the modifier document in find:
https://docs.mongodb.org/manual/reference/operator/query-modifier/

src/mongocxx/hint.cpp
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+#include <mongocxx/hint.hpp>
+
+#include <bsoncxx/stdx/make_unique.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+hint::hint(bsoncxx::document::view index) : index_doc(index) {
```
amidvidy: std::move(index)

src/mongocxx/hint.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+hint::hint(std::string index) : index_string(index) {
```
amidvidy: std::move, also string_view

src/mongocxx/test/hint.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <bsoncxx/json.hpp>
+#include <mongocxx/hint.hpp>
+
+using namespace mongocxx;
+using namespace bsoncxx;
+
+TEST_CASE("Hint", "[hint]") {
+    SECTION("Can be constructed with index name") {
+        std::string index_name = "a";
+        hint index_hint = hint(index_name);
```
amidvidy: no need for assignment, just do hint index_hint{index_name}

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+/// Append this hint to the given stream builder. Intended for use with bsoncxx's
+/// stream builder.
+///
+template <typename T>
+key_context<T> operator<<(key_context<T> builder, const hint& obj) {
+    if (obj.index_doc) {
+        auto hint_doc = concatenate{obj.index_doc.value()};
+        builder << "$hint" << open_document << hint_doc << close_document;
+    } else {
+        builder << "$hint" << obj.index_string.value();
```
amidvidy: prefer * to .value()

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,76 @@
...
+    ///
+    /// Constructs a new hint.
+    ///
+    /// @param index
+    ///   String representing the name of the index to be used.
+    ///
+    explicit hint(std::string index);
+
+    template <typename T>
+    friend key_context<T> operator<<(key_context<T> builder, const hint& obj);
```
amidvidy: instead of this method, could we have an operator document::element() ?

src/bsoncxx/document/view.cpp
samantharitter: Only when there's an issue with the length of the document or the _data ptr.  I ran into this as a problem when I was testing, and it might mean that there is a different bug in this class, or that my tests were doing something really wrong.  Maybe returning end() isn't the best way to handle this but we should still probably check this return value.

src/mongocxx/hint.hpp
samantharitter: Not sure about view vs. value?

acmorrow: A view type doesn't own the data it points to. So, your constructor argument doesn't own its data, nor does the member 'index_doc' (PS: should be _index_doc). So, if the document that you are viewing in the constructor is destroyed, the view in index_doc now points to freed memory. If you make it a value, it will make a copy.

It really depends on how you want hint to work, but I suspect that what you want here is a value.

src/mongocxx/hint.hpp
samantharitter: Why?

amidvidy: I think the API is unnecessarily coupled to the internals of builder::stream.

samantharitter:
Updated, PTAL

src/mongocxx/hint.hpp
amidvidy: Ping

src/mongocxx/options/count.hpp
```diff
@@ -21,36 +21,33 @@
...
     ///
-    void hint(bsoncxx::document::view hint);
+    void hint(class hint index_hint);
 
     ///
-    /// Gets the current index being hinted.
+    /// Gets the current hint.
     ///
-    /// @return The current hint.
+    /// @return The current hint, if one is set
```
Machyne: leave the period in for consistency

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -159,6 +169,20 @@ TEST_CASE("Collection", "[collection]") {
...
 
+        SECTION("Succeeds with hint") {
+            options::count opts;
+            hint index_hint("a_1");
+            opts.hint(index_hint);
+
+            // set our expected_opts so we check against that
+            bsoncxx::document::value doc = bsoncxx::builder::stream::document{}
+                                           << index_hint << bsoncxx::builder::stream::finalize;
+            libbson::scoped_bson_t cmd_opts{doc.view()};
```
Machyne: I think just `libbson::scoped_bson_t cmd_opts{doc};` works.

Machyne:
LGTM barring two small line comments :+1:!

samantharitter:
Updated to fix code review comments from last round, also ditched the operator<< in favor of a to_document() method at @amidvidy 's suggestion

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+    friend bool operator==(const hint& index_hint, std::string index);
+    friend bool operator==(const hint& index_hint, bsoncxx::document::view index);
+
+    ///
+    /// Return a bson document representing this hint.
+    ///
+    /// @return Hint, as a document.
+    ///
+    bsoncxx::document::value to_document();
```
amidvidy: similar to my comment in the create_collection review, it would be nice to add a companion conversion operator that called to_document.

amidvidy: e.g. 
```
MONGOCXX_INLINE operator bsoncxx::document::value() const;
```

Machyne: +!

samantharitter:
ping

src/mongocxx/collection.cpp
```diff
@@ -403,9 +403,21 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
         rp_ptr = options.read_preference()->_impl->read_preference_t;
     }
 
-    auto result = libmongoc::collection_count(
+    // Some options must be added via the options struct
+    bsoncxx::builder::stream::document cmd_opts_builder{};
+
+    if (options.hint()) {
+        auto hint_optional = options.hint();
+        cmd_opts_builder << bsoncxx::builder::stream::concatenate{hint_optional->to_document()};
```
Machyne: would this be better as
`cmd_opts_builder << bsoncxx::builder::stream::concatenate{options.hint()->to_document()};`
then you wouldn't need the `auto` variable

src/mongocxx/collection.cpp
```diff
@@ -403,9 +403,21 @@ std::int64_t collection::count(bsoncxx::document::view filter, const options::co
...
+    }
+
+    bsoncxx::document::value cmd_opts = cmd_opts_builder.extract();
+    scoped_bson_t cmd_opts_bson{cmd_opts};
```
Machyne: Same as above here?

src/mongocxx/hint.hpp
```diff
@@ -0,0 +1,92 @@
...
+///
+class MONGOCXX_API hint {
+   public:
+    ///
+    /// Constructs a new hint.
+    ///
+    /// @param index
+    ///   Document representing the index to be used.
+    ///
+    explicit hint(bsoncxx::document::value index);
```
amidvidy: can this be a document::view? What if the user doesn't want to transfer ownership?

amidvidy:
one question, otherwise LGTM

Machyne:
^ same :)

src/mongocxx/hint.hpp
samantharitter: I suppose so, with a comment about the original index document's lifetime being a superset of the hint's.  Would you opt for just document::view or offering both view and value?

amidvidy: let's offer both for now until we have some kind of type that represents either a view or value. Since this constructor could be called in hot code in a user's application, it makes sense to offer both the zero-copy ownership transfer (for perf) and the copying constructor (for convenience).

amidvidy: see discussion on #269 for more details on the hypothetical view/value variant type

src/mongocxx/collection.cpp
samantharitter: Done, with view() instead of extract()

src/mongocxx/collection.cpp
samantharitter: done

src/mongocxx/hint.hpp
samantharitter: Done

Machyne:
Tests pass locally... so LGTM!

PR #347, opened 2015-10-21
CXX-713: add private macros to c++ private methods

Machyne:


src/bsoncxx/json.cpp
```diff
@@ -209,11 +209,11 @@ class json_visitor {
     std::vector<bool> stack;
     std::size_t padding;
 
-    void pad(std::size_t extra = 0) {
+    BSONCXX_PRIVATE void pad(std::size_t extra = 0) {
```
amidvidy: this should only be necessary in header files

src/bsoncxx/json.cpp
```diff
@@ -209,11 +209,11 @@ class json_visitor {
...
         out << std::setw((stack.size() + -1 + extra + padding) * 4) << "" << std::setw(0);
     }
 
-    void visit_children(const document::view& view) {
+    BSONCXX_PRIVATE void visit_children(const document::view& view) {
```
amidvidy: ditto

src/bsoncxx/json.cpp
acmorrow: More specifically: you should only need to add BSONCXX_PRIVATE or MONGOCXX_PRIVATE if the containing class or structure is marked MONGOCXX_API. Why?

We will build with -fvisibility=hidden, which makes *everything* hidden by default, unless marked with __attribute__((visibility("default"))), which is what MONGOCXX_API expands to.

The issue is that when you have a class that is tagged MONGOCXX_API, then all of its members become part of the ABI. By marking private members of *exported* classes as MONGOCXX_PRIVATE, we override the class wide export, and keep those functions from leaking into the ABI.

That way, if we need to change the signature, types, etc. in the private constructors or methods of these types, then those will not require an ABI bump.

acmorrow:
This LGTM mod Adam's comment about the json.cpp file.

PR #348, opened 2015-10-21
Fix issues from #331

Machyne:
Fix issues from #331

un-expose underlying int values, fix name style, more tests

### DIFF
--- a/src/mongocxx/write_concern.cpp
+++ b/src/mongocxx/write_concern.cpp
@@ -65,9 +65,29 @@ void write_concern::nodes(std::int32_t confirm_from) {
 }
 
 void write_concern::acknowledge_level(write_concern::level confirm_level) {
-    libmongoc::write_concern_set_w(
-        _impl->write_concern_t,
-        static_cast<std::underlying_type<write_concern::level>::type>(confirm_level));
+    std::int32_t w;
+    switch (confirm_level) {
+    case write_concern::level::k_default:
+        w = MONGOC_WRITE_CONCERN_W_DEFAULT;
+        break;
+    case write_concern::level::k_majority:
+        w = MONGOC_WRITE_CONCERN_W_MAJORITY;
+        break;
+    case write_concern::level::k_unacknowledged:
+        w = MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED;
+        break;
+    case write_concern::level::k_tag:
+        // no exception for setting tag if it's set
+        if (libmongoc::write_concern_get_w(_impl->write_concern_t) == MONGOC_WRITE_CONCERN_W_TAG)
+            return;
+    default:
+        // TODO throw a mongocxx exception
+        throw std::invalid_argument(
+            "acknowledge_level can only be used to set levels of k_default, k_majority, or "
+            "k_unacknowledged.");
+        return;
+    }
+    libmongoc::write_concern_set_w(_impl->write_concern_t, w);
 }
 
 void write_concern::tag(stdx::string_view confirm_from) {
@@ -96,9 +116,21 @@ stdx::optional<std::int32_t> write_concern::nodes() const {
 }
 
 stdx::optional<write_concern::level> write_concern::acknowledge_level() const {
+    stdx::optional<write_concern::level> ack_level;
     std::int32_t w = libmongoc::write_concern_get_w(_impl->write_concern_t);
-    return w < 1 ? stdx::optional<write_concern::level>{static_cast<write_concern::level>(w)}
-                 : stdx::nullopt;
+    if (w >= 1) return stdx::nullopt;
+    switch (w) {
+    case MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED:
+        return write_concern::level::k_unacknowledged;
+    case MONGOC_WRITE_CONCERN_W_DEFAULT:
+        return write_concern::level::k_default;
+    case MONGOC_WRITE_CONCERN_W_MAJORITY:
+        return write_concern::level::k_majority;
+    case MONGOC_WRITE_CONCERN_W_TAG:
+        return write_concern::level::k_tag;
+    default:
+        return write_concern::level::k_unknown;
+    }
 }
 
 stdx::optional<std::string> write_concern::tag() const {
--- a/src/mongocxx/write_concern.hpp
+++ b/src/mongocxx/write_concern.hpp
@@ -59,10 +59,12 @@ class MONGOCXX_API write_concern {
     /// A class to represent the special case values for write_concern::nodes.
     /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
     ///
-    enum class level : std::int32_t {
-        kUnacknowledged = 0,
-        kDefault = -2,
-        kMajority = -3
+    enum class level {
+        k_default,
+        k_majority,
+        k_tag,
+        k_unacknowledged,
+        k_unknown,
     };
 
     ///
@@ -148,9 +150,11 @@ class MONGOCXX_API write_concern {
     /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
     ///
     /// @param confirm_level
-    ///   Either level::kUnacknowledged, level::kDefault, or level::kMajority.
+    ///   Either level::k_unacknowledged, level::k_default, or level::k_majority.
     ///
-    /// @warning Setting this to level::kUnacknowledged disables write acknowledgment and all other
+    /// @note the acknowledge level of level::k_tag is set automatically when a tag is set.
+    ///
+    /// @warning Setting this to level::k_unacknowledged disables write acknowledgment and all other
     /// write concern options.
     ///
     void acknowledge_level(level confirm_level);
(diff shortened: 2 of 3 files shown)


samantharitter:
Can you rename the commit so it references CXX-704?  I'd open that ticket back up and have this work be under it.

Machyne:
Yes! Good idea :)

src/mongocxx/write_concern.cpp
```diff
@@ -96,9 +107,25 @@ stdx::optional<std::int32_t> write_concern::nodes() const {
 }
 
 stdx::optional<write_concern::level> write_concern::acknowledge_level() const {
+    stdx::optional<write_concern::level> ack_level;
```
acmorrow: I think you can simplify this a bit, since you have the optional, I don't the ternary helsp?

if (w >= 1)
    return stdx::nullopt;

switch (w) {
    case 0:
        return {write_concern::level::k_un...};
    case -2:
        return {write_concern::level::k_default...};

I'm actually not 100% sure that you need the {} around the value. Maybe not? Could you try it and see how it goes? You don't need the break if you are doing an early return.

src/mongocxx/write_concern.cpp
```diff
@@ -96,9 +107,25 @@ stdx::optional<std::int32_t> write_concern::nodes() const {
...
+    case -3:
+        ack_level = write_concern::level::k_majority;
+        break;
+    case -4:
+        ack_level = write_concern::level::k_tag;
+        break;
+    default:
+        ack_level = write_concern::level::k_unknown;
+    }
+    return w < 1 ? ack_level : stdx::nullopt;
```
acmorrow: My other comment obviates this by eliminating the ternary op, but better to always parenthesize:

return (w < 1) ? ...

src/mongocxx/write_concern.hpp
```diff
@@ -59,10 +59,12 @@ class MONGOCXX_API write_concern {
     /// A class to represent the special case values for write_concern::nodes.
     /// @see http://docs.mongodb.org/manual/reference/write-concern/#w-option
     ///
-    enum class level : std::int32_t {
-        kUnacknowledged = 0,
-        kDefault = -2,
-        kMajority = -3
+    enum class level {
+        k_unacknowledged,
```
acmorrow: Alphabetize?

src/mongocxx/write_concern.cpp
```diff
@@ -65,9 +65,20 @@ void write_concern::nodes(std::int32_t confirm_from) {
 }
 
 void write_concern::acknowledge_level(write_concern::level confirm_level) {
-    libmongoc::write_concern_set_w(
-        _impl->write_concern_t,
-        static_cast<std::underlying_type<write_concern::level>::type>(confirm_level));
+    std::int32_t w;
+    switch (confirm_level) {
+    case write_concern::level::k_unacknowledged:
```
samantharitter: Since we're doing the translation between write_concern::level and integer values in a couple of places, maybe we should pull these out into constants somewhere

src/mongocxx/write_concern.cpp
```diff
@@ -96,9 +107,25 @@ stdx::optional<std::int32_t> write_concern::nodes() const {
 }
 
 stdx::optional<write_concern::level> write_concern::acknowledge_level() const {
+    stdx::optional<write_concern::level> ack_level;
     std::int32_t w = libmongoc::write_concern_get_w(_impl->write_concern_t);
-    return w < 1 ? stdx::optional<write_concern::level>{static_cast<write_concern::level>(w)}
-                 : stdx::nullopt;
+    switch (w) {
+    case 0:
```
samantharitter: See note above

src/mongocxx/write_concern.cpp
```diff
@@ -65,9 +65,20 @@ void write_concern::nodes(std::int32_t confirm_from) {
...
+    std::int32_t w;
+    switch (confirm_level) {
+    case write_concern::level::k_unacknowledged:
+        w = 0;
+        break;
+    case write_concern::level::k_majority:
+        w = -3;
+        break;
+    case write_concern::level::k_default:
+    default:
```
acmorrow: Nice use of fallthrough. However, I wonder if we want setting it here to anything other than k_un..., k_majority, or k_default to actually be an exception? Or, convince me that defaulting to default is the right thing to do.

src/mongocxx/write_concern.cpp
acmorrow: +1

samantharitter:
I understand why we didn't want to leak numeric values for these constants by putting them in the enum, but it might be worth putting those translations somewhere else privately.

src/mongocxx/write_concern.cpp
Machyne: It works without the {}. Looks much nicer :)

src/mongocxx/write_concern.cpp
Machyne: Now using `MONGOC_WRITE_CONCERN_W_...`

Machyne:
Still need to decide if defaulting to k_default or throwing an exception is better... probably exception.

src/mongocxx/write_concern.cpp
```diff
@@ -65,9 +65,27 @@ void write_concern::nodes(std::int32_t confirm_from) {
 }
 
 void write_concern::acknowledge_level(write_concern::level confirm_level) {
-    libmongoc::write_concern_set_w(
-        _impl->write_concern_t,
-        static_cast<std::underlying_type<write_concern::level>::type>(confirm_level));
+    std::int32_t w;
+    switch (confirm_level) {
+    case write_concern::level::k_default:
+        w = MONGOC_WRITE_CONCERN_W_DEFAULT;
```
amidvidy: can you prefix the C enum values with :: (unless they are macros)

src/mongocxx/write_concern.cpp
```diff
@@ -65,9 +65,27 @@ void write_concern::nodes(std::int32_t confirm_from) {
...
+        break;
+    case write_concern::level::k_unacknowledged:
+        w = MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED;
+        break;
+    case write_concern::level::k_tag:
+        // no exception for setting tag if it's set
+        if (libmongoc::write_concern_get_w(_impl->write_concern_t) == MONGOC_WRITE_CONCERN_W_TAG)
+            return;
+    default:
+        throw std::invalid_argument(
```
amidvidy: How is this possible?

src/mongocxx/write_concern.cpp
Machyne: They are macros.

src/mongocxx/write_concern.cpp
Machyne: Not sure if this is your question... but this is reached when calling `wc.acknowledge_level(k_unknown)` or `wc.acknowledge_level(k_tag)` (but I allow setting it to tag if a tag is already specified, I thought it would be nice to have the option of being explicit).

amidvidy: ah ok. I'm not sure how I feel about throwing std::invalid_argument. I would prefer if all exceptions we throw had a common (mongocxx/bsoncxx specific) base class

acmorrow: Do we have a mongocxx base exception yet? Does it derive from std::exception? I'm thinking about:

namespace mongo {
template<typename T>
std_exception : public exeption, virtual public T {;
};

throw std_exception<std::invalid_argument>(...)

or something.

acmorrow: In any event, I think our exceptions story is yet to be resolved. Maybe just TODO this for now and mention it in the ticket we have somewhere about cleaning up exceptions? I'm 99% sure there is one.

amidvidy: we do have a base exception: https://github.com/mongodb/mongo-cxx-driver/blob/master/src/mongocxx/exception/base.hpp

Machyne: [CXX-627](https://jira.mongodb.org/browse/CXX-627) is the ticket about exceptions.

I will change this line to TODO and note it in CXX-627. Sound good?

amidvidy: yup

Machyne:
With the TODO, does this look good now?

samantharitter:
I think this LGTM

src/mongocxx/write_concern.cpp
```diff
@@ -65,9 +65,26 @@ void write_concern::nodes(std::int32_t confirm_from) {
...
+    case write_concern::level::k_unacknowledged:
+        w = MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED;
+        break;
+    case write_concern::level::k_tag:
+        // no exception for setting tag if it's set
+        if (libmongoc::write_concern_get_w(_impl->write_concern_t) == MONGOC_WRITE_CONCERN_W_TAG)
+            return;
+    default:
+        // TODO throw
+        return;
```
acmorrow: I don't like silently eating the users request to set it to a bad value.

We should do *something* here even if it isn't great. I'd prefer terminating, or throwing std::invalid_argument. And keeping a TODO to fix up the exception.

src/mongocxx/write_concern.cpp
acmorrow: Yeah, much better.

acmorrow:
LGTM but we need to do something on setting an invalid ack level.

acmorrow:
@Machyne No need to post for further review, just fix up that one issue and go ahead and push.

PR #349, opened 2015-10-21
CXX-689: Support bypassDocumentValidation

markbenvenuto:
I added support for `bypassDocumentValidation` to insert, update, write, bulk write, and findAndModify. I did not add it to `copyDatabase` or `mapreduce`.

If the user switches on `bypassDocumentValidation`, I force them to use write commands instead of legacy operations.


acmorrow:
@markbenvenuto Could you please move CXX-689 to "in code review" and link to this PR? Thanks...

src/mongo/client/bulk_operation_builder.h
```diff
@@ -68,8 +68,12 @@ class MONGO_CLIENT_API BulkOperationBuilder {
      * @param client The connection to use.
      * @param ns The namespace to apply the operations to.
      * @param ordered Whether or not ordering matters for these operations.
+     * param bypassDocumentValidation Whether to bypass document validation for these operations,
      */
-    BulkOperationBuilder(DBClientBase* const client, const std::string& ns, bool ordered);
+    BulkOperationBuilder(DBClientBase* const client,
+                         const std::string& ns,
+                         bool ordered,
+                         bool bypassDocumentValidation);
```
acmorrow: Maybe default the new parameter to 'false' to avoid source level breakage? That or overload the function.

src/mongo/client/dbclient.cpp
```diff
@@ -2110,16 +2127,18 @@ void DBClientBase::update(
...
+    bool bypassDocumentValidation = flags & UpdateOption_BypassDocumentValidation;
+
     WriteResult writeResult;
-    _write(ns, updates.ops, true, wc, &writeResult);
+    _write(ns, updates.ops, true, bypassDocumentValidation, wc, &writeResult);
 }
 
 BulkOperationBuilder DBClientBase::initializeOrderedBulkOp(const std::string& ns) {
-    return BulkOperationBuilder(this, ns, true);
+    return BulkOperationBuilder(this, ns, true, false);
```
acmorrow: How does one get a bulk operation builder that supports bypass?

src/mongo/client/dbclient.cpp
```diff
@@ -2110,16 +2127,18 @@ void DBClientBase::update(
...
 }
 
 BulkOperationBuilder DBClientBase::initializeUnorderedBulkOp(const std::string& ns) {
-    return BulkOperationBuilder(this, ns, false);
+    return BulkOperationBuilder(this, ns, false, false);
```
acmorrow: Ditto

src/mongo/client/wire_protocol_writer.cpp
```diff
@@ -28,8 +28,11 @@ WireProtocolWriter::WireProtocolWriter(DBClientBase* client) : _client(client) {
 void WireProtocolWriter::write(const StringData& ns,
                                const std::vector<WriteOperation*>& write_operations,
                                bool ordered,
+                               bool bypassDocumentValidation,
                                const WriteConcern* writeConcern,
                                WriteResult* writeResult) {
+    invariant(bypassDocumentValidation == false);
```
acmorrow: By saying 'invariant' here, you are asserting that this code is unreachable by user action. Is that true? Otherwise should be a uassert.

src/mongo/integration/standalone/dbclient_test.cpp
```diff
@@ -112,6 +112,83 @@ const bool kCompiledWithSSL = true;
...
+    c->createCollectionWithOptions(
+        TEST_NS, 0, false, 0, BSON("validator" << BSON("fieldName" << GTE << 1024)));
+
+    c->insert(TEST_NS, BSON("_id" << 1 << "fieldName" << 1024));
+
+    ASSERT_THROWS(
+        c->findAndModify(
+            TEST_NS, BSON("fieldName" << 1024), BSON("$dec" << BSON("fieldName" << 1)), false),
+        OperationException);
+    ASSERT_THROWS(c->findAndModify(TEST_NS,
```
acmorrow: blank line above, maybe

acmorrow:
This looks pretty good to me, and I like the tests. A few small questions and suggestions.

@amidvidy would you please also take a look?

src/mongo/client/dbclient.cpp
markbenvenuto: You can use the constructor directly. i.e. `BulkOperationBuilder bulk(client, ns, true, true)`. See tests.

src/mongo/client/wire_protocol_writer.cpp
markbenvenuto: There is no way for users to reach this if they use the `DBClient` API. If they use `WireProtocolWriter` directly, then they can use. Is `WireProtocolWriter` considered part of the public API of the driver?

src/mongo/client/dbclient.cpp
acmorrow: I guess these were just helper functions for end users? OK then. I suppose we could also add a defaulted parameter for bypass to them? Not sure it is worth it. We can always do it later if people need it. I doubt they will.

src/mongo/client/wire_protocol_writer.cpp
acmorrow: No, it is not part of the public API, so I think this is just fine.

acmorrow:
LGTM. @amidvidy ?

amidvidy:
LGTM

PR #350, opened 2015-10-22
CXX-717: implement C++ version verification

timn:
mongo-cxx-driver crashes if used in an application built with a
different C++ version than the library itself. Add a little compile time
generator that adds the library C++ version and a snippet in config.h
that throws an error if compiling a file against mongo-cxx-driver
headers with a different C++ version.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1696,11 +1696,28 @@ def doConfigure(myenv):
         context.Result(ret)
         return ret
 
+    def GetCXXVersion(context):
+        test_body = """
+        #include <cstdio>
+
+        int
+        main(int argc, char **argv)
+        {
+          printf("%lu", __cplusplus);
+          return 0;
+        }
+        """
+        context.Message('Checking for C++ version... ')
+        ret = context.TryRun(textwrap.dedent(test_body), '.cpp')
+        context.Result(ret[1])
+        return ret[1]
+
     conf = Configure(myenv, help=False, custom_tests = {
         'CheckCXX11Atomics': CheckCXX11Atomics,
         'CheckGCCAtomicBuiltins': CheckGCCAtomicBuiltins,
         'CheckGCCSyncBuiltins': CheckGCCSyncBuiltins,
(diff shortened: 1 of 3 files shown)


kikop11:
who can tell me,how to compile mongo-cxx-driver-master on windows?

acmorrow:
@kikop11 This question would be more appropriate on the mongodb-user group, rather than a random pull request on github.

Briefly, you use CMake to generate a solution file, and then you can build the driver either within VS2015, or on the command line via msbuild.exe.

Please see our appveyor integration file for a demonstration of how to build with msbuild:

https://github.com/mongodb/mongo-cxx-driver/blob/master/appveyor.yml

kikop11:
Hello, build the C driver success,I'm trying to build the C++ driver, from mongo-cxx-driver-master.zip. I'm on Windows 7, 64-bit Intel, with mingw492_32. I run cmake gui configure,error like:

  

CMake Error at E:/MongodbMaster/cmake-3.5.2-win32-x86/cmake-3.5.2-win32-x86/share/cmake-3.5/Modules/ExternalProject.cmake:1757 (message):
error: could not find git for clone of EP_mnmlstc_core
Call Stack (most recent call first):
E:/MongodbMaster/cmake-3.5.2-win32-x86/cmake-3.5.2-win32-x86/share/cmake-3.5/Modules/ExternalProject.cmake:2459 (_ep_add_download_command)
src/bsoncxx/third_party/CMakeLists.txt:19 (ExternalProject_Add)
Configuring incomplete, errors occurred!

Can you tell me what I need to install? Thanks!

acmorrow:
It looks like CMake's ExternalProject_Add can't find the 'git' executable in order to clone the MNMLSTC/core headers.

- Do you have git installed?
- Is it in your path (however that works on MinGW)

Maybe there is a define you can pass in to CMake to tell ExternalProject_Add where to find git?

xdg:
Per comment on Jira about cleaning up long-open tickets where we're not assigning resources, I'm going to close this issue here as well.

PR #352, opened 2015-10-23
CXX-689 Assert bypassDocumentValidation is not used with unacknowledged writes

markbenvenuto:
Per DRIVERS-250, when the a driver uses the legacy write operations to support unacknowledged writes, and the user sets bypassDocumentValidation, the driver should error.

Note: We cannot differentiate between a user setting bypassDocumentValidation to false, or if bypassDocumentValidation was simply initialized to false via default parameter at the moment. This could addressed, but I do not think it is worth the cost to differentiate these two scenarios.


amidvidy:
LGTM

acmorrow:
LGTM

PR #355, opened 2015-10-26
CXX-716: use stdx names in the mongocxx via mongocxx/stdx.hpp

Machyne:
switch all `bsconcxx::stdx` to be `mongocxx::stdx` under the mongocxx namespace

### DIFF
--- a/src/mongocxx/client.hpp
+++ b/src/mongocxx/client.hpp
@@ -164,8 +164,8 @@ class MONGOCXX_API client {
     ///
     /// @return The database
     ///
-    class database database(stdx::string_view name) const &;
-    class database database(stdx::string_view name) const && = delete;
+    class database database(stdx::string_view name) const&;
+    class database database(stdx::string_view name) const&& = delete;
 
     ///
     /// Allows the syntax @c client["db_name"] as a convenient shorthand for the client::database()
@@ -178,8 +178,8 @@ class MONGOCXX_API client {
     ///
     /// @return Client side representation of a server side database
     ///
-    MONGOCXX_INLINE class database operator[](stdx::string_view name) const &;
-    MONGOCXX_INLINE class database operator[](stdx::string_view name) const && = delete;
+    MONGOCXX_INLINE class database operator[](stdx::string_view name) const&;
+    MONGOCXX_INLINE class database operator[](stdx::string_view name) const&& = delete;
 
     ///
     /// Enumerates the databases in the client.
--- a/src/mongocxx/collection.hpp
+++ b/src/mongocxx/collection.hpp
@@ -142,11 +142,9 @@ class MONGOCXX_API collection {
     /// @see mongocxx::bulk_write
     /// @see http://docs.mongodb.org/manual/core/bulk-write-operations/
     ///
-    template<typename container_type>
-    MONGOCXX_INLINE bsoncxx::stdx::optional<result::bulk_write> bulk_write(
-        const container_type& writes,
-        const options::bulk_write& options = options::bulk_write()
-    );
+    template <typename container_type>
+    MONGOCXX_INLINE stdx::optional<result::bulk_write> bulk_write(
+        const container_type& writes, const options::bulk_write& options = options::bulk_write());
 
     ///
     /// Sends writes starting at @c begin and ending at @c end to the server as a bulk write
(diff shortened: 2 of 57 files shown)


src/mongocxx/collection.hpp
```diff
@@ -329,10 +318,8 @@ class MONGOCXX_API collection {
     ///
     /// @see http://docs.mongodb.org/manual/core/read-operations-introduction/
     ///
-    bsoncxx::stdx::optional<bsoncxx::document::value> find_one(
-        bsoncxx::document::view filter,
-        const options::find& options = options::find()
-    );
+    stdx::optional<bsoncxx::document::value> find_one(
+        bsoncxx::document::view filter, const options::find& options = options::find());
```
samantharitter: these closing lines are all clang format changes?

src/mongocxx/options/find_one_and_replace.hpp
```diff
@@ -19,9 +19,11 @@
 #include <cstdint>
 
 #include <bsoncxx/document/view.hpp>
-#include <mongocxx/options/find_one_and_modify.hpp>
 #include <bsoncxx/stdx/optional.hpp>
```
samantharitter: Do we still need this now?  Should headers that use stdx::optional include <mongocxx/stdx.hpp>, <bsoncxx/stdx/optional.hpp>, or both?

src/mongocxx/collection.hpp
Machyne: Yes! After removing the `bsoncxx::` clang-format wanted them on the same line. I guess it shrunk just enough? Either that or it was breaking the format before.

src/mongocxx/options/find_one_and_replace.hpp
Machyne: I just checked, and both are necessary to build. The `<bsoncxx/stdx/*>` defines things such as `optional`, and the `<mongocxx/stdx.hpp>` just says to use all of the `bsoncxx::stdx` namespace inside `mongocxx::stdx`.

Machyne:
ping

samantharitter:
LGTM

amidvidy:
LGTM

src/mongocxx/client.hpp
```diff
@@ -164,8 +164,8 @@ class MONGOCXX_API client {
     ///
     /// @return The database
     ///
-    class database database(stdx::string_view name) const &;
-    class database database(stdx::string_view name) const && = delete;
+    class database database(stdx::string_view name) const&;
```
acmorrow: Seems unrelated, or did clang-format do this?

src/mongocxx/stdx.hpp
```diff
@@ -21,7 +21,7 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace stdx {
 
 // We adopt all the bsoncxx polyfills
-using namespace bsoncxx::stdx;
+using namespace ::bsoncxx::stdx;
```
acmorrow: Nice improvement

src/mongocxx/write_concern.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include <bsoncxx/stdx/optional.hpp>
 #include <bsoncxx/stdx/string_view.hpp>
+
```
acmorrow: Is this change following style elsewhere? I don't remember that we separated the blocks for bsoncxx and mongocxx headers as a general principle, I think we just sorted them together,

src/mongocxx/options/find_one_and_replace.hpp
acmorrow: That may be more of an oversight than a design choice. Can anyone think of a good reason not to include the bsoncxx/stdx headers inside the mongocxx/stdx,hpp header?

src/mongocxx/stdx.hpp
acmorrow: When I wrote this file, I felt a little bad. I wondered if it would have been better to have mongocxx/stdx/optional.hpp, etc., headers, but it just seemed like massive overkill.

However, I do wonder if we should be more restrictive here, and just import name by name as needed by mongocxx.

In other words, rather than saying using namespace in a header (generally not a great idea), just say

using ::bsoncxx::stdx::optional;
using ::bsoncxx::stdx::string_view;

etc.

What do people think?

src/mongocxx/options/find_one_and_replace.hpp
amidvidy: i think this was an explicit choice to avoid facade headers

acmorrow:
I see that this got merged already, which is fine. I do have a few small comments and suggestions above. 
The only real change I see as potentially needed is to include the bsoncxx/stdx headers into mongocxx/stdx.hpp and import type by type with usings.

src/mongocxx/options/find_one_and_replace.hpp
acmorrow: Yeah. Generally a principle I support but maybe working against us here. If we are really only including the bsoncxx/stdx/xyz.hpp headers we need and then individually pulling in the types with using, I think it is something we could live with.

PR #356, opened 2015-10-27
CXX-702 implement create_collection with options

samantharitter:


src/mongocxx/options/create_collection.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+///
+/// Class representing the optional arguments to a MongoDB createCollection command
+///
+class MONGOCXX_API create_collection {
+   public:
+    ///
+    /// To create a capped collection, specify true.
+    ///
+    /// @note If you specify true, you must also set a maximum size using the size() method.
```
samantharitter: Is this something that we want to enforce in the driver?  Currently it does not enforce this.

src/mongocxx/collection.hpp
```diff
@@ -119,10 +118,8 @@ class MONGOCXX_API collection {
     ///
     /// @see http://docs.mongodb.org/manual/reference/command/aggregate/
     ///
-    cursor aggregate(
-        const pipeline& pipeline,
-        const options::aggregate& options = options::aggregate()
-    );
+    cursor aggregate(const pipeline& pipeline,
```
samantharitter: Attack of the clang format.  Is clang format something we want to apply to this codebase, or not?  If not I will fix the style changes here.

src/mongocxx/collection.hpp
```diff
@@ -602,57 +562,51 @@ class MONGOCXX_API collection {
    private:
     friend class database;
 
-    MONGOCXX_PRIVATE collection(const database& database, bsoncxx::stdx::string_view collection_name);
+    MONGOCXX_PRIVATE collection(const database& database,
+                                bsoncxx::stdx::string_view collection_name);
+
+    MONGOCXX_PRIVATE collection(const database& database,
```
samantharitter: This is a new constructor (just pointing out the non-clang-format change in this file)

src/mongocxx/collection.hpp
```diff
@@ -47,6 +47,8 @@
 #include <mongocxx/result/update.hpp>
 #include <mongocxx/write_concern.hpp>
 
+#include <mongoc.h>
```
Machyne: Should use `#include <mongocxx/private/libmongoc.hpp>`.

Is this okay to do in a public header file? I don't know. I think we haven't, though.

samantharitter: Hmm you're right it's probably not ok.  I will think about what to do about that constructor...

src/mongocxx/options/create_collection.hpp
Machyne: Probably. I think we may need some kind of "mongocxx bad argument to options class" exception. It'd be useful for write concern options too :)

src/mongocxx/collection.hpp
Machyne: Instead of creating the mongoc_colletion_t in `database.cpp`, maybe make this constructor take the options class and create it here? The db class could just pass the options along.

src/mongocxx/collection.hpp
Machyne: I think we want to enforce the clang format. Would it be worth doing a full pass sometime?

src/mongocxx/options/create_collection.hpp
```diff
@@ -0,0 +1,119 @@
...
+    ///   When true, disables power of 2 sizing for this collection.
+    ///
+    void no_padding(bool no_padding);
+
+    ///
+    /// Return a bson document representing the options set on this object.
+    ///
+    /// @return Options, as a document.
+    ///
+    bsoncxx::document::value to_document();
```
Machyne: I like this a lot. We don't do this in other options classes (the document creation logic is done in `collection.cpp` e.g.), and I wonder why. Do you know?

Maybe this should be private. Is there a place that we'd want to expose this to users of the driver?

src/mongocxx/collection.hpp
acmorrow: There is a ticket slated for 0.4.0 to productionize clang-format, like we have done for the legacy driver and the server. In the meantime, we should do our best to keep the formatting in line with the current clang-format file in this branch manually.

src/mongocxx/options/create_collection.hpp
Machyne: But if you leave this, one of us can tackle it with [CXX-627](https://jira.mongodb.org/browse/CXX-627).

src/mongocxx/collection.cpp
```diff
@@ -83,6 +83,12 @@ collection::collection(const database& database, bsoncxx::stdx::string_view coll
           database.name(), database._impl->client_impl, collection_name.data())) {
 }
 
+collection::collection(const database& database, bsoncxx::stdx::string_view collection_name,
```
Machyne: Would it be worth switching `mongocxx::stdx` instead of `bsoncxx::stdx` throughout?
I have a pr open to change this everywhere, but as long as you're touching these files, it might be worth it.

src/mongocxx/options/create_collection.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include <mongocxx/stdx.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+namespace options {
+
+///
+/// Class representing the optional arguments to a MongoDB createCollection command
+///
+class MONGOCXX_API create_collection {
```
Machyne: Can you also tackle [CXX-720](https://jira.mongodb.org/browse/CXX-720) here? (adding options for document validation)

samantharitter: I'll take CXX-720, it's definitely related, but I think it should be done in a new PR. I think we'll have to add validators in several places.

src/mongocxx/options/create_collection.hpp
samantharitter: I'm not sure why, I think we had been opting for very minimal options classes that didn't contain any logic, just values.  But I like this, and it avoids implementing lots of getters.  I've also added a to_document() method to the hint PR.

I don't think it's a problem to allow users to call to_document(), I can imagine times when that would be useful.  I don't feel strongly either way.

src/mongocxx/collection.cpp
samantharitter: I'll change these here, and let's coordinate our merges to avoid a rebase nightmare

Machyne: Sounds good! Thanks!

src/mongocxx/options/create_collection.hpp
amidvidy: if you are providing this method, can you also provide a conversion operator, e.g. an (inline) operator bsoncxx:document::value() that calls to_document ?

src/mongocxx/test/database.cpp
```diff
@@ -155,3 +155,45 @@ TEST_CASE("A database", "[database]") {
...
+        }
+
+        SECTION("but raises exception when collection already exists") {
+            collection coll = database[collection_name];
+
+            try {
+                // This operation should throw
+                collection new_collection = database.create_collection(collection_name);
+
+                REQUIRE(false);
```
amidvidy: REQUIRE_THROWS with a lambda?

src/mongocxx/collection.hpp
amidvidy: remove this include, see comment below.

src/mongocxx/collection.hpp
```diff
@@ -602,57 +562,51 @@ class MONGOCXX_API collection {
    private:
     friend class database;
 
-    MONGOCXX_PRIVATE collection(const database& database, bsoncxx::stdx::string_view collection_name);
+    MONGOCXX_PRIVATE collection(const database& database,
+                                bsoncxx::stdx::string_view collection_name);
+
+    MONGOCXX_PRIVATE collection(const database& database,
+                                bsoncxx::stdx::string_view collection_name,
+                                mongoc_collection_t* collection);
```
amidvidy: make this a void*. Then in the CPP static_cast it to mongoc_collection_t*. This is what we do elsewhere when we need to erase the types of mongoc structs in mongocxx headers.

src/mongocxx/database.hpp
```diff
@@ -93,8 +92,7 @@ class MONGOCXX_API database {
     ///
     class collection create_collection(
         bsoncxx::stdx::string_view name,
-        bsoncxx::document::view options
-    );
+        options::create_collection options = options::create_collection());
```
amidvidy: we seem to be passing the options struct elsewhere by const-ref

src/mongocxx/options/create_collection.hpp
samantharitter: On the other hand, this isn't something the c driver enforces, we could just let the server return an error.  In general, I don't have a good idea of how much user input validation we want to do in this driver.

src/mongocxx/options/create_collection.hpp
samantharitter: With the conversion operator, would you opt to still expose the to_document() method, or make that private?

Machyne: Other examples (such as value) expose both:
```
BSONCXX_INLINE document::view view() const noexcept;
BSONCXX_INLINE operator document::view() const noexcept;
```

amidvidy: you can leave to_document i think. An example when it would be useful is if you have an explicit constructor taking a bsoncxx::document::value that you want to call using a hint. saying to_document is a lot nicer than using static_cast

amidvidy: the only think to bikeshed is the naming, should it be `to_document` or `value` to match the other methods?

samantharitter: Good point, will provide both.

`value()` is a bit weird because it's not immediately clear that we mean `bsoncxx::document::value` (I mean it is if you go read the header, but won't be from code).  Slight preference for `to_document`

samantharitter: So actually, looking at this again, the conversion doesn't really help us here:
```
libbson::scoped_bson_t opts_bson{options.to_document()};
```
because scoped_bson_t can only be constructed from a document::view, not a value.  Still worth adding the value conversion for other uses?

amidvidy: it's good that doesn't compile otherwise you would be holding a dangling view :). so I think its still useful

src/mongocxx/options/create_collection.hpp
amidvidy: yes, I think we should try to validate as much as possible client side.

src/mongocxx/options/create_collection.hpp
samantharitter: yes, exactly.  But it feels premature to add an operator that we aren't using anywhere yet.

Machyne: If for some reason, someone is making their own bson to use with a db command, then it could be useful?

amidvidy: yeah, I don't think our use/disuse of a method is necessarily a signal of its necessity given that this is a library.

src/mongocxx/options/create_collection.hpp
samantharitter: After a brief drivers survey, I'm not going to add validation for this now.  Driver authors do a range of things, and since there's no spec nobody really knows what the right behavior is.  The C driver does do some validation but not the validation described in our docs.

Machyne: Agreed!

src/mongocxx/options/create_collection.hpp
samantharitter: Acknowledged

samantharitter:
ping, is this good to go once green on Travis?

src/mongocxx/test/options/create_collection.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <bsoncxx/types.hpp>
+#include <bsoncxx/types/value.hpp>
+#include <mongocxx/options/create_collection.hpp>
+
+using namespace bsoncxx;
+using namespace mongocxx;
+
+TEST_CASE("create_collection", "[create_collection]") {
+    options::create_collection cc;
+
```
Machyne: Can you also add `CHECK_OPTIONAL_ARGUMENT` tests here?

Machyne:
Outside of Travis status and my request for more tests, LGTM!

src/mongocxx/test/options/create_collection.cpp
samantharitter: The CHECK_OPTIONAL_ARGUMENT macro requires that these methods have setters and getters defined.  I haven't implemented getters because we have the to_document() method.  For now I'd like to leave the getters off, we can always add them later.

PR #357, opened 2015-10-27
CXX-585: make public macros hygenic

Machyne:


src/bsoncxx/config/prelude.hpp
```diff
@@ -15,4 +15,12 @@
 #include <bsoncxx/config/config.hpp>
 #include <bsoncxx/export.hpp>
 
-#define BSONCXX_INLINE inline __attribute__ ((__visibility__("hidden"), __always_inline__))
+#pragma push_macro("MONGOCXX_INLINE")
```
acmorrow: I think you need to do these before you include the headers above. Otherwise, you are just pushing our definitions. In the legacy C++ driver, we also undef after we push so that our decl doesn't conflict. Checkout undef_macros and redef_macros in the legacy driver.

src/bsoncxx/builder/core.cpp
```diff
@@ -481,5 +481,3 @@ void core::clear() {
 }  // namespace builder
 BSONCXX_INLINE_NAMESPACE_END
 }  // namespace bsoncxx
-
-#include <bsoncxx/config/postlude.hpp>
```
acmorrow: Good catch. Harmless, but not stylish

src/bsoncxx/config/prelude.hpp
Machyne: Oops! Good catch :)

src/bsoncxx/config/postlude.hpp
```diff
@@ -11,3 +11,35 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
```
amidvidy: given that most of these macros are defined by CMake, it would be greatly preferable if we could generate the postlude automatically from CMake. Can you investigate if there is a feasible way to do this, at least for the macros generated by CMake?

Machyne:
I filed a ticket to cmake, but for now all of the auto-gen'ed macros are handled manually.

acmorrow:
Link to the ticket?

Machyne:
http://public.kitware.com/Bug/view.php?id=15822

src/mongocxx/result/delete.cpp
```diff
@@ -32,5 +34,3 @@ std::int32_t delete_result::deleted_count() const {
 }  // namespace result
 MONGOCXX_INLINE_NAMESPACE_END
 }  // namespace mongocxx
-
```
samantharitter: Why don't we want to be including the postlude in these files?

acmorrow: It isn't needed in .cpp files. The postlude exists to restore the state to as close to pristine as we can get after you have seen a header file as a client. That isn't a concern for .cpp files since clients never include them

src/bsoncxx/config/postlude.hpp
acmorrow: We filed a ticket with CMake about this, right? Can we link the ticket here?

src/bsoncxx/config/postlude.hpp
```diff
@@ -11,3 +11,55 @@
...
+// src/bsoncxx/b64_ntop.h
+#undef BSONCXX_B64_ASSERT
+#pragma pop_macro("BSONCXX_B64_ASSERT")
+
+// src/bsoncxx/document/element.cpp
```
acmorrow: Shouldn't be needed for .cpp files

src/bsoncxx/config/prelude.hpp
```diff
@@ -12,7 +12,58 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// src/bsoncxx/b64_ntop.h
+#pragma push_macro("BSONCXX_B64_ASSERT")
+#undef BSONCXX_B64_ASSERT
+
+// src/bsoncxx/document/element.cpp
```
acmorrow: Shouldn't be needed for .cpp files

src/bsoncxx/config/prelude.hpp
```diff
@@ -12,7 +12,58 @@
...
+#undef BSONCXX_PRIVATE
+
 #include <bsoncxx/config/config.hpp>
+#include <bsoncxx/config/version.hpp>
 #include <bsoncxx/export.hpp>
 
-#define BSONCXX_INLINE inline __attribute__ ((__visibility__("hidden"), __always_inline__))
+#pragma push_macro("BSONCXX_INLINE")
+#undef BSONCXX_INLINE
+#define BSONCXX_INLINE inline __attribute__((__visibility__("hidden"), __always_inline__))
```
acmorrow: This definition never belonged in prelude.cpp. Could you add a TODO to move it to a compiler.hpp header, and file an associated ticket?

src/bsoncxx/config/CMakeLists.txt
```diff
@@ -4,11 +4,6 @@ configure_file(
 )
 
 configure_file(
-    ${CMAKE_CURRENT_SOURCE_DIR}/private/config.hpp.in
```
acmorrow: These actually do have a purpose. I don't think we should remove them. It is fine for them to remain empty for now. If we ever had any configure checks that resulted in HAVE macros that were purely internal, this is where they would go. So, I'd like to keep it.

src/mongocxx/config/CMakeLists.txt
```diff
@@ -8,11 +8,6 @@ configure_file(
     ${CMAKE_CURRENT_BINARY_DIR}/config.hpp
 )
 
-configure_file(
```
acmorrow: Please add these back, per my other comment.

src/mongocxx/config/prelude.hpp
```diff
@@ -12,7 +12,44 @@
...
+#undef MONGOCXX_NO_DEPRECATED
+#pragma push_macro("MONGOCXX_PRIVATE")
+#undef MONGOCXX_PRIVATE
+
 #include <mongocxx/config/config.hpp>
+#include <mongocxx/config/version.hpp>
 #include <mongocxx/export.hpp>
 
+#pragma push_macro("MONGOCXX_INLINE")
+#undef MONGOCXX_INLINE
```
acmorrow: Like in the bson side, this shouldn't be defined here. Add a TODO here as well, can be the same ticket as I suggested in the bson side. Ideally, MONGOCXX_INLINE would just expand to BSONCXX_INLINE, but we probably want a mongocxx/compiler.hpp file as well. It can just include the bson lib one.

src/mongocxx/options/bulk_write.hpp
```diff
@@ -18,6 +18,8 @@
 
 #include <bsoncxx/stdx/optional.hpp>
 
+#include <mongocxx/write_concern.hpp>
```
acmorrow: Unrelated?

src/mongocxx/config/prelude.hpp
Machyne: [CXX-733](https://jira.mongodb.org/browse/CXX-733)

src/bsoncxx/config/prelude.hpp
Machyne: [CXX-733](https://jira.mongodb.org/browse/CXX-733)

src/mongocxx/options/bulk_write.hpp
Machyne: Unrelated, but I needed to move this over here form the cpp file in order to compile. We do use the write concern in our hpp file, so unless we want to forward declare the class, I think we need it here.

Currently investigating why this isn't an issue in other branches...

acmorrow:
LGTM. Ship it!

src/bsoncxx/config/prelude.hpp
```diff
@@ -12,7 +12,51 @@
...
+#undef BSONCXX_NO_DEPRECATED
+#pragma push_macro("BSONCXX_PRIVATE")
+#undef BSONCXX_PRIVATE
+
 #include <bsoncxx/config/config.hpp>
+#include <bsoncxx/config/version.hpp>
 #include <bsoncxx/export.hpp>
 
-#define BSONCXX_INLINE inline __attribute__ ((__visibility__("hidden"), __always_inline__))
+// TODO to move this to a compiler.hpp header
```
acmorrow: If you haven't already, please file a ticket for this in the epic, and schedule for 3.0.0-rc0.

Machyne: [CXX-733](https://jira.mongodb.org/browse/CXX-733)

PR #358, opened 2015-10-27
CXX-629: assign visibility to enum classes where needed

Machyne:


src/mongocxx/collection.cpp
```diff
@@ -47,18 +47,6 @@
 #include <mongocxx/result/update.hpp>
 #include <mongocxx/write_concern.hpp>
 
-namespace {
-enum class cursor_flag : uint32_t {
```
Machyne: This was totally unused.

samantharitter: Hmm I wonder whether that's an indication that the flag was unnecessary, or that our implementation is incomplete...

samantharitter:
LGTM except I am unsure about the cursor_flags

src/mongocxx/collection.cpp
Machyne: Probably the latter... we have a partial duplicate of this in `src/mongocxx/options/find.hpp`

Machyne:
I found that I can access all of the enums when installing and then importing... I need to do more research on this first. Sorry for premature PR! Should have tested more first.

Machyne:
It looks like what's happening is the enum's aren't getting exported, so the real visibility difference is forward declaring vs. putting the whole enum in the public header files.

Do we want the macro's anyway for visibility?

Note, that the one real change this PR makes is to make the cursor type public (which I think we want, as it's a parameter to set the cursor type in find options).

amidvidy:
I'm not sure we need MONGOCXX_API on the enum class types, given that they are (a) purely symbolic, and (b) are generally fully defined in headers. Can you compile with both gcc/clang and see if they give any warnings about the annotations being unnecessary/unenforceable

Machyne:
This creates warnings in gcc
`src/bsoncxx/types.hpp:54:1: warning: type attributes ignored after type is already defined [-Wattributes]`
but not clang.

These visibility attributes don't affect the enum classes. I think that I'll remove the macros, but keep the other minor cleanup changes (especially because I did move one of the enums that we want public into the hpp file) and then resolve the issue. Does that sound good?

amidvidy:
yes!

Machyne:
Done!

amidvidy:
LGTM!

samantharitter:
(still LGTM)

PR #359, opened 2015-10-28
CXX-714: Build systems should run tests with catch

samantharitter:


.mci.yml
```diff
@@ -80,7 +80,7 @@ tasks:
               script: |
                   PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DBUILD_UNIT_TESTS=true -DBUILD_EXAMPLES=true -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
                   make
-                  make test
+                  ./build/src/mongocxx/test/test_driver
```
Machyne: Probably `./src/mongocxx/test/test_driver` if you're already in the build folder for `make`.

.travis.yml
```diff
@@ -70,4 +70,6 @@ before_script:
 
 script:
     - make -j3
-    - make test
+
+    # Run tests with catch
+    - ./build/src/mongocxx/test/test_driver
```
Machyne: Same

.mci.yml
samantharitter: yuuup saw that but forgot to push it to this branch!

Machyne:
As long as you're touching these files, can you change the way we configure the c driver to be
`./configure --enable-tests=no --enable-examples=no --with-libbson=bundled --prefix=...`
(use `./configure` over `./autogen.sh` and add in `--with-libbson=bundled` just to be explicit)

samantharitter:
Because we're cloning from the repository, we won't have the `configure` script until after running `./autogen.sh`

Machyne:
oops! yeah, we'll have to use `autogen.sh` first

LGTM!

samantharitter:
@amidvidy want to give this a quick look so I can push it?  We're green on Travis and Evergreen.

amidvidy:
LGTM. in the future maybe we want CMake to have an option to install the test executables so we don't have to reach in to the build directory to run them, but in the meantime this is better than what we had

PR #360, opened 2015-10-28
CXX-696: Add tests to verify spec compliance with unacknowledged write concerns

markbenvenuto:
According to the [read and write concern spec](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst):

> An `Unacknowledged WriteConcern` is when (`w` equals 0) AND (`journal` is not set or is `false`).

This change adds a few simple unit tests to ensure our compliance. There are no other changes.


amidvidy:
LGTM

PR #361, opened 2015-10-28
CXX-692 Check writeConcern errors in findAndModify

markbenvenuto:
Per an update to DRIVERS-224, I added a test to verify `findAndModify` uses writeConcern correctly. This was important because `findAndModify` goes through a different code path the insert/update/delete.

Changes
1. Make `getMaxWireVersion` so that replica set connections can check it. This uses the primary. It is undefined if customers are using mixed version clusters.
2. Check `WriteResult` in `_findAndModify` to verify the writeConcern in this code path.

### DIFF
--- a/src/mongo/client/dbclient_rs.h
+++ b/src/mongo/client/dbclient_rs.h
@@ -203,6 +203,14 @@ public:
      */
     static void setAuthPooledSecondaryConn(bool setting);
 
+    virtual int getMaxWireVersion() {
+        return checkMaster()->getMaxWireVersion();
+    }
+
+    virtual int getMinWireVersion() {
+        return checkMaster()->getMinWireVersion();
+    }
+
 protected:
     /** Authorize.  Authorizes all nodes as needed
     */
--- a/src/mongo/client/dbclientinterface.h
+++ b/src/mongo/client/dbclientinterface.h
@@ -1463,10 +1463,10 @@ public:
         _maxWireVersion = maxWireVersion;
     }
 
-    int getMinWireVersion() {
+    virtual int getMinWireVersion() {
         return _minWireVersion;
     }
-    int getMaxWireVersion() {
+    virtual int getMaxWireVersion() {
         return _maxWireVersion;
     }
     int getMaxBsonObjectSize() {
--- a/src/mongo/integration/replica_set/basic.cpp
+++ b/src/mongo/integration/replica_set/basic.cpp
@@ -43,6 +43,18 @@ public:
     std::auto_ptr<DBClientReplicaSet> conn;
 };
 
+bool serverGTE(DBClientBase* c, int major, int minor) {
+    BSONObj result;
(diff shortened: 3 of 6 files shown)


src/mongo/integration/replica_set/basic.cpp
```diff
@@ -85,4 +97,22 @@ TEST_F(RSBasicTest, SecondaryQueryIsNotInteruptedByPrimaryFailure) {
     conn->findOne(TEST_NS, Query().readPref(ReadPreference_SecondaryOnly, BSONArray()));
     primary.start();
 }
+
+TEST_F(RSBasicTest, FindAndModifyFailure) {
+    if (serverGTE(conn.get(), 3, 1)) {
+        WriteConcern wcAll = WriteConcern().nodes(5).journal(true);
+        conn->insert(TEST_NS, BSON("_id" << 1 << "i" << 1));
+
+        ASSERT_THROWS(conn->findAndModify(TEST_NS,
```
amidvidy: can you add a comment explaining why we expect this to throw?

markbenvenuto: Done

src/mongo/client/dbclientinterface.h
```diff
@@ -1466,7 +1466,7 @@ class MONGO_CLIENT_API DBClientBase : public DBClientWithCommands, public DBConn
     int getMinWireVersion() {
```
amidvidy: worth doing the same for minWireVersion for symmetry?

markbenvenuto: done

src/mongo/client/dbclient_rs.h
```diff
@@ -203,6 +203,22 @@ class MONGO_CLIENT_API DBClientReplicaSet : public DBClientBase {
      */
     static void setAuthPooledSecondaryConn(bool setting);
 
+    virtual int getMaxWireVersion() {
+        if (!_master.get()) {
+            return checkMaster()->getMaxWireVersion();
```
amidvidy: can we always call checkMaster() ? so the entire body would be `return checkMaster()->getMaxWireVersion();`. If there hasn't been a failover checkMaster() is cheap - if it isn't then we need to call it anyway.

src/mongo/client/dbclient_rs.h
```diff
@@ -203,6 +203,22 @@ class MONGO_CLIENT_API DBClientReplicaSet : public DBClientBase {
...
+        }
+
+        return _master->getMaxWireVersion();
+    }
+
+    virtual int getMinWireVersion() {
+        if (!_master.get()) {
+            return checkMaster()->getMinWireVersion();
```
amidvidy: ditto

src/mongo/client/dbclient_rs.h
markbenvenuto: done

amidvidy:
LGTM

PR #362, opened 2015-10-30
CXX-628: create unreachable macro for bsoncxx

Machyne:


src/bsoncxx/config/prelude.hpp
```diff
@@ -16,3 +16,7 @@
 #include <bsoncxx/export.hpp>
 
 #define BSONCXX_INLINE inline __attribute__ ((__visibility__("hidden"), __always_inline__))
+
+#pragma push_macro("BSONCXX_UNREACHABLE")
+#undef BSONCXX_UNREACHABLE
+#define BSONCXX_UNREACHABLE std::abort();
```
amidvidy: can you take off the semicolon here, and then put it after uses of BSONCXX_UNREACHABLE? I think it looks cleaner that way and we generally prefer to make our macros statement-like

Machyne: Fixed!

amidvidy:
LGTM

Machyne:
@samantharitter - good to merge?

samantharitter:
LGTM

acmorrow:
LGTM

PR #363, opened 2015-10-30
CXX-722: add MaxAwaitTimeMs option to find opts

Machyne:
The changes not immediately related to MaxAwaitTimeMs are to fix bugs that showed up when I added the test. The scope of the enum was changed to prevent ambiguity.


src/mongocxx/options/find.hpp
```diff
@@ -128,6 +129,28 @@ class MONGOCXX_API find {
     const stdx::optional<std::int32_t>& limit() const;
 
     ///
+    /// The maximum amount of time for the server to wait on new documents to satisfy a tailable
+    /// cursor query. This only applies to a TAILABLE_AWAIT cursor. When the cursor is not a
+    /// TAILABLE_AWAIT cursor, this option is ignored. The default is no value.
```
samantharitter: A bit confusing, reword to something like "The default is to wait until the operation completes"

src/mongocxx/options/find.hpp
```diff
@@ -128,6 +129,28 @@ class MONGOCXX_API find {
...
+    ///
+    /// @note On servers < 3.2, this option is ignored.
+    ///
+    /// @param max_await_time_ms
+    ///   The max amount of time (in milliseconds).
```
samantharitter: The max amount of time to wait for new documents

src/mongocxx/options/find.cpp
```diff
@@ -31,10 +31,18 @@ void find::comment(std::string comment) {
     _comment = comment;
 }
 
+void find::cursor_type(enum cursor_type cursor_type) {
```
samantharitter: Can you put all of the new getters and setters that aren't related to MaxAwaitTimeMs into a separate commit?

Machyne: Done!

samantharitter:
Please tag the other commit with either this ticket number or "Minor:"

Otherwise, LGTM once Travis passes!

src/mongocxx/test/options/find.cpp
```diff
@@ -0,0 +1,29 @@
+#include "catch.hpp"
+#include "helpers.hpp"
+
+#include <bsoncxx/document/view.hpp>
+
```
acmorrow: kill this blank line

src/mongocxx/options/find.hpp
```diff
@@ -26,13 +26,14 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
+enum class cursor_type : std::uint8_t { k_non_tailable, k_tailable, k_tailable_await };
```
acmorrow: Why pull this out of the find options?

src/mongocxx/options/find.hpp
```diff
@@ -270,6 +294,7 @@ class MONGOCXX_API find {
     stdx::optional<std::string> _comment;
     stdx::optional<enum cursor_type> _cursor_type;
     stdx::optional<std::int32_t> _limit;
+    stdx::optional<std::int64_t> _max_await_time_ms;
```
acmorrow: I don't see that anything actually uses this new field?

src/mongocxx/options/find.hpp
Machyne: Too many things were named `cursor_type`. At the time, I couldn't get this to work without pulling it out of the same scope as all of the other things named this. I'll try again and let you know the exact issue.

src/mongocxx/options/find.hpp
Machyne: Hmm. Good catch! That is because nothing does.

This will only be used to default the maxTimeMS at the cursor level for getMore commands. I will talk to the C driver team and figure out how they plan to implement this on their end :)

src/mongocxx/options/find.hpp
Machyne: The issue is when trying to get a value (ex: `k_tailable` value). The compiler can't resolve `mongocxx::options::find::cursor_type::k_tailable`.

src/mongocxx/options/find.hpp
Machyne: This depends on [CDRIVER-966](https://jira.mongodb.org/browse/CDRIVER-966).

It will be a straight forward on line change, just need to set the option on the `mongoc_cursor_t` before wrapping it.

src/mongocxx/options/find.hpp
hanumantmk: a liberal sprinkling of typename does the trick:

https://github.com/hanumantmk/mongo-cxx-driver/commit/2719bc0ed6a5462f97182b3d466acfc47cfc9ce9

Machyne: sprinkling added!

Machyne:
The C driver finished this, so it's now good to go here too!

src/mongocxx/options/find.hpp
```diff
@@ -30,7 +30,6 @@ namespace options {
 /// Class representing the optional arguments to a MongoDB query.
 ///
 class MONGOCXX_API find {
-
```
acmorrow: I think this blank line shoudl come back.

src/mongocxx/options/find.hpp
```diff
@@ -128,6 +127,29 @@ class MONGOCXX_API find {
...
+    /// second.
+    ///
+    /// @note On servers < 3.2, this option is ignored.
+    ///
+    /// @param max_await_time_ms
+    ///   The max amount of time (in milliseconds) to wait for new documents.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/operator/meta/maxTimeMS
+    ///
+    void max_await_time_ms(std::int64_t max_await_time_ms);
```
acmorrow: Should the argument here be a std::chrono duration? It seems like it would be much more idiomatic. Do we have other places where we are setting timeouts?

acmorrow:
This LGTM mod a nit and a question. RE the question: not something to fix here, I'd prefer to get this shipped, but if there is agreement that timeouts should be specified with chrono, I'd like to see a ticket for that written before this is pushed so we don't forget about it. I think it would be very much better.

src/mongocxx/options/find.hpp
Machyne: Clang format removes it for me.

PR #364, opened 2015-11-02
CXX-723: make find actually use sort

Machyne:


src/mongocxx/collection.cpp
```diff
@@ -98,11 +98,15 @@ cursor collection::find(bsoncxx::document::view filter, const options::find& opt
     scoped_bson_t filter_bson;
     scoped_bson_t projection(options.projection());
 
-    if (options.modifiers()) {
-        filter_builder << "$query" << bsoncxx::types::b_document{filter}
-                       << builder::stream::concatenate{
-                              options.modifiers().value_or(document::view{})};
-
+    if (options.sort() || options.modifiers()) {
+        filter_builder << "$query" << bsoncxx::types::b_document{filter};
```
samantharitter: Is there a reason we don't want to use $query when we don't have any modifiers?  I'm wondering if we can simplify the logic here.

src/mongocxx/collection.cpp
```diff
@@ -98,11 +98,15 @@ cursor collection::find(bsoncxx::document::view filter, const options::find& opt
...
+        if (options.modifiers()) {
+            filter_builder << builder::stream::concatenate{
+                options.modifiers().value_or(document::view{})};
```
samantharitter: We don't need both `if (options.modifiers()) ...` and `options.modifiers().value_or(...)`, I'd say just use the if statement and concatenate `*options.modifiers()`

src/mongocxx/collection.cpp
```diff
@@ -98,11 +98,15 @@ cursor collection::find(bsoncxx::document::view filter, const options::find& opt
...
+        }
+        if (options.sort()) {
+            filter_builder << "$orderby" << bsoncxx::types::b_document{*options.sort()};
```
samantharitter: this is a nice alternative to
```
builder << "$orderby" << builder::stream::open_document << builder::stream::concatenate{*options.sort()} << builder::stream::close_document;
```
we should try to do this elsewhere.

Machyne: Oh yeah, this is what I wanted in my comment on your most recent PR! I'll add this comment there.

src/mongocxx/collection.cpp
Machyne: No reason, AFAIK. Done!

samantharitter:
LGTM

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -244,7 +244,8 @@ TEST_CASE("Collection", "[collection]") {
             REQUIRE(skip == skip);
             REQUIRE(limit == limit);
             REQUIRE(batch_size == batch_size);
-            REQUIRE(bson_get_data(query) == doc.data());
+            bsoncxx::document::view query_view{bson_get_data(query), query->len};
```
acmorrow: How is this change related to the others?

acmorrow:
LGTM with a question

src/mongocxx/test/collection_mocked.cpp
Machyne: For simplicity, we now send all finds with the `{$query: {}}` form. Previously we would simply send the match document when we didn't have modifiers to the find. This test assumed that we would never send finds with the $query syntax, and would have broken had we ever tested sorts (which this PR implements) or modifiers.

Since we _always_ use the $query form now, this new assumption is safe (and works when testing sorts). I will add a test to the mock as well to ensure this stays functioning.

Machyne:
@acmorrow - I added a test for sorting in the mock. Hopefully that and my explanation answer your question. Good to merge?

acmorrow:
LGTM

PR #365, opened 2015-11-02
Minor: fix cmake download on travis

samantharitter:


.travis.yml
```diff
@@ -25,7 +25,7 @@ before_install:
     - wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
 
     # Get the latest CMake
-    - wget http://www.cmake.org/files/v3.1/${CMAKE_VERSION}.tar.gz
```
Machyne: This has been like this since dc0bc6b4 ... did it stop working? Was it a change by cmake? Just curious.

samantharitter: I haven't finished debugging this yet, changing the url is not a complete fix.  I just need it to keep running on travis as I iterate... there's some kind of certificate issue I'm digging into

samantharitter:
This is ready for review now.  Something about cmake.org's certificate must have changed over the weekend, and now wget can't match the subject alternate name.  Not sure what our travis machines are running, but this is a known wget bug in some linux systems.  Curl handles this fine though!

amidvidy:
LGTM

.travis.yml
```diff
@@ -25,7 +25,7 @@ before_install:
     - wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
```
Machyne: Little late... but would it be worth switching this to curl too for consistency?
(e.g. `curl http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -`)

samantharitter: Yes, sorry, I wanted to unblock all the other branches waiting on travis tests!  We could, I don't feel strongly either way.

Machyne: Probably not worth. :)

PR #366, opened 2015-11-03
CXX-632: de-inline insert_many, create insert many builder

Machyne:


src/mongocxx/insert_many_builder.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <bsoncxx/builder/stream/document.hpp>
+
+#include <mongocxx/collection.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+insert_many_builder::insert_many_builder(const options::insert& options)
+    : _writes{options.ordered().value_or(true)}, _inserted_ids{}, _index{0} {
+    if (options.write_concern()) _writes.write_concern(*options.write_concern());
```
samantharitter: please use { } even though this is a one-line conditional

src/mongocxx/insert_many_builder.cpp
```diff
@@ -0,0 +1,56 @@
...
+};
+
+void insert_many_builder::operator()(const bsoncxx::document::view& doc) {
+    if (!doc["_id"]) {
+        bsoncxx::builder::stream::document new_document;
+        new_document << "_id" << bsoncxx::oid(bsoncxx::oid::init_tag)
+                     << bsoncxx::builder::stream::concatenate{doc};
+
+        _writes.append(model::insert_one{new_document.view()});
```
samantharitter: Are you sure we won't end up with a dangling view here?  new_document will go out-of-scope after this block and the model we append takes a view of it.  There's a TODO in insert_one.hpp that says "should this be a value?" and I suspect the answer might be yes.  Either that, or the bulk builder will have to own its docs somehow.

src/mongocxx/insert_many_builder.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+        _inserted_ids.emplace(_index++, doc["_id"]);
+    }
+};
+
+stdx::optional<result::insert_many> insert_many_builder::insert(collection* col) const {
+    auto val = col->bulk_write(_writes).value();
+    result::bulk_write res{std::move(val)};
+    stdx::optional<result::insert_many> result{
+        result::insert_many{std::move(res), std::move(_inserted_ids)}};
```
samantharitter: You can just do `stdx::optional<result::insert_many> result{{std::move(res), std::move(_inserted_ids)}};` without the result::insert_many bit

src/mongocxx/insert_many_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <mongocxx/result/insert_many.hpp>
+#include <mongocxx/stdx.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+///
+/// Class to build an insert_many bulk write operation.
+///
+/// This is used by mongocxx::collection for mongocxx::collection::insert_many.
```
samantharitter: Do we not want users to use this class directly?  If so let's say that explicitly.

src/mongocxx/insert_many_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    /// @param options
+    ///   The options for this insert_many.
+    ///
+    explicit insert_many_builder(const options::insert& options);
+
+    ///
+    /// Appends a single insert to insert_many bulk write operation. The document is copied into the
+    /// bulk operation completely, so there is no dependency between the life of the document and
+    /// the insert_many bulk write operation itself.
```
samantharitter: This isn't actually true... the model only takes a view of the document.

src/mongocxx/insert_many_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    /// @param doc
+    ///   The next document to be added to the insert many operation.
```
samantharitter: insert_many

src/mongocxx/insert_many_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    void operator()(const bsoncxx::document::view& doc);
+
+    stdx::optional<result::insert_many> insert(collection* col) const;
```
samantharitter: Please document this method as well

samantharitter:
We have to answer some important lifetime questions about views passed into the insert_one models through this builder.  Thoughts?

src/mongocxx/test/collection.cpp
```diff
@@ -42,6 +42,36 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
         REQUIRE(i == 1);
     }
 
+    SECTION("insert and read multiple documents", "[collection]") {
+        bsoncxx::builder::stream::document b1;
+        bsoncxx::builder::stream::document b2;
+        bsoncxx::builder::stream::document b3;
+
+        b1 << "_id" << bsoncxx::oid{bsoncxx::oid::init_tag} << "x" << 1;
```
samantharitter: Can you please add some tests for objects without _id fields?  Then we'll hit the problematic path above in the builder.

src/mongocxx/collection.hpp
```diff
@@ -608,34 +609,9 @@ template <typename document_view_iterator_type>
...
-
-            inserted_ids.emplace(index++, new_document.view()["_id"]);
-        } else {
-            writes.append(model::insert_one(current));
-
-            inserted_ids.emplace(index++, current["_id"]);
-        }
-
-    });
+    insert_many_builder op = std::for_each(begin, end, insert_many_builder{options});
```
acmorrow: This is... subtle. So insert_many_builder has operator(), and accumulates state, and is then moved out of the std::for_each?.

That is kinda amazing. I would not have expected that to work.

My suggestion though would be to capture op by auto:

auto op = ...

src/mongocxx/insert_many_builder.cpp
```diff
@@ -0,0 +1,56 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mongocxx/config/prelude.hpp>
+
+#include <mongocxx/insert_many_builder.hpp>
+
+#include <bsoncxx/builder/stream/document.hpp>
+
```
acmorrow: No blank line between these headers. Order and spacing of headers is:

prelude

self-header <if applicable>

<sorted list of needed mongocxx/bsoncxx headers with no blank lines>

src/mongocxx/insert_many_builder.hpp
```diff
@@ -0,0 +1,65 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/config/prelude.hpp>
+
+#include <bsoncxx/stdx/optional.hpp>
+
```
acmorrow: kill this blank line

src/mongocxx/insert_many_builder.hpp
acmorrow: I think it is harmless if they did, but suggest that they probably want to use the helper in collection.hpp

src/mongocxx/insert_many_builder.cpp
acmorrow: FWIW I've been extremely wary of adding a requirement on a variant type because there is, as yet, no consensus at the standardization level about what semantics a variant type will have. It is definitely a troublesome aspect of working with views and values, but I don't want to make us dependent on MNMLSTC's particular choice of semantics for variants given its lack of forward progress and probably divergence from the standard.

Machyne:
```
    /// Appends a single write to the bulk write operation. The write operation's contents are
    /// copied into the bulk operation completely, so there is no dependency between the life of an
    /// appended write operation and the bulk operation itself.
```
from bulk_write's append. I basically copied this commend for the operator() of the insert many builder. :) This creates a new scoped_bson_t, so it does actually copy the contents over.

src/mongocxx/insert_many_builder.hpp
Machyne: When append is called on the bulk write, it creates a new scoped_bson_t. I am pretty sure that doing that copies all of the contents. It might also be copied again inside libmongoc. We should look into that.

src/mongocxx/collection.hpp
Machyne: Yes! The possible implementation example from cppreference.com:
```
template<class InputIt, class UnaryFunction>
UnaryFunction for_each(InputIt first, InputIt last, UnaryFunction f)
{
    for (; first != last; ++first) {
        f(*first);
    }
    return f;
}
```
So I'm just capturing the return.

src/mongocxx/insert_many_builder.cpp
Machyne: That's so cool!

src/mongocxx/insert_many_builder.cpp
samantharitter: We talked a bit about this in-person on Friday.  I have separate concerns about scoped_bson_t's copying semantics, but for the sake of this review, it's fine!

src/mongocxx/insert_many_builder.hpp
samantharitter: Acknowledged

samantharitter:
LGTM

src/mongocxx/insert_many_builder.cpp
```diff
@@ -0,0 +1,56 @@
...
+        bsoncxx::builder::stream::document new_document;
+        new_document << "_id" << bsoncxx::oid(bsoncxx::oid::init_tag)
+                     << bsoncxx::builder::stream::concatenate{doc};
+
+        _writes.append(model::insert_one{new_document.view()});
+
+        _inserted_ids.emplace(_index++, new_document.view()["_id"]);
+    } else {
+        _writes.append(model::insert_one{doc});
+
```
acmorrow: kill this blank line

acmorrow:
LGTM

PR #367, opened 2015-11-04
CXX-727  Make it possible to rename collections

koraa:


src/mongocxx/collection.hpp
```diff
@@ -477,6 +477,16 @@ class MONGOCXX_API collection {
     stdx::string_view name() const noexcept;
 
     ///
+    /// Rename this collection.
+    ///
+    /// @param new_name The new name to assign to the collection.
+    /// @param overwrite_old Whether to overwrite any
+    ///   existing collections called new_name. Default: false
+    /// @throws exception::operation if the operation fails.
+    void name(stdx::string_view new_name, bool overwrite_old = false);
```
Machyne: Can you rename `overwrite_old` to `drop_target_before_rename` to match the C driver?

src/mongocxx/collection.hpp
```diff
@@ -477,6 +477,16 @@ class MONGOCXX_API collection {
     stdx::string_view name() const noexcept;
 
     ///
+    /// Rename this collection.
+    ///
+    /// @param new_name The new name to assign to the collection.
+    /// @param overwrite_old Whether to overwrite any
+    ///   existing collections called new_name. Default: false
```
Machyne: Add `@see https://docs.mongodb.org/manual/reference/command/renameCollection/`

src/mongocxx/collection.cpp
```diff
@@ -65,6 +65,23 @@ stdx::string_view collection::name() const noexcept {
     return _impl->name;
 }
 
+void collection::name(stdx::string_view new_name, bool overwrite_old) {
+  bson_error_t error;
+
+  _impl->name = (std::string)new_name;
+  auto result = mongoc_collection_rename(
```
Machyne: Please run clang-format on this to follow our style format.

src/mongocxx/database.hpp
```diff
@@ -143,14 +143,6 @@ class MONGOCXX_API database {
     stdx::string_view name() const;
 
     ///
-    /// Renames this database.
-    ///
-    /// @param new_name the new name for the database.
-    /// @param drop_target_before_rename whether to drop existing databases with the new name.
-    ///
-    void rename(stdx::string_view new_name, bool drop_target_before_rename);
```
Machyne: Good catch, thanks!

src/mongocxx/collection.hpp
Machyne: Calling this `rename` would be more clear.

src/mongocxx/collection.hpp
Machyne: Also, please change the last line to "The default is false." for consistency.

Machyne:
Thanks for the PR and the work! :) I'll be happy to merge this in after you fix some small comments.

src/mongocxx/collection.cpp
```diff
@@ -65,6 +65,23 @@ stdx::string_view collection::name() const noexcept {
     return _impl->name;
 }
 
+void collection::name(stdx::string_view new_name, bool overwrite_old) {
+  bson_error_t error;
+
+  _impl->name = (std::string)new_name;
```
amidvidy: move this line to line 83 in case the rename fails. Also, please just write `_impl->name = new_name`. The cast should not be needed as string_view defines a conversion operator to std::string.

amidvidy:
Can you add a unit test to collection_mocked showing that _impl->name changes to the new name if the call to the C-driver succeeds, and that _impl->name does not change if the call to the C-driver fails? @Machyne can help you out with the mocking framework. Thanks again for the PR.

src/mongocxx/collection.cpp
Machyne: Few more comments here: `mongoc_collection_t` doesn't have a name field, I think you meant `->collection`. However, `mongoc_collection_rename` copies the new name to the impl collection field, so you won't need this. (see additional comment below)

src/mongocxx/collection.cpp
```diff
@@ -65,6 +65,23 @@ stdx::string_view collection::name() const noexcept {
...
+  auto result = mongoc_collection_rename(
+      _impl->collection_t
+    , _impl->database_name.c_str()
+    , _impl->name.c_str()
```
Machyne: This should just be `new_name.data()`. :)

src/mongocxx/collection.hpp
koraa: sure, I will change both.
`drop_target_before_rename` hmm. IMO overwrite_old is much more clear.
`rename` Sure about that? I deliberately elected not to call it rename, because this implements a getter/setter pattern. If mongocxx generally uses getters/setters I think it would be more idiomatic to consider the name a gettable and settable property too.

src/mongocxx/collection.cpp
koraa: @amidvidy I initially did not use the cast, but it failed. Further investigation revealed that the string conversion operator in mnmlstc is marked explicit.

https://mnmlstc.github.io/core/string.html#basic_string_view:T:::castto-std::basic_string:T:-operator

koraa:
@amidvidy Is there any particular reason why collection::name() stores the name internally rather than returning a string view on mongoc_collection_get_name()?

If we altered name() to use that we could free ourselves from the added state synchronization and also save on the added test.

src/mongocxx/collection.cpp
koraa: Sure about that?

https://mnmlstc.github.io/core/string.html#basic_string_view:T:::dataCCE

"This pointer is not guaranteed to be null terminated, and should be treated as such."

src/mongocxx/collection.cpp
amidvidy: understood. In that case, please use static_cast, we try to avoid C-style casts.

amidvidy:
good idea @koraa, I do not see why it is needed (some of this code predates me). I think that would be a good change to make. If you do that, we can remove the name member from collection::impl as well.

koraa:
perfect...

src/mongocxx/collection.cpp
Machyne: Interesting... I suggested it because it is used elsewhere (ex: [collection constructor](https://github.com/mongodb/mongo-cxx-driver/blob/master/src/mongocxx/collection.cpp#L70)).
`c_str` is probably safer though.

src/mongocxx/collection.cpp
```diff
@@ -65,6 +65,20 @@ stdx::string_view collection::name() const noexcept {
     return _impl->name;
 }
 
+void collection::rename(stdx::string_view new_name, bool drop_target_before_rename) {
+  bson_error_t error;
+
+  auto result = mongoc_collection_rename(_impl->collection_t, _impl->database_name.c_str(),
```
samantharitter: Hi @koraa , please also add this symbol to the list in `src/mongocxx/private/libmongoc_symbols.hpp` and then call `libmongoc::collection_rename` instead of calling `mongoc_collection_rename` directly.

src/mongocxx/collection.hpp
```diff
@@ -477,6 +477,16 @@ class MONGOCXX_API collection {
     stdx::string_view name() const noexcept;
 
     ///
+    /// Rename this collection.
+    ///
+    /// @param new_name The new name to assign to the collection.
+    /// @param drop_target_before_rename Whether to overwrite any
+    ///   existing collections called new_name. The default is false.
+    /// @throws exception::operation if the operation fails.
+    void name(stdx::string_view new_name, bool drop_target_before_rename = false);
```
samantharitter: Just some style things, please change your doc comment indentation so it reads like:
```
/// @param new_name
///   The new name to assign to the collection.
```
Also, please add an empty `///` line between the `@param`'s and the `@throws`, and one at the end before the function declaration.

samantharitter:
Thanks for your work @koraa !

On the name() vs. rename() debate, I prefer rename().  Yes, we do follow a getter/setter pattern in much of the codebase, but most of those are for attributes that exist in driver classes for convenience.  This is a bit different because it translates directly to a database operation, and since that's called renameCollection() we should use rename() here.

Storing the name and returning it with name() actually makes some sense.  You can't get a valid collection object without knowing its name, and since we have it we can avoid network round trips to the database if we just cache the name.  With renaming, we'll have to update the cached value.

koraa:
Well, then. It's got tests now (let's see it whether the CI successfully runs them because locally they did not work in the first place).

I am also using sring_view::data() now to get the c_string, but strictly speaking this is not correct: the string returned by data() is not guaranteed to be a correctly zero-terminated string. It frequently contains one, when it represents a full std::string or a full c string, but in cases where string_view is most useful – when it references part of a bigger underlying buffer – it wont be.

We could try to solve this by creating a helper function that checks whether string_view::end() references a \0 byte, but this optimization is usually not applicable, because we don't know anything about the character referenced by end(); it could be valid, but it also could be part of a buffer used and written to by another thread or it could be unallocated memory.
Ultimately we need to copy the string.

See https://groups.google.com/a/isocpp.org/forum/#!topic/std-proposals/LEPhIf_msXw

Optimally you would get mongoc/bsonc to support string parameters with sizes.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -466,4 +466,24 @@ TEST_CASE("Collection", "[collection]") {
         REQUIRE(bulk_operation_execute_called);
         REQUIRE(bulk_operation_destroy_called);
     }
+
+    SECTION("Rename Fails, name stays the same", "collection::rename") {
```
Machyne: Can you run clang-format on this too?

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -466,4 +466,24 @@ TEST_CASE("Collection", "[collection]") {
...
+      bool collection_rename_called  = false, collection_rename_error = false;
+      std::string old_name = (std::string)mongo_coll.name();
```
Machyne: Use `std::string old_name{mongo_coll.name()};` instead? That would also avoid the c style cast.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -466,4 +466,24 @@ TEST_CASE("Collection", "[collection]") {
...
+
+      collection_rename->interpose([&](mongoc_collection_t *, const char*, const char*, bool, bson_error_t*) {
+        collection_rename_called = true;
+        return false;
```
Machyne: verify here that the expected name (in this case "bad_name") is getting passed in

Machyne:
Thanks for adding the test!

I agree with your comment on `.data()`; it would be great if libmongoc took lengths instead of require null delimited strings, but for now we'll have to make a copy. I'm glad that you found this!

src/mongocxx/collection.hpp
```diff
@@ -477,6 +477,18 @@ class MONGOCXX_API collection {
     stdx::string_view name() const noexcept;
 
     ///
+    /// Rename this collection.
+    ///
```
Machyne: Old comment got buried, please also add
`/// @see https://docs.mongodb.org/manual/reference/command/renameCollection/`
here.

koraa:
@Machyne Actually I think the correct way to address this is to redesign the API to take strings OR c strings (with the string implementation just extracting the c string). This way copying is only necessary in rare cases and explicit.

Then you should get mongoc to take strings with lengths and then go back to using string_views.

If you just copy unconditionally, you will create lot's of unnecessary overhead.

koraa:
@Machyne As it turns out, `std::string old_name{mongo_coll.name()};` is incorrect. It breaks in clang. Thus I believe the c-style cast is indeed the correct way of invoking an explicit cast operator.

samantharitter:
The reason we discourage C-style casts in the driver is that C++ provides its own casts, and these are safer as they allow the compiler to do more checking.  They also require the programmer to be much more deliberate about the cast and are much easier to search for in code.

However because this isn't a performance-critical path (renaming collections often is kind of an anti-pattern) I think the best solution here is for us to make the extra copy and build a string from the `string_view`.  Doesn't `string_view` provide a `to_string()` method we could use for this?

We don't really want to provide an interface that takes `char*`s.  One thing we've tried very hard to do is to fully separate the user-facing interface from the underlying implementation (the C driver).  Having a `char*` interface leaks that detail in a weird way.  Ideally the C driver would provide an interface taking a `char*` and `len` and we could use that to avoid the copies while still using `string_view`.

Thank you for finding this, we'll have to do an audit of other `string_view` uses in the codebase.  I've opened [CXX-730](https://jira.mongodb.org/browse/CXX-730) to track that work.

koraa:
@samantharitter I am well aware of the c++ casts, but this has nothing to do with runtime polymorphism. This is explicitly a call to the string view's explicit-marked cast operator.

koraa:
By the way, I just looked it up and libmongoc actually does not query the database  in order to retrieve the database name; it is already caching the database name in an internal structure, which kind of makes sense since mongoc_collection_get_name() needs to return a persistent pointer.
Interestingly, since that function just returns a pointer to that internal buffer, all of thosee pointers will automatically reference the new value when it's changed.

https://github.com/mongodb/mongo-c-driver/blob/master/src/mongoc/mongoc-collection-private.h#L33
https://github.com/mongodb/mongo-c-driver/blob/master/src/mongoc/mongoc-collection.c#L1656

samantharitter:
Hi @koraa, sorry for the delay in getting back to you.  Thanks for looking into the libmongoc code, I should not have assumed that it was running a db operation there.

I wasn't aware that string_data had an operator basic_string, that's very handy.  However, since that conversion is still going to copy the underlying string, let's either put a comment explaining the intention there above the call, or use to_string(), which will have the same effect but is in my opinion clearer.

src/mongocxx/private/libmongoc_symbols.hpp
```diff
@@ -44,6 +44,8 @@ MONGOCXX_LIBMONGOC_SYMBOL(client_pool_try_pop)
 MONGOCXX_LIBMONGOC_SYMBOL(client_set_read_prefs)
 MONGOCXX_LIBMONGOC_SYMBOL(client_set_ssl_opts)
 MONGOCXX_LIBMONGOC_SYMBOL(client_set_write_concern)
+MONGOCXX_LIBMONGOC_SYMBOL(collection_get_name)
```
samantharitter: Please sort these new entries into alphabetical order with the other collection_* symbols

src/third_party/catch/include/helpers.hpp
```diff
@@ -103,7 +103,10 @@
     auto collection_count_with_opts = libmongoc::collection_count_with_opts.create_instance();    \
     auto collection_create_index = libmongoc::collection_create_index.create_instance();          \
     auto collection_find = libmongoc::collection_find.create_instance();                          \
-    auto collection_aggregate = libmongoc::collection_aggregate.create_instance();
+    auto collection_aggregate = libmongoc::collection_aggregate.create_instance();                \
+    auto collection_get_name = libmongoc::collection_get_name.create_instance();                  \
+    collection_get_name->interpose([](mongoc_collection_t*){ return "dummy_collection"; });       \
+    auto collection_rename = libmongoc::collection_rename.create_instance();
```
samantharitter: I see that you've added new mocked helpers to test this feature, did you have a chance to write some specific test cases for the operation?  I'd suggest interposing a method on rename() to test it mocked, and I think we should also add a test case for this to our integration tests, in `src/mongocxx/test/collection.cpp`.

koraa: No, the old test code relied on name() to work; since it now relies on a mongoc method I had to mock that.

Personally I don't think we can meaningfully test for rename(); we can check whether the basic interface was used, but the resulting code would be more of a liability than safety.

If we want to test rename() we should do it against a real db.

src/mongocxx/private/libmongoc_symbols.hpp
koraa: (oh, right. Thought I'd done that)

src/third_party/catch/include/helpers.hpp
samantharitter: Right, that's what the integration test in `src/mongocxx/test/collection.cpp` would be for, running the operation against a db.  I think there's still validity to testing both the interface and the end-to-end behavior of the call.

koraa:
Hi @samantharitter,
thanks for getting back to me: The current version of the code does not do any string conversion any more since I entirely removed name() in favor of using the mongoc function; this eradicated the need to store the name somewhere.

Neither did I copy the string_view to fix the problem with the lack of a zero terminator since I thought you'd have some internal discussion about how you're going to deal with that.

src/third_party/catch/include/helpers.hpp
koraa: That would be good indeed. But at this point I am afraid my boss told me not to use too much time on this PR any more.

samantharitter:
I saw that you'd gone back to using .data(), that's totally fine.  Under the umbrella of CXX-730 we're going to go through and fix our uses of string_view, we can address this use there instead.

src/mongocxx/private/collection.hpp
```diff
@@ -38,8 +38,7 @@ class collection::impl {
          const class client::impl* client, stdx::string_view name)
```
Machyne: Since `name` is no longer used here, let's remove it. I think the only place we call this constructor is collection.cpp line 83.

Machyne:
One small comment, otherwise LGTM! Thanks again for the PR :)

PR #368, opened 2015-11-04
CXX-580: add support for reaching into nested bson docs/arrays

Machyne:


src/bsoncxx/array/element.hpp
```diff
@@ -29,44 +29,45 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+    using document::element::get_codewscope;
+    using document::element::get_int32;
+    using document::element::get_timestamp;
+    using document::element::get_int64;
+    using document::element::get_minkey;
+    using document::element::get_maxkey;
+
+    using document::element::get_value;
+
+    using document::element::operator[];
```
Machyne: This line is the only non-clan-format change.

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,106 @@ enum class type : std::uint8_t;
...
+    /// this will throw.
+    ///
+    /// @throws std::runtime_error
+    ///
+    /// @param i
+    ///   The index of the element.
+    ///
+    /// @return The element if it exists, or the invalid element.
+    ///
+    array::element operator[](std::uint32_t i) const;
```
Machyne: Again, lots of clang-format. Happy to remove or keep – I forget what the consensus was for changes until we production clang-format.

Anyway, these two methods are the only real changes in this block.

src/bsoncxx/types.hpp
```diff
@@ -31,36 +31,15 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
-    k_code = 0x0D,
-    k_symbol = 0x0E,
-    k_codewscope = 0x0F,
-    k_int32 = 0x10,
-    k_timestamp = 0x11,
-    k_int64 = 0x12,
-    k_maxkey = 0x7F,
-    k_minkey = 0xFF
+#define BSONCXX_ENUM(name, val) k_##name = val,
+#include <bsoncxx/enums/type.hpp>
```
Machyne: I was nervous about having two copies of the bson hex values. If there's a reason to keep these enums as is, happy to revert.

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,106 @@ enum class type : std::uint8_t;
...
+    ///
+    /// @return The matching element, if found, or the invalid element.
+    ///
+    element operator[](stdx::string_view key) const;
+
+    ///
+    /// If this element is an array, indexes into this BSON array. If the index is out-of-bounds,
+    /// the invalid array::element will be returned. As BSON represents arrays as documents, the
+    /// runtime of operator[] is linear in the length of the array. If this element is not an array,
+    /// this will throw.
```
samantharitter: move the note about when it throws to the @throws bit `@throws std::runtime_error if this element is not an array.`

src/bsoncxx/document/element.hpp
samantharitter: I thought we wanted to gradually clang-format these files?  I say keep it.

src/bsoncxx/test/bson_get_values.cpp
```diff
@@ -0,0 +1,154 @@
...
+              << "bools" << open_array << true << false << close_array;
+
+    auto doc = build_doc.view();
+
+    SECTION("works with one level") {
+        REQUIRE(doc["ints"][0]);
+        REQUIRE(doc["ints"][0].get_int32() == 1);
+
+        REQUIRE(doc["ints"][1]);
+        REQUIRE(doc["ints"][1].get_int32() == 3);
```
samantharitter: it would be so nice if operator== could work without the get_int32() method, maybe some day

src/bsoncxx/test/bson_get_values.cpp
```diff
@@ -0,0 +1,154 @@
...
+        REQUIRE(doc["bools"][1]);
+        REQUIRE(doc["bools"][1].get_bool() == false);
+    }
+
+    SECTION("works with two levels") {
+        REQUIRE(doc["ints"][2][0]);
+        REQUIRE(doc["ints"][2][0].get_int32() == 5);
+    }
+
+    SECTION("throws on non-document") {
```
samantharitter: on non-array

src/bsoncxx/test/bson_get_values.cpp
```diff
@@ -0,0 +1,154 @@
...
+        REQUIRE_THROWS(doc["ints"][0][0]);
+        REQUIRE_THROWS(doc["ints"][2][0][0]);
+        REQUIRE_THROWS(doc["bools"][1][0]);
+    }
+
+    SECTION("returns invalid on not found") {
```
samantharitter: on out-of-bounds?

src/bsoncxx/test/bson_get_values.cpp
```diff
@@ -0,0 +1,154 @@
...
+        REQUIRE(!doc["ints"][9]);
+        REQUIRE(!doc["ints"][2][9]);
+        REQUIRE(!doc["bools"][9]);
+    }
+}
+
+TEST_CASE("[] can reach into nested documents", "[bsoncxx]") {
```
samantharitter: How many levels of nesting can we have?

src/bsoncxx/test/bson_get_values.cpp
```diff
@@ -0,0 +1,154 @@
...
+    SECTION("throws on non-document") {
+        REQUIRE_THROWS(doc["ints"]["x"]["key"]);
+        REQUIRE_THROWS(doc["ints"]["more"]["z"]["key"]);
+        REQUIRE_THROWS(doc["bools"]["f"]["key"]);
+    }
+
+    SECTION("returns invalid on not found") {
+        REQUIRE(!doc["ints"]["badKey"]);
+        REQUIRE(!doc["ints"]["more"]["badKey"]);
+        REQUIRE(!doc["bools"]["badKey"]);
```
samantharitter: Can you add a case for trying to reach into an invalid element?  Like `doc["ints"]["blah"]["blah"]`

src/bsoncxx/types.hpp
samantharitter: @hanumantmk what do you think?

samantharitter:
This is great and so convenient!  Can you please add some examples to `examples/bsoncxx/view_and_value.cpp' ?

src/bsoncxx/test/bson_get_values.cpp
Machyne: MongoDB only supports 100 I think. I can add a stress test though.

src/bsoncxx/test/bson_get_values.cpp
Machyne: Added in the mixed section in order to do invalid array elements also

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,104 @@ enum class type : std::uint8_t;
...
+    types::b_timestamp get_timestamp() const;
+    types::b_int64 get_int64() const;
+    types::b_minkey get_minkey() const;
+    types::b_maxkey get_maxkey() const;
+
+    types::value get_value() const;
+
+    ///
+    /// If this element is a document, finds the first element of the document with the provided
+    /// key. If there is no such element, the invalid document::element will be returned. The
```
acmorrow: 'an invalid document::element'?

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,104 @@ enum class type : std::uint8_t;
...
+    /// @param key
+    ///   The key to search for.
+    ///
+    /// @return The matching element, if found, or the invalid element.
+    ///
+    element operator[](stdx::string_view key) const;
+
+    ///
+    /// If this element is an array, indexes into this BSON array. If the index is out-of-bounds,
+    /// the invalid array::element will be returned. As BSON represents arrays as documents, the
```
acmorrow: ditto

src/bsoncxx/document/element.hpp
acmorrow: Keep it for now. I sort of feel like we should call a moratorium on running clang-format until we are close to done. It just makes reviews harder, right now.

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,104 @@ enum class type : std::uint8_t;
...
+    /// @throws std::runtime_error if this element is not an array.
+    ///
+    /// @param i
+    ///   The index of the element.
+    ///
+    /// @return The element if it exists, or the invalid element.
+    ///
+    array::element operator[](std::uint32_t i) const;
+
+    const std::uint8_t* raw;
```
acmorrow: Wait... these are public? Why?

src/bsoncxx/test/bson_get_values.cpp
acmorrow: +1

src/bsoncxx/test/bson_get_values.cpp
acmorrow: It is very important, while working in the bson library, to be aware of the distinction between BSON the spec, and BSON the document format used by MongoDB. People may want to use our BSON library in contexts that have nothing to do with MongoDB. The BSON library should implement the BSON spec, and be agnostic about database limits. If such limits are to be enforced, they should adhere at mongocxx level, not bsoncxx.

src/bsoncxx/types.hpp
acmorrow: +1, I'd appreciate a clarification from @hanumantmk as well

src/bsoncxx/document/element.hpp
acmorrow: Also, not part of this change as you didn't change it, but lets not forget to look into it?

src/bsoncxx/test/bson_get_values.cpp
Machyne: Good distinction, and totally agreed! I added a test for 2000 levels - seems to work fine :)
Hopefully people do use this for things other than MongoDB, it's a cool format.

Machyne:
Example added!

examples/bsoncxx/getting_values.cpp
```diff
@@ -0,0 +1,74 @@
...
+    if (invalid1 || invalid2) {
+        std::abort(); // Not reached.
+    }
+
+    // The ["key"] can only be used on document views or elements that are of type k_document. This
+    // will throw otherwise.
+    try {
+        doc["_id"]["throws"];
+        std::abort();
+    } catch (std::runtime_error) {
```
acmorrow: catch exceptions by const&

examples/bsoncxx/getting_values.cpp
```diff
@@ -0,0 +1,74 @@
...
+        // Trying to get a value from an int throws.
+    };
+
+    // Similarly, [index] only works on array views or elements that are of type k_array.
+    try {
+        doc["name"][3];
+        std::abort();
+    } catch (std::runtime_error) {
```
acmorrow: ditto

examples/bsoncxx/getting_values.cpp
```diff
@@ -0,0 +1,74 @@
...
+    auto invalid2 = doc["contribs"][1000];
+    if (invalid1 || invalid2) {
+        std::abort(); // Not reached.
+    }
+
+    // The ["key"] can only be used on document views or elements that are of type k_document. This
+    // will throw otherwise.
+    try {
+        doc["_id"]["throws"];
+        std::abort();
```
acmorrow: Use unreachable?

src/bsoncxx/document/element.hpp
```diff
@@ -31,73 +31,104 @@ enum class type : std::uint8_t;
...
+    /// @throws std::runtime_error if this element is not an array.
+    ///
+    /// @param i
+    ///   The index of the element.
+    ///
+    /// @return The element if it exists, or an invalid element.
+    ///
+    array::element operator[](std::uint32_t i) const;
+
+    const std::uint8_t* raw;
```
acmorrow: Why are these public? @hanumantmk?

src/bsoncxx/types.hpp
acmorrow: Ping @hanumantmk

acmorrow:
This LGTM with two questions for @hanumantmk

src/bsoncxx/types.hpp
amidvidy: i think they were that way originally and I changed them to be this way. We wanted to remove the use of X-macros from public headers

samantharitter: Ping @hanumantmk

acmorrow: @amidvidy @hanumantmk Was our aversion to xmacros in public headers based on being wary of having non-hygenic macros used in public headers? I think we have a solution to that in place now with the prelude/postlude headers being populated.

@Machyne - Is the BSONCXX_ENUM macro handled in prelude/postlude?

acmorrow:
@hanumantmk Ping on this, I think it looks good but would like your thoughts on a few of the open questions.

src/bsoncxx/types.hpp
amidvidy: IIRC the reason was to keep the headers as readable as possible. While X-Macros are great for brevity, they are a bit inscrutable to someone unfamiliar with the technique. We decided that X-macros were fine in .cpp files, however. I'm happy to revisit this decision if we think the benefit of using X-Macros in headers outweighs the drawbacks.

hanumantmk: Yep, there was a specific decision to keep xmacros out of headers for readability.

I love xmacros more than the next guy, so I'm in favor of the change.

src/bsoncxx/document/element.hpp
hanumantmk: I think they were public partially out of convenience, and partially because they're exactly the constructor params (so there's not much information hiding going on).

If we want to wrap them up behind accessors, or friend some types in and bury the triplet ctor, that would be a little cleaner

hanumantmk:
LGTM, replies inline

src/bsoncxx/types.hpp
acmorrow: I'm good with it too. Please do add a NOTE so that users are not utterly baffled. I think something like:

Please see the binary_sub_type header for the values that will be used here. The generated enumeration values are of the form k_oid, k_regex, etc.

And do the same NOTE for the other xmacro below.

src/bsoncxx/document/element.hpp
acmorrow: I would like accessors better I think, but lets not change them here, as I'd like to get this pushed. Could you or @Machyne please file a ticket for -rc0 to make them private and add accessors. It seems like it would be straightforward.

Machyne:
updated with comment re: x-macros. PTAL!

src/bsoncxx/config/postlude.hpp
```diff
@@ -42,6 +42,12 @@
 #undef BSONCXX_VERSION_PATCH
 #pragma pop_macro("BSONCXX_VERSION_PATCH")
 
+// src/bsoncxx/types.hpp
+#ifdef BSONCXX_ENUM
+static_assert(false, "BSONCXX_ENUM must be undef'ed");
+#endif
+#pragma push_macro("BSONCXX_ENUM")
```
acmorrow: Don't you mean pop_macro here?

src/bsoncxx/document/element.hpp
acmorrow: @Machyne - Please don't forget to file the ticket for rc0 here.

src/bsoncxx/types.hpp
acmorrow: +1.

acmorrow:
LGTM - with a question about push v pop, a request for a comment as per @hanumantmk comment, and a request for a follow up ticket about making the members private.

No need to repost, feel free to commit after addressing the above.

src/bsoncxx/document/element.hpp
Machyne: https://jira.mongodb.org/browse/CXX-756

src/bsoncxx/config/postlude.hpp
Machyne: Yes :)

PR #370, opened 2015-11-06
CXX-731 Examples must pass uri to client

samantharitter:
A lot of unrelated clang-format changes here as well.

### DIFF
--- a/examples/mongocxx/aggregate.cpp
+++ b/examples/mongocxx/aggregate.cpp
@@ -6,6 +6,7 @@
 #include <mongocxx/client.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/pipeline.hpp>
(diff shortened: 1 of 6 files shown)


examples/mongocxx/aggregate.cpp
```diff
@@ -16,7 +17,7 @@ using bsoncxx::builder::stream::finalize;
 
 int main(int, char**) {
     mongocxx::instance inst{};
-    mongocxx::client conn{};
+    mongocxx::client conn{mongocxx::uri()};
```
Machyne: Do we want to use `mongocxx::uri{}`? Same comment throughout.

Machyne:
Does this fix the segfaults?? :)
One small comment, otherwise, LGTM

PR #371, opened 2015-11-06
CXX-734: 'make examples' will run all examples

Machyne:
Note: cmake must be configured with `-DBUILD_EXAMPLES=true` in order to build the example executables.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -78,7 +78,7 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DBUILD_UNIT_TESTS=true -DBUILD_EXAMPLES=true -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
                   make
                   ./src/mongocxx/test/test_driver
         - func: "stop_mongod"
--- a/.mci.yml
+++ b/.mci.yml
@@ -78,8 +78,9 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install ..
                   make
+                  make run-examples
                   ./src/mongocxx/test/test_driver
         - func: "stop_mongod"
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,10 @@ project(MONGO_CXX_DRIVER LANGUAGES CXX)
 
 cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)
 
+set (CMAKE_SKIP_BUILD_RPATH false)
+set (CMAKE_BUILD_WITH_INSTALL_RPATH false)
+set (CMAKE_INSTALL_RPATH_USE_LINK_PATH true)
+
 # Add in our modules
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/")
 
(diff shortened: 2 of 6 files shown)


samantharitter:
Can you add this target to our evergreen and travis configs so we start running the examples on PRs?  otherwise this LGTM

CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ project(MONGO_CXX_DRIVER LANGUAGES CXX)
 
 cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)
 
+set (CMAKE_SKIP_BUILD_RPATH false)
+set (CMAKE_BUILD_WITH_INSTALL_RPATH false)
+set (CMAKE_INSTALL_RPATH_USE_LINK_PATH true)
+
```
acmorrow: Lets read this and make sure that this is really what we want: https://cmake.org/Wiki/CMake_RPATH_handling

.mci.yml
```diff
@@ -78,8 +78,9 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DBUILD_UNIT_TESTS=true -DBUILD_EXAMPLES=true -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DBUILD_EXAMPLES=true -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
```
acmorrow: Why make building the examples conditional?

examples/CMakeLists.txt
```diff
@@ -9,4 +9,16 @@ add_custom_target(install_libs
 add_subdirectory(bsoncxx)
 add_subdirectory(mongocxx)
 
-add_custom_target(examples DEPENDS ${MONGOCXX_EXAMPLES} ${BSONCXX_EXAMPLES})
+add_custom_target(examples DEPENDS ${MONGOCXX_EXAMPLE_EXECUTABLES} ${BSONCXX_EXAMPLE_EXECUTABLES})
+
+# Run all mongocxx examples on `make examples`.
```
acmorrow: Should we call this target "run-examples" or something? I miss the SCons alias facility when using CMake.

.mci.yml
Machyne: Right now, we need root privileges when making with examples, so I left it as optional.
The change here is just removing an unused definition.

CMakeLists.txt
Machyne: I think this is what we want! Here's what that link has to say about these options:
>With these settings you are able to run the programs directly from the build tree. Once installed, you have to ensure that they find the required shared libraries (e.g. by setting LD_LIBRARY_PATH, or by installing shared libraries into one of the default library directories).

examples/CMakeLists.txt
Machyne: To run the tests, it's `make test`. Maybe `make example` is what we want? I guess "test" is a common verb, while "example" as a verb is less colloquial. Is "test" being used as a verb in "make test"?? How do you read make commands? o.o

.mci.yml
acmorrow: Wait, we do? Why?

amidvidy: I think we do because the examples rely on the driver being installed, and the default install prefix is /usr/local. We could change the default install prefix, but that may be unexpected for users used to a typical `make && sudo make install` workflow. Alternatively we could make the examples not depend on the installed driver, but that would lessen their value as they are our main test that the installation process actually works.

Machyne:
This one should be good to go!

.mci.yml
```diff
@@ -78,8 +78,9 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DBUILD_UNIT_TESTS=true -DBUILD_EXAMPLES=true -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} CMAKE_INSTALL_PREFIX=/data/tmp/cxx-driver-install ..
```
acmorrow: - I don't like this install prefix, as it pollutes the image. Can we make it ./install? As far as I know that will work, since our PWD is working_dir.
- Why remove the -D? I think that is the canonical way to set the install prefix for CMake?

examples/CMakeLists.txt
```diff
@@ -9,4 +9,16 @@ add_custom_target(install_libs
...
+foreach(EXAMPLE ${MONGOCXX_EXAMPLE_EXECUTABLES})
+    get_filename_component(EXAMPLE_EXECUTABLE "${CMAKE_BINARY_DIR}/examples/mongocxx/${EXAMPLE}" ABSOLUTE)
+    add_custom_command(TARGET examples POST_BUILD COMMAND ${EXAMPLE_EXECUTABLE})
```
acmorrow: Lets call the target 'run-examples'. It makes it clear that you are going to run something, not just build it.

.mci.yml
Machyne: - I will change the dir
- the `-D` was gone before this diff... not sure when that happened

examples/CMakeLists.txt
Machyne: Fixed!

acmorrow:
LGTM

PR #373, opened 2015-11-12
CXX-690 Add new options to FAM

Machyne:
This add write concern and bypass document validation to the find and modify helpers.


src/mongocxx/collection.cpp
```diff
@@ -300,29 +300,25 @@ stdx::optional<result::delete_result> collection::delete_one(
...
-    bool r = mongoc_collection_find_and_modify(
-        _impl->collection_t, bson_filter.bson(), bson_sort.bson(), bson_replacement.bson(),
-        bson_projection.bson(), false, options.upsert().value_or(false),
-        rd == options::return_document::k_after, reply.bson(), &error);
+    libmongoc::find_and_modify_opts_destroy(opts);
 
     if (!r) {
-        throw exception::write(std::move(_impl->gle()), std::make_tuple(error.message, error.code));
+        const ::bson_t* gle = libmongoc::collection_get_last_error(collection);
+        libbson::scoped_bson_t error_copy;
```
hanumantmk: I'd fully qualify the calls to scoped_bson_t, or not (I'd lean on the side of not, following the lead for the rest of the file)

src/mongocxx/collection.cpp
```diff
@@ -300,29 +300,25 @@ stdx::optional<result::delete_result> collection::delete_one(
...
+        bson_copy_to(gle, error_copy.bson());
+        throw exception::write(std::move(error_copy.steal()), std::make_tuple(error.message, error.code));
```
hanumantmk: you shouldn't need the std::move, error_copy.steal() is already returning a temporary

src/mongocxx/collection.cpp
```diff
@@ -300,29 +300,25 @@ stdx::optional<result::delete_result> collection::delete_one(
...
-        _impl->collection_t, bson_filter.bson(), bson_sort.bson(), bson_replacement.bson(),
-        bson_projection.bson(), false, options.upsert().value_or(false),
-        rd == options::return_document::k_after, reply.bson(), &error);
+    libmongoc::find_and_modify_opts_destroy(opts);
 
     if (!r) {
-        throw exception::write(std::move(_impl->gle()), std::make_tuple(error.message, error.code));
+        const ::bson_t* gle = libmongoc::collection_get_last_error(collection);
+        libbson::scoped_bson_t error_copy;
+        bson_copy_to(gle, error_copy.bson());
```
hanumantmk: You shouldn't need this whole copy.

Just:

document::view error{bson_get_data(gle), gle->len};

exception::write takes a value, which will implicitly construct from the view and do the copy there

src/mongocxx/collection.cpp
```diff
@@ -335,68 +331,90 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
     return b.extract();
 }
 
+stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
+    bsoncxx::document::view filter, bsoncxx::document::view replacement,
+    const options::find_one_and_replace& options) {
+    auto opts = libmongoc::find_and_modify_opts_new();
+    ::mongoc_find_and_modify_flags_t flags = ::MONGOC_FIND_AND_MODIFY_NONE;
```
hanumantmk: more auto?

src/mongocxx/collection.cpp
```diff
@@ -300,29 +300,25 @@ stdx::optional<result::delete_result> collection::delete_one(
...
 
-    bool r = mongoc_collection_find_and_modify(
-        _impl->collection_t, bson_filter.bson(), bson_sort.bson(), bson_replacement.bson(),
-        bson_projection.bson(), false, options.upsert().value_or(false),
-        rd == options::return_document::k_after, reply.bson(), &error);
+    libmongoc::find_and_modify_opts_destroy(opts);
 
     if (!r) {
-        throw exception::write(std::move(_impl->gle()), std::make_tuple(error.message, error.code));
+        const ::bson_t* gle = libmongoc::collection_get_last_error(collection);
```
hanumantmk: auto gle =

src/mongocxx/options/find_one_and_update.hpp
```diff
@@ -74,11 +88,11 @@ class MONGOCXX_API find_one_and_update {
     const stdx::optional<bsoncxx::document::view>& projection() const;
 
     ///
-    /// Sets the state of the document to be returned by the operation, either the
-    /// original document or its newly-update form.
+    /// Set the desired version of the replaced document to return, either the original
```
hanumantmk: Is it the replaced document for update?

src/mongocxx/options/find_one_and_update.hpp
```diff
@@ -86,9 +100,9 @@ class MONGOCXX_API find_one_and_update {
     void return_document(return_document return_document);
 
     ///
-    /// Gets the state of the document to be returned by the operation.
+    /// Which version of the updated document to return.
```
hanumantmk: upsert changes the version of the document to return?

src/mongocxx/private/collection.hpp
```diff
@@ -45,7 +42,7 @@ class collection::impl {
     ~impl() { libmongoc::collection_destroy(collection_t); }
 
     bsoncxx::document::value gle() {
-        const bson_t* gle = mongoc_collection_get_last_error(collection_t);
+        const bson_t* gle = libmongoc::collection_get_last_error(collection_t);
         libbson::scoped_bson_t error_copy;
         bson_copy_to(gle, error_copy.bson());
         return std::move(error_copy.steal());
```
hanumantmk: You can just

return {bson_get_data(gle), gle->len}

src/mongocxx/test/options/find_one_and_delete.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <bsoncxx/builder/stream/document.hpp>
+#include <mongocxx/options/find_one_and_delete.hpp>
+
+using namespace mongocxx;
+
+TEST_CASE("find_one_and_delete", "[find_one_and_delete][option]") {
+    options::find_one_and_delete opts{};
+
+    std::int64_t ms{400};
+    bsoncxx::builder::stream::document proj{};
```
hanumantmk: I might inline these with finalize and call .view() in CHECK_OPTIONAL_ARGUMENT.

That or add equality operators to compare view's and values to to document::view/value and array::view/value

src/third_party/catch/include/helpers.hpp
```diff
@@ -103,8 +103,61 @@
...
+        bsoncxx::document::view sort_view{bson_get_data(sort), sort->len};                         \
+        REQUIRE(sort_view == expected_find_and_modify_opts_sort);                                  \
+        return true;                                                                               \
+    });                                                                                            \
+    bsoncxx::document::view expected_find_and_modify_opts_update;                                  \
+    auto find_and_modify_opts_set_update =                                                         \
+        libmongoc::find_and_modify_opts_set_update.create_instance();                              \
+    find_and_modify_opts_set_update->interpose([&expected_find_and_modify_opts_update](            \
+        mongoc_find_and_modify_opts_t*, const ::bson_t* update) {                                  \
+        bsoncxx::document::view update_view{bson_get_data(update), update->len};                   \
```
hanumantmk: bleh, we should just add a helper to get a view/value from a bson_t

src/mongocxx/options/find_one_and_update.hpp
Machyne: What do you mean by this?

Machyne:
Updated and also rebased. Ready for more review :)

src/bsoncxx/private/helpers.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <bsoncxx/document/value.hpp>
+#include <bsoncxx/document/view.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+
+namespace helpers {
+
+BSONCXX_INLINE document::view view_from_bson_t(const bson_t* bson) {
```
acmorrow: You don't need BSONCXX_INLINE on private inline functions.

acmorrow: Even more, you shouldn't apply it as it forces the functions inline, even if the compiler doesn't really want to. We want that for public inlines for ABI reasons, but not for private inlines where we want the compiler to decide as appropriate based on its inlining heuristics.

src/mongocxx/collection.cpp
```diff
@@ -327,30 +327,24 @@ stdx::optional<result::delete_result> collection::delete_one(
     return stdx::optional<result::delete_result>(result::delete_result(std::move(result.value())));
 }
 
-stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
-    bsoncxx::document::view filter, bsoncxx::document::view replacement,
-    const options::find_one_and_replace& options) {
+stdx::optional<bsoncxx::document::value> _find_and_modify(::mongoc_collection_t* collection,
```
acmorrow: If this is a free file-scoped function, it should go in an unnamed namespace and probably at the top of the file. You should also drop the _ from the beginning. Global names beginning with _[a-z] are reserved. You aren't quite global here because you are in the mongocxx namespace, but you are too close for comfort.

src/mongocxx/collection.cpp
```diff
@@ -327,30 +327,24 @@ stdx::optional<result::delete_result> collection::delete_one(
...
-        options.return_document().value_or(options::return_document::k_before);
+    bool r = libmongoc::collection_find_and_modify_with_opts(collection, bson_filter.bson(), opts,
+                                                             reply.bson(), &error);
 
-    // TODO handle bypass document validation when CDRIVER-945 is done.
-    bool r = mongoc_collection_find_and_modify(
-        _impl->collection_t, bson_filter.bson(), bson_sort.bson(), bson_replacement.bson(),
-        bson_projection.bson(), false, options.upsert().value_or(false),
-        rd == options::return_document::k_after, reply.bson(), &error);
+    libmongoc::find_and_modify_opts_destroy(opts);
```
acmorrow: I don't really like that this function takes a raw pointer to an opts allocated elsewhere and then destroys it. I think I'd prefer that callers manage the lifetime of their own objects. I think you should just have this function take a const ref to the opts_t. Make the callers manage the lifetime.

src/mongocxx/collection.cpp
```diff
@@ -363,69 +357,88 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
     return b.extract();
 }
 
+stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
+    bsoncxx::document::view filter, bsoncxx::document::view replacement,
+    const options::find_one_and_replace& options) {
+    auto opts = libmongoc::find_and_modify_opts_new();
```
acmorrow: capture the result of find_and_modify_opts_new into a unique_ptr where the deleter is ibmongoc::find_and_modify_opts_destroy

src/mongocxx/collection.cpp
```diff
@@ -363,69 +357,88 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
...
+    scoped_bson_t bson_replacement{replacement};
+    scoped_bson_t bson_sort{options.sort()};
+    scoped_bson_t bson_projection{options.projection()};
+
+    libmongoc::find_and_modify_opts_set_update(opts, bson_replacement.bson());
+    if (options.bypass_document_validation()) {
+        libmongoc::find_and_modify_opts_set_bypass_document_validation(
+            opts, *options.bypass_document_validation());
+    }
+    if (options.sort()) {
```
acmorrow: I'd prefer to see a blank line before each of these if blocks.

src/mongocxx/collection.cpp
```diff
@@ -363,69 +357,88 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
...
+        libmongoc::find_and_modify_opts_set_sort(opts, bson_sort.bson());
+    }
+    if (options.projection()) {
+        libmongoc::find_and_modify_opts_set_fields(opts, bson_projection.bson());
+    }
+    if (options.upsert().value_or(false)) {
+        flags |= ::MONGOC_FIND_AND_MODIFY_UPSERT;
+    }
+    auto rd = options.return_document().value_or(options::return_document::k_before);
```
acmorrow: blank line before auto rd

src/mongocxx/collection.cpp
```diff
@@ -363,69 +357,88 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
...
+                                              static_cast<::mongoc_find_and_modify_flags_t>(flags));
+
+    return _find_and_modify(_impl->collection_t, filter, opts);
+}
+
 stdx::optional<bsoncxx::document::value> collection::find_one_and_update(
     bsoncxx::document::view filter, bsoncxx::document::view update,
     const options::find_one_and_update& options) {
-    scoped_bson_t bson_filter{filter};
+    auto opts = libmongoc::find_and_modify_opts_new();
```
acmorrow: ditto re lifetime

src/mongocxx/options/find_one_and_delete.cpp
```diff
@@ -20,6 +20,10 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
+void find_one_and_delete::max_time_ms(std::int64_t max_time_ms) {
```
acmorrow: should this be signed? what does the crud spec say if anything?

src/mongocxx/options/find_one_and_replace.cpp
```diff
@@ -20,6 +20,14 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
+void find_one_and_replace::bypass_document_validation(bool bypass_document_validation) {
+    _bypass_document_validation = bypass_document_validation;
+}
+
+void find_one_and_replace::max_time_ms(std::int64_t max_time_ms) {
```
acmorrow: ditto re signed/unsigned

src/mongocxx/options/find_one_and_replace.cpp
```diff
@@ -20,6 +20,14 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
+void find_one_and_replace::bypass_document_validation(bool bypass_document_validation) {
```
acmorrow: is this just spillover from another change?

acmorrow: nm, I see it is not.

src/mongocxx/options/find_one_and_replace.hpp
```diff
@@ -139,33 +156,13 @@ class MONGOCXX_API find_one_and_replace {
...
-    ///   Whether or not to bypass document validation
-    ///
-    void bypass_document_validation(bool bypass_document_validation);
-
-    ///
-    /// Gets the current value of the bypass_document_validation option.
-    ///
-    /// @return The optional value of the bypass_document_validation option.
-    ///
-    const stdx::optional<bool>& bypass_document_validation() const;
```
acmorrow: why did these move?

src/mongocxx/options/find_one_and_update.hpp
```diff
@@ -138,33 +156,13 @@ class MONGOCXX_API find_one_and_update {
     ///
     const stdx::optional<bool>& upsert() const;
 
-    ///
-    /// Sets the bypass_document_validation option.
```
acmorrow: why did these move?

src/mongocxx/options/find_one_and_delete.cpp
Machyne: The CRUD spec says:
> `maxAwaitTimeMS: Optional<Int64>;`

src/mongocxx/options/find_one_and_replace.hpp
Machyne: These moved because I had forgotten that I had added them in the agg/bulk opps bypass doc validation pr, then I added them again here in the proper alphabetical order, then I rebased...

I can put them back if you'd prefer? I like them better arranged by name though, feels more consistent.

src/mongocxx/options/find_one_and_update.hpp
Machyne: (see above)

Machyne:
Updated with `make_guard`, spaces, and `inline`!

src/mongocxx/collection.cpp
```diff
@@ -330,102 +380,102 @@ stdx::optional<result::delete_result> collection::delete_one(
...
     scoped_bson_t bson_update{update};
     scoped_bson_t bson_sort{options.sort()};
     scoped_bson_t bson_projection{options.projection()};
 
-    scoped_bson_t reply;
-    reply.flag_init();
+    libmongoc::find_and_modify_opts_set_update(opts, bson_update.bson());
 
-    bson_error_t error;
+    if (options.bypass_document_validation()) {
```
samantharitter: is there a way to do these common setup things somewhere else, so we don't repeat the code?

src/mongocxx/collection.cpp
```diff
@@ -330,102 +380,102 @@ stdx::optional<result::delete_result> collection::delete_one(
...
 
-    if (!r) {
-        throw exception::write(std::move(_impl->gle()), std::make_tuple(error.message, error.code));
+    if (options.upsert().value_or(false)) {
+        flags |= ::MONGOC_FIND_AND_MODIFY_UPSERT;
     }
 
-    bsoncxx::document::view result = reply.view();
+    auto rd = options.return_document().value_or(options::return_document::k_before);
+    if (rd == options::return_document::k_after) {
```
samantharitter: I think you can simplify these lines to `if(options.return_document() == options::return_document::k_after)` if you want.

src/mongocxx/collection.cpp
Machyne: `options` in all three cases are different classes... We could make them super classes of some base class? I don't know what the best way to do it is. Maybe template it?
Suggestions? Is it worth doing?

src/mongocxx/collection.cpp
Machyne: Didn't know optionals could compare to their underlying values. That's awesome!

src/bsoncxx/private/helpers.hpp
```diff
@@ -0,0 +1,41 @@
+// Copyright 2014 MongoDB Inc.
```
acmorrow: 2015?

src/mongocxx/collection.cpp
```diff
@@ -47,6 +47,56 @@
...
+
+    bsoncxx::builder::stream::document b{};
+    b << bsoncxx::builder::stream::concatenate{result["value"].get_document()};
+    return b.extract();
+}
+
+template <typename T>
+class guard {
+   public:
+    guard(T&& t) : _t{t} {
```
acmorrow: std::move

src/mongocxx/collection.cpp
```diff
@@ -47,6 +47,56 @@
...
+    }
+    ~guard() {
```
acmorrow: blank line between methods

src/mongocxx/collection.cpp
```diff
@@ -47,6 +47,56 @@
...
+
+    if (result["value"].type() == bsoncxx::type::k_null)
+        return mongocxx::stdx::optional<bsoncxx::document::value>{};
+
+    bsoncxx::builder::stream::document b{};
+    b << bsoncxx::builder::stream::concatenate{result["value"].get_document()};
+    return b.extract();
+}
+
+template <typename T>
```
acmorrow: Add a todo to move this to a private header. No need to do it now, but we will want this again in another file someday.

src/mongocxx/collection.cpp
acmorrow: A template local to this file in the unnamed namespace sounds like the least bad way to go about it. Maybe try it and make a call on whether it seems better. Overall, I agree with Samantha on keeping things DRY.

src/mongocxx/options/find_one_and_delete.cpp
acmorrow: This is another place I think we should consider making the parameter type a std::chrono, and dropping the 'ms' from the method name. As before, not to be changed here, but lets write up the ticket.

src/mongocxx/options/find_one_and_replace.hpp
acmorrow: Moving it is fine, no need to revert.

acmorrow:
LGTM mod nits

src/mongocxx/collection.cpp
amidvidy: i think you want std::forward here.

acmorrow: That was my original inclination as well, but I don't think you actually do. I think you would only want that if the constructor was declared

`
template <typename U>
guard(U&& u) : _t(std::forward<U>(u)) {}
`

The fact that the class itself is a template doesn't mean the non-template constructor is a case for needing universal references.

I don't have my copy of Effective Modern handy, but I seem to remember he mentions this exact case.

amidvidy: ah yeah, that makes sense. Good point.

src/mongocxx/collection.cpp
Machyne: I tried with a template, and I think it's cleaner as is. Maybe this can be reexamined later.

PR #376, opened 2015-11-17
CXX-727 Make it possible to rename collections

Machyne:
This is the work from @koraa plus one additional mock test.

(this replaces #367)


src/mongocxx/private/libmongoc_symbols.hpp
```diff
@@ -53,6 +53,8 @@ MONGOCXX_LIBMONGOC_SYMBOL(collection_drop)
 MONGOCXX_LIBMONGOC_SYMBOL(collection_find)
 MONGOCXX_LIBMONGOC_SYMBOL(collection_find_and_modify)
 MONGOCXX_LIBMONGOC_SYMBOL(collection_find_indexes)
+MONGOCXX_LIBMONGOC_SYMBOL(collection_get_name)
+MONGOCXX_LIBMONGOC_SYMBOL(collection_rename)
```
samantharitter: please sort this in alphabetically

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -55,6 +56,31 @@ TEST_CASE("Collection", "[collection]") {
     collection mongo_coll = mongo_db[collection_name];
     REQUIRE(mongo_coll);
 
+    SECTION("Collection Rename", "[collection]") {
```
samantharitter: As long as you're adding this test, could you also add an integration test for rename?

Machyne: added!

samantharitter:
@koraa thanks again for all of your work on this!  LGTM

Machyne:
Thanks again for the work @koraa ! This closes #367.

src/mongocxx/collection.hpp
```diff
@@ -466,6 +466,20 @@ class MONGOCXX_API collection {
...
+    ///
+    /// @param new_name The new name to assign to the collection.
+    /// @param drop_target_before_rename Whether to overwrite any
+    ///   existing collections called new_name. The default is false.
+    ///
+    /// @throws exception::operation if the operation fails.
+    ///
+    /// @see https://docs.mongodb.org/manual/reference/command/renameCollection/
+    ///
+    void rename(stdx::string_view new_name, bool drop_target_before_rename = false);
```
acmorrow: I know this predates your changes (and that this was already merged), but I really dislike bool parameters of this form. They make code that uses the API inscrutable.

Is this signature mandated by the CRUD specification? If not, perhaps we would do better with either an enum or two methods, though I'm having difficulty with what the method name should be fore the compound operation.

koraa: I agree. I think it's terrible and I just included it because I thought you'd ask for consistency with the C api.

IMO programmers should just drop the target manually…

acmorrow: Yes. Any method that needs an "and" to describe it is not a method. It frustrates error handling as well: which part failed? However, we do have a spec for the CRUD API, so if it is mandated we should try to adhere to the spec. If the specification does mandate that this method take an argument to describe the drop target behavior, I'd at least prefer that we make it an enum so you can understand it when seen in client code without needing to look up the driver headers.

PR #377, opened 2015-11-19
CXX-691 Find options take a hint

samantharitter:


src/mongocxx/collection.cpp
```diff
@@ -125,6 +124,10 @@ cursor collection::find(bsoncxx::document::view filter, const options::find& opt
         filter_builder << "$orderby" << bsoncxx::types::b_document{*options.sort()};
     }
 
+    if (options.hint()) {
+        filter_builder << builder::stream::concatenate{options.hint()->to_document()};
```
samantharitter: This pattern that we use throughout the codebase is bad, because to_document() returns a temporary value that concatenate{} converts to a view and drops on the floor.  View_or_value will fix it!

src/mongocxx/options/find.hpp
```diff
@@ -113,6 +113,21 @@ class MONGOCXX_API find {
     const stdx::optional<enum cursor_type>& cursor_type() const;
 
     ///
+    /// Sets the index to use for this operation.
+    ///
+    /// @param index_hint
+    ///   Object representing the index to use.
+    ///
```
Machyne: Add `@see https://docs.mongodb.org/manual/reference/operator/meta/hint/`.
Should we add "When an index filter exists for the query shape, MongoDB ignores the $hint." as a note?

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -297,22 +297,31 @@ TEST_CASE("Collection", "[collection]") {
...
             REQUIRE(coll == mongo_coll.implementation());
             REQUIRE(flags == MONGOC_QUERY_NONE);
             REQUIRE(skip == skip);
             REQUIRE(limit == limit);
             REQUIRE(batch_size == batch_size);
+
             bsoncxx::document::view query_view{bson_get_data(query), query->len};
-            REQUIRE(query_view["$query"].get_document() == doc);
+
+            REQUIRE(bsoncxx::to_json(query_view["$query"].get_document()) == bsoncxx::to_json(doc));
```
Machyne: Why was this changed? Did it not work before?

src/mongocxx/options/find.hpp
acmorrow: Yeah, I like the comment suggestion.

src/mongocxx/options/find.hpp
```diff
@@ -30,7 +31,6 @@ namespace options {
 /// Class representing the optional arguments to a MongoDB query.
 ///
 class MONGOCXX_API find {
-
```
acmorrow: Add this back please - we try to add a blank line between the class open and the leading public.

acmorrow:
LGTM

samantharitter:
@acmorrow @tessavitabile this is ready for re-review, it has been updated to use view_or_value.

tessavitabile:
LGTM

src/bsoncxx/view_or_value.hpp
```diff
@@ -123,6 +123,30 @@ class view_or_value {
     View _view;
 };
 
+///
+/// Equality operators for comparison with plain Views.
+///
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator==(view_or_value<View, Value> lhs, View rhs) {
```
acmorrow: Given that view_or_value may transfer ownership, I'm very certain you want to say `const view_or_value<View, Value>&` here. For the `View` type argument, I don't think it matters.

samantharitter:
@acmorrow I saw your LGTM from November 28 and mistakenly merged but I'll add any changes to a new commit!

acmorrow:
I think that is the only one. I maybe missed it last time.

samantharitter:
Nope, you didn't miss it, that change is new

src/bsoncxx/view_or_value.hpp
samantharitter: That makes sense.  The same should be true for the operator== below where view_or_value is on the rhs, I think.

acmorrow: Right. All of them need it.

PR #378, opened 2015-11-19
CXX-695 Support for Read Concern

Machyne:


Machyne:
This doesn't have tests for read concern yet. I wanted to put it up just in case. I'll hopefully add tests tomorrow.

src/mongocxx/client.cpp
```diff
@@ -55,6 +55,16 @@ void* client::implementation() const {
     return _impl->client_t;
 }
 
+void client::read_concern(class read_concern rc) {
+    libmongoc::client_set_read_concern(_impl->client_t, rc._impl->read_concern_t);
+}
+
+class read_concern client::read_concern() const {
+    auto rc = libmongoc::client_get_read_concern(_impl->client_t);
+    return (class read_concern){
```
samantharitter: Do we want to cache the read concern to avoid this copying dance?  I'm leaning towards what you have here, so we can avoid hanging more things on the client object, but just thought I'd raise the question.

src/mongocxx/client.hpp
```diff
@@ -107,6 +108,27 @@ class MONGOCXX_API client {
     MONGOCXX_DEPRECATED void* implementation() const;
 
     ///
+    /// Sets the read concern for this client.
+    ///
+    /// Modifications at this level do not effect existing databases instances that have have been
+    /// created by this client but do effect new ones as databases inherit the @c read_concern
```
samantharitter: affect

src/mongocxx/client.hpp
```diff
@@ -107,6 +108,27 @@ class MONGOCXX_API client {
...
+    /// TODO add link to docs once they exist.
+    ///
+    void read_concern(class read_concern rc);
+
+    ///
+    /// Returns the current read concern for this client.
+    ///
+    /// @return The current @c read_concern
+    ///
+    class read_concern read_concern() const;
```
samantharitter: If the user has not set a read concern for this client, is there a default value that is returned?

src/mongocxx/collection.cpp
```diff
@@ -38,6 +38,7 @@
 #include <mongocxx/private/libmongoc.hpp>
 #include <mongocxx/private/pipeline.hpp>
 #include <mongocxx/private/read_preference.hpp>
+#include <mongocxx/private/read_concern.hpp>
```
samantharitter: nit, read_concern.hpp < read_preference.hpp

src/mongocxx/collection.hpp
```diff
@@ -478,6 +479,23 @@ class MONGOCXX_API collection {
     ///
     void rename(stdx::string_view new_name, bool drop_target_before_rename = false);
 
+    ///
+    /// Sets the read_concern for this collection. Changes will not have any effect on existing
+    /// cursors or other read operations which use the read concern.
```
samantharitter: which use the previously-set read concern?

src/mongocxx/collection.hpp
```diff
@@ -478,6 +479,23 @@ class MONGOCXX_API collection {
...
+    ///
+    /// @param rc
+    ///   The new @c read_concern
+    ///
+    /// TODO add link to docs once they exist.
+    ///
+    void read_concern(class read_concern rc);
```
samantharitter: same question as above about default behavior.

src/mongocxx/collection.hpp
```diff
@@ -478,6 +479,23 @@ class MONGOCXX_API collection {
...
+
+    ///
+    /// Gets the read_concern for the collection.
+    ///
+    /// @return The current read_concern.
+    ///
+    class read_concern read_concern() const;
```
samantharitter: Same question as above about default behavior.

src/mongocxx/read_concern.hpp
```diff
@@ -0,0 +1,138 @@
...
+    /// @TODO link to the docs when they exist.
+    ///
+    enum class level {
+        k_local,
+        k_majority,
+        k_unknown,
+    };
+
+    ///
+    /// Constructs a new read_concern.
```
samantharitter: Does this read concern default to k_local?

src/mongocxx/read_concern.hpp
```diff
@@ -0,0 +1,138 @@
...
+    ///
+    ~read_concern();
+
+    ///
+    /// Sets the read concern level.
+    ///
+    /// @param rc_level
+    ///   Either k_local or k_majority.
+    ///
+    /// @throws std::invalid_argument if rc_level is not k_local or k_majority.
```
samantharitter: Our first mongocxx-generated exception??

src/mongocxx/client.cpp
Machyne: ^ same question for `write_concern` and `read_preference`.

The good thing about not cacheing is that we can return an owned value to the caller. If we cached it, I think  it would have to be a view.

src/mongocxx/read_concern.hpp
```diff
@@ -0,0 +1,138 @@
...
+    ///
+    stdx::optional<level> acknowledge_level() const;
+
+    ///
+    /// Sets the read concern string.
+    ///
+    /// @param rc_string
+    ///   The read concern string.
+    ///
+    void acknowledge_string(stdx::string_view rc_string);
```
samantharitter: When would you want to use this method over using acknowledge_level()?

src/mongocxx/read_concern.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    ///
+    /// Gets the current read concern string. If the read concern level was set with
+    /// acknowledge_level, this will return either "local" or "majority".
+    ///
+    /// @return The read concern string.
+    ///
+    stdx::optional<stdx::string_view> acknowledge_string() const;
```
samantharitter: Same questions

src/mongocxx/client.hpp
Machyne: Gah!!! It was copy paste, I swear! This is one of my biggest pet peeves. Good catch :)

src/mongocxx/client.hpp
Machyne: Yes! libmongoc will default the read concern on uri, clients, dbs, and collections. The default is:
```
read_concern = (mongoc_read_concern_t *)bson_malloc0 (sizeof *read_concern);
```
so it will be empty string. Should I add a `k_empty` to the enum? It might be nice.

src/mongocxx/collection.cpp
Machyne: :)

src/mongocxx/read_concern.hpp
Machyne: Sadly no. I copied this idea `write_concern`.

src/mongocxx/read_concern.cpp
```diff
@@ -0,0 +1,92 @@
...
+        case read_concern::level::k_local:
+            libmongoc::read_concern_set_level(_impl->read_concern_t,
+                                              MONGOC_READ_CONCERN_LEVEL_LOCAL);
+            break;
+        case read_concern::level::k_majority:
+            libmongoc::read_concern_set_level(_impl->read_concern_t,
+                                              MONGOC_READ_CONCERN_LEVEL_MAJORITY);
+            break;
+        default:
+            // TODO throw a mongocxx exception
```
samantharitter: Let's make a ticket for this.  Or are you going to lump this into the existing error-related code review?

src/mongocxx/test/read_concern.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <bsoncxx/stdx/string_view.hpp>
+#include <mongocxx/read_concern.hpp>
+#include <mongocxx/stdx.hpp>
+
+using namespace mongocxx;
+
+TEST_CASE("a default read_concern", "[read_concern]") {
+    read_concern rc{};
+
+    SECTION("has an unset level") {
```
samantharitter: Isn't it set, just to k_unknown?

src/third_party/catch/include/helpers.hpp
```diff
@@ -19,7 +19,7 @@
         REQUIRE(!OBJECT.NAME());                     \
     }                                                \
                                                      \
-    SECTION("has a method to set the upsert") {      \
+    SECTION("has a method to set the NAME") {        \
```
samantharitter: nice catch

src/mongocxx/read_concern.hpp
Machyne: It's in the spec for future compatibility. It's also implemented in the C Driver as a string.

src/mongocxx/read_concern.cpp
Machyne: I was gonna lump it in to the "all exceptions should inherit from the same thing" ticket... want me to make a new one? I can at least comment on that ticket.

Machyne: ([CXX-627](https://jira.mongodb.org/browse/CXX-627))

src/third_party/catch/include/helpers.hpp
Machyne: I was really confused when my broken read_concern test caused the "upsert" test failed :)

Machyne:
Mock tests added! Ready for full review.
One question is the use of `#MONGOC_I_AM_A_DRIVER` in the mock. My thought is that it makes testing easier at little cost, so I'm in favor of doing this more going forward.

src/mongocxx/client.cpp
hanumantmk: I think the copy's fine.  The point of these things is that you set the read/write concern at some level and get re-use that way, not that the objects are magically cheap

src/mongocxx/read_concern.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <mongocxx/private/libmongoc.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+read_concern::read_concern() : _impl{stdx::make_unique<impl>(libmongoc::read_concern_new())} {
+}
+
+read_concern::read_concern(std::unique_ptr<impl>&& implementation) {
+    _impl.reset(implementation.release());
```
acmorrow: does

`
_impl = std::move(implementation)
`

work? That seems more idiomatic.

src/mongocxx/read_concern.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+read_concern::read_concern(read_concern&&) noexcept = default;
+read_concern& read_concern::operator=(read_concern&&) noexcept = default;
+
+read_concern::read_concern(const read_concern& other)
+    : _impl(stdx::make_unique<impl>(libmongoc::read_concern_copy(other._impl->read_concern_t))) {
+}
+
+read_concern& read_concern::operator=(const read_concern& other) {
+    _impl.reset(stdx::make_unique<impl>(libmongoc::read_concern_copy(other._impl->read_concern_t))
```
acmorrow: same question about just using unique_ptr assignment here

src/mongocxx/read_concern.cpp
```diff
@@ -0,0 +1,89 @@
...
+            break;
+        default:
+            // TODO throw a mongocxx exception
+            throw std::invalid_argument(
+                "acknowledge_level can only be used to set levels of k_local or k_majority.");
+    }
+}
+
+void read_concern::acknowledge_string(stdx::string_view rc_string) {
+    libmongoc::read_concern_set_level(_impl->read_concern_t, rc_string.data());
```
acmorrow: Does set_level need a null termianted string?

src/mongocxx/read_concern.cpp
```diff
@@ -0,0 +1,89 @@
...
+        return read_concern::level::k_local;
+    } else if (strcmp(MONGOC_READ_CONCERN_LEVEL_MAJORITY, level) == 0) {
+        return read_concern::level::k_majority;
+    } else {
+        return read_concern::level::k_unknown;
+    }
+}
+
+stdx::string_view read_concern::acknowledge_string() const {
+    auto level = libmongoc::read_concern_get_level(_impl->read_concern_t);
```
acmorrow: What is the lifetime of the result of read_concern_get_level?

src/mongocxx/test/client.cpp
```diff
@@ -15,6 +15,10 @@
 #include "catch.hpp"
 #include "helpers.hpp"
 
+#define MONGOC_I_AM_A_DRIVER true
```
acmorrow: This makes me a little sad. It will mean that packagers won't be able to build the tests against the system libmongoc unless there is also an unstable -priv package. How onerous would it be to do without?

src/mongocxx/uri.cpp
```diff
@@ -65,29 +68,48 @@ void* uri::implementation() const {
     return _impl->uri_t;
 }
 
+bsoncxx::document::view uri::options() const {
+    auto opts_bson = libmongoc::uri_get_options(_impl->uri_t);
```
acmorrow: what is the lifetime of the returned information?

src/mongocxx/client.hpp
acmorrow: Or should the return of this be optional<read_concern>?

src/mongocxx/collection.hpp
```diff
@@ -478,9 +479,26 @@ class MONGOCXX_API collection {
...
 
     ///
-    /// Sets the read_preference for this collection. Changes will not have any effect on existing
+    /// Sets the read_preference for this collection. Changes will not have any affect on existing
```
acmorrow: I disagree

src/mongocxx/read_concern.hpp
```diff
@@ -0,0 +1,141 @@
...
+///
+class MONGOCXX_API read_concern {
+
+   public:
+    ///
+    /// A class to represent the read concern level.
+    /// @TODO link to the docs when they exist.
+    ///
+    enum class level {
+        k_empty,
```
acmorrow: would k_empty still have any meaning if you used optional in the API as suggested above?

src/mongocxx/client.cpp
amidvidy: the syntax for class-qualifying a type to call it's constructor is a bit confusing here as a cursory glance would lead one to believe you are deploying the dreaded c-style cast. Would this compile if we just return the initializer list? If not can you leave a comment explaining what is actually going on?

src/mongocxx/client.hpp
amidvidy: +1 for optional

src/mongocxx/client.cpp
amidvidy: https://jira.mongodb.org/browse/CXX-752

src/mongocxx/read_concern.cpp
Machyne: Yes, it does. Hopefully string_view_or_value fixes this along with collection renaming (etc)? Should I leave this for now?

src/mongocxx/collection.hpp
Machyne: find and replace, at it again... :S
I think it would be better as a verb here anyway, but no sense making the change just to rephrase a comment

src/mongocxx/read_concern.hpp
Machyne: Not that I can think of. Optional it is!

src/mongocxx/read_concern.cpp
Machyne: The lifetime is the same as the read_concern object.

src/mongocxx/uri.cpp
Machyne: The same as the uri object.

Machyne:
Updated! Ready to go? :)

src/mongocxx/client.hpp
acmorrow: @Machyne Ping? This doesn't seem to be done, though you had a comment that indicated that it was. Do you need to push a commit?

Machyne: Yes! Pushed :)

src/mongocxx/read_concern.cpp
acmorrow: Yes, leave it.

src/mongocxx/read_concern.cpp
acmorrow: ping. Can this not just be assignment?

src/mongocxx/test/client.cpp
acmorrow: Lets not worry about this just yet. I'm fine with doing this for now. We can always address later if it is an issue.

acmorrow:
LGTM, mod one question about the unique_ptr. No need to repost either way on that. @samantharitter, any additional feedback?

src/mongocxx/read_concern.cpp
Machyne: My understanding is that reset calls the deleter on the old underlying pointer, but assignment may not? I think we want to call the deleter, so reset is better here. Is that right?

acmorrow: I'd be surprised if that was true. It would seem to violate the intended semantics of unique_ptr. Write a little toy program and try it?

Machyne: Oops! You were totally right :) Fixed!

src/mongocxx/collection.cpp
```diff
@@ -666,6 +667,19 @@ void collection::drop() {
     }
 }
 
+void collection::read_concern(class read_concern rc) {
+    libmongoc::collection_set_read_concern(_impl->collection_t, rc._impl->read_concern_t);
+}
+
+stdx::optional<class read_concern> collection::read_concern() const {
```
samantharitter: So, the spec says: "Unless specified [on a client, database, or collection], the [read_concern] value MUST be inherited from its parent."  We are not currently following this behavior, nor is the C driver.   Because we aren't storing any read concern objects, but are instead calling into the C driver each time we fetch the read_concern, I feel that we shouldn't implement this logic.  We should instead expect the C driver to return the right thing when we call libmongoc::collection_get_read_concern--the "right thing" being a parent's read_concern, if it's not set on the collection.

So I guess I'm not asking you to do anything differently here, I just wanted to bring this up.  I've opened [CXX-758](https://jira.mongodb.org/browse/CXX-758) to track this, there's a corresponding CDRIVER ticket linked.

src/mongocxx/collection.hpp
```diff
@@ -485,6 +486,23 @@ class MONGOCXX_API collection {
...
+    ///
+    /// @param rc
+    ///   The new @c read_concern
+    ///
+    /// TODO add link to docs once they exist.
+    ///
+    void read_concern(class read_concern rc);
+
+    ///
+    /// Gets the read_concern for the collection.
```
samantharitter: Can you add a comment here that mentions CXX-758, so we remember to change this once that is fixed?

src/mongocxx/database.hpp
```diff
@@ -162,9 +162,30 @@ class MONGOCXX_API database {
...
+    /// TODO add link to docs once they exist.
+    ///
+    void read_concern(class read_concern rc);
+
+    ///
+    /// The current read concern for this database.
+    ///
+    /// @return the current read_concern
+    ///
+    stdx::optional<class read_concern> read_concern() const;
```
samantharitter: Can you also add a comment here that mentions CXX-758, so we remember to change this once that is fixed?

samantharitter:
A couple notes about the inheritance thing.  Otherwise LGTM!

Machyne:
Comments added! :)

PR #380, opened 2015-11-30
CXX-745 - Implement mongocxx::database::drop()

tessavitabile:


src/mongocxx/test/database.cpp
```diff
@@ -205,4 +219,13 @@ TEST_CASE("Database integration tests", "[database]") {
...
         clear_collection(database, collection_name);
     }
+
+    SECTION("A database may be dropped") {
+        clear_collection(database, collection_name);
+
+        database.create_collection(collection_name);
+        REQUIRE(database.has_collection(collection_name));
+        database.drop();
+        REQUIRE(!database.has_collection(collection_name));
```
tessavitabile: I wasn't sure of the best way to test whether a database had been dropped. I also considered iterating through list_databases. Any advice on this?

amidvidy: I think this works as is

src/mongocxx/database.cpp
```diff
@@ -97,6 +97,11 @@ class collection database::create_collection(stdx::string_view name,
     return mongocxx::collection(*this, static_cast<void*>(result));
 }
 
+void database::drop() {
+    bson_error_t error;
+    libmongoc::database_drop(_impl->database_t, &error);
```
amidvidy: you are going to want to test the error after this call, and then throw an exception if it is non-OK.

amidvidy: note that the docs for mongoc_database_drop state that the error will always be set if the return value is false.

src/mongocxx/test/database.cpp
```diff
@@ -71,6 +71,20 @@ TEST_CASE("A database", "[database]") {
...
+        database_drop->interpose([&](mongoc_database_t* database, bson_error_t* error) { 
+            drop_called = true; 
+            return true;
+        });
+
+        database database = mongo_client["database"];
+        REQUIRE(!drop_called);
+        database.drop();
+        REQUIRE(drop_called);
+    }
```
amidvidy: can you add a negative test showing that the call will throw if database_drop returns false. See REQUIRE_THROWS

amidvidy:
other than the negative test, LGTM. @samantharitter ptal

src/mongocxx/test/database.cpp
samantharitter: I agree, I think this is fine.

samantharitter:
LGTM!

acmorrow:
LGTM as well

PR #381, opened 2015-12-01
CXX-710 - Allow copying of collections

tessavitabile:
I wasn't sure if/how to write tests for this. I tested by trying the following assignments:

auto collection2 = collection;

mongocxx::collection collection2 {collection};

auto collection2 = mongocxx::collection(collection);
    
auto collection2 = conn["testdb"]["testcollection2"];
collection2 = collection;


src/mongocxx/collection.cpp
```diff
@@ -140,6 +140,17 @@ collection::collection(const database& database, void* collection)
                                     database._impl->client_impl)) {
 }
 
+collection::collection(const collection& c) noexcept : _impl{} {
+    impl tmp {*(c._impl)};
+    _impl = stdx::make_unique<impl>(tmp);
```
amidvidy: no need for the temp impl here - as long as you don't `std::move` c's impl, make_unique will copy, not move from it.. Also, you can do all the work in the initializer list i.e.

`collection::collection(const collection& c) noexcept : _impl{stdx::make_unique<impl>(*c._impl) {}`

src/mongocxx/collection.cpp
```diff
@@ -140,6 +140,17 @@ collection::collection(const database& database, void* collection)
...
+}
+
+collection& collection::operator=(const collection& c) noexcept {
+    impl tmp {*(c._impl)};
```
amidvidy: no need for `tmp`

src/mongocxx/private/collection.hpp
```diff
@@ -39,6 +39,15 @@ class collection::impl {
           client_impl(client) {
     }
 
+    impl(const impl& i) : collection_t{} { 
+        collection_t = libmongoc::collection_copy(i.collection_t);
```
amidvidy: set collection_t in the initializer list

amidvidy:
mostly style suggestions.

amidvidy:
if you want you could write a sanity test that creates a collection named foo, copies it and checks that the name is still foo. But the logic is pretty straightforward so I'm not sure its needed, so I'll leave it up to you. LGTM

src/mongocxx/private/libmongoc_symbols.hpp
```diff
@@ -47,6 +47,7 @@ MONGOCXX_LIBMONGOC_SYMBOL(client_set_ssl_opts)
 MONGOCXX_LIBMONGOC_SYMBOL(client_set_write_concern)
 MONGOCXX_LIBMONGOC_SYMBOL(collection_aggregate)
 MONGOCXX_LIBMONGOC_SYMBOL(collection_count)
+MONGOCXX_LIBMONGOC_SYMBOL(collection_copy)
```
samantharitter: nit, please alphabetize and put before collection_count

samantharitter:
+1 for a test, otherwise LGTM

src/mongocxx/collection.hpp
```diff
@@ -93,6 +91,16 @@ class MONGOCXX_API collection {
     collection& operator=(collection&&) noexcept;
 
     ///
+    /// Copy constructs a collection.
+    ///
+    collection(const collection&) noexcept;
```
acmorrow: I don't think the copy constructor and copy assignment operator should be noexcept. They allocate memory via make_unique, so could possibly throw.

src/mongocxx/private/collection.hpp
```diff
@@ -39,6 +39,13 @@ class collection::impl {
           client_impl(client) {
     }
 
+    impl(const impl& i) : collection_t{libmongoc::collection_copy(i.collection_t)} { }
```
acmorrow: Can collection_copy fail?

src/mongocxx/private/collection.hpp
```diff
@@ -39,6 +39,13 @@ class collection::impl {
...
+
+    impl& operator=(const impl& i) { 
+        collection_t = libmongoc::collection_copy(i.collection_t);
```
acmorrow: Same question? What happens if collection_copy fails?

acmorrow:
Are we doing to want to be able to copy database objects as well?

tessavitabile:
There's a TODO to do this for databases, and database copy is implemented in the C driver, so we could do this now. Should I add it to this pull request?

src/mongocxx/private/collection.hpp
tessavitabile: I'm not sure how to figure out whether collection_copy can fail.

amidvidy: I would look at the source in the c driver. Are there any situations in which it returns a null pointer, for example?

amidvidy:
I would do it in a separate PR

src/mongocxx/private/collection.hpp
amidvidy: I'm pretty sure bson_malloc calls abort() if malloc fails, so the program will just crash if we run out of memory.  Likewise for BSON_ASSERT.

As such I think it's safe to assume `mongoc_collection_copy` can't fail (without immediately crashing).

amidvidy: This does bring up a larger issue @acmorrow, that our memory management is not exactly harmonized with the standard c++ facilities like `set_new_handler`, etc. Perhaps we want to use `bson_mem_set_vtable` so that `bson_malloc` uses `operator new` ?

acmorrow: That function scares me a bit. It is global, and needs to be done during process setup. I'm not sure I want to mess with it. My understanding is that libbson / libmongoc already assume a non-failing allocator, and they can't throw. I don't think it helps us

acmorrow:
LGTM

src/mongocxx/private/collection.hpp
amidvidy: we could potentially expose it somehow in bsoncxx or do it when you create a mongocxx::instance. I think I would find it confusing if I called set_new_handler, but the driver still allocated memory through raw malloc/free internally.

tessavitabile:
When Sam and I were discussing allowing copying for database, we realized there were fields I hadn't copied in collection::impl.

amidvidy:
still lgtm

src/mongocxx/private/collection.hpp
```diff
@@ -39,6 +39,19 @@ class collection::impl {
...
     }
 
+    impl(const impl& i) : 
+      collection_t{libmongoc::collection_copy(i.collection_t)},
+      database_name{i.database_name},
+      client_impl{i.client_impl}
+    {}
+
+    impl& operator=(const impl& i) { 
+        collection_t = libmongoc::collection_copy(i.collection_t);
```
samantharitter: Same thing here actually, we should clean up the original `collection_t` with `libmongoc::collection_destroy`.

amidvidy: as with the database review, this is not a problem in practice as we only copy construct impls, even when copy assigning a `database_t`. I think that is cleaner than manually destroying the `collection_t`

amidvidy: We might just want to make the `collection_t` member of `collection::impl` a `mongoc_collection_t* const` so that we can only copy construct but not copy assign impls, and then just keep the rest of the code the same.

tessavitabile:
I did not make the mongoc_collection_t* a const, because the libmongoc methods that use this expect a nonconstant.

samantharitter:
LGTM again!

PR #382, opened 2015-12-01
CXX-753 - Allow copying of databases

tessavitabile:


src/mongocxx/test/database.cpp
```diff
@@ -38,6 +38,21 @@ TEST_CASE("A default constructed database is false-ish", "[database]") {
...
+    client mongodb_client{uri{}};
+
+    std::string dbname{"foo"};
+    std::string dbname2{"bar"};
+    database db = mongodb_client[dbname];
+
+    database db2 = db;
+    database db3 = mongodb_client[dbname2];
+    db3 = db;
+
```
amidvidy: it looks like you are testing the copy assignment operator, but not the copy constructor. Can you test the copy constructor as well?

tessavitabile: I thought that db2 was testing the copy constructor, since db2 had not been previously assigned. Is that not how it works? Would "database db2 {db}" or "database db2 = database(db)" test the copy constructor?

amidvidy: yes, you're right, my bad. Copy initializing db2 (database db2 = db) will call the copy constructor. However, I think it may be clearer (for this test) to use direct initialization e.g. `database db2{db}`

tessavitabile: Sure, I can change it.

src/mongocxx/test/database.cpp
```diff
@@ -38,6 +38,21 @@ TEST_CASE("A default constructed database is false-ish", "[database]") {
...
 }
 
+TEST_CASE("database copy", "[database]") {
+    client mongodb_client{uri{}};
+
+    std::string dbname{"foo"};
+    std::string dbname2{"bar"};
+    database db = mongodb_client[dbname];
+
+    database db2 {db};
```
samantharitter: no space here between db2 and {db}

src/mongocxx/private/database.hpp
```diff
@@ -34,6 +34,19 @@ class database::impl {
...
     {}
 
+    impl(const impl& i) : 
+        database_t{libmongoc::database_copy(i.database_t)},
+        client_impl{i.client_impl},
+        name{i.name} 
+    {}
+
+    impl& operator=(const impl& i) {
+        database_t = libmongoc::database_copy(i.database_t);
```
samantharitter: We should probably call `libmongoc::database_destroy` on the old `database_t` before assigning to it, otherwise we'll be leaking memory.

amidvidy: i think the destructor for the old impl will be called when we assign to it.

src/mongocxx/database.cpp
```diff
@@ -42,6 +42,13 @@ database::database(const class client& client, stdx::string_view name)
           name.data())) {
 }
 
+database::database(const database& d) : _impl{stdx::make_unique<impl>(*(d._impl))} { }
+
+database& database::operator=(const database& d) {
+    _impl = stdx::make_unique<impl>(*(d._impl));
```
amidvidy: if we wanted to use the copy assignment operator of the impl this line would be

*impl = *d._impl. I think this is better though so our old impl gets cleaned up automatically

src/mongocxx/private/database.hpp
amidvidy: actually - do we even use the copy assignment operator of the impl? it looks like we only ever use the copy constructor

samantharitter: I don't think it's true that the destructor is called when you copy assign.  The impl object isn't going out of scope, and its lifetime isn't ending: http://en.cppreference.com/w/cpp/language/destructor

It does seem like we won't ever be using copy assignment.  But if we are defining a copy constructor, it seems like good practice re: the rule of three to also implement copy assignment.  Otherwise we should explicitly delete it with a good comment explaining why we did that.

amidvidy: not here, but the destructor is called in the copy assignment operator of mongocxx::database (which is what actually matters) since we don't actually use the impl's copy assignment operator there.

amidvidy: I guess my interpretation of the rule of three (or rule of 5) isn't that you necessarily need to define all the lifecycle methods, but that if you define one, then you need to define (or delete) the rest as the compiler generated defaults will not be correct.

I think in this case we can just delete the copy assignment (and move assignment operator) of mongocxx::database::impl, and make the internal `database_t *` a `database_t* const`

samantharitter: Right, but as you said, that's not using this operator.  If we were using the copy assignment operator, we would have to explicitly destroy the database_t.  So if we're going to implement this method, we should do that, even if we won't end up using the code now.

samantharitter: I think we should just delete it, with a good comment

amidvidy: I think we raced to comment but I think we should explicitly delete it if we aren't going to be using it. In either case I don't really feel that strongly about it, as both ways are correct, i just prefer the way that results in less code.

tessavitabile:
I did not make the mongoc_database_t* a const, because the libmongoc methods that use this expect a nonconstant.

samantharitter:
LGTM

acmorrow:
LGTM. However:

I still have a worry about the concurrency semantics. In particular, it is going to be a source of confusion that a copy of a collection object may be handed to another thread and the source and copy accessed concurrently, but that the same is not true of a database object. If you want to transfer a database object to another thread, you need to transfer the originating client object as well. However, client is only moveable, not copyable. So, when would it be useful to copy a database object? If you are 'on thread', there is no need?

Is that inconsistency severe enough that we should prohibit the ability to copy a database object? Or does that not help, since you could easily move a database object, and find yourself in the same bad place.

tessavitabile:
Actually, the same issue occurs for collections, since they also contain a client. I had just made a mistake initially and forgotten to copy that field. So collections are databases are consistent with each other, but I agree that it's not clear when copying is useful, since the source and copy can't be used simultaneously. One use case Sam mentioned was that you might copy a collection and modify its read preference, so that you can use a collection with two different read preferences (in the same thread).

amidvidy:
maybe we should update the wiki to make this clear?

https://github.com/mongodb/mongo-cxx-driver/wiki/Library-Thread-Safety

tessavitabile:
Yes, if we are going with this, then I think updating the wiki makes sense.

acmorrow:
@tessavitabile OK, I'm glad to hear they are consistent, and I agree we need to document it well. Please go ahead and update the wiki when you make this commit.

PR #384, opened 2015-12-03
CXX-631 - collection::create_index returns the index name

tessavitabile:


src/mongocxx/collection.cpp
```diff
@@ -624,9 +624,15 @@ bsoncxx::document::value collection::create_index(bsoncxx::document::view keys,
         throw exception::operation(std::make_tuple(error.message, error.code));
     }
 
-    // TODO: return the response from the server, this is not possible now due to the way
-    // libmongoc works.
-    return bsoncxx::document::value{bsoncxx::document::view{}};
+    if (options.name()) {
+        return bsoncxx::builder::stream::document{}
```
amidvidy: I think this could be a bit cleaner if you use the [value_or](http://mnmlstc.github.io/core/optional.html#optional:T:::value_or__URR) method on optional. It could look something like this:
```
bsoncxx::builder::stream::document{};
document << "name" << options.name.value_or(std::string{libmongoc::collection_keys_to_index_string(bson_keys.bson()));
return document.extract();

amidvidy: i guess the way i suggested always calls collection_keys_to_index_string even if it is unneeded. in that case, I think a few extra lines justify the performance benefit. So maybe better to keep as is

samantharitter:
Your code looks good.  I brought up a question about this on the Jira ticket though, please hold off on merging this until that is resolved!  Sorry for not investigating further before you did all this work.

src/mongocxx/collection.cpp
tessavitabile: That makes sense, and it's good to know about value_or anyway.

tessavitabile:
No worries, and will do!

samantharitter:
LGTM

acmorrow:
LGTM as well

PR #385, opened 2015-12-03
CXX-759 working iterators for empty doc views

hanumantmk:
We have to check the return value of iter_next, or we can hand out a begin() iterator that doesn't equal end(), but that you also can't read out of


samantharitter:
LGTM

amidvidy:
lgtm

acmorrow:
LGTM

PR #387, opened 2015-12-04
CXX-762 - Implement bsoncxx::v0::types::value::value()

tessavitabile:


src/bsoncxx/builder/core.cpp
```diff
@@ -410,7 +410,11 @@ void core::concatenate(const bsoncxx::document::view& view) {
 }
 
 void core::append(const bsoncxx::types::value& value) {
-    switch (static_cast<int>(value.type())) {
+    if (!value.type()) {
+        throw(std::runtime_error("invalid value"));
```
amidvidy: throw is a statement, not a function.

src/bsoncxx/types/value.cpp
```diff
@@ -84,8 +88,12 @@ value::~value() {
     }
 }
 
-bsoncxx::type value::type() const {
-    return _type;
+stdx::optional<bsoncxx::type> value::type() const {
+    if (_type == 0) {
```
amidvidy: I would use the ternary operator here.

src/bsoncxx/builder/core.cpp
samantharitter: aha, something that belongs in the bsoncxx error category!!  Sweet.  Maybe mark this with a todo about the exception being thrown.

src/bsoncxx/json.cpp
```diff
@@ -274,15 +274,16 @@ std::string to_json(types::value value) {
 
     json_visitor v(ss, false, 0);
 
-    switch ((int)value.type()) {
-#define BSONCXX_ENUM(name, val)            \
-    case val:                              \
-        v.visit_value(value.get_##name()); \
-        break;
+    if(value.type()) {
```
amidvidy: how would we get an uninitialized value here?

src/bsoncxx/json.cpp
```diff
@@ -274,15 +274,16 @@ std::string to_json(types::value value) {
...
+        switch ((int)*(value.type())) {
```
samantharitter: In general we prefer to use C++ casts over C-style casts:
```
switch (static_cast<int>(*value.type())) {
```

amidvidy: also, instead of int, please use the `std::underlying_type` type trait to get the actual type (which is currently uint8_t, but may change)

i.e.

`switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*value.type()) {`

amidvidy:
Lets think for a bit. Why should we have an "invalid value" be the default instead of some sensible default (empty string, 0 int, etc.). Seems less error-prone, and it eliminates some error cases

src/bsoncxx/types/value.cpp
```diff
@@ -22,14 +22,18 @@
...
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace types {
 
+// The default constructor does not initialize the value union.
+value::value() : _type(0) {}
+
 #define BSONCXX_ENUM(name, val)                                                 \
     value::value(b_##name value)                                                \
-        : _type(static_cast<bsoncxx::type>(val)), _b_##name(std::move(value)) { \
+        : _type(val), _b_##name(std::move(value)) { \
```
samantharitter: run clang-format?

tessavitabile:
I'm not exactly sure why we want the default to be an invalid value. Perhaps because a default-constructed value should eventually be initialized?

samantharitter:
We discussed with @hanumantmk this morning, we are using 0 to represent the invalid type.  We chose not to implement some b_invalid type to match the 0 value.  The reasoning was that we wanted a default-constructed value to be more like a moved-from value, where attempting to use it would fail rather than work with a b_invalid type.  I don't feel very strongly either way, it seems unlikely that the default constructor will be used other than with containers that require one.

amidvidy:
I'm fine with implementing an 'invalid' state if we go the full 9 yards and implement move constructors and move assignment operators that leave moved-from values in said state. Otherwise I think this results in additional error handling logic for little gain compared with say, making value() return value{b_bool{false}}

src/bsoncxx/types/value.cpp
acmorrow: May not be able to do so if the types are different.

However, maybe the following works?

return _type ? { static_cast<bsoncxx::type>(_type); } : { stdx::nullopt };

src/bsoncxx/json.cpp
tessavitabile: I don't think we can, but since value.type() is an optional, it seems safer to check before taking *(value.type()). Do you think it was the correct choice to skip the value if it's invalid?

src/bsoncxx/json.cpp
acmorrow: +1

src/bsoncxx/json.cpp
amidvidy: this seems like a good usage of an `invariant` -like utility, like the one in the server. Thoughts @samantharitter, @acmorrow ?

samantharitter: This is a public-facing function, so it's possible somebody could try to use it with an invalid value type:
```
types::value my_value{};
std::cout << "my value: " << bsoncxx::to_json(my_value) << " hooray" << std::endl;
```
I think we should throw an exception if the type isn't set, since that's what we're doing in other public-facing methods like append() above.

Aside, invariant functionality would be handy for other things, we should do it.

tessavitabile:
Sorry, it looks like clang-format made a lot of changes, which may be confusing for the code review. Let me know if I should undo the changes.

src/bsoncxx/json.cpp
amidvidy: ah yeah, I forgot they had the overload for a standalone value.

 However, consider the case in which a view is passed in to `to_json`. The view should not have been able to have been created with an 'invalid' value as we check for that when appending to a builder.

So maybe we throw here, but in the implementation of `to_json(document::view view)`

```
try {
.. existing logic
} catch (/* exception thrown for invalid value */) {
  BSONCXX_UNREACHABLE;
}
```

amidvidy:
hmm, if possible it may be good to hold off on clang format, get the commit LGTM'd and then clang-format as a separate (non-reviewed) commit after

amidvidy:
if that's annoying to do as you've already formatted everything don't worry about it.

src/bsoncxx/json.cpp
samantharitter: You could default-construct a document::view and attempt to use it, then it would be invalid in a similar way.

I think that we have a larger decision to make overall here in how default-constructed "invalid" types (like this one, document::view, uninitialized clients, etc.) should behave.  I think all such types in the driver will currently just seg fault if you use them this way, but we have open tickets like [CXX-742](https://jira.mongodb.org/browse/CXX-742) requesting that we fail in a friendlier way.

tessavitabile:
Okay, thanks--it would be a bit of a hassle. In the future, I'll avoid committing large formatting changes as part of a cr.

src/bsoncxx/json.cpp
tessavitabile: For now, do you think I should catch the exception in to_json(document::view
view)?

On Mon, Dec 7, 2015 at 11:47 AM, Samantha Ritter <notifications@github.com>
wrote:

> In src/bsoncxx/json.cpp
> <https://github.com/mongodb/mongo-cxx-driver/pull/387#discussion_r46844121>
> :
>
> > @@ -274,15 +274,16 @@ std::string to_json(types::value value) {
> >
> >      json_visitor v(ss, false, 0);
> >
> > -    switch ((int)value.type()) {
> > -#define BSONCXX_ENUM(name, val)            \
> > -    case val:                              \
> > -        v.visit_value(value.get_##name()); \
> > -        break;
> > +    if(value.type()) {
>
> You could default-construct a document::view and attempt to use it, then
> it would be invalid in a similar way.
>
> I think that we have a larger decision to make overall here in how
> default-constructed "invalid" types (like this one, document::view,
> uninitialized clients, etc.) should behave. I think all such types in the
> driver will currently just seg fault if you use them this way, but we have
> open tickets like CXX-742 <https://jira.mongodb.org/browse/CXX-742>
> requesting that we fail in a friendlier way.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mongodb/mongo-cxx-driver/pull/387/files#r46844121>.
>



-- 

*{* name     : "Tess Avitabile",
  title    : "Software Engineer 2",
  phone    : "518-533-8094",
  location : "New York, NY",
  twitter  : "@MongoDB
<https://www.google.com/url?q=https%3A%2F%2Ftwitter.com%2Fmongodb&sa=D&sntz=1&usg=AFQjCNGEAIAhZyZhF7Z9ORWsRliTuc-2dg>
",
  facebook : "MongoDB
<https://www.google.com/url?q=https%3A%2F%2Fwww.facebook.com%2Fmongodb&sa=D&sntz=1&usg=AFQjCNGPMcaFBzmWsh-MpaWeTH6vMQoDIg>
" *}*

src/bsoncxx/builder/core.cpp
```diff
@@ -410,7 +410,12 @@ void core::concatenate(const bsoncxx::document::view& view) {
 }
 
 void core::append(const bsoncxx::types::value& value) {
-    switch (static_cast<int>(value.type())) {
+    if (!value.type()) {
+        // TODO: use bsoncxx error category
+        throw std::runtime_error("invalid value");
+    }
+
+    switch (static_cast<int>(*(value.type()))) {
```
acmorrow: Why are we casting to int here? It seems unnecessary.

src/bsoncxx/json.cpp
```diff
@@ -274,7 +274,12 @@ std::string to_json(types::value value) {
 
     json_visitor v(ss, false, 0);
 
-    switch ((int)value.type()) {
+    if (!value.type()) {
+        // TODO: use bsoncxx error category
+        throw std::runtime_error("invalid value");
+    }
+
+    switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*value.type())) {
```
acmorrow: I'm also puzzled about why we are casting here. It should be fine to switch on an enum

src/bsoncxx/types/value.cpp
```diff
@@ -84,8 +90,8 @@ value::~value() {
     }
 }
 
-bsoncxx::type value::type() const {
-    return _type;
+stdx::optional<bsoncxx::type> value::type() const {
+    return _type ? stdx::optional<bsoncxx::type>(static_cast<bsoncxx::type>(_type)) : stdx::nullopt;
```
acmorrow: Isn't _type already a bsoncxx::type? I'm also still curious about whether this can be simplified. Does

```
return _type ? {_type} : {stdx::nullopt};
```

not work?

src/bsoncxx/types/value.cpp
```diff
@@ -96,11 +102,19 @@ bsoncxx::type value::type() const {
...
+    if (!lhs.type() || !rhs.type()) {
+        return false;
+    }
+
+    if (*(lhs.type()) != *(rhs.type())) {
         return false;
     }
 
-    switch (static_cast<int>(lhs.type())) {
+    switch (static_cast<int>(*(lhs.type()))) {
```
acmorrow: Same question about casting here

src/bsoncxx/types/value.cpp
```diff
@@ -96,11 +102,19 @@ bsoncxx::type value::type() const {
 #undef BSONCXX_ENUM
 
 bool operator==(const value& lhs, const value& rhs) {
-    if (lhs.type() != rhs.type()) {
+    if (!lhs.type() && !rhs.type()) {
+        return true;
```
acmorrow: So, what this says is that two values with unspecified types compare as equal. Just to make sure we have considered the options, is there any benefit to saying that they don't compare equal (much like NaN != NaN).

src/bsoncxx/types/value.hpp
```diff
@@ -23,377 +23,376 @@
...
+    value& operator=(value&&) noexcept;
+
+    ~value();
+
+    friend BSONCXX_API bool operator==(const value&, const value&);
+    friend BSONCXX_API bool operator!=(const value&, const value&);
+
+    ///
+    /// @return The type of the underlying BSON value stored in this object.
+    /// If the type is invalid, no type is returned.
```
acmorrow: How about 'a disengaged optional' is returned.

src/bsoncxx/types/value.hpp
```diff
@@ -23,377 +23,376 @@
...
+    ///
+    stdx::optional<bsoncxx::type> type() const;
```
acmorrow: Is the change here the only non-whitespace change in this file?

src/bsoncxx/json.cpp
tessavitabile: It's not actually an enum, it's a macro, defined here: https://github.com/mongodb/mongo-cxx-driver/blob/master/src/bsoncxx/enums/type.hpp. So if the *value.type() is not cast to an int, it tries to cast all the ints in that file to value_type, which fails.

src/bsoncxx/types/value.cpp
tessavitabile: _type is now an int, so it doesn't work.

src/bsoncxx/types/value.cpp
tessavitabile: I just assumed two invalid types would be equal. Maybe there is some benefit to them not being equal, though I'm not sure what it would be.

src/bsoncxx/types/value.hpp
tessavitabile: Also that _type now has type uint8_t. Otherwise, that's all.

acmorrow:
I'm not super happy with how this has ended up. My feeling is that making the type byte no longer the enum value has opened too many holes, required a lot of casting.

I wonder if this would all go better if we had value hold an optional<bsoncxx::type>.

@tessavitabile - could you work up that up in a new PR against a clean baseline and without clang-format, so we can see how it goes?

tessavitabile:
Sure, sounds good.

acmorrow:
I think there is agreement that the other approach, using optional, is preferable. Lets close this review out in favor of that one.

acmorrow:
Closed in favor of https://github.com/mongodb/mongo-cxx-driver/pull/394

PR #388, opened 2015-12-04
CXX-756 - Make accessor methods to get at raw pointer for bson

tessavitabile:


amidvidy:
I think we have been using a different naming convention in the driver for setters/getters. See the options classes for an example, but the gist of it is that for a `std::size_t` field '_garply' we would provide:
```
// returns the value of garply
std::size_t garply() const;

// sets the value of garply
void garply(std::size_t garply);
```

src/bsoncxx/document/element.hpp
```diff
@@ -69,6 +69,15 @@ class BSONCXX_API element {
 
     explicit operator bool() const;
 
+    const std::uint8_t * const get_raw() const;
```
amidvidy: I would run clang format instead of fixing it by hand, but always glue the asterisk to the type name e.g.

`const std::uint8_t*`

src/bsoncxx/array/view.cpp
```diff
@@ -47,19 +47,19 @@ view::iterator& view::iterator::operator++() {
...
+    i.raw = _element.raw();
+    i.len = _element.length();
+    i.next_off = _element.offset();
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element.raw = nullptr;
-        _element.length = 0;
-        _element.offset = 0;
+        _element.raw(nullptr);
```
acmorrow: Could we just write this as _element = element{nullptr, 0, 0};

src/bsoncxx/document/view.cpp
```diff
@@ -124,7 +125,8 @@ view::const_iterator view::const_iterator::operator++(int) {
 }
 
 bool operator==(const view::const_iterator& lhs, const view::const_iterator& rhs) {
-    return (lhs._element.raw == rhs._element.raw && lhs._element.offset == rhs._element.offset);
+    return (lhs._element.raw() == rhs._element.raw() &&
```
amidvidy: lets use std::tie here.

return std::tie(lhs._element.raw(), lhs,_element.offset()) == std::tie(rhs._element.raw(), rhs._element.offset());

src/bsoncxx/array/view.cpp
```diff
@@ -47,19 +47,19 @@ view::iterator& view::iterator::operator++() {
...
+        _element.length(0);
+        _element.offset(0);
     } else {
-        _element.raw = i.raw;
-        _element.length = i.len;
-        _element.offset = i.off;
+        _element.raw(i.raw);
```
acmorrow: Same thing, does

_element = element{i.raw, i.len, i.off}

work?

src/bsoncxx/array/view.cpp
```diff
@@ -99,19 +100,19 @@ view::const_iterator& view::const_iterator::operator++() {
...
+    i.raw = _element.raw();
+    i.len = _element.length();
+    i.next_off = _element.offset();
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element.raw = nullptr;
-        _element.length = 0;
-        _element.offset = 0;
+        _element.raw(nullptr);
```
acmorrow: As above

src/bsoncxx/document/view.cpp
```diff
@@ -47,19 +47,19 @@ view::iterator& view::iterator::operator++() {
...
+    i.len = _element.length();
+    i.next_off = _element.offset();
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element.raw = nullptr;
-        _element.length = 0;
-        _element.offset = 0;
+        _element.raw(nullptr);
+        _element.length(0);
```
acmorrow: same question as above, here and below

src/bsoncxx/array/view.cpp
tessavitabile: I don't know why it was written this way. Perhaps it's slightly more efficient? But I'm happy to change it.

acmorrow: It should inline away and end up being equivalent

src/bsoncxx/document/view.cpp
tessavitabile: tie wants a reference, not a pointer. I'm guessing we actually want to compare the pointers here. Is there a way to do that with tie?

tessavitabile:
I didn't use std::tie because element::raw() yields a pointer, and std::tie expects a reference, and I thought we probably want to compare the actual pointers.

amidvidy:
it's a bit confusing as std::tie yields a tuple of references to what was passed in to it, in this case it would be a tuple containing a reference to a pointer, and a reference to an offset. It will actually do the right thing in this case - it will generate the exact same code as what you have, but be a little shorter/more readable

tessavitabile:
Maybe I'm doing something wrong, but the compiler doesn't allow me to put a pointer into std::tie.

src/bsoncxx/document/view.cpp
acmorrow: I'd expect it to work.

```
#include <tuple>
#include <cstdlib>

struct X {
    const int* a;
    int b;
};

bool operator==(const X& lhs, const X& rhs) {
    return std::tie(lhs.a, lhs.b) == std::tie(rhs.a, rhs.b);
}


int main(int argc, char* argv[]) {
    const X x1{nullptr, 2};
    const X x2{&x1.b, 3};
    return (x1 == x2) ? EXIT_FAILURE : EXIT_SUCCESS;
}
```

src/bsoncxx/document/view.cpp
```diff
@@ -72,7 +68,8 @@ view::iterator view::iterator::operator++(int) {
 }
 
 bool operator==(const view::iterator& lhs, const view::iterator& rhs) {
-    return (lhs._element.raw == rhs._element.raw && lhs._element.offset == rhs._element.offset);
+    return (lhs._element.raw() == rhs._element.raw() &&
+            lhs._element.offset() == rhs._element.offset());
```
acmorrow: Use tie (assuming you get it to work) here as well

src/bsoncxx/array/view.cpp
```diff
@@ -124,7 +117,8 @@ view::const_iterator view::const_iterator::operator++(int) {
 }
 
 bool operator==(const view::const_iterator& lhs, const view::const_iterator& rhs) {
-    return (lhs._element.raw == rhs._element.raw && lhs._element.offset == rhs._element.offset);
+    return (lhs._element.raw() == rhs._element.raw() &&
```
acmorrow: Tie

src/bsoncxx/array/view.cpp
```diff
@@ -72,7 +68,8 @@ view::iterator view::iterator::operator++(int) {
 }
 
 bool operator==(const view::iterator& lhs, const view::iterator& rhs) {
-    return (lhs._element.raw == rhs._element.raw && lhs._element.offset == rhs._element.offset);
+    return (lhs._element.raw() == rhs._element.raw() &&
```
acmorrow: Tie

src/bsoncxx/array/view.cpp
```diff
@@ -47,19 +47,15 @@ view::iterator& view::iterator::operator++() {
     }
 
     bson_iter_t i;
-    i.raw = _element.raw;
-    i.len = _element.length;
-    i.next_off = _element.offset;
+    i.raw = _element.raw();
```
acmorrow: We have this code four times. Is there a way we could just have a private to_bson_iter_t method?

src/bsoncxx/document/view.cpp
tessavitabile: Sam helped me figure it out: std::tie requires its arguments to be l-values. Is there is a way to convert to l-values in one line? If not, performing two equality checks is probably cleaner.

acmorrow: Ah... I wonder if this is a job for http://en.cppreference.com/w/cpp/utility/tuple/forward_as_tuple

tessavitabile: Yes, indeed!

tessavitabile:
I created two helpers to_bson_iter_t, one in array/view.cpp and one in document/view.cpp. I considered instead having one helper in mongocxx/private/libbson.hpp.

src/bsoncxx/document/view.cpp
```diff
@@ -27,6 +27,16 @@ namespace bsoncxx {
...
 
+namespace {
+bson_iter_t to_bson_iter_t(element e) {
+    bson_iter_t i;
+    i.raw = e.raw();
+    i.len = e.length();
+    i.next_off = e.offset();
+    return i;
+}
+}
```
acmorrow: add // namespace

src/bsoncxx/document/view.cpp
```diff
@@ -27,6 +27,16 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace document {
 
+namespace {
+bson_iter_t to_bson_iter_t(element e) {
```
acmorrow: If you wanted, I think you could add this to bsoncxx/private/helpers and re-use it in both files and remove the dup definitions. Your call.

acmorrow:
LGTM with a nit and an optional suggestion

tessavitabile:
Thanks for the suggestion. One function uses document::element and the other uses array::element, and array::element inherits privately from document::element, so I'm actually not sure I can write one function for both. So I'll just keep the two functions.

acmorrow:
Ah right. LGTM then

PR #389, opened 2015-12-04
CXX-757 Fix value append on bson builder core

hanumantmk:
Disables append for non-char pointer types
- prevents implicit boolean conversion

removes appends for string literals && enables append for char pointers
- string literals still get their memchr inlined
- regular char*'s work

### DIFF
--- a/src/bsoncxx/builder/core.hpp
+++ b/src/bsoncxx/builder/core.hpp
@@ -18,6 +18,7 @@
 
 #include <memory>
 #include <stdexcept>
+#include <type_traits>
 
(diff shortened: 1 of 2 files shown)


amidvidy:
does this still allow us to pass function pointers as subdoc/subarray callables?

acmorrow:
LGTM. Can you add a test to confirm the answer to @amidvidy's question?

hanumantmk:
@amidvidy Yeah, they're still fine.  The overloads for function pointers are actually up in the stream interface, and get deduced long before we consult any of the core::append() overloads

PR #392, opened 2015-12-09
CXX-770: Add missing headers and fix enum typing

markbenvenuto:
Fix miscellaneous compilations issues for Windows
1. Missing headers for things that have a different include order in other compilers
2. Ensure `enum class` is used in all the right places since `enum` != `enum class`.

### DIFF
--- a/src/bsoncxx/array/view.cpp
+++ b/src/bsoncxx/array/view.cpp
@@ -14,6 +14,7 @@
 
 #include <cstdlib>
 #include <cstring>
+#include <tuple>
 
 #include <bson.h>
 
--- a/src/mongocxx/collection.hpp
+++ b/src/mongocxx/collection.hpp
@@ -23,6 +23,7 @@
 
 #include <bsoncxx/builder/stream/document.hpp>
 #include <bsoncxx/document/view.hpp>
+#include <bsoncxx/stdx/optional.hpp>
 #include <bsoncxx/stdx/string_view.hpp>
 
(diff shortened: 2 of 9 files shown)


acmorrow:
LGTM

acmorrow:
This seems to have broken the clang build https://travis-ci.org/mongodb/mongo-cxx-driver/jobs/95846928

markbenvenuto:
This is murkier then I thought.
1. gcc warns about using "enum class" on forward decls and enum decls
2. clang errors about mixing "enum class" on forward decls and enum decls
3. msvc requires it to be the same across forward decls and enum decls

This may be the straw that broke the camel's back.

GCC
/home/mark/mongo-cxx-driver/src/mongocxx/options/find.hpp:103:22: warning: elaborated-type-specifier for a scoped enum must not use the ‘class’ keyword
     void cursor_type(enum class cursor_type cursor_type);

Clang
/home/mark/mongo-cxx-driver/src/mongocxx/options/find.hpp:103:27: error: reference to scoped enumeration must use 'enum' not 'enum class'
    void cursor_type(enum class cursor_type cursor_type);

MSVC
D:\mongo-cxx-driver\src\mongocxx/options/find.hpp(293): error C3431: 'cursor_type': a scoped enumeration cannot be redeclared as an unscoped enumeration

acmorrow:
@markbenvenuto - Lets close this, since this approach proved unworkable.

PR #394, opened 2015-12-09
CXX-762 - Implement bsoncxx::v0::types::value::value()

tessavitabile:
This PR is to try having _type in value be an optional bsoncxx::type instead of an int.

### DIFF
--- a/src/bsoncxx/json.cpp
+++ b/src/bsoncxx/json.cpp
@@ -274,7 +274,12 @@ std::string to_json(types::value value) {
 
     json_visitor v(ss, false, 0);
 
-    switch ((int)value.type()) {
+    if (!value.type()) {
+        // TODO: use bsoncxx error category
+        throw std::runtime_error("invalid value");
+    }
+
+    switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*(value.type()))) {
 #define BSONCXX_ENUM(name, val)            \
     case val:                              \
         v.visit_value(value.get_##name()); \
--- a/src/bsoncxx/json.cpp
+++ b/src/bsoncxx/json.cpp
@@ -274,7 +274,7 @@ std::string to_json(types::value value) {
 
     json_visitor v(ss, false, 0);
 
-    if (!value.type()) {
+    if (!value) {
         // TODO: use bsoncxx error category
         throw std::runtime_error("invalid value");
     }
--- a/src/bsoncxx/types/value.cpp
+++ b/src/bsoncxx/types/value.cpp
@@ -22,14 +22,19 @@
 #include <bsoncxx/config/prelude.hpp>
 
 #include <bsoncxx/json.hpp>
+#include <bsoncxx/stdx/optional.hpp>
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace types {
 
+// The default constructor does not initialize the value union.
+value::value() : _type(stdx::nullopt) {
+}
+
 #define BSONCXX_ENUM(name, val)                                                 \
     value::value(b_##name value)                                                \
-        : _type(static_cast<bsoncxx::type>(val)), _b_##name(std::move(value)) { \
+        : _type(stdx::optional<bsoncxx::type>(static_cast<bsoncxx::type>(val))), _b_##name(std::move(value)) { \
     }
 
 #include <bsoncxx/enums/type.hpp>
@@ -40,15 +45,28 @@ value::value(const value& rhs) {
 }
 
 value& value::operator=(const value& rhs) {
-    _type = rhs._type;
-
-    switch (static_cast<int>(_type)) {
-#define BSONCXX_ENUM(type, val)       \
-    case val:                         \
-        _b_##type = rhs.get_##type(); \
-        break;
+    if(_type && !rhs._type) {
+            switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)     \
+        case val:                   \
+            _b_##type.~b_##type();  \
+            break;
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
+        }
+    }
+
+    _type = rhs._type;
+ 
+    if(_type) {
+        switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)           \
+        case val:                         \
+            _b_##type = rhs.get_##type(); \
+            break;
+#include <bsoncxx/enums/type.hpp>
+#undef BSONCXX_ENUM
+        }
     }
 
     return *this;
@@ -56,35 +74,52 @@ value& value::operator=(const value& rhs) {
 
 value::value(value&& rhs) noexcept {
     *this = std::move(rhs);
+    rhs._type = stdx::nullopt;
 }
 
 value& value::operator=(value&& rhs) noexcept {
-    _type = rhs._type;
-
-    switch (static_cast<int>(_type)) {
-#define BSONCXX_ENUM(type, val)               \
-    case val:                                 \
-        _b_##type = std::move(rhs._b_##type); \
-        break;
+    if(_type && !rhs._type) {
+            switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)     \
+        case val:                   \
+            _b_##type.~b_##type();  \
+            break;
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
+        }
+    }
+
+    _type = rhs._type;
+    rhs._type = stdx::nullopt;
+
+    if(_type) {
+        switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)                   \
+        case val:                                 \
+            _b_##type = std::move(rhs._b_##type); \
+            break;
+#include <bsoncxx/enums/type.hpp>
+#undef BSONCXX_ENUM
+        }
     }
 
     return *this;
 }
 
 value::~value() {
-    switch (static_cast<int>(_type)) {
-#define BSONCXX_ENUM(type, val) \
-    case val:                   \
-        _b_##type.~b_##type();  \
-        break;
+    if(_type) {
+            switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)     \
+        case val:                   \
+            _b_##type.~b_##type();  \
+            break;
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
+        }
     }
 }
 
-bsoncxx::type value::type() const {
+stdx::optional<bsoncxx::type> value::type() const {
     return _type;
 }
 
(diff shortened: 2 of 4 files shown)


src/bsoncxx/json.cpp
```diff
@@ -274,7 +274,12 @@ std::string to_json(types::value value) {
 
     json_visitor v(ss, false, 0);
 
-    switch ((int)value.type()) {
+    if (!value.type()) {
+        // TODO: use bsoncxx error category
+        throw std::runtime_error("invalid value");
```
tessavitabile: Should this exception be caught in to_json(document::view view)?

src/bsoncxx/types/value.cpp
```diff
@@ -96,11 +131,19 @@ bsoncxx::type value::type() const {
 #undef BSONCXX_ENUM
 
 bool operator==(const value& lhs, const value& rhs) {
+    if(!lhs.type() && !rhs.type()) {
```
tessavitabile: Should two moved-from values be equal?

tessavitabile:
I also noticed that in copy-assign and move-assign, I never cleared lhs's value union, so I addressed that in this PR.

src/bsoncxx/json.cpp
samantharitter: I don't think so, I think it should bubble up to the user.

src/bsoncxx/types/value.cpp
```diff
@@ -22,14 +22,19 @@
...
 namespace types {
 
+// The default constructor does not initialize the value union.
+value::value() : _type(stdx::nullopt) {
+}
+
 #define BSONCXX_ENUM(name, val)                                                 \
     value::value(b_##name value)                                                \
-        : _type(static_cast<bsoncxx::type>(val)), _b_##name(std::move(value)) { \
+        : _type(stdx::optional<bsoncxx::type>(static_cast<bsoncxx::type>(val))), _b_##name(std::move(value)) { \
```
samantharitter: You should be able to just say
```
_type(static_cast<bsoncxx::type>(val))
```
and the compiler will construct the optional for you.  Optional has constructor overloads that take a T directly: http://en.cppreference.com/w/cpp/experimental/optional/optional

src/bsoncxx/types/value.cpp
```diff
@@ -40,15 +45,28 @@ value::value(const value& rhs) {
 }
 
 value& value::operator=(const value& rhs) {
-    _type = rhs._type;
+    if(_type && !rhs._type) {
```
samantharitter: I think this block needs to run regardless of whether rhs has a type set.  If it has a type, we still want to clear out the union in lhs to be able to set it to something new.

src/bsoncxx/types/value.cpp
```diff
@@ -56,35 +74,52 @@ value& value::operator=(const value& rhs) {
 
 value::value(value&& rhs) noexcept {
     *this = std::move(rhs);
+    rhs._type = stdx::nullopt;
 }
 
 value& value::operator=(value&& rhs) noexcept {
-    _type = rhs._type;
+    if(_type && !rhs._type) {
```
samantharitter: same note here as with copy assignment

src/bsoncxx/types/value.cpp
samantharitter: Also, I think we need a check for self-equality before doing anything here.  If `this == rhs` and we execute this code, we will get ourselves into a sticky situation by cleaning up our own union and then attempting to assign from it.

src/bsoncxx/json.cpp
amidvidy: how would a view (containing pre-serialized bson) contain invalid elements? that would seem to indicate a bug in the library itself

samantharitter: A user could default-construct a types::value object, or move from one, then try to pass it to to_json.

amidvidy: i mean specifically in the overload `std::string to_json(document::view view)`. As Tess mentioned it may be worth catching the exception and then doing something (like crashing the process) as that should be impossible, I think.

src/bsoncxx/types/value.cpp
acmorrow: I think yes.

acmorrow: Please do run clang-format before pushing to fix up if( rather than if (

src/bsoncxx/types/value.hpp
```diff
@@ -154,8 +155,9 @@ namespace types {
 
         ///
         /// @return The type of the underlying BSON value stored in this object.
+        /// If the type is invalid, a disengaged optional is returned.
         ///
-        bsoncxx::type type() const;
+        stdx::optional<bsoncxx::type> type() const;
```
acmorrow: What would you think about adding an `BSONCXX_INLINE explicit operator bool() const` to this class that just delegated to calling type and checking the state of the optional. Then you could check whether a value was engaged or not by calling `if (value)`, rather than `if (value.type())`

src/bsoncxx/types/value.cpp
```diff
@@ -22,14 +22,19 @@
...
 
 #include <bsoncxx/json.hpp>
+#include <bsoncxx/stdx/optional.hpp>
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace types {
 
+// The default constructor does not initialize the value union.
+value::value() : _type(stdx::nullopt) {
```
acmorrow: does _type{} work here?

src/bsoncxx/types/value.cpp
```diff
@@ -56,35 +74,52 @@ value& value::operator=(const value& rhs) {
 
 value::value(value&& rhs) noexcept {
     *this = std::move(rhs);
+    rhs._type = stdx::nullopt;
```
acmorrow: Is this line necessary? I'd expect the move assignment to do this already

src/bsoncxx/types/value.cpp
acmorrow: I agree with both of the above comments

src/bsoncxx/types/value.cpp
acmorrow: With the caveat that we should think carefully about self move assign here. @hanumantmk, thoughts?

src/bsoncxx/types/value.cpp
tessavitabile: I wonder if the old union is cleared out automatically when you assign to it, since before we didn't need to clear out the union. But I'm happy to do it explicitly.

tessavitabile: Yes, assigning to a data member of a union makes the other data member unassigned. In that case, I think it might make sense to leave this as is, since then the block that clears out the union won't run in the case of self-assignment.

src/bsoncxx/types/value.cpp
tessavitabile: I checked, and moving out of an optional does not make the rhs a nullopt. I would expect it to also.

src/bsoncxx/types/value.cpp
samantharitter: From [cppreference](http://en.cppreference.com/w/cpp/language/union): "If members of a union are classes with user-defined constructors and destructors, to switch the active member, explicit destructor and placement new are generally needed"

I don't think we've defined any custom dtors for the b_* types now, but we call ~b_*type() in ~value().  It's probably not necessary to call the dtors because we have default destructors but this could change in the future, and it might be better to just add this now.

src/bsoncxx/types/value.cpp
samantharitter: The author of [this stack overflow post](http://stackoverflow.com/questions/9322174/move-assignment-operator-and-if-this-rhs) feels that a check here isn't actually necessary.  The reasoning is that either rhs is a temporary, so it should be a unique reference, or rhs is some other object that the user wants to make look like a temporary.  In the second case, it's a user error if user code causes a self move assign, so we should consider throwing an exception here.

src/bsoncxx/types/value.cpp
acmorrow: You need to call the dtor even if it is trivial.

src/bsoncxx/types/value.cpp
acmorrow: What I meant wast that if value's move assignment operator is correctly implemented, than the result of executing it should be that rhs._type becomes nullopt already. You shouldn't need to do it again.

src/bsoncxx/types/value.hpp
```diff
@@ -154,8 +155,14 @@ namespace types {
...
+
+        ///
+        /// @return Whether the underlying BSON value is engaged.
+        ///
+        BSONCXX_INLINE explicit operator bool() const { return type() ? true : false; }
```
acmorrow: Might be better as : `return static_cast<bool>(type())` since optional already defines an explicit conversion to bool operator.

src/bsoncxx/types/value.cpp
```diff
@@ -56,35 +78,57 @@ value& value::operator=(const value& rhs) {
...
 }
 
 value& value::operator=(value&& rhs) noexcept {
-    _type = rhs._type;
+    if (this == &rhs) {
+        // TODO: use bsoncxx error category
+        throw std::runtime_error("self move assign");
+    }
```
acmorrow: I'm not sure about this. Lets discuss.

src/bsoncxx/types/value.cpp
```diff
@@ -40,15 +45,32 @@ value::value(const value& rhs) {
 }
 
 value& value::operator=(const value& rhs) {
-    _type = rhs._type;
+    if (this == &rhs) {
+        return *this;
+    }
+
+    if (_type) {
```
acmorrow: Can we factor this out (it is repeated below in the move assign operator) into a private method named `destroy` or similar? That method could also take care of restoring _type to nullopt.

src/bsoncxx/types/value.cpp
```diff
@@ -56,35 +78,57 @@ value& value::operator=(const value& rhs) {
...
     return *this;
 }
 
 value::~value() {
-    switch (static_cast<int>(_type)) {
-#define BSONCXX_ENUM(type, val) \
-    case val:                   \
-        _b_##type.~b_##type();  \
-        break;
+    if (_type) {
```
acmorrow: I think this is another instance of the `destroy` method

src/bsoncxx/types/value.cpp
```diff
@@ -56,35 +78,57 @@ value& value::operator=(const value& rhs) {
...
+        case val:                   \
+            _b_##type.~b_##type();  \
+            break;
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
+        }
+    }
+
+    _type = rhs._type;
+    rhs._type = stdx::nullopt;
```
acmorrow: Rather than this, after the code below where you move from rhs, add a `rhs.destroy()` call.

src/bsoncxx/types/value.cpp
tessavitabile: Ah, I understand.

src/bsoncxx/types/value.cpp
```diff
@@ -40,15 +45,32 @@ value::value(const value& rhs) {
...
-    _type = rhs._type;
+    if (this == &rhs) {
+        return *this;
+    }
+
+    if (_type) {
+            switch (static_cast<typename std::underlying_type<bsoncxx::type>::type>(*_type)) {
+#define BSONCXX_ENUM(type, val)     \
+        case val:                   \
+            _b_##type.~b_##type();  \
```
acmorrow: This implementation strategy means that value doesn't offer the strong exception safety guarantee.

tessavitabile: Should I use a temp instead?

acmorrow:
@hanumantmk - why do we want value to be default constructible, in general?

src/bsoncxx/types/value.cpp
acmorrow: Or, we need to assert that all b_* types are noexcept for move, move assign, dtor, etc.

samantharitter:
The request for the default constructor came from a user who wanted to create a map of values, map requires a default constructor.

acmorrow:
So, for associative contains, the mapped_type only needs to be DefaultConstructible to support `operator[]`.

Maybe the right answer is to not do this at all?

That said, having reviewed this class more closely, I think we need to audit its original implementation for correctness, noexcept, etc.

acmorrow:
Furthermore, if you really want to use `operator[]`, make the `mapped_type` of your container be `optional<value>`?

samantharitter:
Seems reasonable.  If we don't want to implement a default constructor, there's still a little work to do under this ticket to remove its declaration in the header file.  Maybe @hanumantmk knows the reason why it was originally included there?

PR #395, opened 2015-12-10
CXX-762 - Audit bsoncxx::v0::types::value

tessavitabile:


src/bsoncxx/types/value.cpp
```diff
@@ -1,3 +1,4 @@
+// audit
```
amidvidy: detritus?

samantharitter: This was just so we could see the .cpp file in the review, to facilitate an audit of the code.

acmorrow:
Sigh. Github pull request fail. You can't comment on lines you didn't change. We should probably move this review to reitveld.

tessavitabile:
https://mongodbcr.appspot.com/42090001

PR #396, opened 2015-12-11
CXX-649 - mongoc and mongocxx error spaces

tessavitabile:


src/mongocxx/exception/private/mongoc_error.hpp
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/config/prelude.hpp>
+
+#include <bsoncxx/document/value.hpp>
+
```
acmorrow: don't need to separate bsoncxx from mongocxx headers

src/mongocxx/exception/exception.cpp
```diff
@@ -10,29 +10,22 @@
...
+#include <utility>
 
-// TODO: figure out if we need this
-// TODO: figure out if bson document should be view/value, etc.
+#include <mongocxx/exception/exception.hpp>
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
-namespace error {
+namespace exception {
```
samantharitter: On the bsoncxx side I chose not to use a namespace exception, for two reasons: first, I wanted the type to read bsoncxx::exception, not bsoncxx::exception::exception; second, the name 'exception' being used as both a type and a namespace seemed to make the compiler unhappy.  I guess it's a matter of taste on the first part.

src/mongocxx/exception/exception.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+namespace exception {
+
+///
+/// A class to be used as the base class for all mongocxx exceptions.
+///
+class MONGOCXX_API exception : public std::system_error {
+   public:
```
samantharitter: since we're no longer virtually inheriting, can you forward the constructor now?

src/mongocxx/exception/mongoc_error.hpp
```diff
@@ -0,0 +1,70 @@
...
+    k_server_selection_bad_wire_version,
+    k_server_selection_failure,
+    k_server_selection_invalid_id,
+    k_stream_connect,
+    k_stream_invalid_state,
+    k_stream_invalid_type,
+    k_stream_name_resolution,
+    k_stream_not_established,
+    k_stream_socket,
+    k_raw_bson,
```
samantharitter: alphabetize

src/mongocxx/exception/mongoc_error_category.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include <mongocxx/config/prelude.hpp>
+
+#include <string>
+
+#include <mongocxx/exception/mongoc_error_category.hpp>
+
+#include <mongocxx/exception/mongoc_error.hpp>
+
+namespace {
+using mongocxx::exception::mongoc_error;
```
samantharitter: nit, blank lines before and after this

src/mongocxx/exception/mongoc_error_category.cpp
```diff
@@ -0,0 +1,135 @@
...
+            return "mongoc protocol_error error";
+        case mongoc_error::k_protocol_invalid_reply:
+            return "mongoc protocol_invalid_reply error";
+        case mongoc_error::k_query_command_not_found:
+            return "mongoc query_command_not_found error";
+        case mongoc_error::k_query_failure:
+            return "mongoc query_failure error";
+        case mongoc_error::k_query_not_tailable:
+            return "mongoc query_not_tailable error";
+        case mongoc_error::k_scram_not_done:
```
samantharitter: no case for k_raw_bson?

src/mongocxx/exception/mongocxx_error.hpp
```diff
@@ -10,25 +10,23 @@
...
 MONGOCXX_INLINE_NAMESPACE_BEGIN
-namespace error {
+namespace exception {
 
-class write {
-    std::int32_t code;
-    std::string message;
-}
+enum class mongocxx_error : std::int32_t {
+};
```
samantharitter: We should probably seed this with something, maybe just a k_unknown?

src/mongocxx/exception/operation.hpp
```diff
@@ -16,18 +16,44 @@
...
+    /// Constructs a new operation exception.
+    ///
+    /// @param ec
+    ///   The error code associated with this exception.
+    /// @param raw_server_error
+    ///   The optional raw bson error document to be associated with this exception.
+    /// @param what_arg
+    ///   An optional message to be returned by `what`.
+    ///
+    explicit operation(std::error_code ec, bsoncxx::document::value raw_server_error,
```
samantharitter: should this take raw_server_error by &&?

src/mongocxx/exception/exception.hpp
tessavitabile: I wasn't sure whether we want to, since maybe we don't want all of the system_error constructors. What do you think?

src/mongocxx/exception/operation.hpp
tessavitabile: I'm not sure. Would you mind explaining why we would do it this way?

src/mongocxx/exception/exception.cpp
tessavitabile: The advantage of the exception namespace is that the derived classes read mongocxx::exception::operation, etc (and in the case of bulk_write, this avoids a clash with the existing mongocxx::bulk_write). But I'm happy to call the derived classes operation_exception, bulk_write_exception, etc.

src/mongocxx/exception/mongoc_error_category.cpp
tessavitabile: Should the message be "mongoc raw_bson error"?

src/mongocxx/exception/exception.cpp
tessavitabile: I'm going to go ahead and change the derived classes to operation_exception, bulk_write_exception, etc.

src/mongocxx/exception/exception.hpp
samantharitter: I don't see a problem with any of its constructors, I think forwarding them is fine

src/mongocxx/exception/mongocxx_error.hpp
```diff
@@ -16,21 +16,24 @@
...
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
-namespace exception {
 
-class MONGOCXX_API operation : public base {
-   public:
-    using base::base;
+enum class mongocxx_error : std::int32_t {
+	k_unknown = std::numeric_limits<std::int32_t>::max(),
```
tessavitabile: Should I not have added the k_unknown? I'm wondering if the same logic applies to the bsoncxx error handling cr, where we don't have k_unknown since we are creating all the errors, so they should all be known.

samantharitter: I think not.  But it does make sense to have one for the libmongoc_errors!

acmorrow:
@tessavitabile please ping this review when it is ready for another look

tessavitabile:
@acmorrow @samantharitter This should be in line with the bsoncxx error handling now.

acmorrow:
@tessavitabile that was quick - I should have waited 5 more minutes. I'll take a look right now

src/mongocxx/exception/operation_exception.hpp
```diff
@@ -0,0 +1,59 @@
...
+    /// Constructs a new operation exception.
+    ///
+    /// @param ec
+    ///   The error code associated with this exception.
+    /// @param raw_server_error
+    ///   The optional raw bson error document to be associated with this exception.
+    /// @param what_arg
+    ///   An optional message to be returned by `what`.
+    ///
+    explicit operation_exception(std::error_code ec, bsoncxx::document::value&& raw_server_error,
```
acmorrow: I don't think this needs to be explicit since there are multiple non-default arguments

src/mongocxx/exception/error_code.hpp
```diff
@@ -10,25 +10,21 @@
...
 
-#include <mongocxx/config/prelude.hpp>
+#pragma once
 
-#include <cstdint>
-#include <string>
+#include <mongocxx/config/prelude.hpp>
 
-// TODO: figure out if we need this
+#include <limits>
```
acmorrow: Why <limits>?

src/mongocxx/exception/authentication_exception.hpp
```diff
@@ -16,50 +16,43 @@
...
 #include <mongocxx/stdx.hpp>
 
 namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
-namespace exception {
-
-using error_and_code_type = std::tuple<std::string, std::int32_t>;
 
-class MONGOCXX_API base : virtual public std::exception {
+class MONGOCXX_API authentication_exception : public exception {
```
acmorrow: Not saying its wrong, but why does authentication_exception re-roll this rather than just being an operation exception? Just curious.

src/mongocxx/error/write_concern.hpp
```diff
@@ -1,38 +0,0 @@
...
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License./**
-
-#include <mongocxx/config/prelude.hpp>
-
-#include <cstdint>
-#include <string>
-
-// TODO: figure out if we need this
```
acmorrow: I guess we figured out we didn't?

acmorrow:
This looks very good. A few small comments and questions. It does look like there are some merge conflicts you need to address?

src/mongocxx/exception/authentication_exception.hpp
tessavitabile: Do you mean inheriting from operation_exception? I'm not sure.

src/mongocxx/error/write_concern.hpp
tessavitabile: It looks like this is an earlier version of error handling. This file was deleted as part of Matt's pull request.

tessavitabile:
How does this team resolve merge conflicts?

src/mongocxx/exception/error_code.hpp
tessavitabile: No idea.

acmorrow:
@tessavitabile anyone on the team should be able to walk you through it. Just grab someone and ask.

src/mongocxx/exception/authentication_exception.hpp
tessavitabile: Is there a reason you might want to catch operation exceptions and not authentication exceptions?

acmorrow: If you wanted to handle them differently, you could always do

```
try {
   // do stuff
}
catch (const authentication_exception&) {
    // We handle auth issues at a higher level, just
    // let the exception through
    throw;
} catch (const operation_exception& op_xcp) {
    // OK, I actually want to do something here, this isn't an auth issue.
}
```

But if you *do* want to handle auth exceptions the same way, then you can just catch `operation_exception`. So I think it can just be a subclass of operation_exception? @samantharitter what do you think?

src/mongocxx/error/write_concern.hpp
acmorrow: ok

src/mongocxx/exception/authentication_exception.hpp
samantharitter: I don't see a problem with having auth exceptions inherit from operation exceptions, that seems much more straightforward

acmorrow:
LGTM - Ship it!

PR #397, opened 2015-12-14
CXX-768 Bsoncxx error handling

samantharitter:


src/bsoncxx/exception/libbson_error.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <limits>
+#include <system_error>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+
+enum class libbson_error : std::int32_t {
```
samantharitter: These are all of the errors that I found in libbson, we're not currently checking for or throwing any of them.

src/bsoncxx/exception/bsoncxx_error.hpp
```diff
@@ -0,0 +1,51 @@
...
+    k_unknown = std::numeric_limits<std::int32_t>::max(),
+};
+
+///
+/// Translate a bsoncxx error into a std::error_code.
+///
+/// @param an error from bsoncxx
+/// @return an error_code
+///
+std::error_code make_error_code(bsoncxx_error error);
```
tessavitabile: My make_error_code is in mongoc_error_category.hpp. Do you think I should move it?

src/bsoncxx/exception/error_category.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <bsoncxx/exception/libbson_error.hpp>
+
+namespace {
+
+using namespace bsoncxx;
+
+///
+/// An error category for errors originating in bsoncxx.
+///
+class bsoncxx_error_category_impl final : public std::error_category {
```
tessavitabile: For mongocxx, these classes are in separate files. Not sure if that matters.

src/bsoncxx/exception/exception.cpp
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <bsoncxx/exception/exception.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace exception {
```
tessavitabile: Sorry, I'm confused. I thought you weren't using namespace exception?

samantharitter: Darn I thought I found them all and deleted them!  This one is left-over.

src/bsoncxx/exception/exception.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+exception::exception(std::error_code ec, std::string what_arg)
+    : std::system_error(ec, what_arg), _ec(ec), _what(what_arg) {
+}
+
+const std::error_code& exception::code() const {
```
tessavitabile: Do you need this, or is it inherited from system_error? (And if not, I should probably include it in mongocxx::exception.)

src/bsoncxx/exception/exception.cpp
```diff
@@ -0,0 +1,37 @@
...
+    return _ec;
+}
+
+const char* exception::what() const noexcept {
```
tessavitabile: Same here

tessavitabile:
Should there be a make_error_condition?

tessavitabile:
A few questions to synchronize this with mongocxx error handling, but otherwise it looks good.

src/bsoncxx/exception/error_category.cpp
samantharitter: It shouldn't matter because they are still in an anonymous namespace, so they won't be leaked out anywhere.  To a caller, the result of libbson_error_category() or bsoncxx_error_category() will just look like any error_category.Though for consistency's sake maybe we should pick one approach?

I split it up like this:
- bsoncxx_error.hpp/cpp - defines driver-specific error types and transforms them into error_codes
- libbson_error.hpp/cpp - ditto above, for libbson
- error_category.hpp/cpp - defines both error categories

It looks like you/Matt split it up like this:
- mongoc_error.hpp - defines libmongoc-specific error types
- mongocxx_error.hpp - ditto above, for mongocxx
- mongoc_error_category.hpp/cpp - defines mongoc error category, and transforms mongoc_error types into error_codes
- mongocxx_error_category.hpp/cpp - ditto above, for mongocxx

Personally, I think it makes more sense to keep the utilities (like make_error_code) with the error type files rather than with the error_categories.  But they are all related to each other so you could argue either way!  Let's pick one.

src/bsoncxx/exception/exception.cpp
samantharitter: I actually don't think I meant to commit this file!

samantharitter:
I wasn't sure about make_error_condition.  We wouldn't be using this method anywhere.  Do we expect a driver user to call it?  Probably not?  You can't throw them with a std::system_error, which makes sense, since a system_error is inherently platform-dependent.  I can't remember why we had this method in the other review.

tessavitabile:
Actually, I think it does not make sense to have this method. My understanding is that error_conditions are used to create groupings of error_codes to test for. You never throw an error_condition--you use them to test whether your error_code matches a particular error_condition. So I don't think it makes sense to create an error_condition out of a mongoc_error--if we wanted to use error_conditions, we would probably write a separate enum with broader conditions (perhaps client error, collection error, etc) and define equivalences between the specific error_codes and the broader error_conditions, so that users could test whether their error_code matched a broader condition. I don't think we care to do that, so I should probably get rid of the method.

src/bsoncxx/exception/error_category.cpp
tessavitabile: I don't feel strongly either way, so I'm happy to match what you did.

src/bsoncxx/document/element.cpp
```diff
@@ -32,11 +34,11 @@
...
 
-#define BSONCXX_TYPE_CHECK(name)                            \
-    do {                                                    \
-        if (type() != bsoncxx::type::name) {                \
-            throw std::runtime_error("type is not " #name); \
-        }                                                   \
+#define BSONCXX_TYPE_CHECK(name)                                                            \
+    do {                                                                                    \
+        if (type() != bsoncxx::type::name) {                                                \
+            throw bsoncxx::exception{make_error_code(bsoncxx_error::k_wrong_element_type)}; \
```
acmorrow: Could we use the 'what' argument to get the type back in?

src/bsoncxx/exception/bsoncxx_error.hpp
```diff
@@ -0,0 +1,51 @@
...
+    k_cannot_close_document_in_sub_array,
+    k_cannot_perform_array_operation_on_document,
+    k_cannot_perform_document_operation_on_array,
+    k_need_key,
+    k_no_array_to_close,
+    k_no_document_to_close,
+    k_unmatched_key_in_builder,
+    k_unset_element,
+    k_wrong_element_type,
+    k_unknown = std::numeric_limits<std::int32_t>::max(),
```
acmorrow: Do we need k_unknown?

samantharitter:
I've opened [CXX-775](https://jira.mongodb.org/browse/CXX-775) in JIRA for the error_condition question and put it into "features we are unsure of."  It seems like we should wait on this work until we have more error codes or have a better idea of what sort of groupings would be useful to driver users.

src/bsoncxx/exception/error_category.hpp
```diff
@@ -0,0 +1,41 @@
+// Copyright 2015 MongoDB Inc.
```
acmorrow: Is this header public or private?

src/bsoncxx/exception/libbson_error.hpp
```diff
@@ -0,0 +1,44 @@
+// Copyright 2015 MongoDB Inc.
```
acmorrow: Is this header public or private? If it is public, it leaks that we are using libbson internally.

src/bsoncxx/exception/bsoncxx_error.hpp
acmorrow: Does this need to be public?

samantharitter: I forget what we decided, did we want the make_error_code utilities to be grouped with the error type stuff or the error category stuff?  I've moved this to a private bsoncxx_error header.

src/bsoncxx/document/element.cpp
samantharitter: I've changed this to spell out different error types for the various value types instead.

src/bsoncxx/exception/bsoncxx_error.hpp
samantharitter: Probably not, because we're generating the bsoncxx errors, so we should always know what they are!  Removing.

src/bsoncxx/exception/bsoncxx_error.hpp
tessavitabile: They got grouped with the error type stuff.

samantharitter: So I should leave these where they are?

tessavitabile: Yeah, that's what I've got. make_error_code(mongocxx_error error) is in mongocxx/exception/mongocxx_error.cpp, and make_error_code(mongoc_error error) is in mongocxx/exception/private/mongoc_error.hpp.

tessavitabile:
LGTM

src/bsoncxx/exception/bsoncxx_error.hpp
acmorrow: If this is a public API method, it should have the export macro on it

src/bsoncxx/exception/bsoncxx_error.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <limits>
+#include <system_error>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+
+enum class bsoncxx_error : std::int32_t {
```
acmorrow: this is already in the namespace bsoncxx. Does the enum name also need to start with bsoncxx? Could we call it error_codes or similar?

src/bsoncxx/private/libbson_error.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <bsoncxx/private/libbson_error.hpp>
+
+#include <bsoncxx/private/error_category.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+
+libbson_error convert_libbson_error_code(int code, int domain) {
+    switch (domain) {
+        case BSON_ERROR_JSON:
```
acmorrow: I think this domain/code thing suggests that we are supposed to treat each domain as its own error_category. Then the enumerations for each category would re-iterate the bson error code values.

src/bsoncxx/private/libbson_error.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+///
+/// Convert an error code and domain from libbson into a libbson_error.
+///
+/// @param code A libbson error code
+/// @param domain A libbson error domain
+///
+/// @return A libbson error
+///
+libbson_error convert_libbson_error_code(int code, int domain);
```
acmorrow: If each domain had a its own enum and category, then you would only need

`std::error_code make_error_code(int code, int domain)`

src/bsoncxx/exception/bsoncxx_error.hpp
samantharitter: I didn't think we wanted it to be public?  I guess it should move to the private header, then.

acmorrow:
This does LGTM too, mod an export and a naming nit. I do wonder whether we would get some benefit by definint a separate enum and error_category for each domain value that the C driver defines. We would need some sort of degenerate error_category to represent "unknown domain". But the upside would be that our system_errors would have error_codes that were 1:1 with the libbson ones. @samantharitter @tessavitabile thoughts on that?

src/bsoncxx/exception/bsoncxx_error.hpp
acmorrow: Thats right, I just forgot we had discussed. Agree, move it to a private header.

tessavitabile:
That seems to be exactly the intention of error_category's, so it makes sense to me.

src/bsoncxx/private/error_category.cpp
```diff
@@ -0,0 +1,153 @@
...
+                return "unset document::element";
+        }
+    }
+};
+
+///
+/// An error category for libbson errors in the BSON_ERROR_JSON domain.
+///
+
+class libbson_json_error final : public std::error_category {
```
samantharitter: @acmorrow  and @tessavitabile this has been updated to use separate error categories for libbson error domains, do we like it?

tessavitabile: I do! Now we are taking advantage of the fact that error categories let you preserve the original error codes.

tessavitabile:
LGTM again.

src/bsoncxx/private/error_category.cpp
acmorrow: +1 I think this is nice. @tessavitabile since we all agree, lets do this for the libmongoc errors in your review, if you haven't already

src/bsoncxx/private/error_category.cpp
```diff
@@ -0,0 +1,154 @@
...
+        default:
+            static const libbson_unknown_error unknown_category{};
+            return unknown_category;
+    }
+}
+
+BSONCXX_INLINE_NAMESPACE_END
+}  // namespace bsoncxx
+
+#include <bsoncxx/config/postlude.hpp>
```
acmorrow: It is harmless, but you don't actually need to include the postlude in .cpp files.

acmorrow:
@samantharitter This looks great. LGTM. @tessavitabile please let me know when your mongocxx side has converged with this design and I will give it a look. Really happy with how this turned out, I think this is a big improvement.

PR #398, opened 2015-12-15
CXX-774: Add dependencies to make paralleized build available

duckie:
When generating makefiles with CMake, a paralleized build such as "make -j8" does not work because the examples try to compile over a local installation which is not finished yet.

This can be solved by adding the right dependencies.

Jira issue: https://jira.mongodb.org/browse/CXX-774

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,9 +1,11 @@
 add_custom_target(install_headers
     COMMAND "${CMAKE_COMMAND}" -DCMAKE_INSTALL_COMPONENT=dev -P "${CMAKE_BINARY_DIR}/cmake_install.cmake"
+    DEPENDS bsoncxx mongocxx
 )
 
 add_custom_target(install_libs
     COMMAND "${CMAKE_COMMAND}" -DCMAKE_INSTALL_COMPONENT=runtime -P "${CMAKE_BINARY_DIR}/cmake_install.cmake"
+    DEPENDS bsoncxx mongocxx
 )
 
 add_subdirectory(bsoncxx)


acmorrow:
@duckie - Thanks for the PR.

As you probably figured out, what we want to do is build the examples against the installed bsoncxx and mongocxx libraries. This ensures that we have installed all of the necessary headers, that we have set rpath's correctly, etc.

Unfortunately though, I'm not convinced that what you have proposed is a complete solution. I think it may be the case that CMake simply can't express the dependency that we need here, since it apparently is impossible to depend on a built in target:

http://public.kitware.com/Bug/view.php?id=8438

The dependency edges that you added will, I think, require that the libraries for mongocxx and bsoncxx be compiled and linked, but not that they be installed. So, you have definitely made the race less likely, but not completely eliminated it.

However, maybe we should just merge this as the practical effect seems to be that the build works better than it does now.

We will need to think about it a bit.

duckie:
Hello @acmorrow and thank you for reviewing my changes.

It seems to me that you missed a little point.  The tests are already depending on `install-headers` and `install-libs`, each of those executing *another* cmake process to install the libraries, and I did not make any change to that.

My change makes sure that this installation step is depending on the build step of the libraries. No built-in target is involved is the dependency chain. If you are thinking about the `install` target, is is contained in the call to a subprocess.

Therefore, the race should be eliminated completely.

acmorrow:
@duckie Yes, I think you are right. I tried out a parallel build and it worked fine. I will merge this. Have you signed the MongoDB contributor agreement?

duckie:
@acmorrow Yes, I just did.

acmorrow:
@duckie merged, thank you!

PR #399, opened 2015-12-15
CXX-736 - Split out tasks in evergreen

tessavitabile:
https://evergreen.mongodb.com/version/56707e533ff1221cc50003fb_0

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -47,42 +47,77 @@ functions:
 #######################################
 
 pre:
-  - command: expansions.fetch
-    params:
-      keys:
-        - local_key: "aws_key"
-          remote_key: "project_aws_key"
-        - local_key: "aws_secret"
-          remote_key: "project_aws_secret"
-  - command: shell.exec
-    params:
-      script: |
-        rm -rf "mongo-cxx-driver"
-        rm -fr "mongo-c-driver"
-        rm -fr mongod
+  - func: "start_mongod"
+
+#######################################
+#              Post Task               #
+#######################################
+
+post:
+  - func: "stop_mongod"
 
 #######################################
 #               Tasks                 #
 #######################################
 
 tasks:
-    - name: compile_and_test
+    - name: compile
+      depends_on: []
       commands:
+        - command: expansions.fetch
+          params:
+            keys:
+              - local_key: "aws_key"
+                remote_key: "project_aws_key"
+              - local_key: "aws_secret"
+                remote_key: "project_aws_secret"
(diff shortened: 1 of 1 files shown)


.mci.yml
```diff
@@ -47,42 +47,77 @@ functions:
...
+            keys:
+              - local_key: "aws_key"
+                remote_key: "project_aws_key"
+              - local_key: "aws_secret"
+                remote_key: "project_aws_secret"
+        - command: shell.exec
+          params:
+            script: |
+              rm -rf "mongo-cxx-driver"
+              rm -fr "mongo-c-driver"
```
markbenvenuto: use `rf` instead of `fr`

.mci.yml
```diff
@@ -100,7 +135,10 @@ buildvariants:
       run_on:
           - ubuntu1404-test
       tasks:
-          - name: compile_and_test
+          - name: compile
```
samantharitter: Can you define a variable with this list of tests and use it for both buildvariants?  Would look something like this:

```
tasks: &all_tasks
- compile
- run_examples
- etc

- name: ubuntu-test
  tasks: *all_tasks

- name: ubuntu-debug
  tasks: *all_tasks
```

tessavitabile:
The non-compile tasks have been failing arbitrarily with the message:
Command failed: chdir /data/mci/mongo-cxx-driver/build: no such file or directory
Does anyone have an idea of why this might be happening, and why it only happens sometimes?

acmorrow:
@tessavitabile did the non-compile tasks execute on the same instance where the compile ran?

tessavitabile:
How can I tell? An example is here:
https://evergreen.mongodb.com/build/mongodb_cpp_driver_dev_ubuntu1404_debug_05459816cae6258bad01d248574629795b7d398e_15_12_16_18_05_11

acmorrow:
They definitely ran on different hosts. Check the host for each task that it ran.

tessavitabile:
@jbreams thought it might not be worthwhile to split up the tasks, since the tasks each run quickly, and if we split up the different tasks, then we need to install the C driver and start a mongod each time. Thoughts? He also suggested that if we don't split up the tasks, then we might want to create a way to see which subtask failed, even though they're all part of the same task, which has been done for server tests.

acmorrow:
OK, I'm convinced. Lets close this, I don't think we want to do this, at least not right now.

samantharitter:
Agreed!

PR #400, opened 2015-12-16
CXX-755 Don't require SSL support in libmongoc

acmorrow:
This change also fixes up some instances where we were calling
mongoc APIs directly, rather than through the shim, and drops
the dependency on the private read concern header.


acmorrow:
@samantharitter @tessavitabile I grabbed this one since it seemed easy. 

@Machyne - I'm unclear on why we needed the MONGOC_I_AM_A_DRIVER and access to the private struct definition in mongoc-read-concern-private.h. Calling [mongoc_read_concern_get_level](http://api.mongodb.org/c/1.3.0/mongoc_read_concern_get_level.html) seems to work just fine?

src/mongocxx/client.cpp
```diff
@@ -33,8 +33,12 @@ client::client() noexcept = default;
 client::client(const class uri& uri, const options::client& options)
     : _impl(stdx::make_unique<impl>(libmongoc::client_new_from_uri(uri._impl->uri_t))) {
     if (options.ssl_opts()) {
+#if defined(MONGOC_HAVE_SSL)
         auto mongoc_opts = options::make_ssl_opts(*options.ssl_opts());
         libmongoc::client_set_ssl_opts(_impl->client_t, &mongoc_opts);
+#else
+        // TODO: For now, just ignoring. Should we throw?
```
samantharitter: I think we should throw, we can add a new error type k_ssl_not_enabled or something.

src/mongocxx/pool.cpp
```diff
@@ -42,11 +42,15 @@ pool::pool(const uri& mongodb_uri, stdx::optional<options::ssl> ssl_options)
...
                                     std::move(ssl_options))} {
     if (_impl->ssl_options) {
+#if defined(MONGOC_HAVE_SSL)
         auto mongoc_opts = options::make_ssl_opts(*_impl->ssl_options);
         // We store ssl options in a member variable because we need the strings to stay alive so
         // libmongoc can use the raw char arrays, however, the mongoc_ssl_opt_t struct can be a
         // temporary as the driver will copy it.
         libmongoc::client_pool_set_ssl_opts(_impl->client_pool_t, &mongoc_opts);
+#else
+        // TODO: For now, just ignoring. Should we throw?
```
samantharitter: ditto here

src/mongocxx/test/pool.cpp
```diff
@@ -96,6 +97,7 @@ TEST_CASE(
     REQUIRE(interposed.crl_file == crl_file);
     REQUIRE(interposed.weak_cert_validation == allow_invalid_certificates);
 }
+#endif
```
samantharitter: Can we also test that without SSL support in libmongoc, trying to pass SSL options to a pool or client fails?  Or throws.

samantharitter: Also, will we eventually want a non-SSL builder to test this?  Or is it something we should just test manually from time to time?

src/mongocxx/client.cpp
acmorrow: I like that idea. Should I leave as a TODO until new error handling facilities are in?

src/mongocxx/test/pool.cpp
acmorrow: It might be tricky because we would need to build the C driver twice. Another way would be to make it so that building with SSL support was a configuration option for us. Could be:
- Yes: use SSL, error if underlying C driver does not offer it
- No: Don't use SSL, even if underlying C driver does offer it
- Auto: SSL if C driver SSL.

Do you think that is worth adding to this commit? That would make it trivial to do a no-SSL build of the C++ driver without needing two C driver builds.

acmorrow: I filed https://jira.mongodb.org/browse/CXX-778 to track the above idea. I think for now we should move forward with this as is.

@samantharitter - Should I leave the error TODOs, or hold this until after you and @tessavitabile  have the updated error handling in?

samantharitter: I think leaving the TODOs is fine!

Machyne:
@acmorrow - great alternative! LGTM :)

PR #402, opened 2015-12-18
CXX-630 - bulk_write::append needs to handle k_uninitialized

tessavitabile:
I wasn't able to test, since I'm not sure it's possible to create a model::write with type k_uninitialized. There is no default constructor, and a moved-from model::write keeps its original type.

### DIFF
--- a/src/mongocxx/exception/private/error_category.cpp
+++ b/src/mongocxx/exception/private/error_category.cpp
@@ -19,6 +19,7 @@
 
 #include <mongocxx/exception/private/error_category.hpp>
 
+#include <mongocxx/exception/error_code.hpp>
 #include <mongocxx/private/libmongoc.hpp>
 
 namespace {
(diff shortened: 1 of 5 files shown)


tessavitabile:
...which is maybe a problem.

acmorrow:
@tessavitabile - You get all the easy tickets. I hadn't looked at this particular class before. It has many of the same flaws as the `value` type, but in this case, we can't easily punt and make it non-movable.

Lets put this on hold for a bit and focus on the testing improvements for a while so we can discuss the best way forward here.

tessavitabile:
Yes, it is very familiar! Sure, we can wait on this.

src/mongocxx/exception/private/error_category.cpp
```diff
@@ -382,7 +383,12 @@ class mongocxx_error_category_impl final : public std::error_category {
     }
 
     std::string message(int code) const noexcept override {
-        return "unknown mongocxx error";
+        switch(static_cast<error_code>(code)) {
+        case error_code::k_bulk_write_type_uninitialized:
+            return "bulk write type was k_uninitialized";
```
samantharitter: I realize we're going to leave this class alone for now and we may not need this error, but if we do end up using it, I'd prefer the message be less opaque, something like "cannot operate on uninitialized bulk write"

PR #403, opened 2015-12-18
CXX-739 string::view_or_value

samantharitter:
A new bsoncxx::string::view_or_value type that provides safe (null-terminated) conversion to const char*.

### DIFF
--- a/src/bsoncxx/string/view_or_value.cpp
+++ b/src/bsoncxx/string/view_or_value.cpp
@@ -1,0 +1,36 @@
+// Copyright 2015 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <bsoncxx/string/view_or_value.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace string {
+
+const char* view_or_value::c_str() {
+    // If we do not own our string, we cannot guarantee that it is null-terminated,
+    // so make an owned copy.
+    if (!_value) {
+        _value = std::move(_view.to_string());
+        _view = *_value;
+    }
+
+    return _value->c_str();
+}
+
+}  // namespace string
+BSONCXX_INLINE_NAMESPACE_END
+}  // namespace bsoncxx
--- a/src/bsoncxx/string/view_or_value.hpp
+++ b/src/bsoncxx/string/view_or_value.hpp
@@ -1,0 +1,104 @@
+// Copyright 2015 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <string>
+
+#include <bsoncxx/stdx/string_view.hpp>
+#include <bsoncxx/view_or_value.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace string {
+
+///
+/// Class representing a view-or-value variant type for strings.
+///
+/// This class adds several string-specific methods to the bsoncxx::view_or_value template:
+/// - a constructor overload for const char*
+/// - a constructor overload for std::string by l-value reference
+/// - a safe c_str() operation to return null-terminated c-style strings.
+///
+class BSONCXX_API view_or_value : public bsoncxx::view_or_value<stdx::string_view, std::string> {
+   public:
+    ///
+    /// Forward all bsoncxx::view_or_value constructors.
+    ///
+    using bsoncxx::view_or_value<stdx::string_view, std::string>::view_or_value;
+
+    ///
+    /// Construct a string::view_or_value using a null-terminated const char *.
+    /// The resulting view_or_value will keep a string_view of 'str', so it is
+    /// important that the passed-in string outlive this object.
+    ///
+    /// @param str A null-terminated string
+    ///
+    BSONCXX_INLINE view_or_value(const char* str)
+        : bsoncxx::view_or_value<stdx::string_view, std::string>(stdx::string_view(str)) {
+    }
+
+    ///
+    /// Allow construction with an l-value reference to a std::string. The resulting
+    /// view_or_value will keep a string_view of 'str', so it is important that the
+    /// passed-in string outlive this object.
+    ///
+    /// Construction calls passing a std::string by r-value reference will use the
+    /// constructor defined in the parent view_or_value class.
+    ///
+    /// @param str A std::string l-value reference.
+    ///
+    BSONCXX_INLINE view_or_value(const std::string& str)
+        : bsoncxx::view_or_value<stdx::string_view, std::string>(stdx::string_view(str)) {
+    }
+
+    ///
+    /// Return a null-terminated string from this string::view_or_value.
+    ///
+    /// @note if we are non-owning, we cannot assume our string_view is null-terminated.
+    /// In this case we must make a new, owning view_or_value with a null-terminated
+    /// copy of our view.
+    ///
+    /// @return A pointer to the original string, or a null-terminated copy.
+    ///
+    const char* c_str();
+};
+
+///
+/// Compare string::view_or_value directly with const char *.
+///
+
+BSONCXX_INLINE bool operator==(const view_or_value& lhs, const char* rhs) {
+    return lhs.view() == stdx::string_view(rhs);
+}
+
+BSONCXX_INLINE bool operator!=(const view_or_value& lhs, const char* rhs) {
+    return !(lhs == rhs);
+}
+
+BSONCXX_INLINE bool operator==(const char* lhs, const view_or_value& rhs) {
+    return rhs == lhs;
+}
+
+BSONCXX_INLINE bool operator!=(const char* lhs, const view_or_value& rhs) {
+    return !(rhs == lhs);
+}
+
+}  // namespace string
+BSONCXX_INLINE_NAMESPACE_END
+}  // namespace bsoncxx
+
+#include <bsoncxx/config/postlude.hpp>
--- a/src/bsoncxx/view_or_value.hpp
+++ b/src/bsoncxx/view_or_value.hpp
@@ -128,6 +128,22 @@ class view_or_value {
 };
 
 ///
+/// Compare view_or_value objects.
+///
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator==(const view_or_value<View, Value>& lhs,
+                               const view_or_value<View, Value>& rhs) {
+    return lhs.view() == rhs.view();
+}
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator!=(const view_or_value<View, Value>& lhs,
+                               const view_or_value<View, Value>& rhs) {
+    return !(lhs == rhs);
+}
+
+///
 /// Equality operators for comparison with plain Views.
 ///
 
@@ -148,6 +164,30 @@ BSONCXX_INLINE bool operator!=(const view_or_value<View, Value>& lhs, View rhs) 
 
 template <typename View, typename Value>
 BSONCXX_INLINE bool operator!=(View lhs, const view_or_value<View, Value>& rhs) {
+    return !(rhs == lhs);
+}
+
+///
+/// Equality operators for comparison with plain Values.
+///
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator==(const view_or_value<View, Value>& lhs, const Value& rhs) {
+    return lhs.view() == View(rhs);
+}
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator==(const Value& lhs, const view_or_value<View, Value>& rhs) {
+    return rhs == lhs;
+}
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator!=(const view_or_value<View, Value>& lhs, const Value& rhs) {
+    return !(lhs == rhs);
+}
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator!=(const Value& lhs, const view_or_value<View, Value>& rhs) {
     return !(rhs == lhs);
 }
 
--- a/src/bsoncxx/view_or_value.hpp
+++ b/src/bsoncxx/view_or_value.hpp
@@ -66,14 +66,14 @@ class view_or_value {
     /// @param value
     ///   A Value type.
     ///
-    BSONCXX_INLINE view_or_value(Value&& value) : _value{std::move(value)}, _view{_value->view()} {
+    BSONCXX_INLINE view_or_value(Value&& value) : _value(std::move(value)), _view(*_value) {
     }
 
     ///
     /// Construct a view_or_value from a copied view_or_value.
     ///
     BSONCXX_INLINE view_or_value(const view_or_value& other)
-        : _value{other._value}, _view{_value ? _value->view() : other._view} {
+        : _value(other._value), _view(_value ? *_value : other._view) {
     }
 
     ///
@@ -81,16 +81,17 @@ class view_or_value {
     ///
     BSONCXX_INLINE view_or_value& operator=(const view_or_value& other) {
         _value = other._value;
-        _view = _value ? _value->view() : other._view;
+        _view = _value ? *_value : other._view;
         return *this;
     }
 
     ///
     /// Construct a view_or_value from a moved-in view_or_value.
     ///
+
     /// TODO CXX-800: Create a noexcept expression to check the conditions that must be met.
     BSONCXX_INLINE view_or_value(view_or_value&& other) noexcept
-        : _value{std::move(other._value)}, _view{_value ? _value->view() : std::move(other._view)} {
+        : _value{std::move(other._value)}, _view(_value ? *_value : std::move(other._view)) {
         other._view = View();
         other._value = stdx::nullopt;
     }
@@ -101,7 +102,7 @@ class view_or_value {
     /// TODO CXX-800: Create a noexcept expression to check the conditions that must be met.
     BSONCXX_INLINE view_or_value& operator=(view_or_value&& other) noexcept {
         _value = std::move(other._value);
-        _view = _value ? _value->view() : std::move(other._view);
+        _view = _value ? *_value : std::move(other._view);
         other._view = View();
         other._value = stdx::nullopt;
         return *this;
@@ -120,7 +121,8 @@ class view_or_value {
         return _view;
     }
 
-   private:
+    // TODO: make these private again (CXX-801)
+   protected:
     stdx::optional<Value> _value;
     View _view;
 };
--- a/src/mongocxx/collection.cpp
+++ b/src/mongocxx/collection.cpp
@@ -74,7 +74,8 @@ mongocxx::stdx::optional<bsoncxx::document::value> find_and_modify(
 
     if (!r) {
         auto gle = mongocxx::libmongoc::collection_get_last_error(collection);
-        mongocxx::throw_exception<mongocxx::write_exception>(bsoncxx::helpers::value_from_bson_t(gle), error);
+        mongocxx::throw_exception<mongocxx::write_exception>(
+            bsoncxx::helpers::value_from_bson_t(gle), error);
     }
 
     bsoncxx::document::view result = reply.view();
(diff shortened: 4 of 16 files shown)


src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+#pragma once
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <string>
+
+#include <bsoncxx/stdx/string_view.hpp>
+#include <bsoncxx/view_or_value.hpp>
+
+#include <iostream>
```
acmorrow: Remove this header

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace string {
+
+using string_variant = bsoncxx::view_or_value<stdx::string_view, std::string>;
+
+///
+/// Class representing a view-or-value variant type for strings.
+///
+class BSONCXX_API view_or_value : public string_variant {
```
acmorrow: My guess is that this class is needed mostly for the `const char*` constructor, to work around the double conversion issue. If I've got that right, could you add some details to the documentation for it explaining that?

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    using string_variant::string_variant;
+
+    ///
+    /// Construct a string::view_or_value using a null-terminated const char *.
+    ///
+    /// @param str A null-terminated string
+    ///
+    BSONCXX_INLINE view_or_value(const char* str)
+        : bsoncxx::view_or_value<stdx::string_view, std::string>(std::string(str)) {
```
acmorrow: can just say : string_variant(...) to name the base class here?

acmorrow: You could chose to take the `const char *` either to a view, or to an owned std::string. The latter requires an allocation, copy, free cycle. Why is that the right choice, as opposed to making a view?

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// Return a null-terminated string from this string::view_or_value.
+    ///
+    /// @note if we are non-owning, we cannot assume our string_view is null-terminated.
+    /// In this case we must make a new, owning view_or_value with a null-terminated
+    /// copy of our view.
+    ///
+    /// @return A pointer to the original string, or a null-terminated copy.
+    ///
+    BSONCXX_INLINE const char* c_str() {
```
acmorrow: This class is concrete, so this method doesn't need to be inline.

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+        // If we own our string, it must be a std::string, return a const char*
+        if (_value) {
```
acmorrow: I think this method might be simpler if you reversed the logic:

if (!_value) {
    // become value-y
}

return _value->c_stR()

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <bsoncxx/stdx/string_view.hpp>
+#include <bsoncxx/view_or_value.hpp>
+
+#include <iostream>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace string {
+
+using string_variant = bsoncxx::view_or_value<stdx::string_view, std::string>;
```
acmorrow: Should anyone be using the string_variant type directly? If not, maybe don't make the public typedef. It will make the declarations below a little worse, but you can always make a private typedef in view_or_value below so you can at least refer to the super class constructors easily.

src/bsoncxx/view_or_value.hpp
```diff
@@ -118,7 +118,7 @@ class view_or_value {
         return _view;
     }
 
-   private:
+   protected:
```
acmorrow: I see why you did this so you can access them to implement c_str above, but I'm not sure you actually need to. I'd prefer to keep these private.

src/mongocxx/collection.cpp
```diff
@@ -119,24 +119,24 @@ collection::operator bool() const noexcept {
     return static_cast<bool>(_impl);
 }
 
-stdx::string_view collection::name() const noexcept {
-    return stdx::string_view{libmongoc::collection_get_name(_impl->collection_t)};
+std::string collection::name() const noexcept {
```
acmorrow: Why did this need to change?

src/mongocxx/private/collection.hpp
```diff
@@ -56,10 +55,10 @@ class collection::impl {
     }
 
     mongoc_collection_t* collection_t;
-    std::string database_name;
+    bsoncxx::string::view_or_value database_name;
```
acmorrow: Is this really what we want? When would we want the ownership of this data to be ambiguous?

src/mongocxx/private/database.hpp
```diff
@@ -48,8 +48,7 @@ class database::impl {
 
     mongoc_database_t* database_t;
     const class client::impl* client_impl;
-    std::string name;
-
+    bsoncxx::string::view_or_value name;
```
acmorrow: Same question here about why we want this

src/mongocxx/test/collection.cpp
```diff
@@ -35,14 +35,14 @@ TEST_CASE("collection copy", "[collection]") {
 
     std::string collname{"foo"};
     std::string collname2{"bar"};
-    collection coll = db[collname];
+    collection coll = db[stdx::string_view{collname}];
```
acmorrow: Is this actually necessary? I don't want it to be - passing a `std::string` to the `database::operator[]` really should work. Otherwise, really hard on the users of the classes.

src/mongocxx/test/collection.cpp
```diff
@@ -192,14 +197,16 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
 
         coll.update_one(b1.view(), update_doc.view(), options);
 
-        auto updated = coll.find_one(bsoncxx::document::view());
+        auto updated = coll.find_one({});
```
acmorrow: +1

src/mongocxx/test/collection.cpp
```diff
@@ -213,11 +220,11 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
...
 
         REQUIRE(updated);
         REQUIRE(updated->view()["changed"].get_bool() == true);
-        REQUIRE(coll.count(bsoncxx::document::view()) == 1);
+        REQUIRE(coll.count({}) == 1);
     }
 
     // SECTION("matching upsert updates document", "[collection]") {
```
acmorrow: Any idea why this block of tests is commented out?

src/mongocxx/test/collection.cpp
```diff
@@ -89,6 +88,8 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
     }
 
     SECTION("insert and read multiple documents", "[collection]") {
+        coll.delete_many({});
```
acmorrow: The additions of delete_many here and below... were those bugs in the tests before? If so, how were we not noticing, or, if not, why does this overall change require this change?

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -61,8 +66,12 @@ TEST_CASE("Collection", "[collection]") {
 
     client mongo_client{uri{}};
     write_concern concern;
-    database mongo_db = mongo_client[database_name];
-    collection mongo_coll = mongo_db[collection_name];
+    database mongo_db = mongo_client[stdx::string_view{database_name}];
```
acmorrow: Yeah, I really don't like that we need to do this now.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -61,8 +66,12 @@ TEST_CASE("Collection", "[collection]") {
...
+
+    // Clear out the database
+    mongo_db.drop();
```
acmorrow: Was this a bug before?

src/bsoncxx/view_or_value.hpp
```diff
@@ -66,14 +66,14 @@ class view_or_value {
     /// @param value
     ///   A Value type.
     ///
-    BSONCXX_INLINE view_or_value(Value&& value) : _value{std::move(value)}, _view{_value->view()} {
```
tessavitabile: I don't understand this change. Would you mind explaining it to me? Also, how were we able to take _value->view() before? I don't see how we know this is defined.

src/mongocxx/test/collection.cpp
samantharitter: Agreed, this isn't ideal.  Right now string::view_or_value has the same semantics as document::view_or_value, you can either move in a value-y type, or use a view.

If we want to allow users to just pass in a std::string without moving it, I think we'd want to copy the string and own the copy in the view_or_value.  Does that sound reasonable?  The other option, converting to a string_view, would lead to the same ownership issues that got us here in the first place.

src/mongocxx/test/collection.cpp
samantharitter: No, will investigate.

src/mongocxx/private/collection.hpp
samantharitter: If you wanted to create many database objects with the same name, you might want to do it using views of one string object, to avoid making copies of it.  It's the same argument for why you might want to use views of a document, except the case for a std::string is weaker.

Specifically for collections, you're probably creating your collections using `db["coll_name"]`, and the `database_name` member here is built from the parent db's stored name.  Before, we copied that std::string into every new collection, now we don't need to.

src/bsoncxx/view_or_value.hpp
samantharitter: Before, all of the type pairs that we used this template with (bsoncxx::document, bsoncxx::array) had Value classes with a .view() method.  This isn't something that we require of View/Value pairings, it was just a coincidence.  std::string doesn't have a .view(), it has no knowledge of the stdx::string_view class.  We do require that View types be constructible from Value types, though, so instead of relying on .view() we can use that property instead to construct Views everywhere in this class.

src/bsoncxx/view_or_value.hpp
samantharitter: I think we need access to _value in order to do c_str() the right way.  If we make a string copy off of the view, which we can access through view(), the copied string is temporary and the char* returned by c_str() ends up dangling.  I could add a public or protected getter and setter for _value and use those instead of accessing directly, would that be better?

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+///
+/// Class representing a view-or-value variant type for strings.
+///
+/// This class adds several string-specific methods to the bsoncxx::view_or_value template:
+/// - a constructor overload for const char*
+/// - a constructor overload for std::string by l-value reference, for convenience
+/// - a safe c_str() operation to return null-terminated c-style strings.
+///
+class BSONCXX_API view_or_value : public bsoncxx::view_or_value<stdx::string_view, std::string> {
```
amidvidy: should this be a template specialization as opposed to a subclass?

src/bsoncxx/string/view_or_value.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <bsoncxx/config/prelude.hpp>
+
+#include <bsoncxx/string/view_or_value.hpp>
+
+namespace bsoncxx {
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace string {
+
+using string_variant = bsoncxx::view_or_value<stdx::string_view, std::string>;
```
amidvidy: where do we use this typedef?

samantharitter: below, to invoke the parent class constructor

src/bsoncxx/string/view_or_value.hpp
samantharitter: I wouldn't make it a full specialization because I don't want to specialize the existing view_or_value functionality, just add some methods.  Defining these methods on the string specialization of view_or_value instead of on a subclass might be nice though, will experiment.

acmorrow:
@samantharitter is this ready for re-review?

samantharitter:
@acmorrow I left some questions for you on the review, I still have some things to fix but could you please re-review those?

src/bsoncxx/string/view_or_value.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+view_or_value::view_or_value(const char* str) : string_variant(std::string(str)) {
+}
+
+view_or_value::view_or_value(const std::string& str) : string_variant(stdx::string_view(str)) {
```
acmorrow: I think this constructor should be inline.

src/bsoncxx/string/view_or_value.hpp
acmorrow: I agree.

samantharitter:
I wanted to write up a little note about the travis failure I was running into because I don't entirely understand it. Clang wasn't able to parse `_view{*_value}` for documents (construct a new `document::view` from an existing `document::value`).  gcc is able to do this fine.

The error from clang:
```
no known conversion from 'value_type' (aka
      'bsoncxx::v0::document::value') to
      'const bsoncxx::v0::document::view
      &' for 1st argument
```
This is odd because the `document::value` class has an `operator document::view()`.  I ran into this issue once before, and looked at it briefly with @hanumantmk but we decided it was best to sidestep it by writing `_view{_value->view()}` and doing the conversion manually.  However, `std::string` doesn't have a `view()` method.

It turns out that this was only an issue because we were attempting to construct the view using a braced initializer list.  If we change from `_view{*_value}` to `_view(*_value)` this is fine on both gcc and clang.

Does anybody know and can explain why this makes a difference?

samantharitter:
So this might actually be a bug in gcc, and a feature in clang?  The compiler isn't supposed to consider user-defined conversions from U -> T when considering `T t{u}`:
```
"when considering the argument of a constructor [...] that is a
candidate by [...] by 13.3.1.7 [...] or when the initializer list has
exactly one element and a conversion to some class T or reference to
(possibly cv-qualified) T is considered for the first parameter of a
constructor of T [...] only standard conversion sequences and ellipsis
conversion sequences are considered."
```
https://gcc.gnu.org/ml/gcc-bugs/2014-12/msg02777.html

src/bsoncxx/string/view_or_value.hpp
acmorrow: Ping?

src/mongocxx/collection.cpp
```diff
@@ -123,25 +124,26 @@ collection::operator bool() const noexcept {
     return static_cast<bool>(_impl);
 }
 
-stdx::string_view collection::name() const {
-    return stdx::string_view{libmongoc::collection_get_name(_get_impl().collection_t)};
+std::string collection::name() const {
```
acmorrow: Why did this need to change? It seems that the underlying mongoc collection object keeps its string around, so it should be fine to return a view of that?

src/mongocxx/database.cpp
```diff
@@ -82,7 +82,8 @@ cursor database::list_collections(bsoncxx::document::view_or_value filter) {
 
     return cursor(result);
 }
-stdx::string_view database::name() const {
+
+bsoncxx::stdx::string_view database::name() const {
```
acmorrow: Doesn't look like this should have changed. We stopped saying bsoncxx::stdx in the mongocxx world a while ago.

src/mongocxx/test/collection.cpp
acmorrow: Ping

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,83 @@
...
+    ///
+    /// @note if we are non-owning, we cannot assume our string_view is null-terminated.
+    /// In this case we must make a new, owning view_or_value with a null-terminated
+    /// copy of our view.
+    ///
+    /// @return A pointer to the original string, or a null-terminated copy.
+    ///
+    const char* c_str();
+};
+
```
acmorrow: Should we add mixed mode comparators, like:

bool operator==(const view_or_value& lhs, const std::string& rhs)
bool operator!=(...)

bool operator==(const std::string& lhs, const view_or_value& rhs)
bool operator!=(...)

What about ones with const char*?

src/mongocxx/database.hpp
```diff
@@ -152,7 +152,7 @@ class MONGOCXX_API database {
     ///
     /// @return the name of this database.
     ///
-    stdx::string_view name() const;
+    bsoncxx::stdx::string_view name() const;
```
acmorrow: I'm puzzled why this one can return a `string_view` but `collection::name` can't

src/mongocxx/test/database.cpp
```diff
@@ -232,8 +232,7 @@ TEST_CASE("A database", "[database]") {
 
     SECTION("may create a collection") {
         MOCK_COLLECTION
-        // dummy_collection is the name the mocked collection_get_name returns
-        stdx::string_view collection_name{"dummy_collection"};
+        std::string collection_name{"dummy_collection"};
```
acmorrow: I think if you made the comparator operators I suggested above, you could just say:

auto collection_name = "dummy_collection";

src/mongocxx/test/database.cpp
```diff
@@ -317,7 +316,7 @@ TEST_CASE("Database integration tests", "[database]") {
 
         auto cursor = database.list_collections();
         for (auto&& coll : cursor) {
-            if (coll["name"].get_utf8().value == collection_name) {
+            if (coll["name"].get_utf8().value == stdx::string_view{collection_name}) {
```
acmorrow: Could maybe simplfiy with comparators as well?

src/mongocxx/test/database.cpp
```diff
@@ -267,7 +266,7 @@ TEST_CASE("Database integration tests", "[database]") {
     client mongo_client{uri{}};
     bsoncxx::stdx::string_view database_name{"database"};
     database database = mongo_client[database_name];
-    bsoncxx::stdx::string_view collection_name{"collection"};
+    std::string collection_name{"collection"};
```
acmorrow: Similarly, I think with the comparators, you could just make this a straightforward const char*.

acmorrow:
Overall I think this looks really good. Only things to consider are:
- I'm still confused about why collection::name returns a std::string now. It really doesn't seem to me like it needs to. Lets go over it in person, perhaps I'm just missing something important.
- I think adding some basic mixed type equality operators for the string view_or_value could be nice.

Other than those two, only small nits. Lets chat tomorrow and get this committed, as it is one of the last major interface changes we have pending (other than the great renaming of 2016).

src/bsoncxx/view_or_value.hpp
acmorrow: Want to discuss this one. I'd like us to try hard to keep them private.

src/bsoncxx/string/view_or_value.hpp
acmorrow: So, I took a look at the base view_or_value template again. I think we are missing some equality operations. In particular, I think we definitely want to add:

```
template <typename View, typename Value>
BSONCXX_INLINE bool operator==(const view_or_value<View, Value>& lhs, const view_or_value<View, Value>& rhs)
```
and the associated `!=`.

However, given that we require that the `View` type `is_constructible` from `Value`, I think it might also be good to add the various permutations around:

```
template <typename View, typename Value>
BSONCXX_INLINE bool operator==(const view_or_value<View, Value>& lhs, const Value& rhs)
```
and the associated rhs/lhs swaps and the `!=` variants.

I still think you want to manually add the (in)equality overloads for the string specialization of view_or_value against `const char*`, given the special relationship between `const char*` and `std::string`

src/bsoncxx/string/view_or_value.hpp
samantharitter: I got rid of the string_variant using statement, I didn't think we wanted it in a public header.  Should I add it back?

Re: whether `const char*` becomes a view or a value, I'm fine with it becoming, you're right that this makes more sense given how we handle `const std::string&`.  So the only way to construct an owning string::view_or_value is to std::move in a `std::string`.

acmorrow: No, you don't need to add back the typedef. I think was an old comment and I'd missed that you had removed it.

src/mongocxx/test/database.cpp
samantharitter: Not this one, there's no built-in comparison between a `std::string_view` to a `const char*`.  I think this has to stay a string_view.

src/mongocxx/test/database.cpp
samantharitter: I don't think so, unfortunately, unless we want to add our own `stdx::string_view` to `const char*` comparison operators.  It's sad that these do not exist.

src/bsoncxx/string/view_or_value.hpp
samantharitter: I added the following sets of comparison operators:

1. `view_or_value` and `view_or_value` - compare equal if their views are equal, regardless of ownership

2. `view_or_value` and `View` (covers the `std::string` cases)

3. `string::view_or_value` and `const char*`

samantharitter:
@acmorrow I've pushed revisions to address all of those areas, but I think we still need to discuss how to deal with c_str(), and whether we can implement it without exposing the members of view_or_value

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -0,0 +1,93 @@
...
+///
+/// Compare string::view_or_value directly with const char *.
+///
+
+bool operator==(view_or_value lhs, const char* rhs);
```
acmorrow: all view_or_value args here should be by const&

src/bsoncxx/string/view_or_value.cpp
```diff
@@ -0,0 +1,52 @@
...
+    // so make an owned copy.
+    if (!_value) {
+        _value = std::move(_view.to_string());
+        _view = *_value;
+    }
+
+    return _value->c_str();
+}
+
+bool operator==(view_or_value lhs, const char* rhs) {
```
acmorrow: I think all of these should be inline. In particular, this one, since if the `const char *` is a literal, we want the compiler to be able to see that we are making a `string_view` of it so it can (hopefully) elide the call to `strlen`.

src/bsoncxx/view_or_value.hpp
```diff
@@ -147,6 +163,30 @@ BSONCXX_INLINE bool operator!=(View lhs, const view_or_value<View, Value>& rhs)
     return !(rhs == lhs);
 }
 
+///
+/// Equality operators for comparison with plain Values.
+///
+
+template <typename View, typename Value>
+BSONCXX_INLINE bool operator==(const view_or_value<View, Value>& lhs, Value rhs) {
```
acmorrow: Definitely const& for the Values here, otherwise you are making a copy of something with ownership

src/bsoncxx/view_or_value.hpp
acmorrow: I'm fine with it, please though add a TODO: make these members private. I don't want to hold up this change over it.

acmorrow:
LGTM mod making some arguments `const &` and inlining a few of the comparators. No need to re-post, just fix those up and push.

src/bsoncxx/view_or_value.hpp
samantharitter: Done, I've opened [CXX-801](https://jira.mongodb.org/browse/CXX-801) to track that change.

acmorrow:
LGTM. Ship it!

PR #405, opened 2015-12-21
CXX-742 database operations throw if database is uninitialized

tessavitabile:


src/mongocxx/database.cpp
```diff
@@ -157,5 +164,12 @@ collection database::collection(stdx::string_view name) const {
     return mongocxx::collection(*this, name);
 }
 
+const std::unique_ptr<database::impl>& database::get_impl() const {
+    if(!_impl) {
+        throw operation_exception{make_error_code(error_code::k_database_uninitialized)};
```
tessavitabile: Do we want to use operation_exception? Also, will it work to use the same type of exception in all cases?

src/mongocxx/database.cpp
```diff
@@ -44,11 +46,16 @@ database::database(const class client& client, stdx::string_view name)
           name.data())) {
 }
 
-database::database(const database& d) : _impl{stdx::make_unique<impl>(*(d._impl))} {
+database::database(const database& d) {
```
tessavitabile: Should default constructed databases be copyable?

src/mongocxx/database.hpp
```diff
@@ -248,6 +248,7 @@ class MONGOCXX_API database {
 
     class MONGOCXX_PRIVATE impl;
     std::unique_ptr<impl> _impl;
+    const std::unique_ptr<impl>& get_impl() const;
```
tessavitabile: I wasn't sure about naming the accessor, since the class is named impl and the field is named _impl.

src/mongocxx/exception/private/error_code.cpp
```diff
@@ -18,7 +18,7 @@
 
 #include <mongocxx/exception/private/error_code.hpp>
 
-#include <mongocxx/exception/error_category.hpp>
+#include <mongocxx/exception/private/error_category.hpp>
```
tessavitabile: This was a mistake from mongocxx error handling.

src/mongocxx/database.hpp
tessavitabile: Is returning a reference okay? Or should we make _impl a shared_ptr?

src/mongocxx/database.hpp
```diff
@@ -248,6 +248,7 @@ class MONGOCXX_API database {
 
     class MONGOCXX_PRIVATE impl;
     std::unique_ptr<impl> _impl;
```
tessavitabile: Nothing prevents us from continuing to use _impl.

src/mongocxx/database.cpp
samantharitter: Yes?

src/mongocxx/database.hpp
samantharitter: I think get_impl() is fine, hooray for moving away from ambiguous names!

Will callers of this method ever want to alter the impl's state?  It seems ok to let them do so, in which case maybe it's better to return an impl* than a const unique_ptr&

src/mongocxx/database.cpp
samantharitter: operation_exception is fine with me.  I'm not sure what you mean by the second question.

tessavitabile: I meant that get_impl() is called by a variety of database methods, and I'm not sure all of these methods would want to throw an operation_exception. Probably for databases, this is okay, but if we do the same thing for client and collection, I don't know whether we would want all methods to throw operation_exception, or have some of them throw query_exception, etc.

samantharitter: As operation_exception is the least common denominator, I think we should throw that here.  If other methods make calls that might trigger this exception, they can always catch and re-throw a new exception with a more specific type.

src/mongocxx/database.hpp
tessavitabile: Then should get_impl() be const? This doesn't seem right, since it gives callers a way to alter the impl's state. However, if it's not const, then it can't be used in const functions.

samantharitter: The most logically correct approach would be to provide a const getter and a separate non-const getter.   Also, it seems strange to return any sort of unique_ptr, I'd rather we just returned a reference or bare ptr to the underlying _impl:
```
// const getter
const impl& get_impl() const;
// non-const getter
impl* get_impl();
```
This seems like overkill, though you could have the non-const getter simply call the const getter.

tessavitabile: Having the const and non-const getters seems fine.

The _impl field is currently a std::unique_ptr<impl>. I'm not sure why--perhaps so we don't have to destroy it in the destructor? Do you think I should change it to an impl*?

samantharitter: The _impl field should definitely be a std::unique_ptr, I just don't think the getters should return that.  We're using std::unique_ptr here for lifetime management, so that _impl will get properly cleaned up when the owning database/collection/other object is destroyed.  Because of these semantics, a std::unique_ptr should only be held/used by one thing.  I'd implement the getters like so:
```
const impl& get_impl() const {
   // raise an error etc.
   return *_impl;
}
```
So instead, we return a reference/pointer to the underlying impl for temporary, non-owning access.  I kind of think of it like you're walking a dog, and only one person can hold the leash, but other people can stop and pet the dog.

src/mongocxx/database.cpp
acmorrow: Definitely should be ok to copy default constructed

src/mongocxx/database.cpp
```diff
@@ -157,5 +164,16 @@ collection database::collection(stdx::string_view name) const {
     return mongocxx::collection(*this, name);
 }
 
+const database::impl& database::get_impl() const {
+    if(!_impl) {
+        throw operation_exception{make_error_code(error_code::k_database_uninitialized)};
```
acmorrow: Is operation_exception really the right type here? I thought that was meant to represent an operation that failed on the server. I think we either want to just throw mongocxx::exception here, or consider making a mongocxx::logic_error or similar that inherits from mongocxx::exception for this type of issue.

src/mongocxx/database.hpp
```diff
@@ -248,6 +248,8 @@ class MONGOCXX_API database {
...
+    const impl& get_impl() const;
```
acmorrow: Private methods should start with _. Also, private members should have MONGOCXX_PRIVATE on them in the declaration.

src/mongocxx/exception/error_code.hpp
```diff
@@ -22,6 +22,7 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 
 enum class error_code : std::int32_t {
+	k_database_uninitialized = 1,
```
acmorrow: 'uninitialized' means something pretty specific in C++ land, and error codes are hard to change once you make them. Can we bikeshed a better name?

src/mongocxx/exception/private/error_category.cpp
```diff
@@ -382,7 +383,12 @@ class mongocxx_error_category_impl final : public std::error_category {
     }
 
     std::string message(int code) const noexcept override {
-        return "unknown mongocxx error";
+        switch (static_cast<error_code>(code)) {
+        case error_code::k_database_uninitialized:
+            return "database is uninitialized";
```
acmorrow: Lets make this a little more detailed. 'database' is ambiguous, overall in our context. I think adding the classname here would be better: "mongocxx::database object used when not initialized". Same overall comment about 'initialized", but I don't immediately have a better suggestion

src/mongocxx/database.hpp
```diff
@@ -248,6 +248,8 @@ class MONGOCXX_API database {
...
+    impl* get_impl();
```
acmorrow: Why does this return a pointer and the above one return a reference?

src/mongocxx/database.cpp
```diff
@@ -157,5 +164,16 @@ collection database::collection(stdx::string_view name) const {
...
+    }
+    return *_impl;
+}
+
+database::impl* database::get_impl() {
+    return &(const_cast<database::impl&>(static_cast<const database&>(*this).get_impl()));
```
acmorrow: Quite an expression!

I think it could work better as multiple lines. I'm assuming here you change the return type to & to match the const get_impl

```
auto cthis = const_cast<const database*>(this);
return const_cast<database::impl&>(cthis->get_impl());
```

acmorrow:
I think this looks pretty good. A few small comments and suggestions. What other types should we do this for? I don't want to do it for *everything*, but at least a handful of things that people are likely to make in the default state and then mis-use seems reasonable, so maybe 'client', 'collection', and 'database'?

src/mongocxx/exception/private/error_category.cpp
tessavitabile: Perhaps "cannot use a default-constructed mongocxx::database object"?

acmorrow: That is good, with a small quibble: you can use one, just not for much. How about "invalid use of default-constructed mongocxx::database object"?. That still isn't perfect because it doesn't cover the moved-from case, but it is close enough, I think. Saying "invalid use of default constructed or moved-from mongocxx::database object" seems awfully wordy. What do you think?

src/mongocxx/database.hpp
amidvidy: +1 it is fairly unexpected to have there return different types (modulo-constness)

src/mongocxx/database.cpp
tessavitabile: I wasn't sure about this, since I didn't know whether all methods that call get_impl would want to throw the same kind of exception. But it makes sense that it is a logical error in all cases, so I like the idea of a mongocxx::logic_error.

src/mongocxx/database.cpp
amidvidy: can you put the implementation of this function in the non-const overload so you can avoid the need for const_cast?

src/mongocxx/exception/private/error_category.cpp
tessavitabile: I'd be fine with "invalid use of default constructed or moved-from mongocxx::database object".

tessavitabile:
'client', 'collection' and 'database' seem reasonable, and that was what was discussed in the ticket.

src/mongocxx/exception/error_code.hpp
tessavitabile: k_missing_database_impl? k_database_default_constructed_or_moved_from?

src/mongocxx/database.cpp
acmorrow: OK, lets add it. Should be a trivial addition.

src/mongocxx/exception/error_code.hpp
acmorrow: How about k_invalid_database_object. The error message will clarify the details.

src/mongocxx/database.cpp
acmorrow: Wasn't there some sort of exception throwing function that we were using in mongocxx?

src/mongocxx/database.hpp
tessavitabile: @samantharitter suggested the return types for these functions. I guessed she made the return type of the const function a reference so that we wouldn't need to make both the pointer and the contents of the pointer const, but I wasn't totally sure. Samantha, would you mind explaining your reasoning?

src/mongocxx/database.cpp
tessavitabile: Only for converting bson_error_t's. So it only applies to libmongoc errors.

src/mongocxx/database.cpp
acmorrow: You would still need const_cast I think. I think I like it how it is with non-const implemented in terms of const. You can't get in as much trouble that way.

src/mongocxx/exception/private/error_category.cpp
acmorrow: OK, lets use that then.

tessavitabile:
Should I do 'client' and 'collection' in the same PR?

src/mongocxx/database.cpp
acmorrow: got it. thanks.

src/mongocxx/database.hpp
```diff
@@ -247,7 +247,9 @@ class MONGOCXX_API database {
     MONGOCXX_PRIVATE database(const class client& client, stdx::string_view name);
 
     class MONGOCXX_PRIVATE impl;
-    std::unique_ptr<impl> _impl;
+    MONGOCXX_PRIVATE std::unique_ptr<impl> _impl;
+    MONGOCXX_PRIVATE const impl& _get_impl() const;
```
acmorrow: Methods before members, with a blank line in between. I think this would be ideal:

```
class MONGOCXX_PRIVATE impl;

MONGOCXX_PRIVATE std::unique_ptr<impl> _impl;
MONGOCXX_PRIVATE const impl& _get_impl() const;

std::unique_ptr<impl> _impl;
```

acmorrow:
LGTM with a request for a minor member ordering change. Should we continue in this review and implement for client and collection, or do those in a separate review?

src/mongocxx/database.hpp
```diff
@@ -247,7 +247,9 @@ class MONGOCXX_API database {
...
+    MONGOCXX_PRIVATE impl* _get_impl();
```
acmorrow: Wait, hold that LGTM: This still needs to have the same return type as the const overload.

acmorrow: But LGTM once the const and non-const overloads have the same return type.

acmorrow:
@tessavitabile - Sure, lets roll client and collection in to this review as well, now that we have an LGTM on the database changes. It should be very straightforward. Just ping back the review when you have them in (and the two other changes I requested to database) and I'll take another look.

src/mongocxx/database.hpp
tessavitabile: Yeah sorry, I held off on making that change, since @samantharitter had suggested the return types, and I wanted to ask for her input.

samantharitter: I defer to @acmorrow on the return types

tessavitabile: Okay. Do you think they should both be pointers or references?

samantharitter: I think references, since returning a ptr could imply it might be null, which should not ever happen.

tessavitabile:
@acmorrow, @samantharitter, @amidvidy, the client and collection parts are ready for review.

src/mongocxx/client.hpp
```diff
@@ -226,7 +226,11 @@ class MONGOCXX_API client {
     MONGOCXX_PRIVATE explicit client(void* implementation);
 
     class MONGOCXX_PRIVATE impl;
-    std::unique_ptr<impl> _impl;
+
+    MONGOCXX_PRIVATE impl& _get_impl();
+    MONGOCXX_PRIVATE const impl& _get_impl() const;
+
+    MONGOCXX_PRIVATE std::unique_ptr<impl> _impl;
```
acmorrow: nit: you don't need the MONGOCXX_PRIVATE on members. Only on types and functions.

src/mongocxx/collection.hpp
```diff
@@ -474,7 +474,7 @@ class MONGOCXX_API collection {
     ///
     /// @return The name of the collection.
     ///
-    stdx::string_view name() const noexcept;
+    stdx::string_view name() const;
```
acmorrow: +1 nice catch.

src/mongocxx/collection.hpp
```diff
@@ -609,7 +609,11 @@ class MONGOCXX_API collection {
     MONGOCXX_PRIVATE collection(const database& database, void* collection);
 
     class MONGOCXX_PRIVATE impl;
-    std::unique_ptr<impl> _impl;
+
+    MONGOCXX_PRIVATE impl& _get_impl();
+    MONGOCXX_PRIVATE const impl& _get_impl() const;
+
+    MONGOCXX_PRIVATE std::unique_ptr<impl> _impl;
```
acmorrow: ditto re private macro

src/mongocxx/database.hpp
```diff
@@ -247,7 +247,11 @@ class MONGOCXX_API database {
     MONGOCXX_PRIVATE database(const class client& client, stdx::string_view name);
 
     class MONGOCXX_PRIVATE impl;
-    std::unique_ptr<impl> _impl;
+
+    MONGOCXX_PRIVATE impl& _get_impl();
+    MONGOCXX_PRIVATE const impl& _get_impl() const;
+
+    MONGOCXX_PRIVATE std::unique_ptr<impl> _impl;
```
acmorrow: here too

acmorrow:
LGTM

src/mongocxx/database.cpp
```diff
@@ -157,5 +165,17 @@ collection database::collection(stdx::string_view name) const {
...
 
+const database::impl& database::_get_impl() const {
+    if(!_impl) {
+        throw logic_error{make_error_code(error_code::k_invalid_database_object)};
+    }
+    return *_impl;
+}
+
+database::impl& database::_get_impl() {
+    auto cthis = const_cast<const database*>(this);
```
amidvidy: is this const_cast-fu worth it? or should we actually just duplicate the implementation?

amidvidy:
LGTM aside from const-cast question

acmorrow:
@amidvidy - I think it is fine. We can always revisit later since these are implementation private if we change our minds.

src/mongocxx/database.cpp
tessavitabile: Scott Meyers recommends having the non-const method call the const method, rather than duplicating implementation.

acmorrow: I agree. Lets go with what we have.

amidvidy:
Cool, i'll defer to Tess on this one

src/mongocxx/client.cpp
```diff
@@ -115,5 +118,17 @@ cursor client::list_databases() const {
...
 
+const client::impl& client::_get_impl() const {
+    if(!_impl) {
+        throw logic_error{make_error_code(error_code::k_invalid_client_object)};
+    }
+    return *_impl;
+}
+
+client::impl& client::_get_impl() {
+    auto cthis = const_cast<const client*>(this);
```
samantharitter: you can use const with auto to capture, and then I don't think you'd need the const_case:
```
const auto const_this = this;
```

src/mongocxx/collection.cpp
```diff
@@ -266,13 +274,13 @@ cursor collection::aggregate(const pipeline& pipeline, const options::aggregate&
         rp_ptr = read_preference()._impl->read_preference_t;
     }
 
-    return cursor(libmongoc::collection_aggregate(_impl->collection_t,
+    return cursor(libmongoc::collection_aggregate(_get_impl().collection_t,
                                                   static_cast<::mongoc_query_flags_t>(0),
                                                   stages.bson(), options_bson.bson(), rp_ptr));
 }
 
 void* collection::implementation() const {
```
samantharitter: I wonder why we have this method

src/mongocxx/client.cpp
acmorrow: Won't that make the pointer const, but not the pointed to type?

tessavitabile:
After rebasing, I had the following error:

/Users/Tess/mongo-cxx-driver/src/mongocxx/test/database.cpp:298: FAILED:
due to unexpected exception with message:
  unknown option to collMod: validator: mongoc query_command_not_found error

I got it on master as well. I tried rebuilding the C driver, but it didn't help. Does anyone know what might be the issue?

src/mongocxx/collection.cpp
acmorrow: It is a backdoor into the C driver in case there is something you can't do. We should probably remove it before GA.

samantharitter:
You are probably using an outdated version of the server, the document
validation options are only available on 3.2 and newer.

tessavitabile:
Ah, thanks.

PR #406, opened 2015-12-27
CXX-788 Fix some -Wall warnings under clang and gcc

acmorrow:
Fixes the warnings identified in CXX-788 and several others.

### DIFF
--- a/src/bsoncxx/private/error_category.cpp
+++ b/src/bsoncxx/private/error_category.cpp
@@ -64,6 +64,8 @@ class bsoncxx_error_category_impl final : public std::error_category {
                 return "can't convert builder to a valid view: unmatched key";
             case error_code::k_unset_element:
                 return "unset document::element";
+            default:
+                return "unknown bsoncxx error code";
         }
     }
 };
@@ -120,7 +122,7 @@ class libbson_unknown_error final : public std::error_category {
         return "libbson unknown error";
     }
 
-    std::string message(int code) const noexcept override {
+    std::string message(int) const noexcept override {
         return "unknown libbson error in unknown domain";
     }
 };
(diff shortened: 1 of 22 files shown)


src/bsoncxx/private/error_category.cpp
```diff
@@ -120,7 +122,7 @@ class libbson_unknown_error final : public std::error_category {
         return "libbson unknown error";
     }
 
-    std::string message(int code) const noexcept override {
```
samantharitter: Was having an unused named parameter giving us an unused variable warning?

src/mongocxx/exception/error_code.hpp
```diff
@@ -25,6 +25,7 @@ enum class error_code : std::int32_t {
 	k_invalid_client_object = 1,
 	k_invalid_collection_object,
 	k_invalid_database_object,
+        k_invalid_parameter,
```
samantharitter: extra space?

samantharitter: Also, can you add a case for `k_invalid_parameter` to `mongocxx_error_category_impl::message()` ?

src/bsoncxx/private/error_category.cpp
acmorrow: Yes, with -Wextra. It is pretty pedantic, but it is fair that you declared a name for a parameter but then didn't use it.

src/mongocxx/exception/error_code.hpp
acmorrow: The weird spacing is an artifact - apparently this list is using leading tabs? I've cleaned it up, but we can leave the real fix for clang-format after rc0

acmorrow: Added to the message enum, good catch, thanks.

acmorrow:
Thanks @samantharitter, updated. PTAL.

src/bsoncxx/private/error_category.cpp
samantharitter: Makes sense, thanks

samantharitter:
LGTM!

PR #408, opened 2015-12-27
CXX-786 Remove deprecated 'implementation' methods

acmorrow:
This weakens some tests. @Machyne I think maybe these were tests you added? I'd like to understand how much we think we have lost.

Also, I'm puzzled why we didn't get deprecation warnings for use in the tests. I will look into that .

### DIFF
--- a/src/mongocxx/bulk_write.cpp
+++ b/src/mongocxx/bulk_write.cpp
@@ -87,10 +87,6 @@ void bulk_write::append(const model::write& operation) {
     }
 }
 
-void* bulk_write::implementation() const {
-    return _impl->operation_t;
-}
-
 void bulk_write::bypass_document_validation(bool bypass_document_validation) {
     libmongoc::bulk_operation_set_bypass_document_validation(_impl->operation_t,
                                                              bypass_document_validation);
--- a/src/mongocxx/bulk_write.hpp
+++ b/src/mongocxx/bulk_write.hpp
@@ -89,18 +89,6 @@ class MONGOCXX_API bulk_write {
     void append(const model::write& operation);
 
     ///
-    /// Gets a handle to the underlying implementation.
-    ///
-    /// Returned pointer is only valid for the lifetime of this object.
-    ///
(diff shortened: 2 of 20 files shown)


amidvidy:
what is the actual reason for removing these? Seems like a good escape hatch to have in the case that the C driver supports something we don't, e.g. GridFS

acmorrow:
My recollection of why we added them was so that things that were completely impossible to do early on (like when @jbreams was experimenting early on and couldn't set a write concern or something). If you wanted to do GridFS with the current driver, you would just issue manual commands via run_command.

If we leave them in when we ship 3.0.0, I'd be really worried that users would come to depend on them, even though they shouldn't.

amidvidy:
We've already labeled these methods deprecated and stated that they can stop working at any time. Until we actually offer a superset of the C driver's functionality, I personally don't think it is in driver users' interests to remove them.

amidvidy:
Also, consider the case that in the future, the C driver releases some new functionality that we don't wrap/support. The implementation methods will allow users to access such features without waiting for us to catch up. Code that depends on implementation() will work as long as the C-driver doesn't break ABI. If the C driver does break ABI, the C++ driver is broken regardless.

samantharitter:
I agree that we should remove these methods, for the sake of keeping the driver clean and releasing a 1.0.0 without cruft.  It's true that in the future the C driver could release new functionality that we didn't support, but it seems unlikely.  Given the C driver's relative feature-sparse-ness and our history of working with them, we tend to request features and they respond, not vice versa.  That said we could always add an implementation() method back to any class that needed it should we get into a difficult situation.

acmorrow:
I'd also worry that if we gave people a back channel into the implementation, that if there were things they wanted to do but couldn't and they fall back to the `implementation` methods, then they won't file bugs asking for the features they need, and also become dependent on the underlying abstraction, which we have gone to great efforts to hide. I'm inclined to do this. @hanumantmk any thoughts?

acmorrow:
While this is controversial, I've decided that we want to do it.

PR #409, opened 2015-12-29
CXX-791 Options for bulk ops should be on the options::bulk_op type

acmorrow:
While looking into making an example for producing bulk writes per the user request in CXX-791, I noticed that the bulk write subsystem was a little broken.

The options for a bulk write belong on options::bulk_write, not on libmongocx::bulk_write, which should simply be an accumulator for model::write objects.

Instead, the options should be provided via options::bulk_write when creating a bulk_write object.

### DIFF
--- a/src/mongocxx/bulk_write.cpp
+++ b/src/mongocxx/bulk_write.cpp
@@ -33,8 +33,16 @@ bulk_write& bulk_write::operator=(bulk_write&&) noexcept = default;
 
 bulk_write::~bulk_write() = default;
 
-bulk_write::bulk_write(bool ordered)
-    : _impl(stdx::make_unique<impl>(libmongoc::bulk_operation_new(ordered))) {
+bulk_write::bulk_write(options::bulk_write options)
+    : _impl(stdx::make_unique<impl>(libmongoc::bulk_operation_new(options.ordered()))) {
+    auto options_wc = options.write_concern();
+    if (options_wc)
+        libmongoc::bulk_operation_set_write_concern(_impl->operation_t,
+                                                    options_wc->_impl->write_concern_t);
+
+    auto options_bdv = options.bypass_document_validation();
+    if (options_bdv)
+        libmongoc::bulk_operation_set_bypass_document_validation(_impl->operation_t, *options_bdv);
 }
 
 void bulk_write::append(const model::write& operation) {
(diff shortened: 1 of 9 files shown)


src/mongocxx/bulk_write.cpp
```diff
@@ -33,8 +33,16 @@ bulk_write& bulk_write::operator=(bulk_write&&) noexcept = default;
 
 bulk_write::~bulk_write() = default;
 
-bulk_write::bulk_write(bool ordered)
-    : _impl(stdx::make_unique<impl>(libmongoc::bulk_operation_new(ordered))) {
+bulk_write::bulk_write(options::bulk_write options)
+    : _impl(stdx::make_unique<impl>(libmongoc::bulk_operation_new(options.ordered()))) {
+    auto options_wc = options.write_concern();
```
tessavitabile: Is this necessary because options.write_concern() returns an optional? Is there any problem with the way options.ordered() is used above, since it also returns an optional?

tessavitabile: Never mind, I saw options.ordered() no longer an optional.

tessavitabile:
LGTM

src/mongocxx/bulk_write.cpp
samantharitter: The "ordered" parameter isn't stored or returned as an optional anymore.  I'm wondering if we want to make the decision about default ordered at the options class level, the way it is now, or defer that decision, the way it was before?

acmorrow:
@samantharitter could you give this a look too?

src/mongocxx/collection.cpp
```diff
@@ -319,17 +323,20 @@ stdx::optional<result::insert_one> collection::insert_one(view_or_value document
 stdx::optional<result::replace_one> collection::replace_one(view_or_value filter,
                                                             view_or_value replacement,
                                                             const options::update& options) {
-    class bulk_write bulk_op(false);
-    model::replace_one replace_op(filter, replacement);
+    options::bulk_write bulk_opts;
+    bulk_opts.ordered(false);
+
+    if (options.bypass_document_validation())
```
samantharitter: is there a way to DRY this?  What if we allowed construction of a bulk_write options object using an existing update/insert/delete options object?

samantharitter: Or, if we allow update/insert/delete options to convert to bulk_write options, so they can be passed directly into the construction of a bulk_write object?

src/mongocxx/options/bulk_write.hpp
```diff
@@ -70,10 +77,26 @@ class MONGOCXX_API bulk_write {
     /// @see http://docs.mongodb.org/manual/core/write-concern/
     ///
     const stdx::optional<class write_concern>& write_concern() const;
+    ///
```
samantharitter: Nit, newline above this comment

samantharitter:
LGTM mod a style nit and question about how we construct bulk write options

src/mongocxx/bulk_write.cpp
acmorrow: I think what we have now is pretty good. The `::bulk_write` constructor will default construct an `options::bulk_write` object, and that object defaults to ordered (which is the only safe default). I think you should know at the time you are constructing the holder for the individual ops whether you are planning those to have ordered or unordered semantics.

src/mongocxx/collection.cpp
acmorrow: I'm going to add a TODO to revisit that idea. I think it is interesting but I don't want to delay this work behind it.

src/mongocxx/options/bulk_write.hpp
acmorrow: Done

PR #410, opened 2015-12-30
WIP Support boost for C++11 driver polyfills

acmorrow:
This is one way around the fact that MSVC can't compile MNMLSTC/core. Things that made me sad:

- Had to change some perfectly fine optional returns to make boost happy.
- Had to drop some static asserts because boost amazingly doesn't require the trivial constructors of string_ref to be noexcept.


src/bsoncxx/types/value.cpp
```diff
@@ -30,8 +30,6 @@ namespace types {
 #define BSONCXX_ENUM(name, val)                                                                \
     value::value(b_##name value) noexcept : _type(static_cast<bsoncxx::type>(val)),            \
                                             _b_##name(std::move(value)) {                      \
-        static_assert(std::is_nothrow_copy_constructible<b_##name>::value, "Copy may throw");  \
```
acmorrow: This is very sad. I really don't want to drop these static asserts but boost::string_ref doesn't declare its copy ops noexcpet

acmorrow: I've fought the good fight so far on avoiding use of macros, but I'm inclined to make the inclusion of the static asserts conditional on the source of polyfills.

samantharitter:
Makes sense, those few things are a bit sad.  LGTM though!

PR #411, opened 2015-12-30
CXX-783 Moved OpenSSL checking before boost

emzar:
Looks like we don't need a CheckLib for SSL, but just need to do CheckLinkSSL before checking boost stuff.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1745,6 +1745,41 @@ def doConfigure(myenv):
 
     conf = Configure(myenv)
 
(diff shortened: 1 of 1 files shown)


acmorrow:
Hi @emzar - Thanks for the PR. I have a similar change working its way through review internally. The current plan is to fix it upstream in the server SConstruct, then backport it to the driver, so that they stay more in sync. If, for some reason, that plan doesn't work out, then I agree we can just do what you have proposed here.

PR #412, opened 2016-01-01
CXX-760 Don't rely on pkg-config and get the basic build working on windows

acmorrow:
Lots of CMake improvements needed to get the basic driver build process up and running on Windows. While doing so, made various required compile fixes as pointed out by MSVC. Also adds support for building and installing CMake compatible find_package support for bsoncxx and mongocxx.

This plus the boost changes in https://github.com/mongodb/mongo-cxx-driver/pull/410 get you a more or less working build environment on Windows. There are still many problems there. In particular VS2015's compiler is ICE-ier than a skating rink and rejects a lot of valid code. The great renaming will fix some of those, but not others, and not the ICEs.

I'm not really expecting anyone to dig into this too closely unless you are really fascinated by CMake. I'll probably push that and the boost changes TBR since this is a long weekend.


cmake/FindLibBSON.cmake
```diff
@@ -0,0 +1,24 @@
...
+# Load up PkgConfig if we have it
+find_package(PkgConfig QUIET)
+
+if (PKG_CONFIG_FOUND)
+  pkg_check_modules(LIBBSON REQUIRED libbson-1.0>=${LibBSON_FIND_VERSION} )
+  # We don't reiterate the version information here because we assume that
+  # pkg_check_modules has honored our request.
+  find_package_handle_standard_args(LIBBSON DEFAULT_MSG LIBBSON_FOUND)
+elseif(LIBBSON_DIR)
+  # The best we can do until libbson starts instaling a libbson-config.cmake file
```
samantharitter: typo, "installing"

cmake/FindLibBSON.cmake
```diff
@@ -0,0 +1,24 @@
...
+  set(LIBBSON_LIBRARIES bson-1.0 CACHE INTERNAL "")
+  set(LIBBSON_LIBRARY_DIRS ${LIBBSON_DIR}/lib CACHE INTERNAL "")
+  set(LIBBSON_INCLUDE_DIRS ${LIBBSON_DIR}/include/libbson-1.0 CACHE INTERNAL "")
+  find_package_handle_standard_args(LIBBSON DEFAULT_MSG LIBBSON_LIBRARIES LIBBSON_LIBRARY_DIRS LIBBSON_INCLUDE_DIRS)
+else()
+  message(FATAL_ERROR "Don't know how to find libbson, please set LIBBSON_DIR")
```
markbenvenuto: what should it be set to?

src/bsoncxx/CMakeLists.txt
```diff
@@ -64,8 +64,17 @@ add_library(bsoncxx_static STATIC
 
 set_target_properties(bsoncxx_static PROPERTIES
     OUTPUT_NAME bsoncxx
+    COMPILE_DEFINITIONS "BSONCXX_STATIC"
```
markbenvenuto: Use target_compile_definitions(bsoncxx_static PUBLIC BSONCXX_STATIC), it is clearer

src/bsoncxx/config/compiler.hpp
```diff
@@ -12,4 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(_MSC_VER)
+
+#pragma warning(push)
+#pragma warning(disable: 4251 4273 4275)
```
markbenvenuto: What do these warnings mean?

cmake/FindLibMongoC.cmake
```diff
@@ -0,0 +1,24 @@
...
+# Load up PkgConfig if we have it
+find_package(PkgConfig QUIET)
+
+if (PKG_CONFIG_FOUND)
+  pkg_check_modules(LIBMONGOC REQUIRED libmongoc-1.0>=${LibMongoC_FIND_VERSION} )
+  # We don't reiterate the version information here because we assume that
+  # pkg_check_modules has honored our request.
+  find_package_handle_standard_args(LIBMONGOC DEFAULT_MSG LIBMONGOC_FOUND)
+elseif(LIBMONGOC_DIR)
+  # The best we can do until libMONGOC starts instaling a libmongoc-config.cmake file
```
samantharitter: ditto typo

src/bsoncxx/config/compiler.hpp
samantharitter: Can you add a comment explaining which warnings we are suppressing and why?

src/bsoncxx/private/error_category.cpp
```diff
@@ -137,15 +137,18 @@ const std::error_category& bsoncxx_error_category() {
 
 const std::error_category& libbson_error_category(int domain) {
     switch (domain) {
-        case BSON_ERROR_JSON:
+        case BSON_ERROR_JSON: {
```
samantharitter: Why do we need { } here?

src/bsoncxx/config/compiler.hpp
samantharitter: Oh I must not have refreshed since you added your comment, Mark

cmake/FindLibBSON.cmake
acmorrow: Done

cmake/FindLibBSON.cmake
acmorrow: Done

cmake/FindLibMongoC.cmake
acmorrow: Done

cmake/FindLibMongoC.cmake
```diff
@@ -0,0 +1,24 @@
...
+  set(LIBMONGOC_LIBRARIES mongoc-1.0 CACHE INTERNAL "")
+  set(LIBMONGOC_LIBRARY_DIRS ${LIBMONGOC_DIR}/lib CACHE INTERNAL "")
+  set(LIBMONGOC_INCLUDE_DIRS ${LIBMONGOC_DIR}/include/libMONGOC-1.0 CACHE INTERNAL "")
+  find_package_handle_standard_args(LIBMONGOC DEFAULT_MSG LIBMONGOC_LIBRARIES LIBMONGOC_LIBRARY_DIRS LIBMONGOC_INCLUDE_DIRS)
+else()
+    message(FATAL_ERROR "Don't know how to find libmongoc, please set LIBMONGOC_DIR")
```
acmorrow: Expanded message here too.

src/bsoncxx/private/error_category.cpp
acmorrow: This was actually a real bug that MSVC told us about. Without braces in each case, the scope of the static category extends to the entire switch statement, but you aren't allowed to jump past the initialization of static.

src/bsoncxx/CMakeLists.txt
acmorrow: Good call, done, here and for mongocxx

src/bsoncxx/config/compiler.hpp
acmorrow: Thanks for making me take another look at this. I think the only one we really need is 4251, and I've added a comment about that here and for mongocxx. I think the other ones were spurious due to misconfigurations addressed elsewhere in this PR.

samantharitter:
LGTM with a pretty shallow swim into the cmake

acmorrow:
Updated, PTAL. All review changes are in the newest commit.

markbenvenuto:
LGTM

PR #413, opened 2016-01-03
CXX-777 Build with Wall, Wextra and Werror to hold the line on warnings

acmorrow:
Note that we currently need -Wno-attributes to avoid being overwhelmed by warnings
generated by MNMLSTC core.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -78,7 +78,7 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror" ..
                   make
                   ./src/bsoncxx/test/test_bson
                   ./src/mongocxx/test/test_driver
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ before_script:
 
     # Build the driver and the tests
     - cd build
-    - ${CMAKE_BINARY} -DCMAKE_BUILD_TYPE=$CONFIG ..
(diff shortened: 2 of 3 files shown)


acmorrow:
I just realized we should do this for travis as well. Also, @tessavitabile this approach can probably work for ASAN and UBSAN builds, where you can just inject the right flags when calling CMake.

tessavitabile:
Ah thanks, that means I won't need to change the CMakeLists.txt files, and I can just change the .mci.yml.

acmorrow:
@tessavitabile Yeah, I think so. The only trouble I can think of is if there isn't a similar way to pass additional link arguments, since I believe you need to pass the sanitize argument to the linker as well. Or, perhaps CMake is with it enough to auto propagate the compile flags to the link step. Give it a try and see what happens?

acmorrow:
Ping for an LGTM or comments

acmorrow:
In particular, @samantharitter, please let me know if you think there is a better evergreen/YAML-ish way to handle these flags, given that @tessavitabile is going to need to add some others when she is has the necessary tools in place to run with ASAN/UBSAN

.mci.yml
```diff
@@ -78,7 +78,7 @@ tasks:
           params:
               working_dir: "mongo-cxx-driver/build"
               script: |
-                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install ..
+                  PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror" ..
```
samantharitter: Will these flags change depending on the buildvariant?  If not then it's fine to keep them here.  Otherwise we might want to pull these into a variable, then attach that to each buildvariant and use mci's expansions:
```
cmake_flags:  &standard_warning_flags '-DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror"'

// in the script:
PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install ${cmake_flags} ..

// buildvariant
- name: ubuntu1404
  expansions:
     cmake_flags: *standard_warning_flags
```

But I think this is overkill unless these flags are different on different systems.  We can always do it later if we find we need it.

PR #414, opened 2016-01-04
CXX-790 - Bulk write upsert example

tessavitabile:


acmorrow:
How do you want to sequence this with https://github.com/mongodb/mongo-cxx-driver/pull/409? The example won't compile after that change. Either we can get this committed first and I can update #409, or we can get #409 reviewed and pushed and then update this example.

tessavitabile:
Perhaps #409 should be committed first, since it would be confusing to users of this example if it changes.

acmorrow:
OK. Want to review #409 and then we can update this one?

tessavitabile:
Sure.

acmorrow:
@tessavitabile - I merged #409 so if you want to update this as needed I'll give it a look.

tessavitabile:
@acmorrow I updated this based on #409

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,46 @@
...
+    auto doc1 = document{} << "a" << 1 << finalize;
+    auto doc2 = document{} << "$set" << open_document << "a" << 2 << close_document << finalize;
+
+    update_one update{std::move(doc1), std::move(doc2)};
+    update.upsert(true);
+
+    bulk_write bulk{};
+    bulk.append(update);
+
+    coll.bulk_write(bulk);
```
acmorrow: Is there a result here that we should be checking?

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // @end: cpp-bulk-write-upsert
```
acmorrow: Main should return either EXIT_SUCCESS or EXIT_FAILURE.

examples/mongocxx/bulk_write.cpp
tessavitabile: Sure, I wasn't sure if we wanted this to be a test case that checks the result, or just an example of using bulk_write with upsert, but I can do the former.

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    // @begin: cpp-bulk-write-upsert
+    auto doc1 = document{} << "a" << 1 << finalize;
+    auto doc2 = document{} << "$set" << open_document << "a" << 2 << close_document << finalize;
+
+    update_one update{std::move(doc1), std::move(doc2)};
+    update.upsert(true);
+
+    bulk_write bulk{};
+    bulk.append(update);
```
samantharitter: Since this is an example of a bulk operation, can you expand it to include several CRUD operations?  It would be good to show that you can mix updates with inserts or deletes, for instance.

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,46 @@
...
+    mongocxx::instance inst{};
+    mongocxx::client conn{mongocxx::uri{}};
+
+    auto coll = conn["test"]["coll"];
+
+    // @begin: cpp-bulk-write-upsert
+    auto doc1 = document{} << "a" << 1 << finalize;
+    auto doc2 = document{} << "$set" << open_document << "a" << 2 << close_document << finalize;
+
+    update_one update{std::move(doc1), std::move(doc2)};
```
samantharitter: I would skip the using statement above and use the full "model::update_one" here, for clarity.

samantharitter: It might also be good to add a few comments.  For instance, to explain why each operation is a "model"--as I understand it these are templates for operations, and they could be used multiple times.  You could create an insert_one model and append it twice to the bulk_write, and two matching documents would be inserted.

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Create a model for a delete_one operation.
+    mongocxx::model::delete_one delete_op{doc1.view()};
+
+    // Create a model for an update_one operation.
+    mongocxx::model::update_one upsert_op{doc1.view(), doc2.view()};
+    // Set upsert to true: if no document matches {"a": 1}, insert {"a": 2}.
+    upsert_op.upsert(true);
+
+    // Create a bulk_write operation. By default, its operations are executed in order.
+    mongocxx::bulk_write bulk{};
```
acmorrow: Split these up with ws lines, pairing each comment and the relevant statement:
```
// Do a thing
do_it();

// And another
another_thing();
```

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // The collection should contain two copies of {"a": 2}.
+    auto cursor = coll.find({});
+    int i = 0;
+    for (auto&& doc : cursor) {
+      i++;
+      if (doc["a"].get_int32() != 2) {
+        return EXIT_FAILURE;
+      }
+    }
+    return (i == 2) ? EXIT_SUCCESS : EXIT_FAILURE;
```
acmorrow: blank line before this final return statement

acmorrow: Please add an include of <cstdlib> to ensure that you have definitions for `EXIT_SUCCESS` and `EXIT_FAILURE`

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Delete {"a": 1}.
+    bulk.append(delete_op);
+    // Since no document matches {"a": 1}, insert {"a": 2}.
+    bulk.append(upsert_op);
+    // Insert {"a": 1}. Note the same model may be appended multiple times.
+    bulk.append(insert_op);
+    // Update {"a": 1} to {"a": 2}.
+    bulk.append(upsert_op);
+
+    coll.bulk_write(bulk);
```
acmorrow: This definitely returns something. Make sure to capture it and evaluate it.

acmorrow: To clarify, it definitely returns an `optional`. Whether that optional is engaged or not is a different question, but we should at least act on those possibilities.

tessavitabile:
I didn't check the inserted and upserted IDs in the bulk_write result, since printing them throws. See https://jira.mongodb.org/browse/CXX-798.

examples/mongocxx/bulk_write.cpp
```diff
@@ -0,0 +1,102 @@
...
+    bulk.append(upsert_op);
+
+    // Insert {"a": 1}. Note the same model may be appended multiple times.
+    bulk.append(insert_op);
+
+    // Update {"a": 1} to {"a": 2}.
+    bulk.append(upsert_op);
+
+    mongocxx::stdx::optional<mongocxx::result::bulk_write> result = coll.bulk_write(bulk);
+    if (result) {
```
acmorrow: Consider switching this to exit early if it failed. That way, you can reduce the nesting by one level:
```
if (!result)
    return EXIT_FAILURE;

if (result->...) {
```

acmorrow:
LGTM with a nit-picky comment. Can you also add a TODO about printing the IDs once CXX-798 is resolved? No need to re-post, you can commit it after adding the TODO and optionally fixing the if ordering.

PR #415, opened 2016-01-04
CXX-795 - Add bsoncxx unit tests to evergreen builds

tessavitabile:
https://evergreen.mongodb.com/version/568ab75a3ff1222e7e0002c6_0

Do I need to do anything separate for travis?

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -80,8 +80,9 @@ tasks:
               script: |
                   PKG_CONFIG_PATH="/data/tmp/c-driver-install/lib/pkgconfig" /opt/cmake/bin/cmake -DCMAKE_BUILD_TYPE=${build_type} -DCMAKE_INSTALL_PREFIX=./install ..
                   make
-                  make run-examples
+                  ./src/bsoncxx/test/test_bson
                   ./src/mongocxx/test/test_driver
+                  make run-examples
         - func: "stop_mongod"
 
 #######################################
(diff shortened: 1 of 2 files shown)


.mci.yml
```diff
@@ -82,6 +82,7 @@ tasks:
                   make
                   make run-examples
                   ./src/mongocxx/test/test_driver
+                  ./src/bsoncxx/test/test_bson
```
acmorrow: I'd change the order here. Run the bson tests first, then the driver tests, and then the examples last. That way if you break BSON you will get an error in the bson tests, rather than in the driver tests, etc. Rougly dependency order.

acmorrow:
@tessavitabile - Yes, please do for travis as well if it is also missing the bson tests.

tessavitabile:
How can I see what tests travis is running?

samantharitter:
You'll need to update the "script" part of [.travis.yml](https://github.com/mongodb/mongo-cxx-driver/blob/master/.travis.yml#L73-L79) to make these changes to travis.

acmorrow:
LGTM

.travis.yml
```diff
@@ -72,8 +72,11 @@ before_script:
 script:
     - make
 
-    # Run all examples
-    - make run-examples
+    # Run tests with catch
```
samantharitter: nit, I think you can either use one comment for both sets of tests, or make the comments more specific ("Run mongocxx tests")

samantharitter:
LGTM mod a nit about the travis comment

PR #416, opened 2016-01-05
CXX-630 - Improve move semantics of model::write

tessavitabile:


src/mongocxx/model/write.cpp
```diff
@@ -21,23 +21,56 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace model {
 
 write::write(insert_one value) : _type(write_type::k_insert_one), _insert_one(std::move(value)) {
+    static_assert(std::is_nothrow_move_constructible<insert_one>::value, "Move may throw");
+    static_assert(std::is_nothrow_move_assignable<insert_one>::value, "Move may throw");
 }
 write::write(delete_one value) : _type(write_type::k_delete_one), _delete_one(std::move(value)) {
+    static_assert(std::is_nothrow_move_constructible<delete_one>::value, "Move may throw");
```
samantharitter: maybe differentiate these error messages, for move-assign and move-copy

src/bsoncxx/view_or_value.hpp
```diff
@@ -88,7 +88,7 @@ class view_or_value {
     ///
     /// Construct a view_or_value from a moved-in view_or_value.
     ///
-    BSONCXX_INLINE view_or_value(view_or_value&& other)
+    BSONCXX_INLINE view_or_value(view_or_value&& other) noexcept
```
acmorrow: Add a TODO in the doc comment linking to the new ticket

src/bsoncxx/view_or_value.hpp
```diff
@@ -97,7 +97,7 @@ class view_or_value {
     ///
     /// Assign to this view_or_value from a moved-in view_or_value.
     ///
-    BSONCXX_INLINE view_or_value& operator=(view_or_value&& other) {
+    BSONCXX_INLINE view_or_value& operator=(view_or_value&& other) noexcept {
```
acmorrow: Here too for link to noexcept ticket

src/mongocxx/model/write.cpp
```diff
@@ -21,23 +21,56 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace model {
 
 write::write(insert_one value) : _type(write_type::k_insert_one), _insert_one(std::move(value)) {
+    static_assert(std::is_nothrow_move_constructible<insert_one>::value, "Move-construct may throw");
```
acmorrow: You should include `<type_traits>` above so that the is_xxx things are definitely defined.

src/mongocxx/model/write.cpp
```diff
@@ -57,36 +90,34 @@ void write::destroy_member() {
...
             break;
-        case write_type::k_uninitialized:
-            break;
     }
-
-    _type = write_type::k_uninitialized;
 }
 
 write& write::operator=(write&& rhs) noexcept {
+    if (this == &rhs) {
```
acmorrow: Do we actually need to guard against self move assign? I consider `a = std::move(a)` to be a bug. However, I would like `std::swap(a, a)` to work.

acmorrow:
This LGTM mod nits and a question about self move assign.

src/mongocxx/model/write.cpp
tessavitabile: How can I ensure that std::swap(a, a) works?

samantharitter:
LGTM as well, I don't have a strong opinion about whether `a = std::move(a)` should be something we protect against.

src/mongocxx/model/write.cpp
acmorrow: Lets leave it for now and keep the check. I want to think about it more.

acmorrow:
LGTM

samantharitter:
That travis failure doesn't look like your fault, maybe push a trivial change to run it again?

tessavitabile:
It was caused by my bulk-write example, though I haven't figured out why yet.

tessavitabile:
Sam figured it out and is going to push a fix.

src/mongocxx/model/write.cpp
tessavitabile: Okay

PR #417, opened 2016-01-06
CXX-638 Fixes required for VS2015 support

acmorrow:
This doesn't quite get VS2015 builds working (that needs the great renaming), but it gets it a lot closer.


acmorrow:
@markbenvenuto @hanumantmk - I'd particularly like both of your thoughts on this one.

src/bsoncxx/builder/stream/value_context.hpp
```diff
@@ -60,8 +60,11 @@ class value_context {
 
     operator single_context();
 
+#if !defined(_MSC_VER)
```
markbenvenuto: If you want to check the exact version of the compiler, use: `#if defined(_MSC_VER) && _MSC_VER > 1900023506`

src/bsoncxx/oid.hpp
```diff
@@ -38,7 +38,10 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 class BSONCXX_API oid {
    public:
     struct init_tag_t {};
-    static constexpr init_tag_t init_tag{};
+
+    // TODO(MSVC): Ideally this would be constexpr, but VS2015U1 can't
```
markbenvenuto: See https://connect.microsoft.com/VisualStudio/feedback/details/2092790

src/bsoncxx/util/functor.hpp
```diff
@@ -22,9 +22,89 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace util {
 
+// TODO(MSVC): VS2015U1 Completely falls apart trying to honor the
```
markbenvenuto: specifically, is_convertible returns the wrong results for conversions to std::function

acmorrow:
@markbenvenuto Note that this isn't entirely working yet. Apparently, you can't have a dllexport class have a dllimport method, which breaks the whole idea of our MONGOCXX_INLINE macro

src/bsoncxx/builder/stream/value_context.hpp
acmorrow: @markbenvenuto Do you think it matters here? I'm not sure. Nobody is going to build with any MSVC earlier than 2015 since it just won't work, and we are hoping to remove these once the bugs are fixed in 2015

src/bsoncxx/oid.hpp
acmorrow: Do you think I should add the link to the comment?

src/bsoncxx/util/functor.hpp
acmorrow: Added some text to this effect

acmorrow:
I have no successfully built the driver on VS2015 with the above changes (please ignore the "minor: " commit, it isn't real). This is with hacked in `enum` vs `enum class` changes not shown here, but I'm not proposing to commit those since we are going to do the great renaming.

acmorrow:
Ping @markbenvenuto @hanumantmk this is ready for re-review.

src/bsoncxx/oid.hpp
markbenvenuto: It makes it easier to track if the underlying bug has been fixed for someone who comes along later and sees the TODO.

acmorrow: Done

acmorrow:
@markbenvenuto @hanumantmk I added the ticket link. PTAL? I'd really like to get this in today if possible.

markbenvenuto:
LGTM

hanumantmk:
LGTM

PR #420, opened 2016-01-06
CXX-798 - Fix result::bulk_write::upserted_ids and remove inserted_ids

tessavitabile:
The C driver does not return the inserted IDs in bulk_write results, so we won't either. It is not required by the CRUD spec (https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst).

I chose to have result::bulk_write::upserted_ids return a std::map<std::size_t, bsoncxx::document::element>, where the first element is the operation index in the bulk write, and the second is {"_id" : ...}. This is analogous to result::insert_many::inserted_ids.

### DIFF
--- a/examples/mongocxx/bulk_write.cpp
+++ b/examples/mongocxx/bulk_write.cpp
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 #include <cstdlib>
+#include <iostream>
 
 #include <bsoncxx/builder/stream/document.hpp>
-
+#include <bsoncxx/json.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/uri.hpp>
@@ -98,7 +99,10 @@ int main(int, char**) {
         return EXIT_FAILURE;
     }
 
-    // TODO: print inserted and upserted IDs once CXX-798 is resolved
+    std::cout << "Upserted IDs" << std::endl;
+    for (const auto &id : result->upserted_ids()) {
+        std::cout << "Bulk write index: " << id.first << std::endl << bsoncxx::to_json(id.second) << std::endl;
+    }
 
     // The collection should contain two copies of {"a": 2}.
     auto cursor = coll.find({});
--- a/src/mongocxx/result/bulk_write.cpp
+++ b/src/mongocxx/result/bulk_write.cpp
@@ -43,12 +43,17 @@ std::int32_t bulk_write::upserted_count() const {
     return view()["nUpserted"].get_int32();
 }
 
-bsoncxx::document::element bulk_write::inserted_ids() const {
-    return view()["inserted_ids"];
-}
+bulk_write::id_map bulk_write::upserted_ids() const {
+    id_map upserted_ids;
 
-bsoncxx::document::element bulk_write::upserted_ids() const {
-    return view()["upserted_ids"];
+    if (!view()["upserted"]) {
+        return upserted_ids;
+    }
+
+    for (auto&& id : view()["upserted"].get_array().value) {
+        upserted_ids.emplace(id["index"].get_int32(), id["_id"]);
+    }
+    return upserted_ids;
 }
 
 bsoncxx::document::view bulk_write::view() const {
--- a/src/mongocxx/result/bulk_write.hpp
+++ b/src/mongocxx/result/bulk_write.hpp
@@ -17,6 +17,7 @@
 #include <mongocxx/config/prelude.hpp>
 
 #include <cstdint>
+#include <map>
 #include <vector>
 
 #include <bsoncxx/document/value.hpp>
@@ -33,6 +34,8 @@ namespace result {
 class MONGOCXX_API bulk_write {
 
    public:
+    using id_map = std::map<std::size_t, bsoncxx::document::element>;
+
     explicit bulk_write(bsoncxx::document::value raw_response);
 
     ///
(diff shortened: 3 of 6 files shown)


src/mongocxx/result/bulk_write.hpp
```diff
@@ -33,6 +34,8 @@ namespace result {
 class MONGOCXX_API bulk_write {
 
    public:
+    using id_map = std::map<std::size_t, bsoncxx::document::element>;
```
acmorrow: Do you need ordering by key, or would unordered_map work here too?

src/mongocxx/result/bulk_write.hpp
```diff
@@ -71,18 +74,11 @@ class MONGOCXX_API bulk_write {
...
-    ///
-    /// @return The values of the _id field for inserted documents.
-    ///
-    bsoncxx::document::element inserted_ids() const;
-
-    ///
     /// Gets the ids of the upserted documents.
     ///
-    /// @return The values of the _id field for upserted documents.
+    /// @return A map from bulk write index to _id field for upserted documents.
```
acmorrow: Since the elements in the values in the map are views into the document owned by the bulk write, if the bulk write object is destroyed, the map is invalidated. You should add a comment stating that the returned id_map must not be accessed after the bulk_write object is destroyed.

src/mongocxx/result/replace_one.cpp
```diff
@@ -36,7 +36,11 @@ std::int32_t replace_one::modified_count() const {
 }
 
 stdx::optional<bsoncxx::document::element> replace_one::upserted_id() const {
-    return _result.upserted_ids();
+	if (_result.upserted_ids().size() == 0) {
+		return stdx::nullopt;
+	} else {
```
acmorrow: Since you have a return, you don't need the else:

if (...) {
    return ...;
}

return ...

src/mongocxx/result/update.cpp
```diff
@@ -35,7 +35,11 @@ std::int32_t update::modified_count() const {
 }
 
 stdx::optional<bsoncxx::document::element> update::upserted_id() const {
-    return _result.upserted_ids();
+    if (_result.upserted_ids().size() == 0) {
```
acmorrow: Same comment about else and return

src/mongocxx/test/collection.cpp
```diff
@@ -196,7 +196,8 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
         options::update options;
         options.upsert(true);
 
-        coll.update_one(b1.view(), update_doc.view(), options);
+        auto result = coll.update_one(b1.view(), update_doc.view(), options);
+        REQUIRE(result->upserted_id());
```
acmorrow: +1 to new tests!

src/mongocxx/result/bulk_write.cpp
```diff
@@ -43,12 +43,17 @@ std::int32_t bulk_write::upserted_count() const {
     return view()["nUpserted"].get_int32();
 }
 
-bsoncxx::document::element bulk_write::inserted_ids() const {
-    return view()["inserted_ids"];
-}
+bulk_write::id_map bulk_write::upserted_ids() const {
+	id_map upserted_ids{};
```
acmorrow: I don't think you need the {} here.

src/mongocxx/result/bulk_write.cpp
```diff
@@ -43,12 +43,17 @@ std::int32_t bulk_write::upserted_count() const {
...
+
+	if (!view()["upserted"]) {
+		return upserted_ids;
```
acmorrow: Please configure your editor to use spaces not tabs.

acmorrow: And run clang-format!

acmorrow: Actually, don't bother with clang-format now. We will do a sweep later

examples/mongocxx/bulk_write.cpp
```diff
@@ -98,7 +98,10 @@ int main(int, char**) {
         return EXIT_FAILURE;
     }
 
-    // TODO: print inserted and upserted IDs once CXX-798 is resolved
+    std::cout << "Upserted IDs" << std::endl;
+    for (auto &id : result->upserted_ids()) {
```
acmorrow: `const auto&` would be a little bit better

src/mongocxx/result/bulk_write.hpp
tessavitabile: I think ordering is nice, since it makes it easier to iterate through the upserted IDs in the order they were created.

acmorrow: ok, that makes sense

examples/mongocxx/bulk_write.cpp
```diff
@@ -15,7 +15,7 @@
 #include <cstdlib>
```
acmorrow: Please add `<iostream>` since you are using std::cout

src/mongocxx/result/bulk_write.hpp
```diff
@@ -71,18 +74,13 @@ class MONGOCXX_API bulk_write {
...
+    ///
+    /// The returned id_map must not be accessed after the bulk_write object is destroyed.
```
samantharitter: nit, just move this comment to above the return value comment, and label with @note

src/mongocxx/result/bulk_write.cpp
```diff
@@ -43,12 +43,17 @@ std::int32_t bulk_write::upserted_count() const {
...
 }
 
-bsoncxx::document::element bulk_write::inserted_ids() const {
-    return view()["inserted_ids"];
-}
+bulk_write::id_map bulk_write::upserted_ids() const {
+	id_map upserted_ids;
+
+	if (!view()["upserted"]) {
+		return upserted_ids;
```
acmorrow: Nit: Please do fix the tabs/spaces here before push

samantharitter:
LGTM

acmorrow:
LGTM

PR #421, opened 2016-01-06
CXX-802 - coll.drop() no longer throws if collection doesn't exist

tessavitabile:


src/mongocxx/test/database.cpp
```diff
@@ -36,9 +36,7 @@ namespace {
 
 // Clean up the given collection if it exists
 void clear_collection(const database& database, bsoncxx::stdx::string_view collection_name) {
-    if (database.has_collection(collection_name)) {
-        database[collection_name].drop();
-    }
+    database[collection_name].drop();
```
samantharitter: Instead of changing this method, can we just remove it now and replace the calls to `clear_collection()` with `database[collection_name].drop()` ?

samantharitter:
Can you please also change the bulk_write example back to the way it had been, without the checks?

src/mongocxx/test/database.cpp
tessavitabile: Good call!

src/mongocxx/collection.cpp
```diff
@@ -735,7 +735,7 @@ void collection::drop() {
 
     auto result = libmongoc::collection_drop(_get_impl().collection_t, &error);
 
-    if (!result) {
+    if (!result && error.code != ::MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST) {
```
acmorrow: I've made a choice in my life to not remember the C++ operator precedence rules :) Can you please parenthesize `(error.code != ...)`

acmorrow:
LGTM!

src/mongocxx/collection.cpp
tessavitabile: Haha yes :)

samantharitter:
LGTM

PR #422, opened 2016-01-06
CXX-801 Use .terminated().data() to extract safe c-strings

samantharitter:
@acmorrow this PR contains the change we discussed earlier.

I also added a `.data()` method to the `string::view_or_value` class.  Doing so allows us a place to explain to use terminated() first, and also makes `view_or_value.terminated().data()` possible instead of having to write `view_or_value.terminated().view().data()`.

### DIFF
--- a/src/bsoncxx/string/view_or_value.cpp
+++ b/src/bsoncxx/string/view_or_value.cpp
@@ -20,15 +20,19 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace string {
 
-const char* view_or_value::c_str() {
+view_or_value view_or_value::terminated() const {
     // If we do not own our string, we cannot guarantee that it is null-terminated,
     // so make an owned copy.
-    if (!_value) {
-        _value = std::move(_view.to_string());
-        _view = *_value;
+    if (!is_owning()) {
+        return view_or_value(std::move(view().to_string()));
     }
 
-    return _value->c_str();
+    // If we are owning, return a view_or_value viewing our string
+    return {view()};
+}
+
+const char* view_or_value::data() const {
+    return view().data();
 }
 
 }  // namespace string
--- a/src/bsoncxx/view_or_value.hpp
+++ b/src/bsoncxx/view_or_value.hpp
@@ -91,7 +91,8 @@ class view_or_value {
 
     /// TODO CXX-800: Create a noexcept expression to check the conditions that must be met.
     BSONCXX_INLINE view_or_value(view_or_value&& other) noexcept
-        : _value{std::move(other._value)}, _view(_value ? *_value : std::move(other._view)) {
+        : _value{std::move(other._value)},
+          _view(_value ? *_value : std::move(other._view)) {
         other._view = View();
         other._value = stdx::nullopt;
     }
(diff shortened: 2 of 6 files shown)


src/bsoncxx/string/view_or_value.cpp
```diff
@@ -15,20 +15,26 @@
...
-    if (!_value) {
-        _value = std::move(_view.to_string());
-        _view = *_value;
+    if (!is_owning()) {
+        return view_or_value(std::move(view().to_string()));
     }
 
-    return _value->c_str();
+    // If we are owning, return a view_or_value viewing our string
+    return view_or_value(view());
```
acmorrow: does `return view()` work? I'd think so.

src/bsoncxx/string/view_or_value.cpp
```diff
@@ -15,20 +15,26 @@
...
 
-const char* view_or_value::c_str() {
+view_or_value view_or_value::terminated() const {
     // If we do not own our string, we cannot guarantee that it is null-terminated,
     // so make an owned copy.
-    if (!_value) {
-        _value = std::move(_view.to_string());
-        _view = *_value;
+    if (!is_owning()) {
+        return view_or_value(std::move(view().to_string()));
```
acmorrow: I think the std::move here is unneeded as the result of `to_string` is a temporary already, but having thought about it, I think I like it for clarity.

src/bsoncxx/view_or_value.hpp
```diff
@@ -109,6 +110,15 @@ class view_or_value {
     }
 
     ///
+    /// Return whether or not this view_or_value owns an underlying Value.
+    ///
+    /// @return bool Whether we are owning.
+    ///
+    BSONCXX_INLINE bool is_owning() const noexcept {
+        return bool(_value);
```
acmorrow: Maybe `static_cast<bool>(_value)` is ever so slightly better?

src/bsoncxx/view_or_value.hpp
```diff
@@ -121,8 +131,7 @@ class view_or_value {
         return _view;
     }
 
-    // TODO: make these private again (CXX-801)
-   protected:
+   private:
```
acmorrow: Hooray!

src/bsoncxx/string/view_or_value.cpp
```diff
@@ -15,20 +15,26 @@
 #include <bsoncxx/config/prelude.hpp>
 
 #include <bsoncxx/string/view_or_value.hpp>
+#include <bsoncxx/exception/error_code.hpp>
```
acmorrow: Needed?

acmorrow:
I think I like it a lot. What do you think?

src/bsoncxx/string/view_or_value.cpp
samantharitter: `return {view()}` ?

acmorrow: Sure

src/bsoncxx/string/view_or_value.cpp
samantharitter: Ah, no, I had thought about trying to make `data()` throw when appropriate but realized there was no good way to do that.

src/bsoncxx/view_or_value.hpp
samantharitter: :tada:

samantharitter:
I like it!  If it looks good then I'll merge it in, I think it's going to be a while before Travis catches up with us today.  It passes our tests locally.

acmorrow:
LGTM. How much work to apply everywhere needed?

samantharitter:
Not sure, I've started that work in another branch.  Shouldn't be too bad!

acmorrow:
OK. Lets get this merged then.

PR #424, opened 2016-01-06
CXX-739 Apply string::view_or_value to index options

samantharitter:
I believe this PR adds string::view_or_value everywhere else it needs to go.

### DIFF
--- a/src/bsoncxx/oid.cpp
+++ b/src/bsoncxx/oid.cpp
@@ -34,10 +34,11 @@ oid::oid(init_tag_t) : _is_valid(true) {
     std::memcpy(_bytes, oid.bytes, sizeof(oid.bytes));
 }
 
-oid::oid(stdx::string_view str) : _is_valid(bson_oid_is_valid(str.data(), str.length())) {
+oid::oid(const string::view_or_value& str)
+    : _is_valid(bson_oid_is_valid(str.data(), str.view().length())) {
     if (_is_valid) {
         bson_oid_t oid;
-        bson_oid_init_from_string(&oid, str.data());
+        bson_oid_init_from_string(&oid, str.terminated().data());
         memcpy(_bytes, oid.bytes, sizeof(_bytes));
     }
 }
--- a/src/bsoncxx/oid.hpp
+++ b/src/bsoncxx/oid.hpp
@@ -20,7 +20,7 @@
 #include <ctime>
 #include <string>
 
-#include <bsoncxx/stdx/string_view.hpp>
+#include <bsoncxx/string/view_or_value.hpp>
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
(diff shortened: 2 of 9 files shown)


src/bsoncxx/string/view_or_value.hpp
```diff
@@ -41,6 +41,11 @@ class BSONCXX_API view_or_value : public bsoncxx::view_or_value<stdx::string_vie
     using bsoncxx::view_or_value<stdx::string_view, std::string>::view_or_value;
 
     ///
+    /// Default constructor, equivalent to using an empty string.
+    ///
+    BSONCXX_INLINE view_or_value() = default;
```
samantharitter: I'm not sure why we didn't get this by forwarding bsoncxx::view_or_value's constructors?  It has a default constructor.

src/mongocxx/collection.cpp
```diff
@@ -18,6 +18,7 @@
 #include <cstdint>
 #include <limits>
 #include <tuple>
+#include <unordered_set>
```
samantharitter: My other idea for keeping the temporary view_or_values alive was to throw them in a set.  Would that be better?

src/bsoncxx/oid.cpp
```diff
@@ -33,10 +33,11 @@ oid::oid(init_tag_t) : _is_valid(true) {
     std::memcpy(_bytes, oid.bytes, sizeof(oid.bytes));
 }
 
-oid::oid(stdx::string_view str) : _is_valid(bson_oid_is_valid(str.data(), str.length())) {
+oid::oid(string::view_or_value str)
```
acmorrow: Seems like this could be `const string::view_or_value&`, since you aren't actually storing this anywhere, just copying it into the bson library.

src/bsoncxx/oid.hpp
```diff
@@ -64,12 +64,12 @@ class BSONCXX_API oid {
     explicit oid(const char* bytes, std::size_t len);
 
     ///
-    /// Constructs and oid an initializes it from the provided hex string.
+    /// Constructs an oid and initializes it from the provided hex string.
```
acmorrow: Nice catch

src/bsoncxx/string/view_or_value.hpp
acmorrow: What broke that made you need to add it? I'm surprised as well.

src/mongocxx/hint.cpp
```diff
@@ -42,7 +42,7 @@ bsoncxx::document::value hint::to_document() const {
 }
 
 bool operator==(const hint& index_hint, std::string index) {
-    return index_hint._index_string && index_hint._index_string->to_string() == index;
+    return index_hint._index_string && *(index_hint._index_string) == index;
```
acmorrow: Please add parens (I know they weren't there before)

src/mongocxx/options/find.cpp
```diff
@@ -30,7 +30,7 @@ void find::batch_size(std::int32_t batch_size) {
     _batch_size = batch_size;
 }
 
-void find::comment(std::string comment) {
+void find::comment(bsoncxx::string::view_or_value comment) {
     _comment = comment;
```
acmorrow: std::move

src/bsoncxx/string/view_or_value.hpp
acmorrow: Per investigation, we now understand this (sort of), so, no need to change it.

acmorrow:
LGTM

acmorrow:
LGTM

PR #425, opened 2016-01-07
CXX-754 exception handling example

tessavitabile:
I wasn't totally sure what we'd want out of an exceptions example, so I gave some examples of catching mongocxx exceptions of different types.

Do you think I should put bsoncxx exception examples in the same file, even though this is the mongocxx folder? Or a separate file in the bsoncxx folder?

One thing I realized in making this example, is that we don't allow you to test  equivalence of mongocxx::exception::code() to known error codes, since our make_error_code functions are private, and our libmongoc and libbson error code enums are private. Is this the behavior we want?


amidvidy:
hmm. I think what I intended in this ticket is an example of how to robustly use the driver in the presence of exceptions, rather than an example of catching each type of exception a few ops can throw.

I think we'd want to show an example of robust 'in the wild' code: e.g. find a document, then find another document in a different collection with a query that depends on the content of the first document. I'd expect all the code to be in one large try block that could have a couple different catch clauses from most specific to least specific error.

acmorrow:
I think I agree with @amidvidy's overall point. Regarding the error codes, that is a good point. I think we will need to make them public. Seems like it should be straightforward? Since writing the example motivated the change, I think it would be fine to fold that work in under this ticket.

acmorrow:
That said, I do think the short examples here are also useful. Maybe keep them, but also add a more elaborate example like @amidvidy suggested?

tessavitabile:
That all sounds good. Do you have any concern about exposing libmongoc error codes, in case users' code comes to depend on them and we change implementations? I suppose this is a reason one might use error_conditions. That way we could make public error_conditions and define them to be equivalent to libmongoc error_codes, and if we change implementations, we could define our new error_codes to be equivalent to the original error_conditions. Users would test against error_conditions rather than error_codes, which is really the system_error design.

amidvidy:
I don't think we need to worry about the libmongoc error codes changing as that would be an ABI break on their part. Maybe check with @ajdavis?

tessavitabile:
I meant if the C++ driver ever stopped being a wrapper of the C driver.

amidvidy:
it might be worth it to define error_conditions in the future if we switch implementations, which doesn't seem likely to happen. Until it does, I think it's an unnecessary amount of complexity.

tessavitabile:
It would be too late to define error_conditions then, since users would already be relying on libmongoc error codes. But I agree that it may not be worth the complexity.

tessavitabile:
I had an idea that instead of making the libmongoc error code enum public, I could create a set of methods that return a static copy of each std::error_code that we make. Then if you want to compare an error_code from an exception against a known std::error_code, you could do something like:

e.code() == client_authenticate_error()

instead of:

e.code() == make_error_code(::MONGOC_ERROR_CLIENT_AUTHENTICATE).

Then we can keep the libmongoc error code enum private, and if we ever change our underlying implementation to not use the C driver, we can just change the implementation of client_authenticate_error(). I'll try this out and put it in this pull request.

tessavitabile:
I added a set of methods return static error_code instances in mongocxx/exception/error_code.hpp. Let me know what you think of this approach!

tessavitabile:
(I still plan to add an example of robust "in the wild" code, but I wanted to get your input on this.)

amidvidy:
seems reasonable, although this might be a good candidate for X-macros for brevity (and ease of adding new codes).

examples/mongocxx/exception.cpp
```diff
@@ -0,0 +1,92 @@
...
+    mongocxx::instance inst{};
+    mongocxx::client conn{mongocxx::uri{}};
+
+    // @begin: cpp-logic-error
+    // Using an uninitialized collection throws a mongocxx::logic_error.
+    // A mongocxx::logic_error is-a mongocxx::exception is-a std::system_error.
+    mongocxx::collection coll;
+    try {
+        coll.name();
+    } catch (mongocxx::logic_error& e) {
```
acmorrow: Always catch exceptions by const&

examples/mongocxx/exception.cpp
```diff
@@ -0,0 +1,92 @@
...
+        std::cout << std::endl;
+    }
+    // @end: cpp-operation-exception
+
+    // @begin: cpp-bulk-write-exception
+    // Adding a document whose "_id" is already present throws a mongocxx::bulk_write_exception.
+    // A mongocxx::bulk_write_exception is-a mongocxx::operation_exception is-a mongocxx::exception
+    // is-a std::system_error.
+    auto doc1 = document{} << "_id" << 1 << finalize;
+    coll.insert_one(doc1.view());
```
acmorrow: Do you need to say .view here?

src/mongocxx/exception/error_code.hpp
```diff
@@ -31,6 +31,92 @@ enum class error_code : std::int32_t {
     k_unknown_write_concern,
 };
 
+MONGOCXX_API const std::error_code& bson_invalid_error();
```
acmorrow: Why is this preferable to exposing `mongocxx::make_error_code`?

tessavitabile: It is possibly preferable to exposing the C driver error codes and error domains, because if the C++ driver were to stop being implemented in terms of the C driver, then we could just change the implementation of bson_invalid_error() and not break users' code. But it may not be worth it.

ajdavis:
I haven't been following closely, but my general comment is that the C Driver's error code suffer from a bad awful design flaw that is no good. At some point we will design a migration path away from the current awfulness, and the less you couple the C++11 driver to the existing C error codes, the better.

https://jira.mongodb.org/browse/CDRIVER-644

I don't know *what* the solution will eventually be to our error codes, however.

src/mongocxx/exception/error_code.hpp
samantharitter: I think I like this better than requiring the user to handle libmongoc error codes and domains.  Also, if as Jesse says they are planning to change the C driver's error primitives, this approach would be a safer bet.

examples/mongocxx/exception.cpp
samantharitter: Yes, otherwise the compiler should be upset

src/mongocxx/exception/error_code.hpp
acmorrow: I want to think about it a bit. I'm not sure either way yet.

examples/mongocxx/exception.cpp
acmorrow: Oh right, needs either `.view` or `std::move`

src/mongocxx/exception/error_code.hpp
acmorrow: I realized what was bugging me about this over the weekend. While this does make us independent of the numeric values of the libbson/libmongoc error codes, it is still exposing the C level error schema. The difference is that it is symbolic rather than numeric. That is a good step, but it still feels like a leaky abstraction.

We also have the issue raised in https://jira.mongodb.org/browse/CXX-808

This suggests to me that we haven't quite arrived where we want to be as far as the error abstraction. Lets try to find some time to talk it through. @hanumantmk @ajdavis @amidvidy @samantharitter - I'd be interested in all of your thoughts on this issue, as we don't have a lot of time to resolve it if we want to make our GA date.

Overall we have so far done a very good job making our use of the C driver completely opaque, but this is definitely a troublesome area.

tessavitabile:
I implemented three error categories, for server, implementation, and inherent errors. The error code enums and make_error_code functions are all public, so you can test against known error codes, as you see in examples/mongocxx/exception.cpp.

The heuristic for distinguishing between server and C driver errors is: If the code is only in one enum, we're fine. If it's in both enums, then guess server if the domain is COMMAND or QUERY, and guess C driver otherwise. This is clearly not perfect, since the MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST error has the QUERY domain, so we end up erroneously reporting the server error NamespaceNotFound with the same integer value. I just happened to find this problem when writing the example, and I have no idea how many others there will be. Open question: is there a satisfactory heuristic?

Open question: what should we do with error codes in neither enum? Call them unknown server errors? Guess server or C driver depending on the domain? Since we're hardcoding the enums, we're going to get behind.

In retrospect, this would have been a good time to learn about macros, and this should probably be rewritten using macros, so that the error code enums are each only in one place. But let me know what you think of the user interface for this, and we can worry about macros next.

acmorrow:
Closing this in favor of https://github.com/mongodb/mongo-cxx-driver/pull/454, for now

PR #426, opened 2016-01-08
CXX-747 Add OSX builder to evergreen project

tessavitabile:
https://evergreen.mongodb.com/version/568febb33ff1224e9a000487_0

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -3,6 +3,31 @@
 #######################################
 
 #######################################
+#            Variables                #
+#######################################
+variables:
+
+    ## source variables
+    source:
+        ubuntu_source: &ubuntu_source https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-latest.tgz
+        osx_source: &osx_source https://fastdl.mongodb.org/osx/mongodb-osx-x86_64-latest.tgz
+
+    ## cmake path variables
+    cmake_path:
+        ubuntu_cmake_path: &ubuntu_cmake_path /opt/cmake/bin/cmake
+        osx_cmake_path: &osx_cmake_path /Applications/Cmake.app/Contents/bin/cmake
+
+    ## tar options variables
+    tar_options:
+        ubuntu_tar_options: &ubuntu_tar_options --wildcards --no-anchored
+
+    ## cmake flag variables
+    cmake_flags:
+        ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
+        osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
+
+
+#######################################
 #            Functions                #
 #######################################
 
@@ -17,8 +42,8 @@ functions:
         params:
             working_dir: "."
             script: |
-                wget -O mongodb.tgz https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-latest.tgz
-                tar --extract --file mongodb.tgz --strip-components=2 --wildcards --no-anchored "*/bin/mongod"
+                curl -o mongodb.tgz ${source}
+                tar --extract --file mongodb.tgz --strip-components=2 ${tar_options} "*/bin/mongod"
                 mkdir -p /data/tmp/db
                 ./mongod --fork --logpath=/data/tmp/log --dbpath=/data/tmp/db --pidfilepath=/data/tmp/pid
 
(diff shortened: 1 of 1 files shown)


.mci.yml
```diff
@@ -3,6 +3,17 @@
 #######################################
 
 #######################################
+#            Variables                #
+#######################################
+
+linux_cmake_path: &linux_cmake_path /opt/cmake/bin/cmake
```
samantharitter: I feel like these will expand in the future, so can we divide them up a bit further?  I'm thinking either by operating system or by feature (like "cmake path").  Also, just for good form, let's put these in some kind of bigger "variables" structure.  It's possible that one day evergreen will be able to parse something like that in a helpful way.

```
## variables
variables: 

     ## linux variables
     linux:
          linux_cmake_path: &name stuff
          linux_source: &name stuff
          ...
     ## osx variables
     osx:
          osx_cmake_path: &name stuff
          ...
```

.mci.yml
```diff
@@ -107,7 +121,32 @@ buildvariants:
...
       expansions:
           build_type: "Debug"
+          source: *linux_source
+          tar_options: *linux_tar_options
+          cmake_path: *linux_cmake_path
       run_on:
           - ubuntu1404-test
       tasks:
           - name: compile_and_test
+    - name: osx-108
```
samantharitter: @acmorrow do we need both a release and debug build for OS X?

samantharitter: also, newline before starting a new buildvariant

.mci.yml
```diff
@@ -107,7 +121,32 @@ buildvariants:
...
+      expansions:
+          build_type: "Release"
+          source: *osx_source
+          cmake_path: *osx_cmake_path
+          cmake_flags: *osx_cmake_flags
+      run_on:
+          - osx-108
+      tasks:
+          - name: compile_and_test
+    - name: osx-108-debug
```
samantharitter: newline before starting a new buildvariant

.mci.yml
samantharitter: Looking at this again maybe it makes more sense to organize these by feature?  Up to you!

.mci.yml
```diff
@@ -98,6 +109,9 @@ buildvariants:
       display_name: "Ubuntu 14.04 Release"
       expansions:
           build_type: "Release"
+          source: *linux_source
```
samantharitter: since we have an open ticket to add a RHEL builder, can you change these to `ubuntu_source`, `ubuntu_tar_options` etc?

src/bsoncxx/CMakeLists.txt
```diff
@@ -21,7 +21,13 @@ set(BSONCXX_HEADER_INSTALL_DIR "include/bsoncxx/${BSONCXX_INLINE_NAMESPACE}" CAC
 option(BSONCXX_POLY_USE_STD_EXPERIMENTAL "Use <experimental/ for stdx polyfills" OFF)
 option(BSONCXX_POLY_USE_MNMLSTC "Use mnmlstc/core for stdx polyfills" ON)
 option(BSONCXX_POLY_USE_SYSTEM_MNMLSTC "Obtain mnmlstc/core from system" OFF)
-option(BSONCXX_POLY_USE_BOOST "Use boost for stdx polyfills" OFF)
+
+# MNMLSTC is broken on the OSX evergreen builder
+if(CMAKE_OSX)
+  option(BSONCXX_POLY_USE_BOOST "Use boost for stdx polyfills" ON)
```
samantharitter: I think it might be better to pass in `-DBSONCXX_POLY_USE_BOOST` from our evergreen script on the osx variants rather than change this here

CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# The OSX evergreen builder requires -stdlib=libc++ for cstdint to be defined
+if(CMAKE_OSX)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
samantharitter: similarly, I think it would be better to do this from the evergreen script.  Logically, it feels like evergreen should know about cmake but cmake shouldn't necessarily know about evergreen.

.mci.yml
acmorrow: No, I think one or the other is sufficient. I don't have a strong opinion on which.

CMakeLists.txt
acmorrow: Agreed. There may be OS X users who want mnmlstc, or the experimental versions

src/bsoncxx/CMakeLists.txt
acmorrow: Agreed

tessavitabile:
This is ready for re-review

.mci.yml
```diff
@@ -3,6 +3,30 @@
 #######################################
 
 #######################################
+#            Variables                #
+#######################################
+variables:
+
+    ## source variables
```
samantharitter: this is definitely better than doing it by os!

CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Provide -stdlib=libc++ if necessary for cstdint to be defined
```
acmorrow: This shouldn't be necessary

CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ endif()
...
+if(CMAKE_CXX_LIBCPP)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
samantharitter: Why not just pass -stdlib=libc++ as one of the CXX_FLAGS from the mci script?  Just wondering if this is overkill.

samantharitter:
Just a question about the CMAKE_CXX_LBCPP variable.  Also, can you rebase this against master?  Some changes have gone into .mci.yml since you began this patch.

CMakeLists.txt
acmorrow: Oh, I get it. The OS X builders are soo old that libc++ isn't the default. I agree with @samantharitter's suggestion to do this externally via CMAKE_CXX_FLAGS in the YAML file.

CMakeLists.txt
acmorrow: +1

tessavitabile:
Thanks, now I understand your comment from earlier, @samantharitter, and I used -DCMAKE_CXX_FLAGS.

samantharitter:
LGTM!

acmorrow:
I'm good with it if @samantharitter is, she knows the evergreen config world much better than I do.

PR #427, opened 2016-01-11
CXX-805 Use options::find::comment()

tessavitabile:


acmorrow:
The code LGTM.

Interesting to note that the $comment operator in a query was deprecated in 3.2:

https://docs.mongodb.org/manual/reference/operator/meta/comment/

However, the CRUD spec still mentions it as a find option:

https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst

tessavitabile:
I talked to the C driver team about this. My understanding is that since OP_QUERY is deprecated, all of its $operators are deprecated. The 'find' command is now used instead, and the $operators have become keys (without the $) in the command document. However, the C driver still takes a query of the form {$query: ..., $comment: ...}, and converts this to the form the 'find' command expects.

C driver find command implementation spec: https://jira.mongodb.org/browse/CDRIVER-838
Example of using $comment in the C driver: https://github.com/mongodb/mongo-c-driver/blob/master/tests/test-mongoc-collection-find.c#L582

tessavitabile:
Is the code freeze over?

acmorrow:
@tessavitabile - thanks for checking with the drivers team and about the code freeze.

PR #428, opened 2016-01-12
CXX-807 Enforce compiler minima

acmorrow:
This enforces GCC-4.9, MSVC 2015 Update 1, XCode 6.2, and clang 3.5 as minima.

I believe these are more or less correct, we can always iterate if we find we need to raise or lower.


CMakeLists.txt
```diff
@@ -1,6 +1,30 @@
...
 project(MONGO_CXX_DRIVER LANGUAGES CXX)
 
-cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
+    message(FATAL_ERROR "Insufficient GCC version - GCC 4.9+ required")
+  endif()
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.0.23506")
+    message(FATAL_ERROR "Insufficient Microsoft Visual C++ version - MSVC 2015U1+ required")
```
markbenvenuto: Please spell it out, `MSVC 2015 Update 1+`

CMakeLists.txt
```diff
@@ -1,6 +1,30 @@
...
+
+if(POLICY CMP0025)
+  cmake_policy(SET CMP0025 NEW)
+endif()
+
 project(MONGO_CXX_DRIVER LANGUAGES CXX)
 
-cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
```
samantharitter: Yup.  If you're running gcc 4.8.x you won't be able to build the driver now, there's a bug with `std::max_align_t` that I ran into yesterday: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56019

CMakeLists.txt
samantharitter: Is it possible to include the current version in this message?  And would that be helpful to users?

acmorrow: I was going to, but I realized that Cmake actually already does that:
```
KG_CONFIG_PATH=~/opt/lib/pkgconfig cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DCMAKE_INSTALL_PREFIX=./install -G Ninja
-- The CXX compiler identification is AppleClang 7.0.2.7000181
```

CMakeLists.txt
acmorrow: Yeah. I feel a _little_ bad about excluding gcc 4.8. We could possibly ifdef it and lower? I think it is only used in one place. But, lets set the bar high and see if we get pushback.

CMakeLists.txt
acmorrow: @markbenvenuto text updated per your request

samantharitter:
LGTM

markbenvenuto:
LGTM

PR #432, opened 2016-01-12
CXX-746 Add ASAN builder to evergreen project

tessavitabile:
Do we want to add the ASAN builder before the memory leaks are resolved?

### DIFF
--- a/src/bsoncxx/document/value.cpp
+++ b/src/bsoncxx/document/value.cpp
@@ -53,6 +53,9 @@ value& value::operator=(const value& rhs) {
     return *this;
 }
 
+value::value(value&& rhs) noexcept: _data{std::move(rhs._data)}, _length{std::move(rhs._length)} {
+}
+
 value::unique_ptr_type value::release() {
     auto x = std::move(_data);
 
--- a/.mci.yml
+++ b/.mci.yml
@@ -25,6 +25,11 @@ variables:
     cmake_flags:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
+        asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
+
+    ## test parameters
+    test_params:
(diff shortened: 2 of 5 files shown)


samantharitter:
What memory leaks need to be resolved?

LGTM otherwise

tessavitabile:
I have two different failures based on two different commits.

Based on CXX-812 Support gcc 4.8.2:
https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_asan_compile_and_test_67dadb7874796c9af794de652c0855c0bdc69d6c_16_01_12_22_01_34

This one has JIRA ticket CXX-813:
Based on CXX-811 Exit after any failure in MCI task:
https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_asan_compile_and_test_4e7c41aaf1ffd6dee8686c690d010444879c5205_16_01_12_21_21_38

tessavitabile:
Installing cmake may not be necessary soon: BUILD-1134 is in code review.

samantharitter:
Yes, I'd say to investigate those errors and put fixes to the memory leaks as separate commits in this PR.  I took a quick look at the first one, it looks like something might be off with value's move constructor?  It's defaulted, but maybe we need to do something else there.  I'm not sure yet how you would end up with an uninitialized _length, though.

samantharitter:
For the second issue, look into the mongocxx::mock::instance class.  It looks like we have a custom destructor for that class, but we don't clean up its `_callbacks` member, we might be leaking that.

src/bsoncxx/document/value.cpp
```diff
@@ -53,6 +53,9 @@ value& value::operator=(const value& rhs) {
     return *this;
 }
 
+value::value(value&& rhs) noexcept: _data{std::move(rhs._data)}, _length{std::move(rhs._length)} {
```
acmorrow: Wait... this is weird. Why did we need to do this?

tessavitabile: I have no idea. It eliminated the error found here: https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_asan_compile_and_test_67dadb7874796c9af794de652c0855c0bdc69d6c_16_01_12_22_01_34

acmorrow: All other things the same, you don't get the same warning without ASAN?

tessavitabile: No, the other versions passed: https://evergreen.mongodb.com/version/569577b43ff12266560005d0_0

(except for OSX, since I hadn't rebased to get the change lowering the compiler minimum)

acmorrow: I wonder if it is this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59908

acmorrow: I think I would like you to revert this part of the change (just the value move constructor), and add -Wno-error=maybe-uninitialized to the ASAN build.

tessavitabile: Okay

acmorrow: Another option would be to still revert that change, but try using clang's ASAN rather than GCCs? It might be interesting to see if you get better results compared to GCC for the stacks you don't understand yet?

tessavitabile: Sure, I can try that.

PR #433, opened 2016-01-14
CXX-425 Add UBSAN builder to evergreen

tessavitabile:
This already passes in evergreen. Yay no undefined behavior!

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -26,10 +26,12 @@ variables:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
         asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers -Wno-error=maybe-uninitialized"
+        ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer -fsanitize-blacklist=$(pwd)/../etc/ubsan.blacklist -fno-sanitize-recover -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
 
     ## test parameters
     test_params:
         asan_test_params: &asan_test_params ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer-3.5 ASAN_OPTIONS="detect_leaks=1"
+        ubsan_test_params: &ubsan_test_params UBSAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer UBSAN_OPTIONS="print_stacktrace=1"
 
 
 #######################################
@@ -56,11 +58,12 @@ functions:
         command: shell.exec
         params:
             working_dir: "."
+            ## Use --disable-extra-align when configuring the C driver to avoid misalignment errors in UBSAN
             script: |
                 git clone -b master https://github.com/mongodb/mongo-c-driver.git
                 cd mongo-c-driver
                 rm -rf /data/tmp/c-driver-install
-                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled
+                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled --disable-extra-align
                 make
                 make install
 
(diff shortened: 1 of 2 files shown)


tessavitabile:
It didn't actually pass, and -fno-santitize-recover makes it exit with failure if it does not. Here is the current failure:

https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_ubsan_compile_and_test_de649bf1acba78db9f744c99f36a3655b540a2eb_16_01_14_16_07_02

.mci.yml
```diff
@@ -26,10 +26,12 @@ variables:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
         asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers -Wno-error=maybe-uninitialized"
+        ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer -fno-sanitize-recover -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
```
markbenvenuto: Add `-fsanitize-trap=undefined` to ensure it will crash on error. The documentation is confusing, and marking it this way is clearer to me then just `-fno-santize-recover`.

acmorrow: @markbenvenuto Can you elaborate on why this is better? It seems to disable the `vptr` check among other things, and the docs suggest that it is more about situations where you don't want the ubsan runtime. Why is that better than just asking it not to try to continue, as `-fno-sanitize-recover` does.

.mci.yml
```diff
@@ -60,7 +62,7 @@ functions:
                 git clone -b master https://github.com/mongodb/mongo-c-driver.git
                 cd mongo-c-driver
                 rm -rf /data/tmp/c-driver-install
-                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled
+                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled --disable-extra-align
```
tessavitabile: Should I only do this for the UBSAN builder?

acmorrow: Can you add a comment above this explaining why we are disabling extra align?

acmorrow:
The failure is because of a bug in libstdc++: http://stackoverflow.com/questions/30122500/is-this-code-really-undefined-as-clang-seems-to-indicate

The solution is to add a sanitizer blacklist file with the following suppressions:

```
fun:_ZStaNRSt13_Ios_FmtflagsS_
fun:_ZStanSt13_Ios_FmtflagsS_
```

.mci.yml
tessavitabile: This avoids misaligned address errors like the following: https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_ubsan_compile_and_test_de649bf1acba78db9f744c99f36a3655b540a2eb_16_01_14_15_57_52

I can add a comment.

acmorrow:
@tessavitabile Please add the blacklist above to a new file etc/ubsan.blacklist. You can also add a comment link above those to https://wiki.documentfoundation.org/Development/-fsanitize where someone else figured out exactly the two names to suppress.

Then, you can add `-fsanitize-blacklist=$(pwd)/../etc/ubsan.blacklist` to the CMake invocation. The blacklist needs to be fed in at compile time.

I have it working locally, let me know if you need some help getting it set up. Also, it doesn't seem that your UBASN_OPTIONS for the stack trace is actually working, as they don't seem to be getting printed out. Not sure why.

.mci.yml
```diff
@@ -26,10 +26,12 @@ variables:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
         asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers -Wno-error=maybe-uninitialized"
+        ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer -fno-sanitize-recover -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
 
     ## test parameters
     test_params:
         asan_test_params: &asan_test_params ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer-3.5 ASAN_OPTIONS="detect_leaks=1"
+        ubsan_test_params: &ubsan_test_params UBSAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer UBSAN_OPTIONS="print_stacktrace"=1
```
acmorrow: Ah... you need to move the =1 inside the quotes, I think

.mci.yml
acmorrow: No, it is fine to do it for all builders. The extra alignment is stupid.

tessavitabile:
This is passing: https://evergreen.mongodb.com/task/mongodb_cpp_driver_dev_ubuntu1410_debug_ubsan_compile_and_test_de649bf1acba78db9f744c99f36a3655b540a2eb_16_01_14_16_59_49

.mci.yml
markbenvenuto: Sorry, I had mis-remembered. The `-fsantizie-trap` is good debugging ubsan failures since it generates faults but lacks good logging. Never mind.

acmorrow:
Awesome! LGTM

etc/ubsan.blacklist
```diff
@@ -0,0 +1,2 @@
+fun:_ZStaNRSt13_Ios_FmtflagsS_
```
acmorrow: Could you add a comment or a link here? The SO post or that other one is fine. Just something so we don't puzzle over it a year from now.

acmorrow:
Thanks. Definitely LGTM

samantharitter:
LGTM

PR #434, opened 2016-01-14
CXX-531 Test agains all sources of stdx polyfills

tessavitabile:
https://evergreen.mongodb.com/version/5697cd8a3ff122756500022b_0

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -157,7 +157,7 @@ buildvariants:
           - name: compile_and_test
 
     - name: ubuntu1404-debug-boost
-      display_name: "Ubuntu 14.04 Debug with boost"
+      display_name: Ubuntu 14.04 Debug (boost)"
       expansions:
           build_type: "Debug"
           source: *ubuntu_source
@@ -170,7 +170,7 @@ buildvariants:
           - name: compile_and_test
 
     - name: ubuntu1404-debug-std-experimental
-      display_name: "Ubuntu 14.04 Debug with std experimental"
+      display_name: "Ubuntu 14.04 Debug (experimental)"
       expansions:
           build_type: "Debug"
           source: *ubuntu_source
--- a/.mci.yml
+++ b/.mci.yml
@@ -25,7 +25,7 @@ variables:
     cmake_flags:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         ubuntu_boost_cmake_flags: &ubuntu_boost_cmake_flags -DBSONCXX_POLY_USE_BOOST=ON -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
-        ubuntu_std_experimental_cmake_flags: &ubuntu_std_experimental_cmake_flags -DBSONCXX_POLY_USE_STD_EXPERIMENTAL=ON -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
+        ubuntu_std_experimental_cmake_flags: &ubuntu_std_experimental_cmake_flags -DBSONCXX_POLY_USE_STD_EXPERIMENTAL=ON -DCMAKE_CXX_STANDARD=14 -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
         osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON -DLIBBSON_DIR=/data/tmp/c-driver-install -DLIBMONGOC_DIR=/data/tmp/c-driver-install
         asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers -Wno-error=maybe-uninitialized"
         ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer -fsanitize-blacklist=$(pwd)/../etc/ubsan.blacklist -fno-sanitize-recover -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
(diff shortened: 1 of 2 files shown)


.mci.yml
```diff
@@ -24,6 +24,8 @@ variables:
     ## cmake flag variables
     cmake_flags:
         ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
+        ubuntu_boost_cmake_flags: &ubuntu_boost_cmake_flags -DBSONCXX_POLY_USE_BOOST=ON -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
```
acmorrow: Can you add `-DBSONCXX_POLY_USE_MNMLSTC=OFF` here as well? Currently, the options aren't well factored (I have something in flight) but for now it is the right thing to do.

.mci.yml
```diff
@@ -24,6 +24,8 @@ variables:
...
+        ubuntu_std_experimental_cmake_flags: &ubuntu_std_experimental_cmake_flags -DBSONCXX_POLY_USE_STD_EXPERIMENTAL=ON -DCMAKE_CXX_STANDARD=14 -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
```
acmorrow: Here as well for explicitly disabling MNMLSTC as well as enabling std::experimental

acmorrow:
@tessavitabile FYI if #435 goes first, you don't need to change anything. So, LGTM if we sequence things right.

.mci.yml
```diff
@@ -151,6 +153,32 @@ buildvariants:
       tasks:
           - name: compile_and_test
 
+    - name: ubuntu1404-debug-boost
+      display_name: "Ubuntu 14.04 Debug with boost"
```
samantharitter: for brevity, maybe just "Ubuntu 14.04 Debug (boost)" and "Ubuntu 14.04 Debug (experimental)".  In some views mci doesn't show long names well.

I'd also be ok with an even shorter version, like "Ubuntu (boost)" and "Ubuntu (experimental)."  Of course with this you'd have to go look in .mci.yml to see the ubuntu version, but meh.  I leave it to your discretion.

samantharitter:
Nit about long buildvariant names, otherwise LGTM

tessavitabile:
When I added -DBSONCXX_POLY_USE_MNMLSTC=OFF, I had compile errors. For boost, it could not find boost/smart_ptr/make_unique.hpp. For std experimental, it could not find experimental/string_view.

https://evergreen.mongodb.com/version/5697d4a43ff12275650002de_0

acmorrow:
Interesting. I've tested them all locally. I think without the options refactor it wasn't actually honoring the request to use boost or std::experimental and was still using mnmlstc.

tessavitabile:
Okay. After #435 is committed, I'll remove -DBSONCXX_POLY_USE_MNMLSTC=OFF, rebase and test on evergreen again.

acmorrow:
Well...

- For boost, it looks like the boost is too old to have support for `make_unique`. Can you find out what the first version of boost to offer that was? Then you can bump the minimum required boost in the `find_package(Boost` call in the bsoncxx CMakeLists.txt file.
- For std::experimental, it looks like GCC 4.8 doesn't offer it. Want to see if we get better mileage with clang and libc++? I'm not sure that is installed either. Maybe try it on OS X?

acmorrow:
FYI #435 is in, so you shouldn't need to make any changes per my earlier comments. The lack of building though may still be an issue :)

tessavitabile:
For boost, the first version including make_unique was 1.56.0. The ubuntu 14.04 machines have 1.54.0. However, our OSX builder is already using boost. Is that a sufficient test?

For std::experimental, I tried using clang and libc++ on ubuntu, but cmake considers that clang insufficient. I tried it on OSX and got the same error: that experimental/string_view is not found. I could try clang++ on ubuntu?

acmorrow:
Well, first thing, lets add the 1.58 requirement to the `find_package(Boost...)` so that we lock that knowledge in. That will help prevent users from getting confused.

As far as the OS X builder, I don't think it was actually using boost:

https://evergreen.mongodb.com/task_log_raw/mongodb_cpp_driver_dev_osx_108_compile_and_test_de649bf1acba78db9f744c99f36a3655b540a2eb_16_01_13_23_26_21/0?type=T

And look for EP_. However, my change to the polyfill options means that it will start using boost Real Soon Now (as soon as the build queue catches up). So, we will find out.

We may just not have a new enough system available on MCI to test with std::experimental at this time. Hopefully we will at some point soon when we upgrade the OS X builders, or introduce new variants.

tessavitabile:
Oh, I see. (Though I wonder why turning on Boost caused it to compile on OS X, if it wasn't actually using Boost? Very confusing.)

.mci.yml
acmorrow: Please disregard since the other commit merged.

.mci.yml
acmorrow: Ditto

tessavitabile:
Should I push the boost minimum in a minor commit and abandon this for now?

src/bsoncxx/CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ include_directories(
 link_directories(${LIBBSON_LIBRARY_DIRS})
 
 if (BSONCXX_POLY_USE_BOOST)
-    find_package(Boost REQUIRED)
+    find_package(Boost 1.56.0 REQUIRED)
```
acmorrow: Thanks

.mci.yml
acmorrow: According to this:

http://stackoverflow.com/questions/30995705/cplusplus-201402l-return-true-in-gcc-even-when-i-specified-std-c14

We need either GCC-5.1 or clang-3.5.

I think you should check in with Jonathan and see if we have /any/ MCI images that offer these toolchains yet.

acmorrow:
@tessavitabile - Yes, please push the boost minimum commit as is under this ticket. We may need to hold off on the reset.

.mci.yml
acmorrow: Great news: Our Ubuntu 14.10 variant has clang-3.5.

tessavitabile: Indeed! Sorry, I should have commented. It has std::experimental, but something is broken, which I was looking at:
https://evergreen.mongodb.com/task_log_raw/mongodb_cpp_driver_dev_ubuntu1404_debug_std_experimental_compile_and_test_3f3499ee80b26fe3c110340520b959dbf7df846a_16_01_14_21_52_21/0?type=T

acmorrow: And it looks like libc++ on that platform has experimental optional and string_view. I'm trying to build on a spawnhost right now.

acmorrow:
We are going to get testing with boost on Windows once these tickets are done:

https://jira.mongodb.org/browse/BUILD-1129
https://jira.mongodb.org/browse/BUILD-1136

It appears that std::experimental just isn't ready yet given our current builders. Hopefully that will not be an issue once @jbreams new toolchain is up and running, as that will give us GCC 5.3

acmorrow:
@tessavitabile I think we should close this PR if we don't plan to do more with it right now.

PR #435, opened 2016-01-14
CXX-817 Improve bsoncxx polyfill options handling

acmorrow:
Make it easier to select the other polyfills by not needing to explicitly disable mnmlstc at the same time. Also validate that the combination of options is sensible.

### DIFF
--- a/src/bsoncxx/CMakeLists.txt
+++ b/src/bsoncxx/CMakeLists.txt
@@ -14,9 +14,6 @@
 
 project(BSONCXX)
 
-set(LIBBSON_REQUIRED_ABI_VERSION 1.0)
-find_package(LibBSON 1.3.0 REQUIRED)
-
 # Update these as needed.
 # TODO: read from file
 set(BSONCXX_VERSION_MAJOR 3)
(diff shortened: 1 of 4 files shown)


samantharitter:
A question about one of the polyfill flags, otherwise LGTM

acmorrow:
@samantharitter - It seemed to make the state space of options larger for no real gain. I would have needed to validate more combinations of options. This way, it just is 'on', unless you take an action that makes it be implicitly off.

tessavitabile:
LGTM. 

One thing I find confusing about the options is that if I run cmake -DBSONCXX_POLY_USE_BOOST=ON, then I run cmake -DBSONCXX_POLY_USE_STD_EXPERIMENTAL=ON, the second time both will be on. The options aren't reset to their defaults each time you run cmake.

acmorrow:
I suspect we are supposed to be using `cmake_dependent_option`.

acmorrow:
@tessavitabile @samantharitter - I updated this, PTAL. I added two new commits. The first improves how the boost header search path is propagated when using boost, and the second fixes two compile issues when using the stdx::experimental polyfills.

tessavitabile:
I don't totally understand, but nothing looks glaringly wrong, so LGTM. I guess we use emplace to set string_views.

acmorrow:
It seems that in std::experimental::optional, the templated operator= only works if U can decay to T. In our case, it can't. Using emplace works everywhere though.

acmorrow:
As for the CMake stuff, yeah, I wouldn't worry too much about it.

samantharitter:
still LGTM

PR #438, opened 2016-01-14
CXX-810 Remove WriteConcern fsync property

tessavitabile:


src/mongocxx/write_concern.hpp
```diff
@@ -52,8 +52,6 @@ class uri;
 ///
 /// @see http://docs.mongodb.org/manual/core/write-concern/
 ///
-/// @todo this interface is terrible -- it's exactly what you get with the c-driver essentially.
```
tessavitabile: Do you think the interface is terrible? I don't know why it's an issue to have it be so similar to the C driver.

acmorrow:
@ajdavis - Has the C driver not deprecated these interfaces? I'm surprised we were able to use them without seeing a deprecation warning.

src/mongocxx/write_concern.hpp
acmorrow: I don't think there is anything particularly wrong with it. Who wrote the comment?

acmorrow:
I love pure negative line count commits. LGTM

src/mongocxx/write_concern.hpp
tessavitabile: It appears to be Tyler Brock.

acmorrow: @TylerBrock - any recollection of what you thought was wrong with this one?

TylerBrock: hey everyone, I don't recall why I thought there was room for improvement, forge ahead!

acmorrow: Thanks @TylerBrock!

acmorrow:
LGTM

src/mongocxx/write_concern.hpp
samantharitter: Hi Tyler!

samantharitter:
LGTM

ajdavis:
@acmorrow good point, I've opened https://jira.mongodb.org/browse/CDRIVER-1061 to deprecate get/set_fsync.

PR #439, opened 2016-01-16
CXX-794 Clean up header ordering, grouping, and prelude/postlude

acmorrow:
- The prelude/postlude should be scoped narrowly
- Use the private prelude/postlude for private files
- Group bsoncxx and mongocxx includes together
- Fix some header sorting where noticed
- Add some missing postludes
- Add some missing preludes exposed by adding missing postludes

### DIFF
--- a/src/bsoncxx/string/view_or_value.hpp
+++ b/src/bsoncxx/string/view_or_value.hpp
@@ -14,12 +14,12 @@
 
 #pragma once
 
-#include <bsoncxx/config/prelude.hpp>
-
 #include <string>
 
 #include <bsoncxx/stdx/string_view.hpp>
 #include <bsoncxx/view_or_value.hpp>
+
+#include <bsoncxx/config/prelude.hpp>
 
 namespace bsoncxx {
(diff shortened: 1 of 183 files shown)


acmorrow:
FYI I don't really expect anyone to review all of this.

src/bsoncxx/string/view_or_value.hpp
```diff
@@ -14,13 +14,13 @@
...
 #pragma once
 
-#include <bsoncxx/config/prelude.hpp>
-
 #include <string>
 
 #include <bsoncxx/stdx/string_view.hpp>
 #include <bsoncxx/view_or_value.hpp>
 
+#include <bsoncxx/config/prelude.hpp>
```
samantharitter: not private/prelude.hpp?

samantharitter:
A question about when to use which header, otherwise LGTM

src/bsoncxx/string/view_or_value.hpp
acmorrow: This is a publicly visible header, so it should get the public prelude header. .cpp files and private headers get the private prelude.

PR #440, opened 2016-01-17
CXX-522 Add calling convention tags where needed

acmorrow:
On Windows, non-inline non-member functions of non-template classes
need to have a calling convention explicitly specified, so that if the
consumer of the library isn't using cdecl as the default calling
convention, they can still link with the library.

Also a few drive-by fixes:

- Remove anon namespaces in headers, they don't do what is expected
- Remove some duplicated declarations when a 'friend' already declared
- Fix some visibility nits (missing API, missing PRIVATE)
- Remove trailing class comments - we don't do those
- Make read_preference comparator a free function as is our style

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -22,6 +22,12 @@ add_custom_target(install_libs
     DEPENDS bsoncxx_built mongocxx_built
 )
 
+# Build the examples with vectorcall as the default on Windows
+# so that we shake out missing _CALL macros.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Gv")
+endif()
+
 add_subdirectory(bsoncxx)
 add_subdirectory(mongocxx)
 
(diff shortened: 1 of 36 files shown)


acmorrow:
@markbenvenuto would like your thoughts on this one

markbenvenuto:
Have you tried compiling the examples with `Gv`? That will force the default calling convention to be `_vectorcall` and will help find any issues. The WiredTiger Windows MCI builder has an alternate builder where it uses `Gv` to help root out any issues with missing `_cdecl`.

acmorrow:
@markbenvenuto trying /Gv now. Do you think that is worth just unconditionally doing for the Windows example builds?

examples/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ add_custom_target(install_libs
     DEPENDS bsoncxx_built mongocxx_built
 )
 
+# Build the examples with vectorcall as the default on Windows
+# so that we shake out missing _CALL macros.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Gv")
+endif()
```
acmorrow: @markbenvenuto I tested this out, the examples link just fine. Should we push it that way?

PR #441, opened 2016-01-18
CXX-813 Fix memory errors exposed by ASAN and Valgrind

acmorrow:
- We can't assume that libmongoc has automatically called init/cleanup
  for us. It only does that on some platforms. That makes it mandatory
  to have an instance object. Many of our tests didn't do that, and it
  would be painful for users. Set things up so that an instance is
  implicitly created as needed, and provide ways to retrieve the
  instance. This required moving the tests for instance to its own
  process, among other nuisances.

- The collection::create_index method was not cleaning up the keys that
  had been allocated by libbson. Add the needed bson_free call.

- The collection::distinct method was not cleaning up the temorary
  database object that it constructs. Add the needed database_destroy
  call.

- Fix some mocks so that they write to the bson_error_t out parameter
  when returning a non-successful code, as otherwise these lead to
  read-from-uninit errors when we promote the uninitialized bson_error_t
  to an exception.

### DIFF
--- a/src/mongocxx/exception/error_code.hpp
+++ b/src/mongocxx/exception/error_code.hpp
@@ -22,7 +22,8 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 
 enum class error_code : std::int32_t {
-    k_invalid_client_object = 1,
+    k_instance_already_exists = 1,
+    k_invalid_client_object,
     k_invalid_collection_object,
     k_invalid_database_object,
     k_invalid_parameter,
--- a/src/mongocxx/instance.cpp
+++ b/src/mongocxx/instance.cpp
@@ -14,10 +14,13 @@
 
 #include <mongocxx/instance.hpp>
 
+#include <mutex>
 #include <utility>
 
 #include <bsoncxx/stdx/make_unique.hpp>
+#include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/logger.hpp>
+#include <mongocxx/exception/private/error_code.hpp>
 #include <mongocxx/private/libmongoc.hpp>
 
 #include <mongocxx/config/private/prelude.hpp>
@@ -57,6 +60,10 @@ void user_log_handler(::mongoc_log_level_t mongoc_log_level, const char *log_dom
                                         stdx::string_view{log_domain}, stdx::string_view{message});
 }
 
+std::recursive_mutex instance_mutex;
+instance *current_instance = nullptr;
+std::unique_ptr<instance> global_instance;
+
 }  // namespace
 
 class instance::impl {
@@ -85,13 +92,33 @@ class instance::impl {
 instance::instance() : instance(nullptr) {
 }
 
-instance::instance(std::unique_ptr<logger> logger)
-    : _impl(stdx::make_unique<impl>(std::move(logger))) {
+instance::instance(std::unique_ptr<logger> logger) {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (current_instance) {
+        throw logic_error(make_error_code(error_code::k_instance_already_exists));
+    }
+    _impl = stdx::make_unique<impl>(std::move(logger));
+    current_instance = this;
 }
 
 instance::instance(instance &&) noexcept = default;
 instance &instance::operator=(instance &&) noexcept = default;
-instance::~instance() = default;
+
+instance::~instance() {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (current_instance != this) std::abort();
+    _impl.reset();
+    current_instance = nullptr;
+}
+
+instance &instance::current() {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (!current_instance) {
+        global_instance.reset(new instance);
+        current_instance = global_instance.get();
+    }
+    return *current_instance;
+}
 
 MONGOCXX_INLINE_NAMESPACE_END
 }  // namespace mongocxx
--- a/src/mongocxx/test/bulk_write.cpp
+++ b/src/mongocxx/test/bulk_write.cpp
@@ -17,7 +17,7 @@
 
 #include <bsoncxx/builder/stream/document.hpp>
 #include <bsoncxx/types.hpp>
-
+#include <mongocxx/instance.hpp>
 #include <mongocxx/private/libmongoc.hpp>
 #include <mongocxx/bulk_write.hpp>
 #include <mongocxx/write_concern.hpp>
@@ -25,6 +25,8 @@
 using namespace mongocxx;
 
 TEST_CASE("a bulk_write will setup a mongoc bulk operation", "[bulk_write]") {
+    instance::current();
+
     auto construct = libmongoc::bulk_operation_new.create_instance();
     bool construct_called = false;
     bool ordered_value = false;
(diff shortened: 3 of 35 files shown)


acmorrow:
@hanumantmk @amidvidy @ajdavis I'd be interested in your thoughts on the changes around `instance`. I don't love them, but we do need a better story on how we initialize and terminate the driver. The other option would be to unroll these changes, and figure out how to get all of the catch tests to share an instance object.

acmorrow:
Ping?

src/mongocxx/instance.cpp
```diff
@@ -84,13 +89,33 @@ class instance::impl {
 instance::instance() : instance(nullptr) {
 }
 
-instance::instance(std::unique_ptr<logger> logger)
-    : _impl(stdx::make_unique<impl>(std::move(logger))) {
+instance::instance(std::unique_ptr<logger> logger) {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (current_instance)
+        std::abort();
```
samantharitter: Possible to error with a helpful message?

src/mongocxx/instance.cpp
```diff
@@ -84,13 +89,33 @@ class instance::impl {
...
 }
 
 instance::instance(instance &&) noexcept = default;
 instance &instance::operator=(instance &&) noexcept = default;
-instance::~instance() = default;
+
+instance::~instance() {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (current_instance != this)
+        std::abort();
```
samantharitter: Possible to error with a helpful message?

acmorrow: Not really, since we may be in static destruction here. This should be impossible. If we had `invariant`, I'd write that instead.

src/mongocxx/instance.cpp
acmorrow: Yes, good suggestion. Thanks.

acmorrow:
One thing to note in this review is that it is still possible to fail to initialize the C driver correctly. For instance, if the first thing you constructed was a read_preference object, you might end up calling `mongoc_read_prefs_new` before an `instance` was constructed.

Another way to go with this change would be to unwind my auto-instance work in `client` and `pool`, and fix all of the tests and examples to manually create an instance (we could keep the automatic global instance though).

Thoughts?

src/mongocxx/instance.cpp
```diff
@@ -84,13 +89,33 @@ class instance::impl {
...
+    current_instance = this;
 }
 
 instance::instance(instance &&) noexcept = default;
 instance &instance::operator=(instance &&) noexcept = default;
-instance::~instance() = default;
+
+instance::~instance() {
+    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
+    if (current_instance != this)
```
samantharitter: What about default constructed instances?  Unlike constructing one with a logger, the default ctor won't set `current_instance`.  Is it possible/legal to have a default-constructed instance alive while there is another `current_instance`?

ajdavis:
To the extent I understand C++, the code looks fine to me.

But I dislike the philosophy of this approach: the driver should require users to explicitly call init and cleanup.

The C Driver *should* require you to call mongoc_init and mongoc_cleanup, and on non-GCC it *does* require you to call them. (There is a misfeature that uses GCC tricks to do call init and cleanup automatically, I plan to remove that in the C Driver 2.0.)

So, for the sake of correctness, clarity, simplicity, I propose requiring the user to init and cleanup the C++ driver explicitly.

acmorrow:
@ajdavis I think I agree. However, requiring them to drag around an object forever seems cruel. I think what I will do is to unwind the magic auto-initialization in `client` and `pool`, but retain the ability for 'instance' to manage a singleton. Then, I'll plumb calls to `instance::current()` through all the tests, ensuring that the global instance has been constructed. Does that sound reasonable?

ajdavis:
Oh, I misunderstood, partly. I was arguing that the C++ driver should make the user do something manually that executes the C Driver's global initialization and cleanup routines, the same way the C Driver requires users to call mongoc_init and mongoc_cleanup themselves.

I don't have an opinion about the "instance" object, I'm not familiar with what it does.

acmorrow:
@ajdavis The `instance` object is exactly that. When constructed, it causes `mongoc_init` to be called. When destroyed, it causes `mongoc_cleanup` to be called.

acmorrow:
Updated, PTAL. No longer automatically creates an instance for you, so all the tests now call `instance::current()` to ensure that one exists. That is overkill, but safe. Also, made it so that `instance` now throws if you attempt to construct more than one.

ajdavis:
LGTM.

In the docs, do you or will you recommend that C++ driver users make an
"instance" on the stack at the top of main()? If that's all a user must do
in C++ that's great.

Our examples show mongoc_init and mongoc_cleanup at the top and bottom of
main().

acmorrow:
@ajdavis Yes that is the idea. I doubt we have it documented anywhere yet though.

src/mongocxx/instance.cpp
acmorrow: Done

src/mongocxx/exception/error_code.hpp
```diff
@@ -29,6 +29,7 @@ enum class error_code : std::int32_t {
     k_ssl_not_supported,
     k_unknown_read_concern,
     k_unknown_write_concern,
+    k_instance_already_exists,
```
samantharitter: alphabetize?

src/mongocxx/test/bulk_write.cpp
```diff
@@ -25,6 +25,8 @@
 using namespace mongocxx;
 
 TEST_CASE("a bulk_write will setup a mongoc bulk operation", "[bulk_write]") {
+    instance::current();
```
samantharitter: Is there any way that catch could help us not have to do this in every single test?  It seems like the only way to do that would be to change these TEST_CASEs into SECTIONs within one overarching bulk_write TEST_CASE.  Probably not worth it.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -196,19 +201,29 @@ TEST_CASE("Collection", "[collection]") {
...
+            REQUIRE(bson_get_data(query) == filter_doc.view().data());
+            REQUIRE(skip == expected_skip);
+            REQUIRE(limit == expected_limit);
+            if (expected_opts) {
+                REQUIRE(bson_equal(cmd_opts, expected_opts));
+            }
+
+            if (success) return 123;
+
+            // The caller expects the bson_error_to have been
```
samantharitter: bson_error_t to

samantharitter:
A few nits, also consider adding a test that we throw if you attempt to use the driver without creating an instance, otherwise LGTM

samantharitter:
Oh actually why don't the examples also need `instance::current()` ?

samantharitter:
LGTM after in-person discussion

src/mongocxx/exception/error_code.hpp
acmorrow: Per discussion, we can't do with enums (or lots of other things to) because ABI. However, we haven't set an ABI yet, so I will put it in the right place.

src/mongocxx/test/bulk_write.cpp
acmorrow: I can't think of one. It was a pain to add, but hopefully it just becomes part of how you write a test.

src/mongocxx/test/collection_mocked.cpp
acmorrow: Done

acmorrow:
- re the adding a test: we can't detect if you fail to set up, so I don't think we can test that. I could write a test that making an `instance` twice throws. Do you think that is worth it?
- Re the examples, they already do it, by declaring an instance object as `mongocxx::instance instance{}`, so they don't need to call `instance:current()`.

samantharitter:
Eh, I'm fine without the test too

PR #442, opened 2016-01-18
CXX-819 Default to boost polyfills on MSVC

acmorrow:
No point making things hard for users. Right now the only way to build the driver on Windows is with Boost, so default to that.


markbenvenuto:
:+1:

samantharitter:
LGTM

PR #443, opened 2016-01-18
CXX-820 Add valgrind support on MCI

acmorrow:
Currently red (https://evergreen.mongodb.com/version/569d4a2f3ff1227dfb000040_0) but has fixes for many of those issues pending in #441.


.mci.yml
```diff
@@ -116,7 +117,20 @@ tasks:
                   make
                   ${test_params} ./src/bsoncxx/test/test_bson
                   ${test_params} ./src/mongocxx/test/test_driver
-                  make run-examples
+                  ${test_params} ./examples/bsoncxx/builder_basic
```
samantharitter: it's too bad we don't have any kind of looping functionality in yaml

markbenvenuto: Actually, this should be a `find` with `xargs` so that we do not have update .mci.yml if new examples are added.

acmorrow:
@markbenvenuto updated, PTAL

markbenvenuto:
LGTM

PR #444, opened 2016-01-19
CXX-680: Clang-Format integration

markbenvenuto:
Clang-Format Integration
- Uses a modified version of clang-format.py from mongo master branch (merged in globstar.py from resmoke,py, adjustment for clang-format.py location)
- I filter out `examples` and `third-party`.
- I added `lint` integration into Travis and MCI.

Note: I will merge the changes before merge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,20 +79,13 @@ add_custom_target(docs
     VERBATIM
 )
 
-add_custom_target(modernize
-    clang-modernize
-        -p ${CMAKE_BINARY_DIR}/compile_commands.json
-        -include ${CMAKE_CURRENT_SOURCE_DIR}
-        -exclude ${CMAKE_CURRENT_SOURCE_DIR}/mongocxx/test ${CMAKE_CURRENT_SOURCE_DIR}/bsoncxx/test
-        -format
+add_custom_target(format
+    python ${CMAKE_SOURCE_DIR}/etc/clang_format.py format
     VERBATIM
 )
 
-add_custom_target(format
-    perl -nle "/file/ or next; print ((split /: (.*)/)[1])" ${CMAKE_BINARY_DIR}/compile_commands.json |
-    grep -v "catch.hpp" |
-    grep -v "examples" |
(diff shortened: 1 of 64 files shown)


src/mongocxx/test/private/write_concern.cpp
```diff
@@ -23,7 +23,6 @@ using namespace mongocxx;
 
 TEST_CASE("creation of write_concern passes universal parameters to c-driver's methods",
           "[write_concern][base][c-driver]") {
```
acmorrow: I don't get why this line went away. If this line went away, I'd expect hundreds of others to also go away. Can you check that the script didn't fall over half way through? There are other formatting issues that I expected to see fixed by this commit.

CMakeLists.txt
```diff
@@ -89,10 +89,12 @@ add_custom_target(modernize
...
 add_custom_target(format
-    perl -nle "/file/ or next; print ((split /: (.*)/)[1])" ${CMAKE_BINARY_DIR}/compile_commands.json |
-    grep -v "catch.hpp" |
-    grep -v "examples" |
-    xargs clang-format -i
+    python ${CMAKE_SOURCE_DIR}/etc/clang_format.py format
+    VERBATIM
+)
+
+add_custom_target(lint
```
acmorrow: call this format-lint? It isn't really a traditional lint.

src/mongocxx/test/private/write_concern.cpp
markbenvenuto: This is because `KeepEmptyLinesAtTheStartOfBlocks: false`. Since this is the start of a block, the blank line is removed. The same setting is used in the server.

CMakeLists.txt
markbenvenuto: done

acmorrow:
LGTM

etc/clang_format.py
```diff
@@ -0,0 +1,769 @@
+#! /usr/bin/env python
```
acmorrow: copyright notice?

PR #446, opened 2016-01-21
CXX-825 Use bigobj for the mocked library as well

acmorrow:
Also, don't build the examples as part of the 'all' target. Instead
require users to build them explicitly by building the 'examples'
target.


samantharitter:
LGTM

markbenvenuto:
LGTM

PR #447, opened 2016-01-22
Add explicit path to PKG_CONFIG_PATH for BSON and MongoC libraries

pavel-odintsov:
Hello, folks!

I'm using Mongo C++ module with my own manually compiled version of mongo-c and bson driver.

I specified path to they manually with cmake flags:
```bash
/opt/fastnetmon/libraries/cmake-3.4.2/bin/cmake -DLIBMONGOC_DIR=/opt/fastnetmon/libraries/mongo_c_driver_1_3_0 -DLIBBSON_DIR=/opt/fastnetmon/libraries/mongo_c_driver_1_3_0
```

As you can see it should work. But in your code pkgconfig call are located before any attempts to locate bson and mongo-c from custom path.

So I have added option to set path for pkg-config toolkit to custom place if it's specified. 

Thanks!

### DIFF
--- a/cmake/FindLibBSON.cmake
+++ b/cmake/FindLibBSON.cmake
@@ -23,6 +23,11 @@ include(FindPackageHandleStandardArgs)
 find_package(PkgConfig QUIET)
 
 if (PKG_CONFIG_FOUND)
+  
+  if (LIBBSON_DIR) 
+      set( ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${LIBBSON_DIR}/lib/pkgconfig" )
+  endif()
+
   pkg_check_modules(LIBBSON REQUIRED libbson-1.0>=${LibBSON_FIND_VERSION} )
   # We don't reiterate the version information here because we assume that
   # pkg_check_modules has honored our request.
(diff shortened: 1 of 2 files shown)


acmorrow:
Hi @pavel-odintsov -

Thanks for the pull request. Typically, I achieve building against a custom C driver install by building with PKG_CONFIG_PATH set in the shell environment when I call CMake:

`PKG_CONFIG_PATH=~/opt/lib/pkgconfig cmake ...`

Can you see if that works for you, without this path? I think it is the better way to go when you have pkgconfig. I view the _DIR variables as a worst case fallback for when it is the only option, so I'm a little reluctant to extend them further.

pavel-odintsov:
Hello!

Thanks! I'm using same approach right now. It's working well!

acmorrow:
@pavel-odintsov OK, should I close this then?

pavel-odintsov:
Yep, you could. But actually explicit LIBMONGOC_DIR are useless without PKG_CONFIG_PATH environment variable. Maybe you could drop they and use only environment variable.

What do you think about it?

acmorrow:
Right, they are only intended for use on systems where pkgconfig isn't an option, like windows.

pavel-odintsov:
Yes it make sense. I will use pkg-config env option.

PR #448, opened 2016-01-22
CXX-758 Fix read_concern

samantharitter:


src/mongocxx/read_concern.cpp
```diff
@@ -80,10 +84,10 @@ stdx::optional<read_concern::level> read_concern::acknowledge_level() const {
     }
 }
 
-stdx::optional<stdx::string_view> read_concern::acknowledge_string() const {
+stdx::string_view read_concern::acknowledge_string() const {
     auto level = libmongoc::read_concern_get_level(_impl->read_concern_t);
     if (!level) {
-        return stdx::nullopt;
+        return "";
```
acmorrow: Do we want to return an empty string here or a default constructed string_view?

src/mongocxx/test/read_concern.cpp
```diff
@@ -24,12 +24,12 @@ using namespace mongocxx;
...
-    SECTION("does not have a level") {
-        REQUIRE(!rc.acknowledge_level());
+    SECTION("has level k_server_default") {
+        REQUIRE(rc.acknowledge_level() == read_concern::level::k_server_default);
     }
 
-    SECTION("does not have a string") {
-        REQUIRE(!rc.acknowledge_string());
+    SECTION("has an empty string") {
+        REQUIRE(rc.acknowledge_string() == "");
```
acmorrow: Same question here

acmorrow:
This LGTM, only with a question about whether we want to return `""` or `{}`

PR #449, opened 2016-01-22
CXX-827 Uri constructor takes a string::view_or_value

samantharitter:


acmorrow:
LGTM. I pulled the ticket into rc1 as this seems like an important fix and very safe.

PR #450, opened 2016-01-22
CXX-821 Add links to document validation documentation

samantharitter:
Note: the todos related to read_concern are fixed in the read_concern pull request.


acmorrow:
These changes look good. However, I think there are some other "link to docs" todos client/collection/database, mostly about read_concern

acmorrow:
Also the read_concern class itself

samantharitter:
Right, the read_concern ones are addressed in the read_concern pull request.  I don't think there were any others besides read_concern, did you find any?

acmorrow:
Oh, sorry. I hadn't looked at that one so didn't notice you had picked them up yet. LGTM then

PR #451, opened 2016-01-22
CXX-828 Clean up handling of versions in pc files

acmorrow:
Fixes the bug in CXX-828, and also tightens up the version relationships.


samantharitter:
LGTM

PR #454, opened 2016-01-28
CXX-754 Radically simplify bsoncxx/mongocxx exceptions and error_codes

acmorrow:
This is the first proposed change in the error handling refactor. While investigating what we needed to do for mongocxx, it was clear we should start with bsoncxx.

I'll be stacking more commits above this, but this describes the first one. It does several things:

- It eliminates the other error_categories. We care about only one error coming back from libbson, which is "I couldn't parse your JSON". By removing the other error_categories and just adding a new code of our own to express that, we make it so we don't leak any of libbson's error handling abstractions
- It makes public the make_error_code overload, and specializes the is_error_code_enum, allowing ADL-ified construction and comparison. Nowhere do you now need to invoke make_error_code.
- Removes all of the exception stuff that was previously private.

If, later, we find that we need to express other libbson errors, we will write our own new codes for them in the one error_category.

I plan to pursue a similar strategy to this in mongocxx, though it will necessarily be more complex.


acmorrow:
I forgot to mention one other thing this does: it changes the from_json function to throw on error, rather than return a disengaged optional. Otherwise, there is no way to get libbson's string describing what went wrong!

acmorrow:
OK, updated. This is what I would like to ship for Friday.

For bsoncxx, we just pull directly the one type of error that we care about from libbson into our error space. If, in the future, we need to do this more, we keep doing that. That way the libbson error abstraction never leaks out.

For mongocxx, we have two error spaces. One for the mongocxx library and its local implementation (libmongoc), and a second for server errors. Any locally generated error will eventually have an entry in the mongocxx library error space. Server errors will be represented as symbols in the server_error_space.

However, since we can't currently reliably distinguish between libmongoc locally generated errors and server errors, we make no effort to do so. Instead, we chose to treat any error coming back from libmongoc as a "server error", and we make the server_error_code enum empty.

Users must read the tea leaves for the numeric codes held in the std::system_error contained in the exception.

This isn't great, but it should let users react to the most common sources of errors. Later, when libmongoc gives us a way to distinguish server errors from local errors, we will map the local errors to new symbolic values in the mongocxx::error_codes enum, and we will populate the server_error_codes enum with the symbols that are 1:1 with the error codes from mongo/base/error_codes.err.

I've also taken @tessavitabile's original example from #425 and added it here, with some changes.

I feel like this gives us the best outcome. We don't need any heuristics. Our users get numeric codes for now, but we can symbolize later. We haven't painted ourselves into any sort of corner.

examples/mongocxx/exception.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <mongocxx/exception/operation_exception.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/uri.hpp>
+
+using bsoncxx::builder::stream::close_document;
+using bsoncxx::builder::stream::document;
+using bsoncxx::builder::stream::finalize;
+using bsoncxx::builder::stream::open_document;
+
+int main(int, char**) {
```
tessavitabile: Should this example also include how you would do error-handling in "real code"? That was something I was planning to add.

examples/mongocxx/exception.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // @begin: cpp-logic-error
+    // Using an uninitialized collection throws a mongocxx::logic_error.
+    // A mongocxx::logic_error is-a mongocxx::exception is-a std::system_error.
+    mongocxx::collection coll;
+    try {
+        coll.name();
+    } catch (const mongocxx::logic_error& e) {
+        std::cout << "Using an uninitialized collection throws:" << std::endl;
+
+        if (e.code().category() != mongocxx::error_category()) {
```
samantharitter: Add a note explaining what other error categories there might be here and why checking is necessary/helpful?

examples/mongocxx/exception.cpp
```diff
@@ -0,0 +1,119 @@
...
+    } catch (const mongocxx::operation_exception& e) {
+        std::cout << "Renaming a collection that does not exist throws:" << std::endl;
+
+        // We expect this error to arise from the server
+        if (e.code().category() != mongocxx::server_error_category()) {
+            return EXIT_FAILURE;
+        }
+
+        // We can compare the error_code to a known server side error code number
+        if (e.code().value() != 26) {
```
samantharitter: Perhaps include a link to [error_codes.err](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err) to explain where 26 came from, and to show users the other codes they might use

src/bsoncxx/builder/core.cpp
```diff
@@ -71,8 +70,7 @@ class core::impl {
 
     bsoncxx::document::value steal_document() {
         if (_root_is_array) {
-            throw bsoncxx::exception{
-                make_error_code(error_code::k_cannot_perform_document_operation_on_array)};
+            throw bsoncxx::exception{error_code::k_cannot_perform_document_operation_on_array};
```
samantharitter: this is so much nicer!

src/bsoncxx/exception/error_code.hpp
```diff
@@ -37,10 +38,34 @@ enum class error_code : std::int32_t {
...
+    return {static_cast<int>(error), error_category()};
+}
+
 BSONCXX_INLINE_NAMESPACE_END
 }  // namespace bsoncxx
 
 #include <bsoncxx/config/postlude.hpp>
+
+namespace std {
+// Specialize is_error_code_enum so we get simpler std::error_code construction
```
samantharitter: Can you explain how this works?

src/bsoncxx/json.hpp
```diff
@@ -63,9 +63,9 @@ BSONCXX_API std::string BSONCXX_CALL to_json(types::value value);
 /// @param 'json'
 ///  A string_view into a JSON document
 ///
-/// @returns An engaged optional containing a document::value if conversion worked.
+/// @returns A document::value if conversion worked, or throws.
```
samantharitter: @throws tag?

src/bsoncxx/builder/core.cpp
tessavitabile: +1

tessavitabile:
LGTM

src/mongocxx/exception/server_error_code.cpp
```diff
@@ -12,26 +12,34 @@
...
-/// @param error A mongocxx::error_code
-///
-/// @return A std::error_code
-///
-std::error_code make_error_code(mongocxx::error_code error) {
-    return std::error_code(static_cast<int>(error), mongocxx_error_category());
+namespace {
+
+//
+// An error_category for codes returned by the server.
```
samantharitter: by the server or the underlying implementation?  Or do we not want to expose that even in a comment?

src/mongocxx/exception/server_error_code.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <mongocxx/config/prelude.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+///
+/// @todo document this enum and the values
+///
+enum class server_error_code : std::int32_t {
+    // Intentionally empty at this time!
```
samantharitter: Should we make a placeholder JIRA ticket and link here?

samantharitter:
Some small things and questions, but overall LGTM

examples/mongocxx/exception.cpp
acmorrow: Maybe later. I think this is good enough as a starting point though.

examples/mongocxx/exception.cpp
acmorrow: Done

examples/mongocxx/exception.cpp
acmorrow: Done

src/bsoncxx/builder/core.cpp
acmorrow: Yeah I was really happy when I figured out how to make it work.

src/bsoncxx/exception/error_code.hpp
acmorrow: Having this specialization enables the third constructor for std::error_code http://en.cppreference.com/w/cpp/error/error_code/error_code. This is the mechanism that makes it so that we don't need to call make_error_code explicitly.

src/bsoncxx/json.hpp
acmorrow: Done

src/mongocxx/exception/server_error_code.cpp
acmorrow: Ultimately, this error space will only reflect errors returned by the server. When we are able to distinguish them, I plan to roll the mongoc local errors into the mongocx::error_code enumeration.

src/mongocxx/exception/server_error_code.hpp
acmorrow: Done, in https://jira.mongodb.org/browse/CXX-834

acmorrow:
Updated, PTAL

samantharitter:
LGTM

PR #455, opened 2016-01-28
CXX-835 Support lint when run from non-repository root

markbenvenuto:
Since Travis runs on Ubuntu 12.04, we cannot use `-C` with `git` so we must continue to work around it.

Mac OS X 10.6 `tar` does not support `bzip2` compressed files.


markbenvenuto:
Ping @samantharitter @hanumantmk

samantharitter:
LGTM though I know nothing about the actual clang-format script changes

hanumantmk:
lgtm

PR #459, opened 2016-02-15
Use snprintf instead of std::to_string while compiling on CYGWIN

w1ndy:
std::to_string is not supported by the latest GCC on CYGWIN right now, as mentioned in a stackflow answer http://stackoverflow.com/a/33930814 . This patch uses __CYGWIN__ macro for determining CYGWIN platform and replaces std::to_string by snprintf to pass compilation.

### DIFF
--- a/src/bsoncxx/test/bson_util_itoa.cpp
+++ b/src/bsoncxx/test/bson_util_itoa.cpp
@@ -16,10 +16,23 @@
 
 #include <bsoncxx/private/itoa.hpp>
 
+#if defined(__CYGWIN__)
+#include <cstdio>
+#endif // __CYGWIN__
+
 TEST_CASE("util::itoa is equivalent to to_string(int)", "[bsoncxx::util::itoa]") {
+#if defined(__CYGWIN__)
+    char _cygwin_to_string_buf[128];
+#endif //__CYGWIN__
+    
     for (int i = 0; i <= 10000; i++) {
         bsoncxx::itoa val(i);
+#if defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


src/bsoncxx/test/bson_util_itoa.cpp
```diff
@@ -16,10 +16,23 @@
 
 #include <bsoncxx/private/itoa.hpp>
 
+#if defined(__CYGWIN__)
```
acmorrow: Rather than going to all of this effort, how about just disabling the entire test if __CYGWIN__ is defined? I'm confident that if bson_itoa functions on all other platforms, that it will work just fine on cygwin. Additionally, Please add a comment with a link to the upstream bug so that someday in the future someone can notice that it is no longer required.

acmorrow:
@w1ndy - Could you please also sign the MongoDB contributors agreement (https://www.mongodb.com/legal/contributor-agreement). Thanks!

w1ndy:
@acmorrow 
The issue is mentioned in [Cygwin mailing list](https://sourceware.org/ml/cygwin/2015-10/msg00446.html). Cygwin does not have an active bug tracker, but I submitted [the issue](https://sourceware.org/bugzilla/show_bug.cgi?id=19686) to sourceware's bug tracker anyway hoping someone might notice someday.

The patch itself is trying to ensure the same behavior is preserved across platform. But if you see it's appropriate you can just comment out the test by using CYGWIN macro. I've signed the contributors agreement. Thanks for reviewing the patch.

acmorrow:
Merged in bfe2f7a370cd9efc555de1ae383df3a46814f207

PR #460, opened 2016-02-16
CXX-843 add the $sample operator to the aggregation pipeline.

duckie:
See JIRA issue : https://jira.mongodb.org/browse/CXX-843.

There is no tests attached because a complete test series for aggregation is needed and should be dealt with in another ticket.

### DIFF
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -125,7 +125,7 @@ class MONGOCXX_API pipeline {
     ///
     /// @see http://docs.mongodb.org/manual/reference/operator/aggregation/sample/#pipe._S_sample
     ///
-    /// @param set the number of input documents to select.
+    /// @param size the number of input documents to select.
     ///
     pipeline& sample(std::int32_t size);
 
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -129,7 +129,6 @@ class MONGOCXX_API pipeline {
     ///
     pipeline& sample(std::int32_t size);
 
-
     ///
     /// Skips over the specified number of documents that pass into the stage and passes the
     /// remaining documents to the next stage in the pipeline.
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -120,6 +120,17 @@ class MONGOCXX_API pipeline {
     pipeline& redact(bsoncxx::document::view_or_value restrictions);
 
     ///
+    /// Randomly selects the specified number of documents that pass into the stage and passes the
+    /// remaining documents to the next stage in the pipeline.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/operator/aggregation/sample/#pipe._S_sample
+    ///
+    /// @param set the number of input documents to select.
+    ///
+    pipeline& sample(std::int32_t size);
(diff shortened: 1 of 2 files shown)


duckie:
It seems the travis build is broken : the logs show no error at all. Other pull requests suffer from the same issue.

src/mongocxx/pipeline.hpp
```diff
@@ -120,6 +120,16 @@ class MONGOCXX_API pipeline {
     pipeline& redact(bsoncxx::document::view_or_value restrictions);
 
     ///
+    /// Randomly selects the specified number of documents that pass into the stage and passes the
+    /// remaining documents to the next stage in the pipeline.
+    ///
+    /// @see http://docs.mongodb.org/manual/reference/operator/aggregation/sample/#pipe._S_sample
+    ///
+    /// @param set the number of input documents to select.
```
acmorrow: This param doc comment should name the parameter

src/mongocxx/pipeline.hpp
```diff
@@ -120,6 +120,16 @@ class MONGOCXX_API pipeline {
...
+    ///
+    pipeline& sample(std::int32_t size);
```
acmorrow: Why int32_t here, rather than, say, size_t? Does the $sample operator specify a width for this parameter?

acmorrow: Please note on the above that I'm really asking - I don't happen to know.

acmorrow:
@duckie - The code looks pretty fine, with one commenting nit and a question. Also, you can drop the : in the comment message after the ticket ID.

acmorrow:
@Also, @duckie, could you please sign the MongoDB contributors agreement (https://www.mongodb.com/legal/contributor-agreement) if you have not already done so? Thanks.

src/mongocxx/pipeline.hpp
duckie: The main reason is to stay consistent with the $skip operator. Plus, there is no point in debating about an unsigned integer since it will be encoded in BSON which do not support unsigned integers. Maybe we could change those int32_t for int64_t, but that should be done in one shot for all operators using them, in my opinion.

src/mongocxx/pipeline.hpp
duckie: Right, will fix it.

duckie:
@acmorrow I though I have already signed it, but whatever, I did it again, it should be ok now.

I dont understand what you mean by "dropping the ':' in the comment message". Do you mean in the title of the pull request ?

src/mongocxx/pipeline.hpp
acmorrow: OK, that makes sense

duckie:
Travis build says

```
W: Failed to fetch http://repo.mongodb.com/apt/ubuntu/dists/precise/mongodb-enterprise/stable/multiverse/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
```

I think there is nothing I could do about it.

acmorrow:
@duckie Yeah, I don't think that is your fault in any way. I'll validate that it works locally. Thanks again for the PR.

acmorrow:
Merged in 1d4834ba1168c61d2c8cceec361fbad4c901de9f

PR #462, opened 2016-03-11
CXX-868 Fails to build on VS2015 with error C2719

quatmax:
Hi!

Without these changes I got C2719 compiling mongocxx on Windows 10 with Visual Studio 2015 Update 1.

(full message: error C2719: 'error': formal parameter with requested alignment of 8 won't be aligned)

Maybe this is interesting for you.

BR,
Max

### DIFF
--- a/src/mongocxx/exception/private/mongoc_error.hpp
+++ b/src/mongocxx/exception/private/mongoc_error.hpp
@@ -28,17 +28,17 @@ inline std::error_code make_error_code(int code, int) {
     return {code, server_error_category()};
 }
 
-inline std::error_code make_error_code(::bson_error_t error) {
+inline std::error_code make_error_code(const ::bson_error_t &error) {
     return make_error_code(error.code, error.domain);
 }
 
 template <typename exception_type>
-void throw_exception(::bson_error_t error) {
+void throw_exception(const ::bson_error_t &error) {
     throw exception_type{make_error_code(error), error.message};
 }
 
 template <typename exception_type>
-void throw_exception(bsoncxx::document::value raw_server_error, ::bson_error_t error) {
+void throw_exception(bsoncxx::document::value raw_server_error, const ::bson_error_t &error) {
     throw exception_type{make_error_code(error), std::move(raw_server_error), error.message};
 }
 


acmorrow:
@quatmax Thanks for the PR. This is really an issue with the C driver. You can disable the excess alignment by building libbson with a flag to disable its excess alignment stuff. That would be my recommended approach.

quatmax:
OK!
Thanks a lot for the quick feedback!

acmorrow:
Lets see if the C driver team has any thoughts about how to handle it best. Perhaps there is a fix that the C driver can make that doesn't break ABI, but also avoids the C2719 error? I suspect you will not be the only person to get tripped up by this when building the new C++11 driver on Windows, and I want it to be a good experience.

@ajdavis @bjori any thoughts?

bjori:
I suppose we could highlight it better that if you are compiling libbson for the first time you should compile it with `--disable-extra-align` ? It'll only catch 10% of users though, reading the nitty gritty INSTALL details isn't the normal path :)

I recommend that the cxx driver check for BSON_EXTRA_ALIGN during its configure stage, and error out there and then. Should be easily added after the libbson dependency check?
That would avoid the user hitting a compiler error without any reasonable messaging on how to fix it.

acmorrow:
@bjori Maybe. I think what MSVC is complaining about is that the declaration of the formal parameter expands to include the __declspec align stuff. I think what it wants is that it only applies to the declaration of the type, not each mention of it. Perhaps that could just be fixed in the C driver. I know you want to remove the extra alignment, but you can't do that without an ABI break, so it isn't happening now. But maybe this issue can be solved by just tweaking the way the C driver macros work?

acmorrow:
@bjori - OK, now I'm even more confused. Read https://msdn.microsoft.com/en-us/library/373ak2y1.aspx, and then tell me how we are getting a __declspec(align(...)) in a formal parameter. It doesn't look like libbson is set up to have the declspec end up in the preprocessed source when a bson_error_t is used.

Maybe this is an MSVC bug?

bjori:
That doesn't make any sense.
The types aren't macroed at all.

Weather or not you use --disable-extra-align, we'll still do `__declspec(align(..)) typedef....;` when we declare the types btw.
Are we sure this was the first and only error? not some quirky cascading failure?

hanumantmk:
I wonder if this is one of those terrible moments when something works in C mode for MSVC, but it's broken in C++ mode

acmorrow:
@bjori - No, it is real. This is why our CI build of the C driver on windows uses the --disable-extra-align flag

acmorrow:
And there is this to ponder: https://llvm.org/bugs/show_bug.cgi?id=18140

acmorrow:
@quatmax - Would you mind filing a ticket in the C++ driver about this? I think there is more here to understand: https://jira.mongodb.org/browse/CXX

I'm going to re-open the PR too, since it might make sense to merge this. I don't want to require people to build the C driver in a strange way, especially if we determine that the error is legit.

acmorrow:
@quatmax - I've opened https://jira.mongodb.org/browse/CXX-868. Could you please amend your commit to reference this ticket (see the commit log for the master branch for examples). Also, could you please sign the MongoDB contributors agreement if you have not already done so:

https://www.mongodb.com/legal/contributor-agreement

Thanks

acmorrow:
@quatmax we are hoping to release C++11 driver 3.0.1 this week, and I'd like to merge this before then. Could you please amend the commit and sign the contributor agreement?

Thanks!

quatmax:
Excuse me, have been away a while.

I will do as you wanted me!

BR

quatmax:
I signed the contributor agreement and amend the last commit.

I hope this is OK.

acmorrow:
Merged in 090f51ed3edc0d866e0f750029dc5d4e31da3111. Thanks for the fix!

PR #464, opened 2016-03-16
CXX-872 - stream customization

hanumantmk:
Adds an example for adapting iterator pairs to the stream api

### DIFF
--- a/examples/bsoncxx/CMakeLists.txt
+++ b/examples/bsoncxx/CMakeLists.txt
@@ -24,6 +24,7 @@ set(BSONCXX_EXAMPLES
     builder_core.cpp
     builder_basic.cpp
     builder_stream.cpp
+    builder_stream_fancy.cpp
     getting_values.cpp
     view_and_value.cpp
 )
--- a/examples/bsoncxx/CMakeLists.txt
+++ b/examples/bsoncxx/CMakeLists.txt
@@ -24,7 +24,7 @@ set(BSONCXX_EXAMPLES
     builder_core.cpp
     builder_basic.cpp
     builder_stream.cpp
-    builder_stream_fancy.cpp
+    builder_stream_customization.cpp
     getting_values.cpp
     view_and_value.cpp
 )
--- a/examples/bsoncxx/builder_stream_fancy.cpp
+++ b/examples/bsoncxx/builder_stream_fancy.cpp
@@ -1,0 +1,106 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <map>
+#include <vector>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/builder/stream/array.hpp>
+#include <bsoncxx/json.hpp>
+
+using namespace bsoncxx;
+
+// concatenates arbitrary ranges into an array context
+template <typename begin_t, typename end_t>
+class range_array_appender {
+   public:
+    range_array_appender(begin_t begin, end_t end) : _begin(begin), _end(end) {
+    }
+
+    void operator()(bsoncxx::builder::stream::array_context<> ac) const {
+        for (auto iter = _begin; iter != _end; ++iter) {
+            ac << *iter;
+        }
+    }
+
+   private:
+    begin_t _begin;
+    end_t _end;
+};
+
+template <typename begin_t, typename end_t>
+range_array_appender<begin_t, end_t> make_range_array_appender(begin_t&& begin, end_t&& end) {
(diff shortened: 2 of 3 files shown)


acmorrow:
@hanumantmk Can we make a ticket for this? I'd like to track work that derives from the same origin as this.

examples/bsoncxx/builder_stream_fancy.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+// concatenates arbitrary ranges into an array context
+template <typename begin_t, typename end_t>
+class range_array_appender {
+   public:
+    range_array_appender(begin_t begin, end_t end) : _begin(begin), _end(end) {
+    }
+
+    void operator()(bsoncxx::builder::stream::array_context<> ac) const {
+        for (auto iter = _begin; iter != _end; ++iter) {
```
acmorrow: Should begin_t and end_t be different types?

examples/bsoncxx/builder_stream_fancy.cpp
```diff
@@ -0,0 +1,106 @@
...
+            ac << *iter;
+        }
+    }
+
+   private:
+    begin_t _begin;
+    end_t _end;
+};
+
+template <typename begin_t, typename end_t>
```
acmorrow: Ditto: should they be different types?

examples/bsoncxx/builder_stream_fancy.cpp
```diff
@@ -0,0 +1,106 @@
...
+// concatenates arbitrary ranges into an key context
+template <typename begin_t, typename end_t>
+class range_kvp_appender {
+   public:
+    range_kvp_appender(begin_t begin, end_t end) : _begin(begin), _end(end) {
+    }
+
+    void operator()(bsoncxx::builder::stream::key_context<> ac) const {
+        for (auto iter = _begin; iter != _end; ++iter) {
+            ac << iter->first << iter->second;
```
acmorrow: So, the requirements on *begin_t and _end_t is that they offer members `first` and `second`, so this would only work with things that dereferce to `std::pair`. I guess that is fine. It wouldn't work with a `vector<bsoncxx::builder::basic::kvp>` or `vector<std::tuple<std::string, int>>`, but you could write a different callable for that, right?

examples/bsoncxx/builder_stream_fancy.cpp
```diff
@@ -0,0 +1,106 @@
...
+    using builder::stream::document;
+    using builder::stream::array;
+
+    // bsoncxx::builder::stream presents an iostream like interface for succinctly
+    // constructing complex BSON objects.  It also allows for interesting
+    // primitives by writing your own callables with interesting signatures.
+
+    // stream::document builds a BSON document
+    auto doc = document{};
+    // stream::array builds a BSON array
```
acmorrow: add a blank line here, please

examples/bsoncxx/builder_stream_fancy.cpp
```diff
@@ -0,0 +1,106 @@
...
+    //     "c" : 3
+    // }
+
+    // Adapt our values
+    arr << make_range_array_appender(some_numbers.begin(), some_numbers.end());
+    // Now arr = {
+    //     "0" : 1,
+    //     "1" : 2,
+    //     "2" : 3
+    // }
```
acmorrow: EXIT_SUCCESS

examples/bsoncxx/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(BSONCXX_EXAMPLES
     builder_core.cpp
     builder_basic.cpp
     builder_stream.cpp
+    builder_stream_fancy.cpp
```
acmorrow: Maybe 'customization` is a better suffix for this test? More discoverable, etc.

acmorrow:
@hanumantmk This is really neat. A few small questions and comments but I think you can just merge it when you are ready.

acmorrow:
@hanumantmk But please do make a ticket for it,

examples/bsoncxx/builder_stream_fancy.cpp
hanumantmk: See the new C++17 for loop body:

```
{
    auto && __range = range_expression ; 
    auto __begin = begin_expr ;
    auto __end = end_expr ;
    for ( ; __begin != __end; ++__begin) { 
        range_declaration = *__begin; 
        loop_statement 
    } 
} 
```

I think the point is that end sentinels shouldn't have to be as fat as active iterators just to make the type system happy

examples/bsoncxx/builder_stream_fancy.cpp
hanumantmk: See above

examples/bsoncxx/builder_stream_fancy.cpp
hanumantmk: Absolutely.  If I was dropping it into the main interface, I think I'd add a dispatch type parameter which worked automatically with pairs and size 2 tuples, but that you could modify yourself.

But for an example I think this is sufficient

examples/bsoncxx/builder_stream_fancy.cpp
hanumantmk: done

examples/bsoncxx/builder_stream_fancy.cpp
hanumantmk: done

examples/bsoncxx/CMakeLists.txt
hanumantmk: done

acmorrow:
@hanumantmk Neat. LGTM. Please do make a ticket and squash/retitle.

PR #465, opened 2016-03-16
CXX-866 check for cursor_more before iterating

Machyne:


Machyne:
never mind! mognoc_cursor_more doesn't give us more information.

PR #467, opened 2016-03-21
CXX-822 Swap reversed comments for bulk operations

acmorrow:


Machyne:
LGMT! Much less confusing this way

PR #468, opened 2016-03-21
CXX-865 Enable building with clang with libstdc++ from GCC 4.8

acmorrow:
This is hideous, but necessary when building with clang against libstdc++ from GCC 4.8.2. Fortunately it is:

- Only in a test
- Only needed in one place

So, I haven't bothered to try to macro it out.

### DIFF
--- a/src/mongocxx/test/pool.cpp
+++ b/src/mongocxx/test/pool.cpp
@@ -140,13 +140,13 @@ TEST_CASE(
 
     instance::current();
 
-// GCC before 4.9.0 doesn't place max_align_t in the std namespace.
-#if defined(__clang__) || !defined(__GNUC__) || (__GNUC__ > 4) || \
-    (__GNUC__ == 4 && __GNUC_MINOR__ >= 9)
-    std::max_align_t dummy_address;
-#else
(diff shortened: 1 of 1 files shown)


acmorrow:
Ping?

src/mongocxx/test/pool.cpp
```diff
@@ -141,13 +141,43 @@ TEST_CASE(
...
+    // Using libstdc++, but compiler is neither clang nor GCC; assume
+    // we have it in the right place
+    std::max_align_t dummy_address;
+#endif
+
+#else
+    // Not using libstdc++, assume it is in the right place.
+    std::max_align_t dummy_address;
+#endif
+
```
hanumantmk: How do you feel about:

auto dummy_address = []{ using namespace std; return max_align_t{}; }();

That'll work regardless of where max_align_t is in std or not.  and it skips the preprocessor

acmorrow: Totally works. Great suggestion.

acmorrow:
Updated, PTAL.

Machyne:
Much cleaner! LGTM! :+1:

PR #469, opened 2016-03-21
CXX-864 Enable SSL and SASL in builders

acmorrow:
Also, a drive by to add concurrency to our build.


acmorrow:
Ping?

Machyne:
The concurrency for the compiles seems unrelated, but I think it's good to have. Maybe separate ticket?
Other than that, LGMT! :+1:

PR #470, opened 2016-03-21
CXX-847 Document the model/write class

acmorrow:


src/mongocxx/model/write.hpp
```diff
@@ -32,32 +32,105 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
+    ///
     write(delete_many value);
+
+    ///
+    /// Constructs a write from a @replace_one.
+    ///
     write(replace_one value);
 
+    ///
+    /// Move constructs a write
```
Machyne: End with a `.`

src/mongocxx/model/write.hpp
```diff
@@ -32,32 +32,105 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
+    ///
     write(write&& rhs) noexcept;
+
+    ///
+    /// Move assigns a write
```
Machyne: End with a `.`

src/mongocxx/model/write.hpp
```diff
@@ -32,32 +32,105 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
+    ///
+    /// Returns the current type of this write. You must call this
+    /// method before calling any of the get methods below.
+    ///
     write_type type() const;
 
+    ///
+    /// Accesses the write as an @insert_one. It is illegal to call
+    /// this method if the type returned by the type method above does
+    /// not indicate that it is the proper type.
```
Machyne: The repeated:

> It is illegal to call this method if the type returned by the type method above does not indicate that it is the proper type.

Is a little hard to follow. Maybe:

> It is illegal to call this method if the return of @type (above) does not indicate that <insert_one e.g.> is the proper type of this write model.

src/mongocxx/model/write.hpp
acmorrow: Done

src/mongocxx/model/write.hpp
acmorrow: Done

src/mongocxx/model/write.hpp
acmorrow: Done, more or less. See what you think of the new language.

acmorrow:
Thanks for the review @Machyne. Updated, PTAL.

src/mongocxx/model/write.hpp
Machyne: I think this way is more clear. Nice!

Machyne:
LGTM! 🚢

PR #471, opened 2016-03-21
CXX-862 Don't throw on invalid element access by index or key

acmorrow:


acmorrow:
Please note that this is a subtle behavior change and we will need to relnote it.

src/bsoncxx/array/view.cpp
```diff
@@ -175,9 +175,9 @@ view::iterator view::find(std::uint32_t i) const {
 
     if (bson_iter_init_find(&iter, &b, key.c_str())) {
         return iterator(element(iter.raw, iter.len, iter.off));
-    } else {
-        return end();
     }
+
+    return end();
```
Machyne: Unrelated? Same in the document view iterator.

Machyne:
Small nit, but LGTM!

src/bsoncxx/array/view.cpp
acmorrow: Just clearer than else { return ... };

Machyne: Agreed. :+1:

PR #475, opened 2016-03-24
CXX-877 add example for nested bson document

Machyne:


examples/bsoncxx/builder_stream.cpp
```diff
@@ -58,4 +58,17 @@ int main(int, char**) {
     using builder::stream::finalize;
     auto myQuery = document{} << "foo"
                               << "bar" << finalize;
+
+    // There is a special concatenate_doc helper to add all keys and corresponding values from one
+    // document into another.
+    using bsoncxx::builder::concatenate_doc;
+    doc << concatenate_doc{myQuery.view()};  // `doc` now has "foo": "bar".
```
acmorrow: Maybe the comment would be better if it showed the actual document. "now has" could mean a few different things.

examples/bsoncxx/builder_stream.cpp
```diff
@@ -58,4 +58,17 @@ int main(int, char**) {
...
+
+    // To nested an existing bsoncxx::document::value into a builder stream, you can create a
```
acmorrow: nest

examples/bsoncxx/builder_stream.cpp
```diff
@@ -58,4 +58,17 @@ int main(int, char**) {
...
+    // types::b_document and append it. Alternatively you can open a new document and concatenate
+    // the value in.
+    bsoncxx::document::value nestedValue = document{} << "nested" << true << finalize;
+    document topLevelDoc{};
+    topLevelDoc << "subDoc1" << types::b_document{nestedValue.view()} << "subDoc2" << open_document
```
acmorrow: Show what the resulting document looks like here as well

acmorrow:
LGTM mod nits.

bazineta:
This example uses concatenate_doc, but there's also bsoncxx::builder::stream::concatenate; it'd be nice to have that also described, as it's not clear when or if one would be preferred over the other.

hanumantmk:
actually, I'm not sure we should ever be using concatenate_doc or concatenate_array directly.

The point of the free function concatenate overloads is that they'll do the right thing, and that you won't have to repeat yourself in mentioning array/doc.  I guess the one advantage is that concatenate_doc(array::view{}) will fail, where concatenate(array::view{}) would succeed (assuming the array view was elsewhere in the code and the type signature was less obvious).

I'm not sure how much you get out of that though

Machyne:
Thanks for the feedback! I fixed the typo, added comments to show what the actual documents look like, and switched to using the free function for `concatenate`. PTAL! :)

hanumantmk:
lgtm

PR #479, opened 2016-04-12
CXX-886 Implement lookup operator for aggregation

samantharitter:
@xdg PTAL?

### DIFF
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -80,6 +80,20 @@ class MONGOCXX_API pipeline {
     pipeline& limit(std::int32_t limit);
 
     ///
+    /// Performs a left outer join to an unsharded collection in the same database to filter in
+    /// documents from the "joined" collection for processing.
+    ///
+    /// @see https://docs.mongodb.org/manual/reference/operator/aggregation/lookup/
+    ///
+    /// @param lookup the lookup expression, as a document with the following fields:
+    ///   from: <collection to join>
+    ///   localField: <field from the input documents>
+    ///   foreignField: <field from the documents of the "from" collection>
+    ///   as: <output array field>
+    ///
+    pipeline& lookup(bsoncxx::document::view_or_value lookup);
+
+    ///
     /// Filters the documents. Only the documents that match the condition(s) specified by the
     /// @c filter will continue to the next pipeline stage.
     ///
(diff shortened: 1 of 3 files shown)


src/mongocxx/pipeline.hpp
```diff
@@ -80,6 +80,20 @@ class MONGOCXX_API pipeline {
     pipeline& limit(std::int32_t limit);
 
     ///
+    /// Performs a left outer join to an unsharded collection in the same database to filter in
+    /// documents from the “joined” collection for processing.
```
Machyne: I don't know if we care, but these quotes look non-ASCII to me. Does that matter?

src/mongocxx/test/collection.cpp
```diff
@@ -532,6 +532,69 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
...
     }
 
+    SECTION("aggregation $lookup operator", "[collection]") {
+        auto people_coll_name = "people_on_the_block";
+        auto people_coll = db.create_collection(people_coll_name);
+        auto houses_coll_name = "houses_on_the_block";
+        auto houses_coll = db.create_collection(houses_coll_name);
+
+        // populate one collection with names
+        document name1;
```
hanumantmk: You don't have to rewrite it, but more stylish:

auto name1 = document{} << "key" << "value" << finalize;

src/mongocxx/pipeline.hpp
hanumantmk: Yeah, those are unicode (bytes):

�       226
�       128
�       156
j       106
o       111
i       105
n       110
e       101
d       100
�       226
�       128
�       157

Let's make those ascii

hanumantmk:
LGTM % nit

acmorrow:
LGTM as well, same nit. Any idea how those unicode quotes got in there?

samantharitter:
Weird!  Possibly from copying and pasting from our online documentation about $lookup?  Fixed!

Machyne:
^ definitely a copy paste from the docs [here](https://docs.mongodb.org/manual/reference/operator/aggregation/lookup/#pipe._S_lookup)

From the page source:

```
<p>Performs a left outer join to an unsharded collection in the <em>same</em>
database to filter in documents from the &#8220;joined&#8221; collection for
```

Which (thanks to markdown support of html) ends up looking like:
> <p>Performs a left outer join to an unsharded collection in the <em>same</em>
> database to filter in documents from the &#8220;joined&#8221; collection for

Crazy!

PR #484, opened 2016-05-12
CXX-893 Removed BSONCXX_API macro from non exported classes to fix link errors

aurelienrb:
Linking to bsoncxx.dll (with VC++ 2015) fails because the `BSONCXX_API` macro instructs VC++ to `dllimport` the classes sub_array and sub_document while these classes are header only (and thus not exported by the dll).
Removing the macro fixes the issue.

### DIFF
--- a/src/bsoncxx/builder/basic/sub_array.hpp
+++ b/src/bsoncxx/builder/basic/sub_array.hpp
@@ -36,7 +36,7 @@ void value_append(core* core, T&& t);
 /// An internal class of builder::basic.
 /// Users should almost always construct a builder::basic::array instead.
 ///
-class BSONCXX_API sub_array {
+class sub_array {
    public:
     ///
     /// Default constructor
--- a/src/bsoncxx/builder/basic/sub_document.hpp
+++ b/src/bsoncxx/builder/basic/sub_document.hpp
@@ -35,7 +35,7 @@ void value_append(core* core, T&& t);
 /// An internal class of builder::basic.
 /// Users should almost always construct a builder::basic::document instead.
 ///
-class BSONCXX_API sub_document {
+class sub_document {
    public:
     BSONCXX_INLINE sub_document(core* core) : _core(core) {
     }


acmorrow:
@aurelienrb - Thanks for the PR. I suspect this is the same issue reported here: https://jira.mongodb.org/browse/CXX-893

It does only seem to affect Debug builds, which is interesting. I'd also argue that this is a flaw in the Windows compiler, but it wouldn't be the first time.

- Could you please sign the MongoDB contributor agreement: https://www.mongodb.com/legal/contributor-agreement
- Could you please amend the commit to mention CXX-893 at the beginning of the commit line (see the recent commit log for examples)?

Thanks for helping to make the new C++11 driver better...

aurelienrb:
I just signed the agreement.
To me, the behavior of the compiler is correct: the class is tagged as being imported from a dll, so the compiler tried to get it from there. Indeed, it should be the cause of the bug CXX-893.

Regards,

Aurelien

acmorrow:
@aurelienrb - Yes, I see your point, and in practice it doesn't matter, since we aren't going to get Microsoft to change it. However, my view is that the constructor has been explicitly marked __forceinline. That should be enough to make the compiler not go to the DLL, even if the 'class' has been declared as exported. Exporting a concrete class shouldn't have any effects independently of exporting members, and there aren't any to export here. That said, I'll get this merged.

aurelienrb:
Oh ok I get it now ! I missed that point, thanks for sharing more details. The [doc about __forceinline](https://msdn.microsoft.com/fr-fr/library/bw1hbe6y.aspx) warns about the fact it's ignored in Debug:

> Even with __forceinline [...] the compiler cannot inline a function if the function or its caller is compiled with /Ob0 (the default option for debug builds).

or, as Yoda says:

> /Ob1 you must follow, young padawan!

:-)

acmorrow:
No need to close it, I was going to merge it! While we can rely on building the library with /Ob1, that isn't sufficient - clients shouldn't be required to provide it (they will forget), and then we are back in the same place. I think the sad reality is that MS's strange interpratation of __forceinline and the DLL attributes leaves little choice but to strip it here.

acmorrow:
@aurelienrb - Merged, thanks for the fix!

PR #485, opened 2016-05-12
CXX-896 Use chrono for TTL expiration options

acmorrow:


acmorrow:
@xdg @hanumantmk @samantharitter Please take a look.

src/mongocxx/collection.cpp
```diff
@@ -650,8 +650,11 @@ bsoncxx::document::value collection::create_index(view_or_value keys,
         }
     }
 
-    if (options.expire_after_seconds()) {
-        opt.expire_after_seconds = *options.expire_after_seconds();
+    if (options.expire_after()) {
+        const auto count = options.expire_after()->count();
+        if ((count < 0) || (count > std::numeric_limits<int32_t>::max()))
+            throw logic_error{error_code::k_invalid_parameter};;
```
xdg: Double-semicolons.

xdg:
And here's a test case: https://gist.github.com/xdg/f0ff32984218b0297df2dab989ec080b

samantharitter:
LGTM mod David's semicolon note and additional test case

Machyne:
Can we add an example or test showing how to set a ttl of 2 hours (e,g,) using `std::chrono::hours`?

xdg:
@Machyne, you mean like `std::chrono::duration_cast<std::chrono::seconds>(std::chrono::hours(2))`?

I don't object, but I think it's clutter we don't really need.  Shouldn't we assume that users know how to convert durations (or can do the long-hand math from hours to seconds) on their own?

acmorrow:
@Machyne @xdg They don't even need to do the conversion - it is baked in to std::chrono.

Machyne:
@xdg @acmorrow - my thought was it would be nice to have an explicit example of something like:
```
options::index options{};
options.expire_after(std::chrono::hours{2});
coll.create_index(index_doc.view(), options);
```

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -272,7 +272,7 @@ TEST_CASE("Collection", "[collection]") {
         bool collection_create_index_called = false;
         bool success;
         bool expected_unique = true;
-        std::int32_t expected_expire_after_seconds = 500;
+        auto expected_expire_after = std::chrono::seconds(500);
```
Machyne: Why not `std::chrono::seconds{500};`?

Machyne: (with `{}` instead of `()`)

acmorrow: Use of {} vs () is more a matter of style in cases like this. I don't feel a need to change it.

src/mongocxx/collection.cpp
acmorrow: Thanks, fixed

acmorrow:
Updated. I merged in the test case from @xdg (thanks!), removed the double semi, and fixed up a minor formatting nit in the new test cases that clang-format pointed out (thanks clang-format!). PTAL.

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -38,7 +38,7 @@
 #include <mongocxx/read_preference.hpp>
 #include <mongocxx/stdx.hpp>
 #include <mongocxx/options/update.hpp>
-#include <mongocxx/exception/operation_exception.hpp>
+#include <mongocxx/exception/logic_error.hpp>
```
xdg: For clarity: this is really two changes.  operation_exception was already imported above and thus removed here, and I added logic_error.

xdg:
lgtm

src/mongocxx/test/collection_mocked.cpp
acmorrow: Oh, actually, thanks for pointing that out. This should be sorted in with the other includes. I will fix that before pushing.

PR #486, opened 2016-05-16
CXX-911 Add clang-format version check

xdg:
This commit adds the same version check as in the server build script.

Note: this only suppresses a warning if the version is correct; there is
still no assertion if the version is wrong.

### DIFF
--- a/etc/clang_format.py
+++ b/etc/clang_format.py
@@ -429,6 +429,9 @@ class ClangFormat(object):
         except CalledProcessError:
             cf_version = "clang-format call failed."
 
+        if CLANG_FORMAT_VERSION in cf_version:
+            return True
+
         if warn:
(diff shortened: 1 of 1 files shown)


xdg:
@acmorrow @markbenvenuto Please take a look

acmorrow:
LGTM

acmorrow:
@xdg Feel free to get this committed, with a reminder that we use a rebase workflow to keep a linear history, so please don't just hit the merge button.

acmorrow:
@markbenvenuto Is this also an issue in the server repo copy of this? Or was it fixed there but never backported?

xdg:
@acmorrow The merge button is evil.  I wish GH would let repo admins turn it off.

xdg:
Rebased to master as 86738ee

PR #487, opened 2016-05-19
CXX-880 Don't mask lifecycle methods in b_ types

hanumantmk:
Some b_ types had overly binding universal forwarding ctors.  We need to sfinae those out if the
passed arg is of the type, so that we don't mask lifecycle methods.


acmorrow:
LGTM

PR #488, opened 2016-05-21
CXX-850 Backport server r3.0.8..r3.0.12 changes

acmorrow:


acmorrow:
ping?

src/mongo/base/error_codes.err
```diff
@@ -117,6 +117,7 @@ error_code("InitialSyncOplogSourceMissing", 114)
 error_code("CommandNotSupported", 115)
 error_code("DocTooLargeForCapped", 116)
 error_code("ConflictingOperationInProgress", 117)
+error_code("FailedToSatisfyReadPreference", 133)
```
samantharitter: why here and not under the "backported" section?

src/mongo/client/replica_set_monitor.h
```diff
@@ -124,6 +124,11 @@ class ReplicaSetMonitor {
     void appendInfo(BSONObjBuilder& b) const;
 
     /**
+     * Returns true if the monitor knows a usable primary from it's interal view.
```
samantharitter: it's -> its

src/mongo/util/assert_util.h
```diff
@@ -353,3 +353,17 @@ std::string demangleName(const std::type_info& typeinfo);
...
+ * Example:
+ *     // code above checks that expr can only be FOO or BAR
+ *     switch (expr) {
+ *     case FOO: { ... }
+ *     case BAR: { ... }
+ *     default:
+ *         MONGO_UNREACHABLE;
+ */
+
+#define MONGO_UNREACHABLE ::mongo::invariantFailed("Hit a MONGO_UNREACHABLE!", __FILE__, __LINE__);
```
samantharitter: Do we need to backport this if we aren't using it anywhere?  Do we plan to use it anywhere?

samantharitter:
A couple nits and questions, otherwise LGTM

src/mongo/base/error_codes.err
acmorrow: Because it came in here through a cherry-pick. I don't actually know what the 'backported' section is about, really.

src/mongo/client/replica_set_monitor.h
acmorrow: @kaloianm want to fix these in the server and I'll pick them up on next backport.
@samantharitter good catch

src/mongo/util/assert_util.h
acmorrow: I thought about it, and I think yes, because now that this is in the v3.0 branch, future backports from the server may include code that uses it.

acmorrow:
Merged in dc6e44bbc3209b5554571bc6fc9647a55bc10300

PR #490, opened 2016-06-04
CXX-917 Make SSL actually work

acmorrow:
@Machyne @xdg - please review.
@ajdavis - apparently, the C driver doesn't make owned copies of the strings passed in via http://api.mongodb.com/c/current/mongoc_client_set_ssl_opts.html. That is apparently documented, but pretty user hostile and definitely caused me some grief and time in the debugger. Any chance of changing that?

### DIFF
--- a/examples/mongocxx/connect.cpp
+++ b/examples/mongocxx/connect.cpp
@@ -1,0 +1,84 @@
+// Copyright 2015 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cstdlib>
+#include <iostream>
+#include <string>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <bsoncxx/stdx/make_unique.hpp>
+
+#include <mongocxx/client.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/logger.hpp>
+#include <mongocxx/options/client.hpp>
+#include <mongocxx/uri.hpp>
+
+namespace {
+
+class logger final : public mongocxx::logger {
+   public:
+    explicit logger(std::ostream* stream) : _stream(stream) {
+    }
+
+    void operator()(mongocxx::log_level level, mongocxx::stdx::string_view domain,
+                    mongocxx::stdx::string_view message) noexcept override {
+        if (level >= mongocxx::log_level::k_trace) return;
+        *_stream << '[' << mongocxx::to_string(level) << '@' << domain << "] " << message << '\n';
+    }
+
+   private:
+    std::ostream* const _stream;
+};
+
+}  // namespace
+
+int main(int argc, char* argv[]) {
+    using bsoncxx::builder::stream::document;
+
+    mongocxx::instance inst{bsoncxx::stdx::make_unique<logger>(&std::cout)};
+
+    if (argc != 2) {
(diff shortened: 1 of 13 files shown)


ajdavis:
Seems like this could be fixed without risking backwards compatibility:

https://jira.mongodb.org/browse/CDRIVER-1316

src/mongocxx/pool.cpp
```diff
@@ -43,12 +43,12 @@ pool::pool(const uri& mongodb_uri, stdx::optional<options::ssl> ssl_options)
     : _impl{stdx::make_unique<impl>(libmongoc::client_pool_new(mongodb_uri._impl->uri_t),
                                     std::move(ssl_options))} {
     if (_impl->ssl_options) {
-#if defined(MONGOC_HAVE_SSL)
+#if defined(MONGOC_ENABLE_SSL)
         auto mongoc_opts = options::make_ssl_opts(*_impl->ssl_options);
         // We store ssl options in a member variable because we need the strings to stay alive so
```
acmorrow: This also needs to be changed to stash the vector of SSL strings.

examples/mongocxx/connect.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    if (argc != 2) {
+        std::cout << "usage: " << argv[0] << " MONGODB_URI" << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    try {
+        const auto uri = mongocxx::uri{argv[1]};
+
+        auto client = mongocxx::client{mongocxx::uri{argv[1]}, [] {
```
Machyne: Why did you do this with immediately invoked lambdas?

src/mongocxx/instance.cpp
```diff
@@ -72,6 +72,7 @@ class instance::impl {
         libmongoc::init();
         if (_user_logger) {
             libmongoc::log_set_handler(user_log_handler, _user_logger.get());
+            mongoc_log(MONGOC_LOG_LEVEL_INFO, "mongocxx", "libmongoc logging callback enabled");
```
Machyne: Should this be `libmongoc::log()`?

src/mongocxx/private/client.hpp
```diff
@@ -33,6 +33,7 @@ class client::impl {
     }
 
     mongoc_client_t* client_t;
+    std::vector<bsoncxx::string::view_or_value> ssl_options;
```
Machyne: Would this be worth a `typedef` somewhere? It's used 3 times.

src/third_party/catch/include/helpers.hpp
```diff
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <mongocxx/private/libmongoc.hpp>
```
Machyne: Why do we need this here now?

Machyne:
I had a question and a few nits, otherwise LGTM

examples/mongocxx/connect.cpp
acmorrow: Because we don't have a fix for https://jira.mongodb.org/browse/CXX-881 yet, which would allow a fluent style usage here.

src/mongocxx/instance.cpp
acmorrow: It can't be, due to mongoc_log being varargs

src/third_party/catch/include/helpers.hpp
acmorrow: Because otherwise we don't have MONGOC_ENABLE_SSL defined correctly below.

src/mongocxx/private/client.hpp
acmorrow: @Machyne. Its just a vector :) I think it is OK to say it more than once. In fact, it can be annoying to need to chase down typedefs for every vector.

acmorrow:
@Machyne @xdg I've replied to the comments (I don't think any changes are needed). I'd like to get this pushed sooner rather than later as it is blocking validating a DRIVERS ticket.

src/mongocxx/private/client.hpp
Machyne: Fair enough :)

Machyne:
LGTM!
If CXX-881 goes in first, I'd be in favor of removing the lambdas.

acmorrow:
@Machyne I updated the connect example per CXX-881

PR #491, opened 2016-06-06
CXX-887 Add inserted_id example

xdg:
This expands on the example in the ticket to show result/type checking.

### DIFF
--- a/examples/mongocxx/inserted_id.cpp
+++ b/examples/mongocxx/inserted_id.cpp
@@ -1,0 +1,56 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+#include <bsoncxx/types.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/exception/exception.hpp>
+#include <mongocxx/instance.hpp>
+
+int main(int, char**) {
+    mongocxx::instance inst{};
+    mongocxx::client conn{mongocxx::uri{}};
+
+    bsoncxx::builder::stream::document document{};
+
+    auto collection = conn["test"]["col"];
+    document << "test" << 1;
+
+    try {
+        auto result = collection.insert_one(document.view());
+
+        // Currently, result will always be true (or an exception will be
+        // thrown).  Eventually, unacknowledged writes will give a false
+        // result. See https://jira.mongodb.org/browse/CXX-894
+
+        if (!result) {
+            std::cout << "Unacknowledged write. No id available." << std::endl;
+            return EXIT_SUCCESS;
+        }
+
+        if (result->inserted_id().type() == bsoncxx::type::k_oid) {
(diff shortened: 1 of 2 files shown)


xdg:
@acmorrow @Machyne PTAL.  I'd like to have an `else` clause for `result` being false, but I'm not sure how that could happen, so I left it out.

Machyne:
LGTM

examples/mongocxx/inserted_id.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    if (result) {
+        if (result->inserted_id().type() == bsoncxx::type::k_oid) {
+            bsoncxx::oid id = result->inserted_id().get_oid().value;
+            std::string id_str = id.to_string();
+            std::cout << "Inserted id: " << id_str << std::endl;
+        } else {
+            std::cout << "Inserted id was not an OID type" << std::endl;
+        }
+    }
```
acmorrow: This example should handle exceptions. In the event of an exception, it should print it out, and return EXIT_FAILURE. Otherwise, it should return EXIT_SUCCESS

examples/mongocxx/inserted_id.cpp
```diff
@@ -0,0 +1,41 @@
...
+    mongocxx::client conn{mongocxx::uri{}};
+
+    bsoncxx::builder::stream::document document{};
+
+    auto collection = conn["test"]["col"];
+    document << "test" << 1;
+
+    auto result = collection.insert_one(document.view());
+
+    if (result) {
```
acmorrow: Personally, I'm an anti-deep-nesting sort of person, so I'd reverse the sense here:

if (!result) {
    std::cout << ...
    return EXIT_FAILURE; // presumably this indicates something very wrong
};

// No need for an else clause, since otherwise we have returned.

xdg:
@acmorrow PTAL at 0b1a08d

examples/mongocxx/inserted_id.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <mongocxx/exception/exception.hpp>
+#include <mongocxx/instance.hpp>
+
+int main(int, char**) {
+    mongocxx::instance inst{};
+    mongocxx::client conn{mongocxx::uri{}};
+
+    bsoncxx::builder::stream::document document{};
+
+    auto collection = conn["test"]["col"];
```
acmorrow: I'd argue that you should pull the construction of mongocxx::client and obtaining the client inside the try block - they can fail!

Building the BSON outside the try seems fine if you want.

examples/mongocxx/inserted_id.cpp
```diff
@@ -0,0 +1,53 @@
...
+            return EXIT_SUCCESS;
+        }
+
+        if (result->inserted_id().type() == bsoncxx::type::k_oid) {
+            bsoncxx::oid id = result->inserted_id().get_oid().value;
+            std::string id_str = id.to_string();
+            std::cout << "Inserted id: " << id_str << std::endl;
+        } else {
+            std::cout << "Inserted id was not an OID type" << std::endl;
+        }
```
acmorrow: Should still return EXIT_SUCCESS here

examples/mongocxx/inserted_id.cpp
```diff
@@ -0,0 +1,53 @@
...
+    } catch (const mongocxx::exception& e) {
+        std::cout << "An exception occurred: " << e.what() << std::endl;
```
acmorrow: Should return EXIT_FAILURE here

examples/mongocxx/inserted_id.cpp
xdg: If they are up to spec, they should not fail except in extraordinary circumstances (as they should not be doing any network activity).

acmorrow: Good point.

acmorrow:
LGTM

PR #492, opened 2016-06-07
CXX-878 Wrap std::chrono command options as b_int64

xdg:


xdg:
@acmorrow @Machyne PTAL

Machyne:
LGTM

src/mongocxx/options/modify_collection.cpp
```diff
@@ -35,7 +35,8 @@ void modify_collection::no_padding(bool no_padding) {
 void modify_collection::index(bsoncxx::document::view_or_value index_spec,
                               std::chrono::seconds seconds) {
     _index.emplace(document{} << "keyPattern" << bsoncxx::types::b_document{std::move(index_spec)}
-                              << "expireAfterSeconds" << seconds.count() << finalize);
+                              << "expireAfterSeconds" << bsoncxx::types::b_int64{seconds.count()}
```
acmorrow: Is the indentation here correct? What does clang-format want to do here?

xdg: clang-format wrapped it that way.  I had the `finalize` on the same line, originally.

src/mongocxx/collection.cpp
```diff
@@ -207,7 +207,7 @@ cursor collection::find(view_or_value filter, const options::find& options) {
     }
 
     if (options.max_time()) {
-        filter_builder << "$maxTimeMS" << options.max_time()->count();
+        filter_builder << "$maxTimeMS" << bsoncxx::types::b_int64{options.max_time()->count()};
```
acmorrow: Is the result of max_time()->count always assured to be able to fit within an int64?

src/mongocxx/options/modify_collection.cpp
acmorrow: ok, weird, but it is what it is.

src/mongocxx/collection.cpp
xdg: Technically, std::chrono::milliseconds could be anything larger than 45 bits.  In practice, even if it were > 64bits, I suspect no user is going to put a larger value there, so the question is what happens when the compiler is told to initialize `b_int64` from a > 64-bit int type?  Cast and warn?  Error?

acmorrow: Perhaps you could static_assert that the rep typedef from the duration type is <= 64 bits

acmorrow: @xdg - Per discussion, I think we agreed on a runtime check here?

xdg: I think it would make more sense to put the runtime check in the option class, so the exception triggers when set.

acmorrow: Sure, that sounds reasonable too.

xdg:
Per our verbal discussion, we agreed to defer runtime size checks to [CXX-849](https://jira.mongodb.org/browse/CXX-849)

PR #493, opened 2016-06-07
CXX-906 Fix null deref in getCollectionNames()/getCollectionInfos()

jrassi:
I manually tested this by issuing a kill -9 to mongod right before a call to getCollectionNames(), and ditto with getCollectionInfos().  I assume that this particular fix should be committed without tests, but let me know.

The error message is copy-pasted from elsewhere in this file.

@acmorrow, PTAL.

### DIFF
--- a/src/mongo/client/dbclient.cpp
+++ b/src/mongo/client/dbclient.cpp
@@ -1471,6 +1471,7 @@ list<string> DBClientWithCommands::getDatabaseNames() {
 
 list<string> DBClientWithCommands::getCollectionNames(const string& db, const BSONObj& filter) {
     auto_ptr<DBClientCursor> infos = enumerateCollections(db, filter);
+    uassert(0, "failed to read server response from socket when listing collections", infos.get());
     list<string> names;
 
     while (infos->more()) {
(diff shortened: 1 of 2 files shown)


src/mongo/client/dbclient.cpp
```diff
@@ -1471,6 +1471,7 @@ list<string> DBClientWithCommands::getDatabaseNames() {
 
 list<string> DBClientWithCommands::getCollectionNames(const string& db, const BSONObj& filter) {
     auto_ptr<DBClientCursor> infos = enumerateCollections(db, filter);
+    uassert(0, "socket error for mapping query", infos.get());
```
acmorrow: @jrassi Yeah but in those instances the things were actually queries! I mean, I guess these are too, but I'm sure you can craft a more appropriate error message for here and below.

acmorrow: I also suspect that DBClientWithCommands::exists is similarly afflicted. Can you just check all callers of enumerateCollections?

jrassi: Error message improved, all callers of enumerateCollections() fixed.  PTAL at 5bf73da3.

acmorrow:
LGTM. Please rebase, squash, and commit on top of current tip of legacy (no merges, etc.). Thanks!

PR #494, opened 2016-06-08
CXX-861 Make options setters return a self reference

acmorrow:


acmorrow:
@Machyne @xdg @jrassi Please take a look. Probably good to read JIRA CXX-861 for background.

acmorrow:
FYI @bazineta

Machyne:
Love it, much cleaner API. Do you think we should change any of the examples / tests to make use of this? If not, LGTM.

PR #495, opened 2016-06-14
CXX-784 Allow stream in of doc or array views

acmorrow:
@Machyne @xdg @hanumantmk This is a minor feature enhancement

### DIFF
--- a/examples/bsoncxx/builder_stream.cpp
+++ b/examples/bsoncxx/builder_stream.cpp
@@ -73,17 +73,22 @@ int main(int, char**) {
 
     // To nest an existing bsoncxx::document::value into a builder stream, you can create a
     // types::b_document and append it. Alternatively you can open a new document and concatenate
-    // the value in.
+    // the value in. Or, simplest, is just to stream in the view.
     bsoncxx::document::value nestedValue = document{} << "nested" << true << finalize;
     document topLevelDoc{};
     topLevelDoc << "subDoc1" << types::b_document{nestedValue.view()} << "subDoc2" << open_document
-                << concatenate(nestedValue.view()) << close_document;
+                << concatenate(nestedValue.view()) << close_document << "subDoc3"
+                << nestedValue << finalize;
+
     // `topLevelDoc` now looks like:
     // {
     //     "subDoc1": {
     //         "nested": true
     //     },
     //     "subDoc2": {
+    //         "nested": true
+    //     },
+    //     "subDoc3": {
     //         "nested": true
     //     }
     // }
--- a/src/bsoncxx/builder/core.cpp
+++ b/src/bsoncxx/builder/core.cpp
@@ -375,6 +375,14 @@ void core::append(const oid& value) {
     append(types::b_oid{value});
 }
 
+void core::append(const document::view view) {
+    append(types::b_document{view});
+}
+
+void core::append(const array::view view) {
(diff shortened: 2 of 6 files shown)


Machyne:
Can you update the builder stream example? I think this removes the need to show the
```c++
topLevelDoc << "subDoc1" << types::b_document{nestedValue.view()} << "subDoc2" << open_document
            << concatenate(nestedValue.view()) << close_document;
```
case.

acmorrow:
@Machyne @xdg @hanumantmk Updated, PTAL

Along with updating the example, I also removed a stray <iostream> include that snuck in during testing, and applied some clang-format goodness.

examples/bsoncxx/builder_stream.cpp
```diff
@@ -73,18 +73,24 @@ int main(int, char**) {
...
-    // the value in.
+    // the value in. Or, simplest, is just to stream in the view.
     bsoncxx::document::value nestedValue = document{} << "nested" << true << finalize;
     document topLevelDoc{};
-    topLevelDoc << "subDoc1" << types::b_document{nestedValue.view()} << "subDoc2" << open_document
-                << concatenate(nestedValue.view()) << close_document;
+    auto state1 = topLevelDoc << "subDoc1" << types::b_document{nestedValue.view()};
+    auto state2 = state1 << "subDoc2" << open_document << concatenate(nestedValue.view()) << close_document;
+    auto state3 = state2 << "subDoc3" << nestedValue.view();
+    auto final = state2 << finalize;
```
xdg: I don't think breaking this up into multiple states is the clearest way to convey this.   Having multiple states means I'm wondering why there are multiple states instead of focusing on three techniques.  While not "proper" formatting, I think something like this would be better:
```
topLevelDoc
    << "subDoc1" << types::b_document{nestedValue.view()}
    << "subDoc2" << open_document << concatenate(nestedValue.view())
    << "subDoc3" << nestedValue.view()
    << close_document;
```

This is sort of a "Tufte" take on the example (i.e. 'reduce non-data ink').

xdg:
What about doing this with view_or_value?  I.e. why only take views instead of rvalue references:

```
topLevelDoc << "subDoc1" << getDoc() << finalize;
```

examples/bsoncxx/builder_stream.cpp
acmorrow: Done, though not the formatting.

acmorrow:
There is no need here for view_or_value, since there is no possibility of ownership transfer. However, since document::value is implicitly convertible to document::view, the call to .view in the example is unnecessary. I'll remove it.

Machyne:
Could streaming in a temporary Value cause issues? Could the Value could get cleaned up after the operator View was called?

acmorrow:
@Machyne - I don't think so, because the temporary ```document::value``` would last until the end of the full expression, and value like puts into the stream state are immediately applied. @hanumantmk can you confirm?

hanumantmk:
@Machyne - Streaming in a temporary is fine, we consume the bytes immediately
@acmorrow - I think you do need the .view(), otherwise you're asking for multiple implicit conversions

acmorrow:
@hanumantmk the multiple conversions thing isn't a problem since we added overloads taking views. Thanks for the reviews. I'm going to commit this.

PR #496, opened 2016-06-14
CXX-928 Fix bad merge of SConstruct in dc6e44b

jrassi:
Initialization of `env["MONGO_SSL"]` on non-SSL builds was accidentally deleted when backporting `e5f5b4e` from the server repository, as part of legacy commit dc6e44b.  See the old initialization on the left-hand side of the diff at [SConstruct:935](https://github.com/mongodb/mongo-cxx-driver/commit/dc6e44bb#diff-cc8093822b396591a5530a463c87d048L935).

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1734,7 +1734,8 @@ def doConfigure(myenv):
 
     conf = Configure(myenv)
 
-    if has_option( "ssl" ):
+    conf.env["MONGO_SSL"] = bool(has_option("ssl"))
+    if conf.env["MONGO_SSL"]:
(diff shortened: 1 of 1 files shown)


acmorrow:
LGTM

acmorrow:
Please reply in CXX-928 and let the submitter know that we have addressed this issue. Also, there is another open issue caused by this same backport (re calling env.ConfError). Since you are now familiar with the logic here, can you just take that one down next?

jrassi:
Done, and submitted #497 for the env.ConfError issue (CXX-920).

PR #497, opened 2016-06-14
CXX-920 Add error reporting methods to env

jrassi:
Server SHAs cherry-picked (with modifications) into this commit:

1bb3be7

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -510,14 +510,26 @@ if windows:
 env = Environment(variables=env_vars, **envDict)
 del envDict
 
+def fatal_error(env, msg, *args):
+    print msg.format(*args)
+    Exit(1)
+
+def conf_error(env, msg, *args):
+    print msg.format(*args)
+    print "See {0} for details".format(env['CONFIGURELOG'].abspath)
+
+    Exit(1)
+
+env.AddMethod(fatal_error, 'FatalError')
+env.AddMethod(conf_error, 'ConfError')
+
 if has_option('variables-help'):
     print env_vars.GenerateHelpText(env)
     Exit(0)
 
 unknown_vars = env_vars.UnknownVariables()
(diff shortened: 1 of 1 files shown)


jrassi:
Only a few hunks applied automatically, so the majority of this merge was manual.

@acmorrow, PTAL.

SConstruct
```diff
@@ -599,11 +608,9 @@ else:
     env['OS_FAMILY'] = 'posix'
 
 if has_option( "cc-use-shell-environment" ) and has_option( "cc" ):
-    print("Cannot specify both --cc-use-shell-environment and --cc")
-    Exit(1)
+    env.FatalError("Cannot specify both --cc-use-shell-environment and --cc")
 elif has_option( "cxx-use-shell-environment" ) and has_option( "cxx" ):
-    print("Cannot specify both --cxx-use-shell-environment and --cxx")
-    Exit(1)
+    env.FatalError("Cannot specify both --cxx-use-shell-environment and --cxx")
```
acmorrow: Can't comment on the right line, but please see 628 and 633 below

acmorrow:
I think I see two places that didn't get updated.

Also, did you check the various SConscript files as well to see if any of them have any Exit/exit branches that could be patched up?

acmorrow:
LGTM mod checking other SConscripts

jrassi:
Those two calls fixed in bd73709.  The lowercasing of 'exit' there threw off my grep invocation.

Yep, checked other SConscript files. `find . -name 'SC*' | xargs grep -n -i 'exit('` returns only lines from SConstruct above line 530 (around where the declaration of `env` is).

acmorrow:
@jrassi - Thanks, definitely LGTM. Should definitely make future backports safer.

PR #498, opened 2016-06-14
Fix compilation on Debian with non Linux kernels

giuliopaci:


src/mongo/platform/random.cpp
```diff
@@ -103,7 +103,7 @@ SecureRandom* SecureRandom::create() {
     return new WinSecureRandom();
 }
 
-#elif defined(__linux__) || defined(__sunos__) || defined(__APPLE__) || defined(__freebsd__)
+#elif defined(__linux__) || defined(__sunos__) || defined(__APPLE__) || defined(__freebsd__) || defined(__FreeBSD__) || defined(__FreeBSD_kernel__) || defined(__gnu_hurd__)
```
acmorrow: Are all three of _freebsd__ __FreeBSD__ and __FreeBSD_kernel__ required? What does each of them represent?

giuliopaci:
`__freebsd__` was already there, so I supposed it worked for someone.
`__FreeBSD__` is what is reported to work on https://sourceforge.net/p/predef/wiki/OperatingSystems/, but was not working in my case.
`__FreeBSD_kernel__` is what is reported to work for kernel >= 8.3 and actually worked in my case.

So yes: I think they are all necessary.

jrassi:
Hmm.  `__freebsd__` is an internal define (no longer in use in the server codebase since https://jira.mongodb.org/browse/SERVER-17552) that I would expect to be set on your build at [SConstruct:709](https://github.com/mongodb/mongo-cxx-driver/blob/50ce454/SConstruct#L709).  However, cursory research suggests that `os.sys.platform` is "gnukfreebsd\<VERSION\>" on GNU/kFreeBSD, whereas it is "freebsd\<VERSION\>" on normal FreeBSD.

I wonder whether an approach like https://github.com/jrassi/mongo-cxx-driver/commit/b924b97375203969db9af3d2d90b09f2f6922f96 would work instead?  Though I wouldn't be surprised if the legacy C++ driver build process makes assumptions of a BSD userland if the `freebsd` environment flag is set.

Curious to hear your thoughts, @acmorrow.

jrassi:
Ah, furthermore, https://sources.debian.net/patches/glib2.0/2.33.12%2Breally2.32.4-5/92_kfreebsd_credentials.patch/ suggests that the reason that `__FreeBSD__` is not defined on your system is because "The `__FreeBSD__` macro means we have both a FreeBSD kernel and FreeBSD
libc, which isn't the case on GNU/kFreeBSD (GNU libc and userland on the FreeBSD kernel)".

I'm going to discuss in person tomorrow with @acmorrow.  I'll post here after we talk with a recommendation of a path forward.

giuliopaci:
Maybe it would work, but:
1. it will add libkvm dependency (that apparently is not used as the compilation and tests are passing without this library. Maybe it is needed when FreeBSD libc is used instead of the GNU one);
2. it adds `-fno-omit-frame-pointer` (why?).

jrassi:
I discussed with @acmorrow this afternoon.  Please submit a follow-up patch that:
1. Keeps `__FreeBSD__` and `__FreeBSD_kernel__` in the `#elif` list in line 106, but removes `__freebsd__`.  This is in line with changes made in the server at SERVER-17552.  Let's keep the `__freebsd__` define in the `SConstruct` file though (even though it will no longer be used), so that we avoid merge conflicts in this file when backporting future changes from the server repository.
2. Is formatted with our code formatting tool (run `site_scons/buildscripts/clang_format.py format` from the root of the source repository).

The FreeBSD platform-specific logic in the SConstruct file is many years old, and has since been rewritten in the server.  As far as `-fno-frame-pointer` and libkvm goes, I can point you to https://github.com/mongodb/mongo/commit/57f64fe508006d82b5af4c1460a938bfc4396331 and https://github.com/mongodb/mongo/commit/f710e51faa3085fe879e39717bc2f160bcb9ecd9 respectively, but can't provide much more context.

jrassi:
Hi @giuliopaci, just pinging you on this pull request to see if you have any questions about my feedback above.

giuliopaci:
Hi! Your feedback is perfectly clear. The only issue is to find a couple of minutes to update it. I will try to do that as soon as I get in front of a pc.

giuliopaci:
Updated.

acmorrow:
LGTM

PR #499, opened 2016-06-16
CXX-832 Friendlier error message when decoding Decimal128 BSONElements

jrassi:


jrassi:
@xdg: in reference to your comment on the ticket ("So we don't do it again for future types..."), I'm not thrilled by the idea of using this message for non-decimal unrecognized type ids.  See [this JIRA search](https://jira.mongodb.org/issues/?jql=text%20~%20%22%5C%22BSONElement%20bad%20type%5C%22%22) for a list of hundreds of tickets containing the string "BSONElement: bad type".  Turning all future instances of this error into a message about upgrading one's driver seems like it will generate misleading advice, more than anything.

xdg:
I don't feel that strongly about it.  LGTM.

xdg:
I take that back.  There's a formatting error: https://travis-ci.org/mongodb/mongo-cxx-driver/jobs/138165163

jrassi:
Formatting error fixed in b095af3.

xdg:
LGTM

PR #500, opened 2016-06-17
CXX-849 Verify range of std::chrono::duration options

xdg:
In many places, libmongoc only takes int32_t for certain duration options
even if the server could take an int64_t.  This commit verifies that
std::chrono::duration options are non-negative integers less than the
maximum int32_t or raises a parameter exception.

While some options and code paths could, in theory, take 64-bit values, in
practice, we don't expect users to set index expirations above billions of
seconds or operation timeouts above billions of milliseconds, so the
positive int32_t range should be more than sufficient and applying that
constraint everywhere provides a consistent, predictable interface for
users.

### DIFF
--- a/src/mongocxx/options/aggregate.cpp
+++ b/src/mongocxx/options/aggregate.cpp
@@ -14,6 +14,9 @@
 
 #include <mongocxx/options/aggregate.hpp>
 
+#include <mongocxx/exception/error_code.hpp>
+#include <mongocxx/exception/logic_error.hpp>
+#include <mongocxx/private/constraints.hpp>
 #include <mongocxx/private/read_preference.hpp>
 
 #include <mongocxx/config/private/prelude.hpp>
@@ -33,6 +36,8 @@ aggregate& aggregate::batch_size(std::int32_t batch_size) {
 }
 
 aggregate& aggregate::max_time(std::chrono::milliseconds max_time) {
+    if (!is_int32_duration<std::chrono::milliseconds>(max_time))
+        throw logic_error{error_code::k_invalid_parameter};
     _max_time = std::move(max_time);
     return *this;
 }
--- a/src/mongocxx/options/count.cpp
+++ b/src/mongocxx/options/count.cpp
@@ -14,6 +14,9 @@
 
 #include <mongocxx/options/count.hpp>
 
+#include <mongocxx/exception/error_code.hpp>
+#include <mongocxx/exception/logic_error.hpp>
+#include <mongocxx/private/constraints.hpp>
 #include <mongocxx/private/read_preference.hpp>
 
 #include <mongocxx/config/private/prelude.hpp>
@@ -33,6 +36,8 @@ count& count::limit(std::int64_t limit) {
 }
 
 count& count::max_time(std::chrono::milliseconds max_time) {
+    if (!is_int32_duration<std::chrono::milliseconds>(max_time))
+        throw logic_error{error_code::k_invalid_parameter};
     _max_time = std::move(max_time);
     return *this;
 }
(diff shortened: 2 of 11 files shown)


xdg:
@acmorrow @jrassi PTAL

Dyslexics of the world untie!

src/mongocxx/options/aggregate.cpp
```diff
@@ -33,6 +36,8 @@ aggregate& aggregate::batch_size(std::int32_t batch_size) {
 }
 
 aggregate& aggregate::max_time(std::chrono::milliseconds max_time) {
+    if (!is_int32_duration<std::chrono::milliseconds>(max_time))
```
acmorrow: You actually shouldn't need to write the template type here. It should be able to be inferred from the argument.

acmorrow: In other words:

is_int32_duration(max_time)

should just work.

src/mongocxx/options/aggregate.cpp
```diff
@@ -33,6 +36,8 @@ aggregate& aggregate::batch_size(std::int32_t batch_size) {
...
+        throw logic_error{error_code::k_invalid_parameter};
```
acmorrow: You can provide a message with logic_error, and it might be good to do so. If you don't want to repeat the text, just make a constant string in the header where you define is_int32_duration:

const char kDurationOverflow[] = "A chrono duration was provided that exceeds the implementation limits";

src/mongocxx/private/constraints.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <mongocxx/config/private/prelude.hpp>
+
+#include <cstdint>
+#include <limits>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+template <typename T>
+bool is_int32_duration(const T& duration) {
```
acmorrow: Instead of making the integral type part of then name, make it part of the template:

```
template<typename Integral, typename Duration>
bool is_positive_and_representable_as(const T& duration) {
    const auto count = duration.count();
    return (count >= 0) && (count <= std::numeric_limits<Integral>>::max());
}
```

Then you can call it as:

```
is_positive_and_representable_as<std::int32_t>(argument);
```

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -322,14 +322,8 @@ TEST_CASE("Collection", "[collection]") {
 
         SECTION("Fails with Options") {
             success = false;
-            expected_expire_after =
-                std::chrono::seconds(static_cast<int64_t>(std::numeric_limits<int32_t>::max()) + 1);
```
acmorrow: Why remove the test that tests overflow?

src/mongocxx/options/count.cpp
```diff
@@ -33,6 +36,8 @@ count& count::limit(std::int64_t limit) {
 }
 
 count& count::max_time(std::chrono::milliseconds max_time) {
+    if (!is_int32_duration<std::chrono::milliseconds>(max_time))
```
acmorrow: I'm pretty OK with the plan of validating at option construction time. I have one *small* nit. Which is that really, the problem isn't when you set the options, it is when you try to apply them against an API that can't support them, being the C driver.

With things this way, that relationship is masked: we are enforcing a constraint *here*, that is really an implementation detail relevant over *there*.

I'm not sure how much it matters really though. Just pointing it out in case you have any thoughts.

xdg: I agree that the problem is when they are applied over *there*, but the point at which the user can do anything about it is when the options are set.  I think it's more user friendly to raise the exception early rather than when the option gets used, which would require the user to backtrack to find where the option was set anyway.

src/mongocxx/test/collection_mocked.cpp
xdg: Good catch.  That was left over from the initial change that was allowing it to go up to 64 bits before I realized the C driver wasn't going to allow that.

xdg:
I'm going to close this as we're not actively working on it.  I'll reopen or refile when we get back to it.

PR #504, opened 2016-06-29
CXX-961 32-bit integration tests should use mmapv1

jrassi:


jrassi:
@acmorrow, PTAL.

acmorrow:
LGTM - Thanks for fixing this!

PR #505, opened 2016-07-07
CXX-966 Make bsoncxx::types::b_ types default constructible

adamchel:


Machyne:
LGMT. @hanumantmk @xdg ptal

src/bsoncxx/types.hpp
```diff
@@ -309,6 +314,11 @@ struct BSONCXX_API b_date {
         : value(std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch())) {
     }
 
+    ///
+    /// Default constructor for b_date.
+    ///
+    BSONCXX_INLINE b_date() = default;
+
     std::chrono::milliseconds value;
```
hanumantmk: I might go a little safer on this one and zero the value here, rather than leaving it uninitialized memory.

adamchel: Done.

src/bsoncxx/types.hpp
```diff
@@ -122,6 +122,11 @@ struct BSONCXX_API b_utf8 {
         : value(std::forward<T>(t)) {
     }
 
+    ///
+    /// Default constructor for b_utf8.
+    ///
+    BSONCXX_INLINE b_utf8() = default;
```
acmorrow: Here and everywhere else, make the default constructor the first constructor

src/bsoncxx/types.hpp
```diff
@@ -309,6 +314,12 @@ struct BSONCXX_API b_date {
         : value(std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch())) {
     }
 
+    ///
+    /// Default constructor for b_date.
+    ///
+    BSONCXX_INLINE b_date() : value(0) {
```
acmorrow: Why doesn't = default work here too?

adamchel: `= default` does work, but Jason recommended we zero it out just to be safe.

src/bsoncxx/types.hpp
```diff
@@ -530,6 +556,11 @@ struct BSONCXX_API b_codewscope {
         : code(std::forward<T>(code)), scope(std::forward<U>(scope)) {
     }
 
+    ///
+    /// Default constructor for b_codewscope.
+    ///
+    BSONCXX_INLINE b_codewscope() = default;
+
```
acmorrow: This is the only one that leaves me a little cold. The default constructor for this is going to leave you with a string_view pointing to nothing (not the empty string!), but a view that will view the default empty document. That seems a little weird. I think it is a side effect of our not allowing views to not view something. Is that something we should consider changing.

Also, will you please fix a nit for me, that I was reminded of while thinking about this?

In `view.cpp`, the `k_default_view` object is not const! Could you please fix that?

src/bsoncxx/types.hpp
acmorrow: Hmm. I just read his comment. I don't see why it would be uninitialized? std::chromo::milliseconds has a default constructor?

adamchel: I just tested this and it seems like the default constructor for `std::chrono::milliseconds` doesn't set it to 0, so I'm leaving this as is.

src/bsoncxx/types.hpp
adamchel: Fixing the nit in a separate PR. Per our discussion we'll leave this fix in review so we have more time to think about it.

acmorrow:
Per in-person discussion, we are going to hold off on this while we think about the consequences a little harder.

acmorrow:
Closing, per the decision in https://jira.mongodb.org/browse/CXX-966

PR #506, opened 2016-07-08
CXX-940 Remove bsoncxx::oid::init_tag

adamchel:


Machyne:
LGTM! @acmorrow @xdg - is this what you had in mind?

hanumantmk:
lgtm

src/bsoncxx/oid.hpp
```diff
@@ -37,27 +37,10 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
-    ///
-    oid();
-
     ///
     /// Constructs an oid and initializes it to a newly generated ObjectId.
     ///
-    /// @param tag
-    ///   A bsoncxx::oid::init_tag used to dispatch this overload.
-    ///
-    explicit oid(init_tag_t tag);
```
acmorrow: Perhaps, rather than removing this method, we should simply deprecate it? That would be friendlier.

src/bsoncxx/oid.hpp
```diff
@@ -37,27 +37,10 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
 class BSONCXX_API oid {
    public:
-    struct init_tag_t {};
-
-    // TODO(MSVC): Ideally this would be constexpr, but VS2015U1 can't
-    // handle it.
-    //
-    // See https://connect.microsoft.com/VisualStudio/feedback/details/2092790
-    //
-    static const init_tag_t init_tag;
```
acmorrow: Ditto re deprecation

src/bsoncxx/oid.cpp
```diff
@@ -23,12 +23,7 @@
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 
-const oid::init_tag_t oid::init_tag{};
-
-oid::oid() : _is_valid(false) {
-}
-
-oid::oid(init_tag_t) : _is_valid(true) {
+oid::oid() : _is_valid(true) {
```
acmorrow: In this mode, what is the purpose of _is_valid? Can we still have an invalid oid?

src/bsoncxx/oid.hpp
```diff
@@ -37,27 +37,10 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
     /// Constructs an oid and initializes it to a newly generated ObjectId.
     ///
-    /// @param tag
-    ///   A bsoncxx::oid::init_tag used to dispatch this overload.
-    ///
-    explicit oid(init_tag_t tag);
+    oid();
 
     ///
     /// Constructs an oid initializes it to the contents of the provided buffer.
```
acmorrow: For dumb GitHub review reasons, I can't comment on lines you didn't change. But, see below where we take a view_or_value in the constructor? That is pointless. Just change it to stdx::string_view.

src/bsoncxx/oid.cpp
```diff
@@ -23,12 +23,7 @@
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 
-const oid::init_tag_t oid::init_tag{};
-
-oid::oid() : _is_valid(false) {
-}
-
-oid::oid(init_tag_t) : _is_valid(true) {
```
acmorrow: Ditto re inability to comment where needed, but note that oid_compare gets a lot simpler if you drop the _is_valid stuff.

src/bsoncxx/oid.cpp
Machyne: What should happen when you do `oid my_id{"foobarbaz"};`?
Right now (both in this PR and as is), this is fine and just sets `_is_valid` to false. Maybe it should throw? Anyway, in the current state, `_is_valid` is still needed.

acmorrow: It should throw. Call `bson_oid_is_valid` on the passed in `string_view` and throw (maybe with a new error code) if it is in't valid.

Machyne: If we throw in that scenario, then I agree with getting rid of `_is_valid`.

src/bsoncxx/oid.hpp
adamchel: Good idea, done.

src/bsoncxx/oid.hpp
adamchel: Done.

src/bsoncxx/oid.cpp
adamchel: In that case, what should I do with operator bool()? Should I have it always return true and deprecate it or just get rid of it entirely?

acmorrow: I like deprecate and make it always return true.

src/bsoncxx/oid.cpp
adamchel: Done.

src/bsoncxx/oid.hpp
adamchel: Done, though do we have to be worried at all about the fact that we're no longer adding a null terminator to what we pass into `bson_oid_init_from_string`?

acmorrow: That is a great question. @ajdavis, can you help out here? It would be unfortunate if we needed to copy the string just to pass it in with a null terminator.

acmorrow: Also interesting to consider: since we have already called bson_oid_is_valid, and have not proceeded, can we assume that it is OK to pass non-null terminated? Again, @ajdavis, thoughts?

ajdavis: Use the source Luke =)

https://github.com/ajdavis/libbson/blob/c7c8109f87e3626e7dc5ce4281ceeda48ebf9762/src/bson/bson-oid.h#L214

`bson_oid_init_from_string` doesn't care whether you null terminate its input

adamchel: Perfect, in this case the way I've implemented it should be safe.

src/bsoncxx/oid.hpp
```diff
@@ -104,7 +107,10 @@ class BSONCXX_API oid {
     ///
     /// Conversion operator that indicates that the oid is initialized.
     ///
-    /// @return True if the oid is initialized.
+    /// @deprecated
```
acmorrow: Thanks for doing the doxygen deprecation. Please note that there is also a macro MONGOCXX_DEPRECATED that will expand to __attribute__((__deprecated__))), which will cause deprecation warnings for users. Can we also add that here, and on the deprecated constructor?

adamchel: Done, and since this caught a use of the operator bool() in the JSON library I got rid of that too.

src/bsoncxx/oid.cpp
```diff
@@ -23,31 +25,32 @@
...
-        bson_oid_init_from_string(&oid, str.terminated().data());
-        memcpy(_bytes.data(), oid.bytes, _bytes.size());
+const oid::init_tag_t oid::init_tag{};
+
+oid::oid(init_tag_t) : oid::oid() {
+}
+
+oid::oid(const bsoncxx::stdx::string_view& str) {
+    if (!bson_oid_is_valid(str.data(), str.size())) {
+        throw bsoncxx::exception{error_code::k_invalid_bson};
```
acmorrow: Optional: Make a new code called k_invalid_oid.

acmorrow:
LGTM with an optional suggestion (no need to repost if you decide to make the change).

src/bsoncxx/oid.cpp
adamchel: This actually makes more sense, I'll just rename `k_invalid_bson` to `k_invalid_oid` since we're not using `k_invalid_bson` anywhere.

acmorrow: Just add a new one. Changing error enums like that is dangerous. A user might be using that value, even if we aren't.

adamchel: `k_invalid_bson` didn't actually exist before this commit

acmorrow: Ah, never mind then!

acmorrow:
@adamchel - Did this get merged? If so, can you note the commit and close the PR?

adamchel:
Yeah, this was merged as c25e50b.

PR #509, opened 2016-07-27
CXX-978 registers driver name and version

puppyofkosh:


acmorrow:
LGTM

bjori:
(he doesn't have push karma, so someone will need to merge it for him) :)

xdg:
@acm, @ajdavis, @puppyofkosh – on reflection, my concern about this approach is that when libmongoc makes the feature available by default (not experimental), then this will stop working. 

That means that the C++11 driver needs to update the conditional in a very coordinated way with libmongoc's change from experimental status.  E.g. "If libmongoc version >= 1.5 *or* (libmongoc version < 1.5  and MONGOC_EXPERIMENTAL_FEATURES), then ...register name and version..."

That's a lot of code churn at every place with experimental features.  (I just realized I think I've arrived at the same concern as in PHPC-753.)

I'd much rather independent feature flags, e.g. `MONGOC_FEATURE_CLIENT_METADATA`, which can be enabled by libmongoc now with some "--experimental" configuration option, or enabled by default *later* when no longer experimental.

That means the C++11 driver doesn't need to know about how libmongoc is managing experimentals, only whether it's exposing features or not.  (This is the same approach already taken with `MONGOC_ERROR_API_VERSION_2`)

acmorrow:
@xdg - Good points. I think I largely agree. Fine grained feature flags would be better here. At some point the C++11 driver will bump its minimum required version, and then it could at least drop the #ifdef'ed in checks, but one big flag is sort of awkward. Another approach that the C++11 driver could take though, would be to probe for the existence of metadata_append at configure time, and use its own preprocessor HAVE_XYZ.

ajdavis:
Let's move the discussion here since we're talking about a C Driver feature:

https://jira.mongodb.org/browse/CDRIVER-1393

xdg:
Per my comments on CDRIVER-1337, this should be held off until we create the 3.1 development branch and then can be merged in without the conditional compilation.

xdg:
N.B. handshake function was renamed in https://github.com/mongodb/mongo-c-driver/commit/af286086a0db8c5a2bc66a0b35b1a91d42d520b4

xdg:
Rebased, amended and merged as f43fb9b3d907de8c4bbf4ac7355c0d289bf65423

PR #510, opened 2016-07-27
CXX-919 BSON Null should not be discouraged

xdg:
This removes the unnecessary warning.

### DIFF
--- a/src/bsoncxx/types.hpp
+++ b/src/bsoncxx/types.hpp
@@ -222,9 +222,8 @@ BSONCXX_INLINE bool operator==(const b_binary& lhs, const b_binary& rhs) {
 ///
 /// A BSON undefined value.
 ///
-/// @warning
-///   This BSON type is used internally by the MongoDB server - use by clients
-///   is discouraged.
+/// @deprecated
+///   This BSON type is deprecated and use by clients is discouraged.
 ///
 struct BSONCXX_API b_undefined {
     static constexpr auto type_id = type::k_undefined;
@@ -455,9 +454,8 @@ BSONCXX_INLINE bool operator==(const b_code& lhs, const b_code& rhs) {
 ///
 /// A BSON Symbol value.
 ///
-/// @warning
-///   This BSON type is used internally by the MongoDB server - use by clients
-///   is discouraged.
+/// @deprecated
+///   This BSON type is deprecated and use by clients is discouraged.
 ///
 struct BSONCXX_API b_symbol {
     static constexpr auto type_id = type::k_symbol;
--- a/src/bsoncxx/types.hpp
+++ b/src/bsoncxx/types.hpp
@@ -346,10 +346,6 @@ BSONCXX_INLINE bool operator==(const b_date& lhs, const b_date& rhs) {
 ///
 /// A BSON null value.
 ///
-/// @warning
-///   This BSON type is used internally by the MongoDB server - use by clients
-///   is discouraged.
-///
 struct BSONCXX_API b_null {
     static constexpr auto type_id = type::k_null;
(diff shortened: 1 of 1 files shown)


src/bsoncxx/types.hpp
```diff
@@ -346,10 +346,6 @@ BSONCXX_INLINE bool operator==(const b_date& lhs, const b_date& rhs) {
 ///
 /// A BSON null value.
 ///
-/// @warning
-///   This BSON type is used internally by the MongoDB server - use by clients
-///   is discouraged.
-///
```
acmorrow: This is copy-pasted several other places in this file. Lets get the others while we are at it.

xdg: bsonspec.org isn't clear what's "discouraged".  Should I remove the boilerplate everywhere?  Or are there actually discouraged (but not deprecated) types?

acmorrow: I believe that the intention was that the warning only apply to the deprecated types: undefined, dbpointer, and symbol.

I think the best thing to do is to @ deprecate them with doxygen in the comments, and then officially deprecate the types with BSONCXX_DEPRECATED tags on the types.

This might create some problems when we tackle https://jira.mongodb.org/browse/CXX-936, since then we will trip on our own deprecations.

But I think we should tackle that later, rather than sooner.

acmorrow: To clarify - I think we should also remove the warning for the places were it is improperly applied.

The deprecation thing is going to be tricky. We need to think of a strategy that:

- Allows us to deprecate a class/method/enum/symbol for consumers of the library
- Allows us to continue to use the symbol internally while using -Wall -Werror
- Does not break ABI.

The traditional way to do this, without regard to ABI, is to rename the (lets say class X) to X_no_deprecated, then using it in as X with a deprecation attribute. Then our code, when it needs to use X, uses X_no_deprecated so that it doesn't cause deprecation warnings when we use it (or, when a user of the library explicitly decides to sidestep the deprecation warning). However, doing that is an ABI breaking change, since now any methods of X have moved to the method X_no_deprecated. So, we can't do that. Something we will need to think about during 3.1 development as we move towards stabilizing ABI.

xdg:
@acmorrow, I've removed additional warnings.  I left in the warning for timestamp since -- while not deprecated -- it really *is* only intended for internal use.

xdg:
Note that DBPointer is already marked deprecated

acmorrow:
@xdg - Note that the @deprecated tag is different from the BSONCXX_DEPRECATED macro. The former goes in doxygen comments, the latter goes on method/class/function declarations, and causes a user facing warning to be emitted when the entity is referenced. I'm suggesting that for things that are truly deprecated, we want both. So, dbpointer, undefined, and symbol should all have both @deprecated and BSONCXX_DEPRECATED. I think, as you point out, that the timestamp one should retain the warning, but not have any deprecation tags or macros.

xdg:
CXX-919 is a docs ticket.  Let's do that now, and later discuss the deprecation macro so I understand it better rather than applying it blindly.

acmorrow:
@xdg Sounds reasonable - can you file a 3.1.0-rc0 ticket to do the deprecation macro so we don't forget? We should use it as an opportunity to figure out an ABI neutral deprecation strategy.

acmorrow:
LGTM. Please update he fixVersion for CXX-919 when you resolve it since this will land in the as-yet-unreleased 3.0.2.

xdg:
Sorry for juggling the branch.  This was closed with beb4efe3e33ed4cd884bef570bc5f098072ea42b

PR #511, opened 2016-07-30
CXX-939 Allow default constructed document and array views

acmorrow:
@xdg @hanumantmk @Machyne - Adds the sharp edges to document::value and array::value. If you default construct one, you get a moved from state.

What do we think? See https://jira.mongodb.org/browse/CXX-939 for background.

CC @bazineta

### DIFF
--- a/src/bsoncxx/array/value.cpp
+++ b/src/bsoncxx/array/value.cpp
@@ -23,6 +23,15 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace array {
 
+namespace {
+const auto nop_deleter = [](uint8_t*){};
+}  // namespace
+
+
+value::value() noexcept
+    : _data(nullptr, nop_deleter), _length() {
+}
+
 value::value(std::uint8_t* data, std::size_t length, deleter_type dtor)
     : _data(data, dtor), _length(length) {
 }
--- a/src/bsoncxx/array/value.cpp
+++ b/src/bsoncxx/array/value.cpp
@@ -24,12 +24,10 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace array {
 
 namespace {
-const auto nop_deleter = [](uint8_t*){};
+const auto nop_deleter = [](uint8_t*) {};
 }  // namespace
 
-
-value::value() noexcept
-    : _data(nullptr, nop_deleter), _length() {
+value::value() noexcept : _data(nullptr, nop_deleter), _length() {
 }
 
 value::value(std::uint8_t* data, std::size_t length, deleter_type dtor)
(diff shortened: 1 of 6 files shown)


acmorrow:
Also, @jrassi

Machyne:
I think how this is written right now, a default `value` that gets copied becomes a valid ("true") `value`, because the default view gives you real bson.

src/bsoncxx/array/value.cpp
```diff
@@ -23,6 +23,13 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace array {
 
+namespace {
+const auto nop_deleter = [](uint8_t*) {};
+}  // namespace
+
+value::value() noexcept : _data(nullptr, nop_deleter), _length() {
```
hanumantmk: Does this technically involve static initialization order fiasco's if default constructed views are created during dynamic initialization?

Free function is possibly safer?

src/bsoncxx/document/value.cpp
```diff
@@ -23,6 +23,13 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace document {
 
+namespace {
+const auto nop_deleter = [](uint8_t*) {};
```
hanumantmk: same thought about initialization order fiascos

src/bsoncxx/array/value.cpp
acmorrow: Very interesting question, and I don't know, so I'm going to switch it. Too bad we don't have constexpr lambdas yet.

hanumantmk:
Beyond using a free function nop-deleter over a lambda, LGTM

src/bsoncxx/document/value.cpp
acmorrow: Done

acmorrow:
@Machyne - Good catch. I think it is actually a little worse than that, in that the copy constructor, invoked against a default constructed `value` will call `.view` on it, which isn't well defined.

On the other hand, as currently documented, a default constructed `value` has moved-from semantics, so, should we expect the copy constructor to work?

Or, is this indicating that we should switch paths and make a default constructed value look like an empty document?

Machyne:
@acmorrow - not only that, but the `view.empty()` will be `false` because the length won't be 0, making it even harder to debug. Could we add something like `assert(*this);` in `value::view()`?

I'm also fine with saying "this has moved from semantics, don't copy it".

acmorrow:
I'm tempted to also put up the alternative, which makes a default constructed value look like an empty document, so we can decide which we dislike least.

Note that we can also do nothing! The original motivation here was that it wasn't possible to write:

```
document::value x;
try {
   x = ...
}
catch(...) {
}
```

Do we actually care? Especially given that you can just as easily write:

```
stdx::optional<document::value> x;
try {
    x = ...
}
```

acmorrow:
Closing this issue per discussion in https://jira.mongodb.org/browse/CXX-939

PR #512, opened 2016-07-30
CXX-941 Use libbson for converting to JSON

acmorrow:
Instead of rolling our own hand-rolled BSON to JSON converter, just use the one in libbson.


acmorrow:
@xdg @jrassi @Machyne @hanumantmk 

This fixes an issue where we weren't properly escaping according to the JSON spec. Rather than trying to do that with out own visitor, it seemed easiest to just leverage the work that libbson has done.

That does, however, tie our tests to the particular format that libbson emits. @ajdavis does that seem problematic to you in any way?

ajdavis:
I could use a little more context. The C++ driver does *not* have a public API for converting between BSON and JSON, right? So the question is whether your tests should use libbson's JSON codec, versus a hand-rolled one?

Regardless, I'd recommend you do depend on libbson's JSON codec, and this PR looks like a good idea to me. Our JSON output is quite stable.

acmorrow:
No, we do have a public API, in the bsoncxx/json.h header. It was just that the ```to_json``` function was implemented directly, rather than using libbson, and it was buggy.

ajdavis:
Got it. I still recommend libbson's JSON codec.

src/bsoncxx/test/json.cpp
```diff
@@ -18,6 +18,8 @@
 #include <bsoncxx/exception/exception.hpp>
 #include <bsoncxx/json.hpp>
 
+#include <iostream>
```
hanumantmk: what pulled the iostream dependency?

src/bsoncxx/json.cpp
```diff
@@ -42,247 +42,20 @@ void bson_free_deleter(std::uint8_t* ptr) {
...
-#undef BSONCXX_ENUM
-    }
-
-    return ss.str();
-}
-
-std::string to_json(types::value value) {
-    std::stringstream ss;
+    bson_t bson;
+    bson_init_static(&bson, view.data(), view.length());
```
hanumantmk: check the return value on init_static and throw if the view isn't valid?  That'll catch invalid and mismatched lengths.

Otherwise an invalid view will lead you to bson_as_json some uninitialized bytes

src/bsoncxx/test/json.cpp
acmorrow: Debugging. I've removed it.

src/bsoncxx/json.cpp
acmorrow: We don't seem to actually validate our calls to bson_init_static consistently elsewhere in the code. Should we be?

acmorrow:
@hanumantmk Lets fix up the bson_init_static in a separate ticket. Given that, ping on the rest of this?

src/bsoncxx/test/json.cpp
Machyne: Looks like it's still here?

src/bsoncxx/json.hpp
```diff
@@ -36,28 +36,6 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
 ///
-/// Converts an element (key-value pair) to a JSON key-value pair.
-/// This is not actually a valid JSON document, but it can be useful for debugging.
-///
-/// @param element
-///   A BSON element.
-///
-/// @returns A JSON key-value pair.
-///
-BSONCXX_API std::string BSONCXX_CALL to_json(document::element element);
```
Machyne: Is there a reason that we got rid of this? As the comment says, I think this could be useful for debugging.

src/bsoncxx/test/json.cpp
acmorrow: I didn't post an update, but it is in my tree.

src/bsoncxx/json.hpp
acmorrow: Can't be supported on the libbson implementation, that I am aware of. @ajdavis, LMK if I'm wrong about that.

Machyne: I like how this looks in your test anyway, so not a big issue if it can't be done.

Machyne:
I'd like to hear back from @ajdavis first, but LGTM

src/bsoncxx/json.cpp
acmorrow: I've filed https://jira.mongodb.org/browse/CXX-984 to track this work.

src/bsoncxx/json.hpp
ajdavis: Yeah, we only provide to_json for documents, we have no API for working with elements.

xdg: What about creating a builder, appending the single element, then dumping that to JSON?  It changes the semantics (e.g. now a valid JSON document) but still has debugging utility.

acmorrow: Ideally, given a doc ```doc``` that would be as simple as

```
bsoncxx::document::element elt = ...;
auto elt_jas_json = to_json(builder::stream::document << elt << finalize);
```

But, interestingly, there doesn't seem to be an stream/basic/core overload for the builders that takes an element. Or am I just overlooking it somehow? @hanumantmk?

ajdavis:
I'm unsubscribing from this, let me know if you need anything more from me.

src/bsoncxx/json.hpp
acmorrow: @xdg @Machyne @hanumantmk  I filed https://jira.mongodb.org/browse/CXX-997 to capture this, which would make the feature request here trivial.

I think we should commit this as is. Any further comments?

xdg: LGTM

PR #513, opened 2016-07-30
CXX-890 Don't use objects requiring dynamic init/fini for instance::c…

acmorrow:
…urrent

### DIFF
--- a/examples/mongocxx/instance_management.cpp
+++ b/examples/mongocxx/instance_management.cpp
@@ -1,0 +1,99 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cstdlib>
+#include <memory>
+
+#include <bsoncxx/stdx/make_unique.hpp>
+#include <bsoncxx/stdx/optional.hpp>
+#include <bsoncxx/stdx/string_view.hpp>
+
+#include <mongocxx/instance.hpp>
+#include <mongocxx/logger.hpp>
+#include <mongocxx/pool.hpp>
+#include <mongocxx/stdx.hpp>
+#include <mongocxx/uri.hpp>
+
+namespace {
+
+// This example demonstrates how one might keep a heap allocated mongocxx::instance and
+// mongocxx::pool object associated in a way that allows dynamic configuration. Most of the examples
+// simply create a stack allocated mongocxx::instance object, which is fine for small examples, but
+// real programs will typically require shared access to a commonly configured instance and
+// connection pool across different translation units and components. By providing a singleton which
+// owns the instance and pool objects, we can defer configuration until we are ready to use MongoDB,
+// and share the instance and pool objects between multiple functions.
+
+class mongo_access {
+public:
+
+    static mongo_access& instance() {
+        static mongo_access instance;
+        return instance;
+    }
+
+    void configure(std::unique_ptr<mongocxx::instance> instance,
+                   std::unique_ptr<mongocxx::pool> pool) {
+        _instance = std::move(instance);
+        _pool = std::move(pool);
+    }
+
+    using connection = mongocxx::pool::entry;
+
+    connection get_connection() {
+        return _pool->acquire();
+    }
+
+    mongocxx::stdx::optional<connection> try_get_connection() {
+        return _pool->try_acquire();
+    }
+
+private:
+    mongo_access() = default;
+
+    std::unique_ptr<mongocxx::instance> _instance = nullptr;
+    std::unique_ptr<mongocxx::pool> _pool = nullptr;
+};
+
+}  // namespace
+
+// The 'configure' and 'do_work' functions use the same mongocxx::instance and mongocxx::pool
+// objects by way of the mongo_access singleton.
+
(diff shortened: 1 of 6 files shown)


acmorrow:
@xdg @jrassi @hanumantmk @Machyne 

This is sort of a work-in-progress. Please read CXX-890 for background on why the existing approach for implementing ```instance::current``` was problematic in a static-linking context.

I'm not 100% sure that we want to go this way. If we go the other way, we would just remove instance::current entirely.

To close the ticket we will still need to write up an example demo'ing how to manage a long lived instance object, per @bazineta's request.

src/mongocxx/instance.cpp
```diff
@@ -94,31 +96,25 @@ instance::instance() : instance(nullptr) {
...
 
 instance::instance(instance &&) noexcept = default;
 instance &instance::operator=(instance &&) noexcept = default;
 
 instance::~instance() {
-    std::lock_guard<std::recursive_mutex> lock(instance_mutex);
-    if (current_instance != this) std::abort();
     _impl.reset();
-    current_instance = nullptr;
+    current_instance.store(nullptr);
```
acmorrow: This should be moved up a line, above _impl.reset()

acmorrow: And, actually... it can't be. There is no proper sequencing of the reset on the _impl and nulling out the current_instance pointer. I think this may be the end for `instance::current`

hanumantmk: I wonder if you can handle this with a little state machine.

I.e.  Create a synthetic instance* that implies the shutting down state and make instance::current spin on it if current_instance.load() returns that specific value.

That way you can:

1. Set the current_instance over to the shutting down state (which forces a spin)
2. Run the libmongoc cleanup code
3. Set the current_instance to null
4. Allow the caller of instance::current in and allow it to create a new instance (which re-initializes)

xdg:
I don't think the complexity is worth it.  If users want global state, let's let them design it how they want rather than force them to use/work-around a singleton design.

src/mongocxx/instance.cpp
acmorrow: @hanumantmk I tried out your idea, PTAL. I'm not sure if it is better or not at this point.

acmorrow:
@xdg - Having `instance::current` does two very helpful things: it makes it much easier to write unit tests without requiring a fixture to ensure initialization, and it allows a user who has already built a facility to properly initialize the driver, via their own mechanism, to retrieve elsewhere in their code the initialized instance of the driver, without that code needing to depend on (or even know about) the initialization mechanism used. If it is possible to preserve it, I think it has value, even if it costs a little bit of complexity.

acmorrow:
@jrassi @xdg @hanumantmk - I've updated this with a full examples showing how to manage an instance/pool pair via a singleton, as was requested by @bazineta in CXX-890. I've also verified that with the existing implementation of mongocxx::instance::current on master, I can reproduce the crash when the recursive mutex is accessed during static destruction, by building the new example against a static libmongocxx.a. After the changes in this commit, the example no longer crashes when statically linked.

acmorrow:
Please note that this is the last open issue blocking a C++11 driver 3.0.2 release, so prompt feed back would be appreciated.

examples/mongocxx/instance_management.cpp
```diff
@@ -0,0 +1,110 @@
...
+public:
+
+    static mongo_access& instance() {
+        std::lock_guard<std::mutex> lock(_singleton_mutex);
+        if (!_singleton)
+            _singleton.reset(new mongo_access);
+        return *_singleton;
+    }
+
+public:
```
xdg: Why are there duplicate `public` and `private` labels?

acmorrow: No good reason, really. I'll remove them.

acmorrow: Removed

examples/mongocxx/instance_management.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+class mongo_access {
+public:
+
+    static mongo_access& instance() {
+        std::lock_guard<std::mutex> lock(_singleton_mutex);
+        if (!_singleton)
+            _singleton.reset(new mongo_access);
+        return *_singleton;
+    }
```
xdg: This incurs the lock cost on every access.  How about the cheaper double-checked locking pattern instead?
```
static mongo_access& instance() {
    if (!_singleton) {
        std::lock_guard<std::mutex> lock(_singleton_mutex);
         if (!_singleton)
             _singleton.reset(new mongo_access);
    }
    return *_singleton;
}
```

xdg:
I want to think more about `instance::current`.  My inclination is still to leave it to users to implement (possibly based on the example) if they want.  If the major use is for testing, then we can handle that via something private, not something public.

examples/mongocxx/instance_management.cpp
acmorrow: Double checked locking is unsafe in C++11 without fences: https://en.wikipedia.org/wiki/Double-checked_locking

acmorrow:
We can re-evaluate whether we want to keep it in 3.0.3 or 3.1. For now, this fixes a bug that makes use of mongocxx in static libraries problematic, even if you aren't using `instance::current`. I think it is better to fix the bug now, and evaluate the feature separately, and later.

examples/mongocxx/instance_management.cpp
acmorrow: We could easily enough re-write it to use the Meyer's singleton via magic-statics, if you think that is clearer

acmorrow: Some more details on that: http://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

acmorrow: I updated it to use a magic-statics based singleton.

xdg:
I'm fine with this fix going into 3.0.2.  I do think we should re-evaluate.  I think we may want to separate mongoc init/cleanup from setting/changing the logging function and that might be a cleaner model.  I'll try to articulate something in a ticket soonish.

PR #514, opened 2016-08-01
CXX-939 Default constructed values view the empty view

acmorrow:
Here is an alternative to https://github.com/mongodb/mongo-cxx-driver/pull/511. It represents a compromise:

- Default constructed values for document and array can exist.
- The view from a default constructed document or array is to the same object as a default constructed array::view or document::view
- Moved from document or array values are invalid to use.

In addition, I've plumbed through constexpr and noexcept where appropriate, such that constexpr array and document views can be constructed against static buffers at constexpr time. You can't actually *do* that much with the resulting view, but it is a nice PoC.

### DIFF
--- a/src/bsoncxx/array/value.hpp
+++ b/src/bsoncxx/array/value.hpp
@@ -33,8 +33,11 @@ namespace array {
 ///
 class BSONCXX_API value {
    public:
-    using deleter_type = void (*)(std::uint8_t*);
-    using unique_ptr_type = std::unique_ptr<uint8_t, deleter_type>;
+    using deleter_type = void (*)(const std::uint8_t*);
+    using unique_ptr_type = std::unique_ptr<const uint8_t, deleter_type>;
+
+
+    value() noexcept;
 
     ///
     /// Constructs a value from a buffer.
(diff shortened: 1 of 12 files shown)


acmorrow:
This also tightens up the type of buffer that a value owns, making it const. That was always the intention, but we had overlooked it before.

acmorrow:
@Machyne @xdg @jrassi @hanumantmk PTAL.

acmorrow:
@Machyne @xdg @jrassi @hanumantmk ... ping?

src/bsoncxx/array/value.hpp
```diff
@@ -106,7 +108,7 @@ class BSONCXX_API value {
 };
 
 BSONCXX_INLINE array::view value::view() const noexcept {
-    return array::view{static_cast<uint8_t*>(_data.get()), _length};
+    return array::view{static_cast<const uint8_t*>(_data.get()), _length};
```
Machyne: Do you need this static cast after changing the `unique_ptr_type`?

Machyne:
Can you add a test case to check something like `value{}.view() == view{}`?

acmorrow:
Closing this issue without merging, per discussion in https://jira.mongodb.org/browse/CXX-939

PR #515, opened 2016-08-10
CXX-986 Throw mongocxx::exception if last_error document is null

rustyx:
Fixes [CXX-986](https://jira.mongodb.org/browse/CXX-986)


acmorrow:
@rustyx - Thank you for the ticket and the PR. I think we are planning to fix the ticket in a slightly different way, however, by using the information that the libmongoc call gives us. We would be interested in your thoughts as well. Please see https://github.com/mongodb/mongo-cxx-driver/pull/529

PR #516, opened 2016-08-11
CXX-989 Simplify bsoncxx::document::value::release

xdg:


xdg:
PTAL: @acmorrow, @jrassi

src/bsoncxx/document/value.cpp
```diff
@@ -53,11 +54,7 @@ value& value::operator=(const value& rhs) {
 }
 
 value::unique_ptr_type value::release() {
-    auto x = std::move(_data);
-
-    _data.release();
-
-    return x;
+    return std::move(_data);
```
acmorrow: Per discussion, is this std::move required? I can't remember which way we decided it.

xdg: It is required, because `_data` is a class member.  It's not necessary for `x` because that's a local variable which gets copy-elision (or in the absence of copy-elision support, the return checks first for a constructor consuming an rvalue before one consuming an lvalue).

acmorrow: @xdg right, thanks for reminding me. LGTM

Machyne:
LGTM!

PR #518, opened 2016-08-15
CXX-861 Add example iterating cursor and getting data out

xdg:
I wrote out the types long-hand instead of using 'auto' into order
to better educate readers about what types are being used.

I also omitted most error handling to keep the focus on working with
the cursor, documents, arrays and elements.

### DIFF
--- a/examples/mongocxx/get_values_from_documents.cpp
+++ b/examples/mongocxx/get_values_from_documents.cpp
@@ -1,0 +1,135 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/client.hpp>
+
+using bsoncxx::builder::stream::document;
+using bsoncxx::builder::stream::close_array;
+using bsoncxx::builder::stream::finalize;
+using bsoncxx::builder::stream::open_array;
+using bsoncxx::type::k_array;
+using bsoncxx::type::k_document;
+using bsoncxx::type::k_int32;
+using bsoncxx::type::k_int64;
+using bsoncxx::type::k_oid;
+using bsoncxx::type::k_utf8;
+
+// Document model, showing array with nested documents:
+//
+// {
+//     "_id" : ObjectId(...),
+//     "messagelist" : [
+//       {
+//         "uid" : 413098706,
+//         "status" : 2,
+//         "msg": "..."
+//       },
+//       ...
+//     ]
+// }
+
+// Construct a document in the format of 'messagelist'.
+bsoncxx::document::value new_message(int64_t uid, int32_t status, std::string msg) {
+    document doc;
+    return doc << "uid" << uid << "status" << status << "msg" << msg << finalize;
+}
+
+// Insert a document into the database.
+void insert_test_data(mongocxx::collection& coll) {
+    document builder{};
+    builder << "messagelist" << open_array << new_message(413098706, 3, "Lorem ipsum...")
+            << new_message(413098707, 2, "Lorem ipsum...")
+            << new_message(413098708, 1, "Lorem ipsum...") << close_array;
+
+    bsoncxx::document::value doc = builder << finalize;
+
+    // Normally, one should check the return value for success.
+    coll.insert_one(doc.view());
+}
+
+// Iterate over contents of messagelist.
+void iterate_messagelist(const bsoncxx::document::element& ele) {
+    // Check validity and type before trying to iterate.
+    if (ele.type() == k_array) {
+        bsoncxx::array::view subarray{ele.get_array().value};
+        for (const bsoncxx::array::element& msg : subarray) {
+            // Check correct type before trying to access elements.
+            // Only print out fields if they exist; don't report missing fields.
+            if (msg.type() == k_document) {
+                bsoncxx::document::view subdoc = msg.get_document().value;
+                bsoncxx::document::element uid = subdoc["uid"];
+                bsoncxx::document::element status = subdoc["status"];
+                bsoncxx::document::element msg = subdoc["msg"];
+                if (uid && uid.type() == k_int64) {
+                    std::cout << "uid: " << uid.get_int64().value << std::endl;
+                }
+                if (status && status.type() == k_int32) {
+                    std::cout << "status: " << status.get_int32().value << std::endl;
+                }
+                if (msg && msg.type() == k_utf8) {
+                    std::cout << "msg: " << msg.get_utf8().value << std::endl;
+                }
+            } else {
+                std::cout << "Message is not a document" << std::endl;
+            }
+        }
+    } else {
+        std::cout << "messagelist is not an array" << std::endl;
+    }
+}
+
+// Print document parts to standard output.
+void print_document(const bsoncxx::document::view& doc) {
+    // Extract _id element as a string.
+    bsoncxx::document::element id_ele = doc["_id"];
+    if (id_ele.type() == k_oid) {
+        std::string oid = id_ele.get_oid().value.to_string();
+        std::cout << "OID: " << oid << std::endl;
+    } else {
+        std::cout << "Error: _id was not an object ID." << std::endl;
+    }
+
+    // Extract "messagelist" element, which could be the 'invalid' (false)
+    // element if it doesn't exist.
+    bsoncxx::document::element msgs_ele = doc["messagelist"];
+    if (msgs_ele) {
+        iterate_messagelist(msgs_ele);
+    } else {
+        std::cout << "Error: messagelist field missing." << std::endl;
+    }
+}
+
+void iterate_documents(mongocxx::collection& coll) {
+    // Execute a query with an empty filter (i.e. get all documents).
+    mongocxx::cursor cursor = coll.find({});
+
+    // Iterate the cursor into bsoncxx::document::view objects.
+    for (const bsoncxx::document::view& doc : cursor) {
+        print_document(doc);
+    }
+}
+
+int main() {
+    mongocxx::instance inst{};
+    mongocxx::client client{mongocxx::uri{}};
+    mongocxx::collection coll = client["test"]["events"];
+
+    coll.drop();
+    insert_test_data(coll);
+    iterate_documents(coll);
+}
--- a/examples/mongocxx/CMakeLists.txt
+++ b/examples/mongocxx/CMakeLists.txt
@@ -28,6 +28,7 @@ set(MONGOCXX_EXAMPLES
     connect.cpp
     document_validation.cpp
     exception.cpp
+    get_values_from_documents.cpp
     index.cpp
     inserted_id.cpp
     instance_management.cpp


xdg:
PTAL: @jrassi, @acmorrow

xdg:
Updated to fix a failing type conversion on old Clang.

@jrassi please look this over when you get a chance.  I'd like to merge it and close the ticket.
Thanks!

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/client.hpp>
+
+using namespace bsoncxx;
```
jrassi: Suggest pulling in the types you need from bsoncxx individually, instead of adding a using-directive.  This suggestion is mostly for consistency with the other examples in this directory.  Note also that the Google C++ style guide bans using-directives entirely (see <https://google.github.io/styleguide/cppguide.html#Namespaces), although it looks like the driver uses them moderately.  Mostly in test code, it seems.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+//       {
+//         "uid" : 413098706,
+//         "status" : 2,
+//         "msg": "..."
+//       },
+//       ...
+//     ]
+// }
+
+// Construct a document in the format of 'messagelist'
```
jrassi: Suggest ending comments in periods.  Ditto throughout.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+document::value new_message(int64_t uid, int32_t status, std::string msg) {
+    builder::stream::document doc;
+    return doc << "uid" << uid << "status" << status << "msg" << msg << builder::stream::finalize;
+}
+
+// Insert a document into the database
+void insert_test_data(mongocxx::collection& coll) {
```
jrassi: Suggest changing `coll` to be a pointer type, instead of a reference type.  This is just straight from the Google C++ style guide; see <https://google.github.io/styleguide/cppguide.html#Reference_Arguments>.

Same thing for the declaration of `iterate_documents()`.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+    builder::stream::document builder{};
+    builder << "messagelist" << builder::stream::open_array
+            << new_message(413098706, 3, "Lorem ipsum...")
+            << new_message(413098707, 2, "Lorem ipsum...")
+            << new_message(413098708, 1, "Lorem ipsum...") << builder::stream::close_array;
+
+    document::value doc = builder << builder::stream::finalize;
+
+    // Normally, one should check the return value for success
```
jrassi: I'm of the opinion that all examples should feature production-quality error handling, though I see your dissent in #519.  Perhaps let's discuss further offline.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+    coll.insert_one(doc.view());
+}
+
+// Iterate over contents of messagelist
+void iterate_messagelist(document::element& ele) {
```
jrassi: Suggest declaring `ele` as const.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // Check validity && type before trying to iterate
```
jrassi: Nit: suggest spelling out "and" instead of using "&&" in a comment.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (ele.type() == type::k_array) {
+        array::view subarray{ele.get_array().value};
+        for (array::element msg : subarray) {
```
jrassi: I'm somewhat conflicted as far as whether a detailed exploration of the document_view/element methods belongs here, or in an existing bsoncxx example like getting_values.cpp.  What do you think?

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+            // Check correct type before trying to access elements.
+            // Only print out fields if they exist; don't report missing fields.
+            if (msg.type() == type::k_document) {
+                document::view subdoc{msg.get_document().value};
+                if (subdoc["uid"] && subdoc["uid"].type() == type::k_int64) {
```
jrassi: Suggest pulling out the `subdoc["uid"]` element into a local variable (perhaps `uidEle`?), instead of calculating it a few times.  Same with the other elements in this block.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+            } else {
+                std::cout << "Message is not a document" << std::endl;
+            }
+        }
+    } else {
+        std::cout << "messagelist is not an array" << std::endl;
+    }
+}
+
+// Print document parts to STDOUT
```
jrassi: Nit: suggest "standard out" instead.  STDOUT makes me think of Perl.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+void print_document(const bsoncxx::document::view& doc) {
```
jrassi: Nix the `bsoncxx::` prefix here, as you have a using-directive in scope.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // Define an element view for inspecting different fields
+    document::element ele{};
+
+    // Extract _id element as a string
+    ele = doc["_id"];
+    if (ele.type() == type::k_oid) {
+        std::string oid = ele.get_oid().value.to_string();
```
jrassi: Suggest omitting the to_string() call, for consistency with how you print `msg` at 74, and because you don't need an owned string here.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -0,0 +1,127 @@
...
+        std::cout << "OID: " << oid << std::endl;
+    } else {
+        std::cout << "Error: _id was not an object ID." << std::endl;
+    }
+
+    // Extract "messagelist" element, which could be the 'invalid' (false)
+    // element if it doesn't exist
+    ele = doc["messagelist"];
```
jrassi: I find it slightly more stylish / easier to read to declare a new document::element variable here with a descriptive variable name, instead of reusing the one declared above at 88.  Suggest something like: deleting 88, replacing 91 with `document::element idEle = doc["_id"];`, and replacing 101 with `document::element messageListEle = doc["messageList"];`.

jrassi:
Left a round of feedback, which is mostly style nits.  See also feedback at [get_values_from_documents.cpp:62](https://github.com/mongodb/mongo-cxx-driver/pull/518#discussion_r77737522), where I express uncertainty about where this example should live.

examples/mongocxx/get_values_from_documents.cpp
xdg: It could be `const` -- we don't modify the *object*, though we do modify data on MongoDB.  My view on reference/pointer is "pass reference when it can't be null" (I think that's one of Meyer's prescriptions.)

acmorrow: @jrassi @xdg - There is a general kernel team avoidance of using non-const reference types for parameters, since it makes it hard at the call site to know whether a given parameter may be being changed by the call. If you make "in parameters" by-pointer, then it is clearer at the call sites which parameters are expected to be changed. I'm not taking a position either way on whether we want this guidance to apply in the C++11 driver, at least not quite yet.

xdg: I agree it's a good convention. I suspect when I wrote this, I hadn't internalized the importance of "const" yet and was just getting the code to work.  I'll go through and amend this and other parameters accordingly.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: I see what happened.  Things like "insert_one" aren't currently const methods.  I think that should change as the *objects* are const.  Yes, they carry out operations with side effects that mutate database state, but that's entirely external to the C++ code.

xdg: I'm going to leave these as non-const references as making them pointers implies mutation, which we don't do either.

examples/mongocxx/get_values_from_documents.cpp
xdg: I've put it on the team meeting discussion list

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: The original user question has the data model described so the example code answered their question which included getting elements out of the nested array.  I think giving end-to-end examples like this are useful and will forestall additional questions.  Saying it differently: I think task-focused examples may be more immediately useful to users than narrow API examples that require them to stitch together concepts from multiple example files.

examples/mongocxx/get_values_from_documents.cpp
xdg: Good point given that `operator[]` is O(N)!  Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done

examples/mongocxx/get_values_from_documents.cpp
xdg: I think it can't be done, as `bsoncxx::oid` doesn't provide a helper for streaming to output and doesn't provide a string_view option.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done

xdg:
@jrassi @acmorrow I've pushed a commit with changes. PTAL.

examples/mongocxx/get_values_from_documents.cpp
jrassi: I'm not in agreement about making insert_one const, but I'd like to defer discussion of that.

I do feel more strongly about having a consistent rule across the codebase for reference versus pointer function parameters, though.  From a quick-and-dirty grep of the source I can find 1 file containing function declarations/definitions with non-const reference parameters, versus 38 files containing function declarations/definitions with non-const pointer parameters, and I'd love to maintain conformance with the existing style where possible.  I'd be happy take a look at the Meyers' recommendation and consider it for the future, but I'd say the project currently inherits the "All parameters passed by reference must be labeled const" rule from claiming to follow Google C++ style.

examples/mongocxx/get_values_from_documents.cpp
jrassi: Acknowledged.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -18,7 +18,22 @@
...
+using bsoncxx::builder::stream::finalize;
+using bsoncxx::builder::stream::open_array;
+using bsoncxx::type::k_array;
+using bsoncxx::type::k_document;
+using bsoncxx::type::k_int32;
+using bsoncxx::type::k_int64;
+using bsoncxx::type::k_oid;
+using bsoncxx::type::k_utf8;
+
+using doc_builder = bsoncxx::builder::stream::document;
```
jrassi: Instead of seeing type aliases used for these, I'd much rather either see any of the following, in decreasing order of preference: 1) spelling them out, like the other mongocxx examples do, 2) putting back the mongocxx using-declaration that you had in your first patch set, or 3) using auto.  We should change the names for these classes if we dislike them, but until then I don't like encouraging users to use different names for them.

Not sure if I was unclear in my earlier suggestion, apologies if so.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -34,44 +49,47 @@ using namespace bsoncxx;
...
-    // Check validity && type before trying to iterate
-    if (ele.type() == type::k_array) {
-        array::view subarray{ele.get_array().value};
-        for (array::element msg : subarray) {
+// Iterate over contents of messagelist.
+void iterate_messagelist(const doc_element& ele) {
+    // Check validity and type before trying to iterate.
+    if (ele.type() == k_array) {
+        array_view subarray{ele.get_array().value};
+        for (array_element& msg : subarray) {
```
jrassi: Nit: suggest making `msg` a non-reference type (`element` is only 16 bytes and is trivially destructible) or a const reference, if you don't need to modify it.

examples/mongocxx/get_values_from_documents.cpp
```diff
@@ -82,25 +100,22 @@ void iterate_messagelist(document::element& ele) {
...
+    doc_element id_ele = doc["_id"];
+    if (id_ele.type() == k_oid) {
+        std::string oid = id_ele.get_oid().value.to_string();
         std::cout << "OID: " << oid << std::endl;
     } else {
         std::cout << "Error: _id was not an object ID." << std::endl;
     }
 
     // Extract "messagelist" element, which could be the 'invalid' (false)
     // element if it doesn't exist
```
jrassi: Nit: period here, and on 125, and on 128.

jrassi:
One more round, all style nits this time.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: Done.

examples/mongocxx/get_values_from_documents.cpp
xdg: done.

jrassi:
Let's discuss the thread at <
https://github.com/mongodb/mongo-cxx-driver/pull/518#discussion_r78251165>
at some point (I also think it's possible that you missed my last comment
on it, I'm personally finding GitHub PRs still a bit hard to follow).

PR #519, opened 2016-08-23
minor: add query projection example

xdg:


examples/mongocxx/query_projection.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <mongocxx/instance.hpp>
+#include <mongocxx/uri.hpp>
+
+using bsoncxx::builder::stream::document;
+using bsoncxx::builder::stream::open_document;
+using bsoncxx::builder::stream::close_document;
+using bsoncxx::builder::stream::finalize;
+
+int main(int, char **) {
+  mongocxx::instance inst{};
```
acmorrow: Perhaps this should be the first example that uses pool?

examples/mongocxx/query_projection.cpp
```diff
@@ -0,0 +1,43 @@
...
+  mongocxx::client conn{mongocxx::uri{}};
+
+  auto coll = conn["test"]["foo"];
```
acmorrow: Maybe slightly better to use the name of the test binary as the collection, in case we want to run examples concurrently during test.

examples/mongocxx/query_projection.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  // Create the find options with the projection
+  mongocxx::options::find opts{};
+  opts.projection(document{} << "_id" << 1 << finalize);
+
+  // Execute find with options
+  auto cursor = coll.find(filter.view(), opts);
+  for (auto &&doc : cursor) {
+    std::cout << bsoncxx::to_json(doc) << std::endl;
+  }
```
acmorrow: Main should return. I think you want to return EXIT_SUCCESS

examples/mongocxx/query_projection.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
```
acmorrow: Consider wrapping the whole example in a try/catch and printing an exception to stdout if it occurs, then using EXIT_FAILURE.

acmorrow:
Looks pretty good, a few small comments.

examples/mongocxx/query_projection.cpp
xdg: My quick experiments led to problems.  Let's do a standalone pool example per CXX-995 and then we can retro fit that.

examples/mongocxx/query_projection.cpp
xdg: Will fix.

examples/mongocxx/query_projection.cpp
xdg: Will fix.

examples/mongocxx/query_projection.cpp
xdg: I think try/catch clutters examples and obscures the main point being illustrated.  Generally I think should have one example for exception handling and then only add it to examples when it's extremely relevant (e.g. write failure).  So I'd prefer to omit it here.

jrassi:
One last style nit: I'd love to maintain the existing convention that all comments end in periods in this repository.  No need to change here, but if you agree, I'd love to get in the habit of always doing this going forward.

xdg:
That's fine with me for sentence-style comments.

PR #520, opened 2016-08-24
CXX-1011: Support finding libbson and mongo_c on Windows when pkg-config is present

aseering:
This PR enables a workaround for an upstream bug:  libbson and mongo_c's builds don't create .pc files on Windows.  As a result, pkg-config can't find them even if they exist and it is told where to look.

The workaround is as follows:  The existing cmake files always trust pkg-config if it is found -- if a user specifies a correct path to libbson, but pkg-config can't find libbson, then the cmake file reports that libbson can't be found.  This fix prioritizes the user-specified path so that it is always used if specified, even if pkg-config runs and can't find the package.

### DIFF
--- a/cmake/FindLibBSON.cmake
+++ b/cmake/FindLibBSON.cmake
@@ -22,17 +22,18 @@ include(FindPackageHandleStandardArgs)
 # Load up PkgConfig if we have it
 find_package(PkgConfig QUIET)
 
-if (PKG_CONFIG_FOUND)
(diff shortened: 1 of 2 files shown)


xdg:
Thanks!  Assigned [CXX-1011](https://jira.mongodb.org/browse/CXX-1011) for tracking.

acmorrow:
LGTM, I'm merging this.

acmorrow:
Merged in b3061e3061bc81467bcd008eef8f681d5d45cb52

aseering:
Thanks!

PR #521, opened 2016-08-24
CXX-1010 Use a regular string rather than a raw string; fix Visual Studio 2015 build

aseering:
Visual Studio finds a bunch of syntax errors on this line with the raw string, so it fails to compile mongo_cxx's test suite.  It seems to start failing at "id1"; I'm guessing that it thinks the string ends at that point, maybe because of the escaped quote or because of the macro expansion or some combination of the above.  (There are several other raw strings in the file; they are all simpler strings, but they all compile fine.)

This fixes the test-build on Visual Studio by rewriting this string as a regular (non-raw) string.


xdg:
Thanks.  Assigned [CXX-1010](https://jira.mongodb.org/browse/CXX-1010) for tracking.

xdg:
Cherry picked to master (and clang-formatted).  Thank you!

aseering:
Thanks for merging!  This will help our builds.

PR #523, opened 2016-09-08
CXX-778 Allow SSL configuration independently of the C driver

acmorrow:


src/mongocxx/config/postlude.hpp
```diff
@@ -1,3 +1,4 @@
+
```
xdg: White space?

xdg:
This seems fine, but I thought the original ticket was about CMake: refuse to build if SSL is requested and libmongoc doesn't support it.

src/mongocxx/config/postlude.hpp
acmorrow: Oops, fixed thanks.

acmorrow:
@xdg - We could probably do that, but I'm not sure it is worth it. The real point of the original ticket (which I wrote), was to make sure that we could easily test building the C++ driver with and without SSL without needing to build the C driver twice (one with and once without SSL), because at one point we had broken the non-ssl build but didn't notice since the underlying C driver had SSL.

acmorrow:
Ping?

xdg:
I don't object to what you've done. I do think that letting users detect that libmongoc doesn't have SSL support at build time is useful, but I don't think it's a priority.

PR #524, opened 2016-09-08
CXX-938 Declare polyfills in the inline ABI namespace

acmorrow:


xdg:
Weird travis failure, but otherwise LGTM

Machyne:
LGTM

PR #525, opened 2016-09-08
CXX-957 Honor BUILD_SHARED_LIBS to a sufficient degree

acmorrow:
Not an ideal solution: we still actually build the shared libraries, we just don't install them. Otherwise, we would need to conditionalize every block that references the shared library target with if (BUILD_SHARED_LIBS), which seems pretty bad. Not wanting the shared library is a pretty special case, so I'm OK with requiring them to build them, but avoiding installing.

### DIFF
--- a/src/bsoncxx/CMakeLists.txt
+++ b/src/bsoncxx/CMakeLists.txt
@@ -170,12 +170,14 @@ install(FILES
     COMPONENT dev
 )
 
-install(TARGETS
-    bsoncxx
-    RUNTIME DESTINATION bin COMPONENT runtime
-    LIBRARY DESTINATION lib COMPONENT runtime
-    ARCHIVE DESTINATION lib COMPONENT dev
-)
+if (BUILD_SHARED_LIBS)
+    install(TARGETS
+        bsoncxx
+        RUNTIME DESTINATION bin COMPONENT runtime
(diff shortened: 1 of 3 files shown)


xdg:
I'm fine with that approach.  I have no opinion on the cmake changes themselves.

src/bsoncxx/CMakeLists.txt
```diff
@@ -170,12 +170,14 @@ install(FILES
...
-    RUNTIME DESTINATION bin COMPONENT runtime
-    LIBRARY DESTINATION lib COMPONENT runtime
-    ARCHIVE DESTINATION lib COMPONENT dev
-)
+if (BUILD_SHARED_LIBS)
+    install(TARGETS
+        bsoncxx
+        RUNTIME DESTINATION bin COMPONENT runtime
+        LIBRARY DESTINATION lib COMPONENT runtime
+        ARCHIVE DESTINATION lib COMPONENT dev
```
acmorrow: See https://cmake.org/cmake/help/v3.2/command/install.html#installing-targets

For DLL platforms you need RUNTIME and ARCHIVE, for non-dll platforms you need LIBRARY.

acmorrow: Welcome to CMake :)

PR #527, opened 2016-09-10
CXX-1025 CXX-829 Move special builders to 16.04 and make the ASAN build green

acmorrow:


acmorrow:
https://evergreen.mongodb.com/version/57d46d303ff1227549008f00

acmorrow:
Gentle ping? This gets ASAN green again which is very nice, and also gets us off the Ubuntu 1410 builder, which is EOL, and will make the build team happy (they are trying to retire that distro).

.mci.yml
```diff
@@ -33,15 +33,15 @@ variables:
 
     ## cmake flag variables
     cmake_flags:
-        ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
-        osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers" -DBSONCXX_POLY_USE_BOOST=ON
-        asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_FLAGS="-fsanitize=address -O1 -fno-omit-frame-pointer -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers -Wno-error=maybe-uninitialized"
-        ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer -fsanitize-blacklist=$(pwd)/../etc/ubsan.blacklist -fno-sanitize-recover -Wall -Wextra -Wno-attributes -Werror -Wno-error=missing-field-initializers"
+        ubuntu_cmake_flags: &ubuntu_cmake_flags -DCMAKE_CXX_FLAGS="-Wall -Wextra -Werror -Wno-error=missing-field-initializers"
```
jrassi: Could you clarify the following:
- Do we want `-Wno-attributes`?
- Do we need `-Wno-error=missing-field-initializers`?
- Do the defaults for these change flags over the different compiler versions we're building with?

.mci.yml
```diff
@@ -33,15 +33,15 @@ variables:
...
+        osx_cmake_flags: &osx_cmake_flags -DCMAKE_CXX_FLAGS="-stdlib=libc++ -Wall -Wextra -Werror" -DBSONCXX_POLY_USE_BOOST=ON
+        asan_cmake_flags: &asan_cmake_flags -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0 -fsanitize=address -O1 -g -fno-omit-frame-pointer -Wall -Wextra -Werror"
```
jrassi: A couple of questions:
- What are the version numbers for the old and new versions of the gcc toolchain (the one we were using and the one we will be using, respectively)?
- From reading <https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html>, I'm guessing that you're setting `_GLIBCXX_USE_CXX11_ABI` to zero to fix linker errors when linking against a third-party library that the driver depends on.  Which library, out of curiosity, if you know offhand?  Relatedly, do most C++ library packages that ship with Ubuntu 16.04 link against the old ABI or the new ABI?

src/mongocxx/instance.cpp
```diff
@@ -96,7 +100,15 @@ class instance::impl {
         if (_user_logger) {
             libmongoc::log_set_handler(null_log_handler, nullptr);
         }
+
+        // Under ASAN, we don't want to clean up libmongoc, because it causes libraries to become
+        // unloaded, and then ASAN sees non-rooted allocations that it consideres leaks. These are
+        // also inscrutable, because the stack refers into an unloaded library, which ASAN can't
+        // report. Note that this only works if we have built mongoc so that it doesn't do its
+        // unfortunate automatic invocation of 'cleanup'.
+#if !__has_feature(address_sanitizer)
```
jrassi: I'm curious if there are any related C driver feature requests we could make, such that we would be able to get rid of this `#if` were they to be implemented.

.mci.yml
```diff
@@ -79,7 +79,7 @@ functions:
                 cd mongo-c-driver
                 git checkout 1.3.4
                 rm -rf /data/tmp/c-driver-install
-                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled --disable-extra-align ${cdriver_configure_flags}
+                ./autogen.sh --prefix="/data/tmp/c-driver-install" --enable-tests=no --enable-examples=no --with-libbson=bundled --disable-extra-align --enable-debug --enable-optimizations --disable-shm-counters --disable-static --disable-dependency-tracking --with-pic --disable-automatic-init-and-cleanup ${cdriver_configure_flags}
```
jrassi: Assuming this change is mostly to increase diagnosability of C driver issues during test failures?

jrassi:
LGTM.  Questions included for the pieces I'm less familiar with.

jrassi:
Actually, two more questions:
- Does this resolve CXX-829?  If so, please update the title of that ticket.
- How did you figure out to look at the C driver as a leak source?  Was it just a hunch, or was it obvious to look at each of the driver dependencies just based on the fact that the ASAN stack traces were obfuscated?

.mci.yml
acmorrow: - Re -Wno-attributes: I'm not sure why it was here, but, empirically, it seems not needed, so I removed it.
- We do seem to need the one about missing field initializers, since the 14.04 GCC seems to have a bug.
- I'm not sure about the defaults, but I'd expect no. Ideally, we would not build with the -Wno-error=missing-field-initializers, because I'd like to have that warning. But we seem to get false positives with it on 14.04 GCC.

.mci.yml
acmorrow: - From distrowatch, 14.10 used GCC 4.9.1, and 16.4 uses GCC 	5.3.1.
- Actually, I'm setting that because what we are doing here is using clang to build against libstdc++. The clang versions on these systems does not understand the attributes that newer libstdc++ uses to tag the ABI, so we need to force our view of the libstdc++ headers to be the old ABI.

acmorrow:
- It does resolve CXX-829, I've updated the description
- The stack traces were obfuscated, and said "unknown module" in the stack, which almost certainly meant that some library had been unloaded from the address space. The C++ library doesn't itself unload anything, and I doubted that libstdc++ did either. Our only other dependencies in the valgrind builder were the c driver, so, it was the obvious place to look.

src/mongocxx/instance.cpp
acmorrow: I don't think so. I think what is happening is that the cleanup routines for one of the libraries that the C driver is calling is itself dlclose'ing some other library. We aren't actually building the C driver under ASAN, either. Perhaps we should? But not in this ticket.

.mci.yml
acmorrow: Mostly, yes. Also to make sure that we control the behavior of cleanup programmatically. And a few flags to make the C driver build smaller and leaner.

jrassi:
Thanks for your answers!

.mci.yml
xdg: The disable-automatic-init-and-cleanup turns off mongoc's use of gcc's constructor/destructor attributes, so that cleanup happens only when *we* call it, not at some indeterminate time (possibly before we call it, leading to a runtime exception when we call it again).

PR #529, opened 2016-09-13
CXX-986 Improve find_and_modify error handling

acmorrow:
@xdg @jrassi @Machyne @bjori -

This improves the error handling after calling mongoc_collection_find_and_modify_with_opts. We need to handle two cases:

- The function returned false, but the 'reply' bson is not populated, in which case we simply throw an error wrapping the bson_error_t.
- The function returned false, but the reply bson IS populated, in which case we throw the raw server reply along with the bson_error_t.

I also added a regression test to verify the improvement, based on the repro in the original ticket. I was able to reproduce the original crash when running the test without the new error handling, and the fix does make the test pass and throw a meaningful exception.

### DIFF
--- a/src/mongocxx/collection.cpp
+++ b/src/mongocxx/collection.cpp
@@ -66,7 +66,12 @@ mongocxx::stdx::optional<bsoncxx::document::value> find_and_modify(
     const ::mongoc_find_and_modify_opts_t* opts) {
     mongocxx::libbson::scoped_bson_t bson_filter{filter};
     mongocxx::libbson::scoped_bson_t reply;
-    reply.flag_init();
+
+    // TODO: There is a bug in the 1.3 vintage C driver where collection_find_and_modify_with_opts
+    // does not always initialize its 'reply' parameter in the event of errors. So, explicitly init
+    // the underlying bson_t here.  This can be changed to ''flag_init' once the C driver minimum is
+    // bumped to or past 1.4.0.
+    reply.init();
 
     ::bson_error_t error;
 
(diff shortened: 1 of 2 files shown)


src/mongocxx/collection.cpp
```diff
@@ -74,9 +74,9 @@ mongocxx::stdx::optional<bsoncxx::document::value> find_and_modify(
         collection, bson_filter.bson(), opts, reply.bson(), &error);
 
     if (!r) {
-        auto gle = mongocxx::libmongoc::collection_get_last_error(collection);
-        mongocxx::throw_exception<mongocxx::write_exception>(
-            bsoncxx::helpers::value_from_bson_t(gle), error);
+        if (!reply.view().empty())
+            mongocxx::throw_exception<mongocxx::write_exception>(reply.steal(), error);
+        mongocxx::throw_exception<mongocxx::write_exception>(error);
```
Machyne: Should it be an operation exception in the case where the server doesn't send an error document? What else can go wrong here, do we know?

acmorrow: The documentation states that it throws write_exception if it fails, so I think this is correct, for now (until we refactor error handling).

Machyne: Got it.

Machyne:
LGTM

xdg:
LGTM

acmorrow:
@xdg This has been updated to work around a previously unknown bug in the 1.3.x series C driver.

PR #531, opened 2016-09-15
CXX-1006 clang_format.py shouldn't skip examples/ directory

jrassi:


jrassi:
Curious how the AppVeyor results are going to turn out.  Not sure about the direction of the slashes for `git ls-files` on Windows (do we build within Cygwin there?).

jrassi:
Closing in favor of #532 (same commits, but on top of master), after I realized that this ticket is marked 3.0.3, not 3.1-required.

PR #533, opened 2016-09-15
CXX-1009 pool ctor should take options::pool, not options::ssl

jrassi:


jrassi:
@acm @xdg, PTAL.

src/mongocxx/pool.cpp
```diff
@@ -39,20 +39,20 @@ void pool::_release(client* client) {
 
 pool::~pool() = default;
 
-pool::pool(const uri& uri, stdx::optional<options::ssl> ssl_options)
+pool::pool(const uri& uri, const options::client& options)
```
acmorrow: Is it worth creating an options::pool type? Perhaps in the future the pool will be more configurable.

acm:
could you guys stop cc'ing me (@acm) on stuff?  thanks,
andy

src/mongocxx/pool.cpp
xdg: I don't think we need to require `options::client`.  So if we go this way, we also need a constructor overload for just a URI.

Regarding `options::pool`, I have mixed feelings.  In Jesse's words: "defy augury".  Let's not add it until we actually know we need it.

src/mongocxx/pool.hpp
```diff
@@ -53,14 +53,13 @@ class MONGOCXX_API pool {
...
-    ///  Optional SSL options to use when connecting to the MongoDB deployment.
-    ///
-    /// @throws mongocxx::exception if SSL is enabled and ssl_options are
-    /// invalid, or if SSL is not enabled and ssl_options is engaged.
+    /// @param options
+    ///  Client options to use when connecting to the MongoDB deployment.
     ///
+    /// @throws mongocxx::exception if invalid options are provided (whether from the URI or
+    ///  provided client options).
     pool(const uri& mongodb_uri = mongocxx::uri(),
```
acmorrow: Since this ctor can legally be called with one argument, I think it should be explicit.

src/mongocxx/pool.cpp
acmorrow: I'm sympathetic to that argument. However, keep ABI in mind. If we use options::client here, and later need to move to options::pool, that would be an ABI break. However, if we start with options::pool, and later need to add behavior to options::pool, that at least has the *possibility* to not be an ABI break, depending on how we implement it.

xdg: Most configuration arguments are already in the URI (see [Connection Pool Options](https://docs.mongodb.com/manual/reference/connection-string/#connection-pool-options)).  We only need separate SSL options because there are things like paths that aren't supported by the URI format. I think the likelihood of similar things for pool options are low, so I'm willing to risk an ABI break if we can't find another way to do it.

acmorrow: Wouldn't that then argue for going the other way, and just have both mongocxx::client and mongocxx::pool take only an options::ssl object, and kill of options::client entirely?

xdg: I can more easily imagine some future client option that can't be expressed in a URI than I can imagine a pool option.  But I don't have a very strong opinion either way.  All I really want  is some degree of symmetry between client/pool construction.

jrassi:
Andy: oh, apologies!  @acmorrow goes by the "acm" handle on other systems, and I hadn't realized my mistake of mentioning the wrong user until just now.

jrassi:
@acmorrow, updated with your requested change.

src/mongocxx/pool.hpp
jrassi: Done.

src/mongocxx/pool.cpp
jrassi: I have no particularly strong opinion here.

xdg: Drew and I had a hallway conversation.  Originally, there was an intent of having every object type foo have a matching options::foo type.  If we want to stick with that, we could have `options::pool` and have that constructable from `options::client`.  Let's put this on our list to discuss tomorrow.

acm:
thanks!

acmorrow:
LGTM as an implementation of the "pool takes client options" approach.

@xdg do we have any useful insight back from C driver team re the future direction of options for pool vs client? I'd like us to hold this until we have an affirmative response one way or the other in case we need to revisit this decision and implement an options::pool type.

xdg:
Jesse affirmed that pool options will always be a superset of client options.  I think modeling as options::pool containing options::client is the right way to go, and we can later add pool-specific options like min/max size.  Once this work is done, we should open a CXX ticket to add that to options::pool.

acmorrow:
@xdg Thanks for getting that sorted out with C driver team. @jrassi, can you update this with a new options::pool type that wraps an options::client?

jrassi:
@acmorrow: done, PTAL.

Due to the fact that this new class logically extends an existing options class, its design deviates somewhat from the typical options class.  Notably:
- `options::pool` is implicitly convertible from `options::client`.
- `options::pool` does not have a setter for client options.
- `options::pool::client_opts()` returns a non-optional type.

Happy to change the design if you folks prefer.  The change I'm most tempted to make is to use inheritance; curious if you folks think that better fits the mongocxx design philosophy.

src/mongocxx/pool.cpp
```diff
@@ -39,20 +39,20 @@ void pool::_release(client* client) {
 
 pool::~pool() = default;
 
-pool::pool(const uri& uri, stdx::optional<options::ssl> ssl_options)
+pool::pool(const uri& uri, const options::pool& options)
```
xdg: Does this make options required? I think we should allow constructing a pool without options.

jrassi: Nope, `options` has a default argument of `options::pool()`.  See changes to `mongocxx/pool.hpp` in the same commit:

```
+    explicit pool(const uri& mongodb_uri = mongocxx::uri(),
+                  const options::pool& options = options::pool());
```

This makes the signature of the pool constructor more in line with the [signature of the client constructor](https://github.com/mongodb/mongo-cxx-driver/blob/r3.0.2/src/mongocxx/client.hpp#L69).

xdg: Right.  I wish GH could be taught to put .hpp files before .cpp files.  :-)

LGTM.

jrassi:
After today's conversations regarding deprecations, I'm left unsure whether or not I should restore the original pool constructor as a deprecated constructor.  Thoughts, @xdg?  Or should I push as-is?

xdg:
I think it can be pushed as is.  The pool is likely to be set up in just one place in the code so it's an easy change for end users.

PR #535, opened 2016-09-15
CXX-971 (part 1) Minor cleanup of options classes

jrassi:
Improved consistency of code style, added additional test coverage.

### DIFF
--- a/src/mongocxx/test/options/find_one_and_delete.cpp
+++ b/src/mongocxx/test/options/find_one_and_delete.cpp
@@ -30,10 +30,10 @@ TEST_CASE("find_one_and_delete", "[find_one_and_delete][option]") {
     options::find_one_and_delete opts{};
 
     std::chrono::milliseconds ms{400};
-    auto proj = document{} << "_id" << false << finalize;
+    auto projection = document{} << "_id" << false << finalize;
     auto sort = document{} << "x" << -1 << finalize;
 
     CHECK_OPTIONAL_ARGUMENT(opts, max_time, ms);
-    CHECK_OPTIONAL_ARGUMENT(opts, projection, proj.view());
+    CHECK_OPTIONAL_ARGUMENT(opts, projection, projection.view());
     CHECK_OPTIONAL_ARGUMENT(opts, sort, sort.view());
 }
--- a/src/mongocxx/test/options/index.cpp
+++ b/src/mongocxx/test/options/index.cpp
@@ -15,20 +15,24 @@
 #include "catch.hpp"
 #include "helpers.hpp"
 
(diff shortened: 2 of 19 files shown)


jrassi:
@acmorrow @xdg, PTAL.

acmorrow:
@jrassi LGTM. Note that changing the order of methods in a class can be an ABI break, if those methods are virtual. In this case, none are, so I see no problem. Just something to keep in mind for the future. The vtable layout is part of the ABI.

src/mongocxx/test/options/index.cpp
```diff
@@ -15,20 +15,24 @@
...
 
+#include <bsoncxx/builder/stream/document.hpp>
 #include <bsoncxx/stdx/make_unique.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/options/index.hpp>
 #include <mongocxx/stdx.hpp>
 
+using namespace bsoncxx::builder::stream;
 using namespace mongocxx;
-using namespace mongocxx::options;
```
xdg: Another example of micro change I'm not sure is worth it.

src/mongocxx/test/options/find_one_and_delete.cpp
```diff
@@ -30,10 +30,10 @@ TEST_CASE("find_one_and_delete", "[find_one_and_delete][option]") {
     options::find_one_and_delete opts{};
 
     std::chrono::milliseconds ms{400};
-    auto proj = document{} << "_id" << false << finalize;
+    auto projection = document{} << "_id" << false << finalize;
```
jrassi: Fair.  Note that I made this change since the variable name `projection` is used instead of `proj` everywhere else in the repository.

src/mongocxx/test/options/index.cpp
jrassi: Note five lines below where we qualify the `index` type as `options::index`, even though we had a `mongocxx::options` using-directive in scope.  I made this change out of the belief that it is poor style to have a using-directive in scope unless it is used for the entire scope.

jrassi:
I've split this into four commits and merged them into 3.1-dev, and am happy to take your suggestion of small commits going forward.

Regarding your "unnecessary changes" comment: I do agree that, in general, commits including changes unrelated to the work described in the commit message creates an undue context-switching burden on the reviewer.  However, I do think that janitorial changes like these by themselves are important though, even changes that just focus on improving consistency across files in the same directory.  Consider the fact that this pull request is groundwork for a bulk-addition of the collation option in every class in multiple directories.  It's much easier to perform such a bulk-addition if the author doesn't have to deal with inconsistent use of using-directives across files, members/methods in different orders between classes, etc.

PR #536, opened 2016-09-15
CXX-971 (part 2) Add collation to options::find

jrassi:
On top of #535.  Supersedes https://github.com/jrassi/mongo-cxx-driver/pull/1.

@acmorrow, @xdg: PTAL.  Also, let me know also if you'd like me to use some other workflow for submitting reviews on top of work that hasn't been pushed yet (like stacking many commits in a single pull request, for example).

### DIFF
--- a/src/mongocxx/options/find.hpp
+++ b/src/mongocxx/options/find.hpp
@@ -75,6 +75,28 @@ class MONGOCXX_API find {
     const stdx::optional<std::int32_t>& batch_size() const;
 
     ///
+    /// Sets the collation for this operation.
+    ///
+    /// @param collation
+    ///   The new collation.
+    ///
+    /// @see
+    ///   https://docs.mongodb.com/manual/release-notes/3.3-dev-series-collation/#collation-option
+    ///
+    find& collation(bsoncxx::document::view_or_value collation);
+
+    ///
+    /// Retrieves the current collation for this operation.
+    ///
+    /// @return
+    ///   The current collation.
+    ///
+    /// @see
+    ///   https://docs.mongodb.com/manual/release-notes/3.3-dev-series-collation/#collation-option
+    ///
+    const stdx::optional<bsoncxx::document::view_or_value>& collation() const;
+
+    ///
     /// Attaches a comment to the query. If $comment also exists in the modifiers document then
     /// the comment field overwrites $comment.
     ///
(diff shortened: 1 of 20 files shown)


src/mongocxx/options/find.hpp
```diff
@@ -308,6 +330,7 @@ class MONGOCXX_API find {
    private:
     stdx::optional<bool> _allow_partial_results;
     stdx::optional<std::int32_t> _batch_size;
+    stdx::optional<bsoncxx::document::view_or_value> _collation;
```
acmorrow: So, note that while the addition of the above *methods* wouldn't constitute an ABI break, adding this field would. We aren't promising ABI yet, but, I feel it is important to identify places were it would happen so we gain familiarity. I suspect that at some point we are going to want to give all options an "extension" parameter slot, which is a unique_ptr to some opaque internal extensions object. Or we may find that we need to pimpl all options fields? I'm not sure.

src/mongocxx/options/find.hpp
```diff
@@ -75,6 +75,28 @@ class MONGOCXX_API find {
...
+    find& collation(bsoncxx::document::view_or_value collation);
+
+    ///
+    /// Retrieves the current collation for this operation.
+    ///
+    /// @return
+    ///   The current collation.
+    ///
+    /// @see
+    ///   https://docs.mongodb.com/manual/release-notes/3.3-dev-series-collation/#collation-option
```
acmorrow: Do we need a ticket to go through and update all references to 'dev' docs before release?

xdg: Already there: [CXX-1039](https://jira.mongodb.org/browse/CXX-1039)

src/mongocxx/options/find.hpp
xdg: I'm more inclined to pimpl all options fields because we can expect new ones from the server over time and having half pimpl and half not seems inelegant.  That might be another thing for 2.0, however.

acmorrow: Thats fair. Lets file it as such. I'm not sure why we didn't do it originally, TBH. Maybe we were just tired of pimpl'ing things at that point.

jrassi: Filed https://jira.mongodb.org/browse/CXX-1046.

jrassi:
Merged into 3.1-dev as b73345e.

PR #537, opened 2016-09-16
CXX-971 (part 3) Minor cleanup of model classes

jrassi:
Improved consistency of code style, added additional test coverage.


jrassi:
@acmorrow @xdg, PTAL.

jrassi:
Similarly to #535, I've split this into four commits and merged them into 3.1-dev.

In regard to your "Assuming this all squishes down later" comment: no, I'll keep them split.  Doing so will allow later readers (and blamers) to reap the same readability benefits that the work's official reviewers do.  Also, keeping them split is useful for accelerating the diagnosis of any issues that may come up with this work (like with Evergreen's bisect feature or git bisect).

PR #538, opened 2016-09-16
CXX-971 (part 4) Add collation to model::update_one

jrassi:


jrassi:
@acmorrow @xdg, PTAL.

src/mongocxx/model/update_one.hpp
```diff
@@ -79,6 +101,7 @@ class MONGOCXX_API update_one {
     bsoncxx::document::view_or_value _filter;
     bsoncxx::document::view_or_value _update;
 
+    stdx::optional<bsoncxx::document::view_or_value> _collation;
```
jrassi: AFAICT, the original author of the model classes sequenced the public methods and private members into an alphabetical list of required options, followed by an alphabetical list of optional options.  The blank at 103 was to separate the two sections for the private members.

I'd be happy to remove the blank at 103 and re-sort the members/methods in proper alphabetical order, without regard to whether they are required/optional.  Would you prefer this?

xdg: Let's leave it as is.  LGTM

jrassi:
Merged in d6db05283e26958d190d4df579bdc63254b45fc3.

PR #539, opened 2016-09-21
CXX-1000 Migrate documentation to GitHub Pages

xdg:
NOTE: the base URLs for these are still the staging site.  They need to
be changed to the final deployment URL.

The staging site can be viewed at https://mongodb.github.io/docs-cxx/

This commit creates a docs/ directory containing a general set
of documentation files for use with the hugo static site generator.
The site generated from these files will be deployed to GitHub Pages
and supersede the current wiki.

It also provides CMake targets to generate and deploy both this static
manual and the doxygen generated API documentation sites for all legacy and
mongocxx releases.

The doxygen configurations are generated dynamically to ensure that
documentation from old releases uses the same formatting and "front page"
content as current releases.

The site deployment targets do not create symlinks, which are necessary for
proper "permalinks" to the latest documentation.  These must be added
manually:

* /api/content     -> /api/mongocxx-3.0.x (for latest x in v3 series)
* /api/mongocxx-v3 -> /api/mongocxx-3.0.x (for latest x in v3 series)
* /api/legacy-v1   -> /api/legacy-1.1.x (for latest x in v1 series)


xdg:
I don't expect actual "code review" of these files.  Please look at the staging site and react to that -- particularly the structure and ease of navigation.

Don't worry about inaccurate legacy files as these have been migrated from the wiki with minimal revision. The important part of this ticket is "migrate to GitHub pages" and we can open tickets to fix content later (or just do minor fixes as we find them).

xdg:
Something that already jumps out at me is to have a link to the releases page with all the change notes. And the mongocxx installation instructions need to have 3.0.1 replaced with 3.0.2.

PR #540, opened 2016-09-21
CXX-1044 collection::find should handle partial/no_timeout options

jrassi:


jrassi:
@acmorrow @xdg, PTAL.

acmorrow:
Code LGTM. Per discussion, I propose that we rebase this against master and commit it there, as this is a real bug in the shipping version of the driver.

xdg:
LGTM.  I don't know if we'll ship a 3.0.3 before 3.1 but agree this should go against master and be merged/cherry-picked to 3.1-dev.

jrassi:
Merged into master as 809aa9a.  I'll sync up 3.1-dev in the near future.

PR #541, opened 2016-09-22
CXX-1045 collection::count() serializes options::count::hint incorrectly

jrassi:
First included commit is an ABI-breaking/API-breaking small refactor.  Second commit is the bugfix itself.

@acmorrow @xdg, PTAL.

### DIFF
--- a/src/mongocxx/hint.cpp
+++ b/src/mongocxx/hint.cpp
@@ -27,6 +27,14 @@ hint::hint(bsoncxx::document::view_or_value index) : _index_doc(std::move(index)
 hint::hint(bsoncxx::string::view_or_value index) : _index_string(std::move(index)) {
 }
 
+bsoncxx::types::value hint::to_value() const {
+    if (_index_doc) {
+        return bsoncxx::types::value{bsoncxx::types::b_document{_index_doc->view()}};
+    }
+
+    return bsoncxx::types::value{bsoncxx::types::b_utf8{*_index_string}};
+}
+
 bsoncxx::document::value hint::to_document() const {
     auto doc = bsoncxx::builder::stream::document{};
 
--- a/src/mongocxx/test/hint.cpp
+++ b/src/mongocxx/test/hint.cpp
@@ -32,16 +32,10 @@ TEST_CASE("Hint", "[hint]") {
         std::string index_name = "a_1";
         hint index_hint{index_name};
 
-        SECTION("Can be applied to a query") {
-            document::value filter = builder::stream::document{}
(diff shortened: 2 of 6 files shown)


xdg:
Why is the ABI/API break necessary?  I think you need to elaborate more. Think about the 7th rule of [great commit messages](http://chris.beams.io/posts/git-commit/): _Use the body to explain what and why vs. how_

Refactoring and breaking changes probably deserve some explanation.

jrassi:
Updated with expanded commit message bodies, and rebased against master.

xdg:
I really wish we had a stdx::variant, because we could just replace mongocxx::hint entirely.

Unrelated to this PR, I wonder if there's any point trying to compare strings and documents using the key-name-mangling algorithm.

For v4, I wonder if the member should just become a boxed element, rather than two optionals.  Or if mongocxx::hint should be replaced with a boxed element. Thoughts?

src/mongocxx/hint.cpp
```diff
@@ -27,18 +27,12 @@ hint::hint(bsoncxx::document::view_or_value index) : _index_doc(std::move(index)
 hint::hint(bsoncxx::string::view_or_value index) : _index_string(std::move(index)) {
 }
 
-bsoncxx::document::value hint::to_document() const {
```
jrassi: Done, see latest commit eb0ea89.

src/mongocxx/test/hint.cpp
```diff
@@ -52,6 +52,22 @@ TEST_CASE("Hint", "[hint]") {
             auto index_doc = builder::stream::document{} << "a" << 1 << builder::stream::finalize;
             REQUIRE(index_hint != index_doc);
         }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
jrassi: Alternatively, I could keep the tests and remove the MONGOCXX_DEPRECATED declaration until said ticket is resolved.

xdg: let's keep the tests and remove MONGOCXX_DEPRECATED and have a ticket for doing that later after we figure out how to suppress deprecation warnings in testing.

jrassi: Done, see latest commit 54ea3f23e27049f412d0f79f50b6055092bf88f2.

jrassi: Filed https://jira.mongodb.org/browse/CXX-1056 for suppressing deprecation warnings in testing.

PR #542, opened 2016-09-27
CXX-1051 Add new options to options::find

jrassi:
New options added:
- max
- max_scan
- min
- return_key
- show_record_id
- snapshot

These will replace the "modifiers" option.

### DIFF
--- a/src/mongocxx/options/find.cpp
+++ b/src/mongocxx/options/find.cpp
@@ -57,13 +57,28 @@ find& find::limit(std::int32_t limit) {
     return *this;
 }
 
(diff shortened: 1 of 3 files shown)


jrassi:
@acmorrow @xdg, PTAL.

This implements changes to the CRUD API that are currently in review (https://github.com/mongodb/specifications/pull/102), and is groundwork for upgrading the C++ driver CRUD API to use the C driver 1.5 CRUD API.

I will submit separately the work to hook these options up to collection::find(), and to deprecate options::find::modifiers().

PR #544, opened 2016-09-28
CXX-1052 (part 1) Add options::find public internal method to convert modifiers

jrassi:
@acmorrow @xdg, PTAL.

This is groundwork for CXX-1052 ("Port CRUD methods to use C driver 1.5 CRUD API").  Per discussion from last week, the C driver 1.5 CRUD API does not accept "modifiers", but we have decided to continue to support C++ driver users with existing applications that make use of the "modifiers" options.  Thus, we need to implement logic to convert each legacy modifier to its new respective find option, before we can upgrade to use the new C driver API.  This logic will be invoked in `collection::find()`, before the call to `mongoc_collection_find_with_opts()` is made.

This is a quick work-in-progress patch that I'm submitting for feedback before progressing further.  If we agree this the direction we want to go in, I'll re-submit with a unit test suite.

Note that this builds on top of #542.


src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
 
+find& find::convert_all_modifiers() {
+    if (!_modifiers) {
+        return *this;
+    }
+
+    for (auto&& ele : _modifiers->view()) {
+        if (ele.key() == stdx::string_view("$comment")) {
+            if (ele.type() != bsoncxx::type::k_utf8) {
+                throw exception{error_code::k_invalid_parameter,
```
jrassi: Is there a better exception type to use here?

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
+                                        "non-negative value required for $maxScan"};
+                    }
+                    if (!_max_scan) {
+                        _max_scan = max_scan_int32;
+                    }
+                    break;
+                case bsoncxx::type::k_int64:
+                    max_scan_int64 = ele.get_int64().value;
+                    if (max_scan_int64 < 0 ||
+                        max_scan_int64 > std::numeric_limits<std::int32_t>::max()) {
```
jrassi: I couldn't find any utility functions in the repository for supporting safe numeric conversions.  Should I start something new?

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
+            if (ele.type() != bsoncxx::type::k_bool) {
+                throw exception{error_code::k_invalid_parameter,
+                                "bool type required for $snapshot"};
+            }
+            if (!_snapshot) {
+                _snapshot = ele.get_bool().value;
+            }
+        } else {
+            throw exception{error_code::k_invalid_parameter,
+                            std::string{"unrecognized key in modifiers: "} + ele.key().to_string()};
```
jrassi: We don't have any stream-style string building classes, do we?

src/mongocxx/options/find.hpp
```diff
@@ -455,6 +455,16 @@ class MONGOCXX_API find {
     ///
     const stdx::optional<bsoncxx::document::view_or_value>& sort() const;
 
+    ///
+    /// Modifies the object to convert all modifiers into their respective top-level options. For
+    /// internal use only.
+    ///
+    /// @throws
+    ///   mongocxx::exception if any modifiers are invalid, or if the unsupported "$query" or
```
jrassi: I've decided here to break applications that use `$query` or `$explain` in modifiers.  Is this acceptable?

src/mongocxx/options/find.hpp
```diff
@@ -455,6 +455,16 @@ class MONGOCXX_API find {
...
+    ///   "$explain" modifiers are set.
+    ///
+    MONGOCXX_PRIVATE find& convert_all_modifiers();
```
jrassi: Happy to receive opinions on both of the following:
1. Should this mutate an options::find object, or should it return a copy of the options::find object it's invoked on?
2. Should this be a public non-static class method, or a public static class method, or a free function?  It doesn't particularly need access to any internal state.

src/mongocxx/options/find.cpp
jrassi: I've posted a fixup commit that uses `mongocxx::logic_error` here instead of `mongocxx::exception`.

src/mongocxx/options/find.hpp
xdg: Setting aside issues of ownership, I think it should return a copy.  We're passing options via `const`, so find() can't call this to mutate the options – it needs a clean copy anyway.

src/mongocxx/options/find.cpp
xdg: OK

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
     return _read_preference;
 }
 
+find& find::convert_all_modifiers() {
+    if (!_modifiers) {
+        return *this;
```
xdg: If we change API to return a copy, I think we need have to return a copy here, too, for consistenecy.

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
+    }
+
+    for (auto&& ele : _modifiers->view()) {
```
xdg: I don't like multi-page functions because they impede readability.  How about breaking up type-specific code to separate functions. E.g.:

```
if (ele.key() == sdx::string_view("$comment")) {
    _convert_comment(ele);
} else if (ele.key() == stdx::string_view("$hint")) {
    _convert_hint(ele);
}...
```

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
+        if (ele.key() == stdx::string_view("$comment")) {
```
xdg: Would `ele.key() == "$comment"` work, since string_view is supposed to be implicitly constructible from a `const char *`?

src/mongocxx/options/find.cpp
xdg: Yes, I think so.

src/mongocxx/options/find.cpp
```diff
@@ -211,6 +214,179 @@ const stdx::optional<class read_preference>& find::read_preference() const {
...
+                    break;
+                default:
+                    throw exception{error_code::k_invalid_parameter,
+                                    "numberic type required for $maxScan"};
+            }
+        } else if (ele.key() == stdx::string_view("$maxTimeMS")) {
+            std::int32_t max_time_ms_int32;
+            std::int64_t max_time_ms_int64;
+            double max_time_ms_double;
+            switch (ele.type()) {
```
xdg: Same issue here and having such long code duplicated makes understanding what's going on harder.

src/mongocxx/options/find.hpp
xdg: Yes.

src/mongocxx/options/find.hpp
jrassi: Done, see commit 3 in set: 63533cc829b96853f914eeba6e3ceea53cfe8644.

src/mongocxx/options/find.cpp
jrassi: Yep.

src/mongocxx/options/find.cpp
jrassi: Done, see commit 3 in set: 63533cc829b96853f914eeba6e3ceea53cfe8644.

src/mongocxx/options/find.cpp
jrassi: Nope, and I'm not sure why not.

```
/Users/rassi/work/mongo-cxx-driver/src/mongocxx/options/find.cpp:395:23: error: invalid operands to
      binary expression ('stdx::string_view' (aka 'basic_string_view<char>') and 'const char *')
        if (ele.key() == "$comment") {
            ~~~~~~~~~ ^  ~~~~~~~~~~
```

jrassi:
Posted 63533cc829b96853f914eeba6e3ceea53cfe8644, which addresses comments and adds a suite of unit tests.

src/mongocxx/test/options/find.cpp
```diff
@@ -60,3 +60,213 @@ TEST_CASE("find", "[find][option]") {
...
+    REQUIRE_THROWS(options::find{}
+                       .modifiers(document{} << "$explain" << true << finalize)
+                       .convert_all_modifiers());
+}
+
+TEST_CASE("options::find::convert_all_modifiers() with $hint", "[find][option]") {
+    instance::current();
+    options::find find_opts;
+
+    find_opts = options::find{}
```
xdg: Should have a commen like `// Test hint as string.`

src/mongocxx/test/options/find.cpp
```diff
@@ -60,3 +60,213 @@ TEST_CASE("find", "[find][option]") {
...
+                    .modifiers(document{} << "$hint"
+                                          << "index" << finalize)
+                    .convert_all_modifiers();
+    REQUIRE(!find_opts.modifiers());
+    REQUIRE(find_opts.hint());
+    REQUIRE(*find_opts.hint() == "index");
+
+    find_opts = options::find{}
```
xdg: Add commend: `// Test hint as document.`

jrassi:
PTAL at 5e9394e, which adds `SECTION` blocks to each test case added.  I thought this was a reasonable idea, since almost all of these `TEST_CASE` blocks contains up to four actual "test cases".

Unrelatedly, I also updated the exception tests to use `REQUIRE_THROWS_AS`.

src/mongocxx/options/private/rewriter.cpp
```diff
@@ -0,0 +1,213 @@
...
+                std::string{"unrecognized key in modifiers: "} + ele.key().to_string()};
+        }
+    }
+
+    // Ideally, we'd like to return an options::find object with the 'modifiers' option not engaged,
+    // but the current options::find API does not provide a mechanism to clear an existing option.
+    // While not ideal, it's sufficient for our purposes to set 'modifiers' to an empty view here
+    // (the alternative would be to build up a new options object from scratch, but this would be
+    // verbose and error-prone).
+    converted_options.modifiers(bsoncxx::document::view{});
```
jrassi: The only functional changes in this commit are:
- This line here, which previously read `converted_options._modifiers = stdx::nullopt;` (options/find.cpp:392).
- Updating the test assertions to reflect this change.

jrassi:
Posted a revert of fixup commit 1e6671f, and opened #560 for consideration of those suggested improvements.

Merged this branch into 3.1-dev with https://github.com/mongodb/mongo-cxx-driver/commit/02be4fd34baf53772120fd9a0c7d819fb7587065.

PR #545, opened 2016-09-29
CXX-1052 (part 2) Port collection::find() to use collection_find_with_opts()

jrassi:
@acmorrow @xdg, PTAL.

This ports collection::find() from using [libmongoc::collection_find()](https://s3.amazonaws.com/mciuploads/mongo-c-driver/docs/latest/mongoc_collection_find.html) to using the new-in-1.5 API function [libmongoc::collection_find_with_opts()](https://s3.amazonaws.com/mciuploads/mongo-c-driver/docs/latest/mongoc_collection_find_with_opts.html).

This commit also ports the existing mock suite to use the new API.  I've noticed that the existing coverage is rather light in this area, so I'll be following this up with expanded test coverage.

On top of #541, #542, #544.


src/mongocxx/collection.cpp
```diff
@@ -191,64 +191,105 @@ stdx::optional<result::bulk_write> collection::bulk_write(const class bulk_write
...
-    scoped_bson_t filter_bson;
-    scoped_bson_t projection{options.projection()};
+    builder::stream::document opts_builder;
+    scoped_bson_t filter_bson{filter};
+    scoped_bson_t opts_bson;
 
-    filter_builder << "$query" << bsoncxx::types::b_document{filter};
+    // libmongoc::collection_find_with_opts does not support the legacy "modifiers" options, so we
+    // must copy the options struct and convert all of the modifiers options to their modern
+    // equivalents.
```
xdg: I suggest pulling all the filter/opt builder code into a separate private function so that `find` itself is shorter and focused on sending the filter and getting the cursor ready.   E.g. pass filter and options and a reference to a filter_builder and opts_builder and let the function modify the builders as appropriate.

src/mongocxx/collection.cpp
```diff
@@ -191,64 +191,105 @@ stdx::optional<result::bulk_write> collection::bulk_write(const class bulk_write
...
+    options::find opts(options);
+    opts.convert_all_modifiers();
```
xdg: I think we discussed having this return a copy rather than modify.

jrassi: Rebased against commit on #544.

jrassi:
Rebased against latest #544 and posted an additional patch addressing feedback.  PTAL.

src/mongocxx/collection.cpp
```diff
@@ -189,66 +189,111 @@ stdx::optional<result::bulk_write> collection::bulk_write(const class bulk_write
...
+}  // namespace
+
+cursor collection::find(view_or_value filter, const options::find& options) {
+    // libmongoc::collection_find_with_opts does not support the legacy "modifiers" options, so we
+    // must copy the options struct and convert all of the modifiers options to their modern
+    // equivalents.
+    options::find options_converted = options.convert_all_modifiers();
+
+    scoped_bson_t filter_bson{std::move(filter)};
+    scoped_bson_t options_bson{build_find_options_document(options_converted)};
```
xdg: If there is a read preference, it winds up in options_bson and it shouldn't.  I think you need to pull read preferences out before constructing options_bson.

src/mongocxx/collection.cpp
```diff
@@ -189,66 +189,111 @@ stdx::optional<result::bulk_write> collection::bulk_write(const class bulk_write
...
+
+    const mongoc_read_prefs_t* rp_ptr = NULL;
+    if (options_converted.read_preference()) {
+        rp_ptr = options_converted.read_preference()->_impl->read_preference_t;
     }
```
xdg: I'd suggest moving this  up above 285 and clearing read_preference after extracting the rp_ptr.

src/mongocxx/collection.cpp
jrassi: Mmm, I think that's not correct.  `options_bson` is generated from `build_find_options_document()`, and `build_find_options_document()` doesn't add a read preference because it's not a find command option.

Would it be clearer if I renamed `build_find_options_document()` to `build_find_command_options_document()`?

src/mongocxx/collection.cpp
jrassi: See reply above.

jrassi:
Replied to comments, no new commits.

src/mongocxx/collection.cpp
xdg: I see.  I must have missed the nesting.  Let's leave it as is.

xdg:
LGTM

jrassi:
Merged in https://github.com/mongodb/mongo-cxx-driver/commit/7349684fc9c274421b17bbc5ebfe82b027b49f1b.

PR #547, opened 2016-09-30
CXX-859 Implement Decimal128 support

xdg:
@jrassi @acmorrow PTAL


src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,100 @@
...
+    friend BSONCXX_API bool BSONCXX_CALL operator==(const decimal128& lhs, const decimal128& rhs);
+    friend BSONCXX_API bool BSONCXX_CALL operator!=(const decimal128& lhs, const decimal128& rhs);
+    ///
+    /// @}
+    ///
+
+    ///
+    /// Accessor for high 64 bits
+    ///
+    uint64_t high() const {
```
jrassi: Do you have a sense for what guidelines to use when deciding whether or not to inline methods in `BSONCXX_API`/`MONGOCXX_API` classes?  I'm not really sure yet.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+///
+/// Represents an IEEE 754-2008 Decimal128 value in a platform-independent way.
+///
+class BSONCXX_API decimal128 {
+   public:
+    ///
+    /// Constructs a Decimal128 value representing zero.
+    ///
+    decimal128() : _high(0), _low(0){};
```
xdg: Done.  Lack of space before block seems to be a clang-format issue.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    ///
+    /// Constructs a Decimal128 from high and low 64-bit big-endian parts.
+    ///
+    /// @param high
+    ///   High bits
```
xdg: We have no consistent style on this.  The way it renders, it might make more sense to write it as a sentence starting with the param name.  I'll try that and then putting periods at the end is more sensible.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,100 @@
...
+    /// @param low
+    ///   Low bits
+    ///
+    decimal128(const uint64_t h, const uint64_t l) noexcept : _high(h), _low(l){};
```
xdg: Nixed const.  Here, I expanded the variable names from single-letter because it leads to better documentation with doxygen.  Elsewhere, I'm a firm believer in [short variables in small scopes]{https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-name-length}.

The idea behind noexcept here is that it can't throw an exception.  That's different than initialization from string, where the string might be bad.  I think the OID ones lacking noexcept is either because they are documented as possibly throwing or else an omission.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,100 @@
...
+    ///
+    /// Constructs a Decimal128 from a string.
+    ///
+    /// @param str
+    ///   A string representation of a decimal number
+    ///
+    /// @throws bsoncxx::exception if the string isn't a valid Decimal128
+    /// representation.
+    ///
+    explicit decimal128(const stdx::string_view& str);
```
xdg: Done.

src/bsoncxx/decimal128.hpp
xdg: I think for most accessors, they probably ought to be inline if we're confident the member structure isn't changing.  Generally, though, from a lot of the reading I've done, I'm opposed to inlining either on premature optimization grounds or on the grounds that it makes changing the internal implementation harder.

I'll change these to add BSONCXX_INLINE as I think I'm pretty confident in the design.

src/bsoncxx/types.hpp
```diff
@@ -599,6 +600,42 @@ BSONCXX_INLINE bool operator==(const b_int64& lhs, const b_int64& rhs) {
 }
 
 ///
+/// A BSON Decimal128 value.
+///
+struct BSONCXX_API b_decimal128 {
+    static constexpr auto type_id = type::k_decimal128;
+
+    bsoncxx::decimal128 value;
```
xdg: done

src/bsoncxx/types.hpp
```diff
@@ -599,6 +600,42 @@ BSONCXX_INLINE bool operator==(const b_int64& lhs, const b_int64& rhs) {
...
+              typename std::enable_if<
+                  !std::is_same<b_decimal128, typename std::decay<T>::type>::value, int>::type = 0>
+    BSONCXX_INLINE explicit b_decimal128(T&& t)
+        : value(std::forward<T>(t)) {
+    }
+
+    ///
+    /// Conversion operator giving a b_decimal128 as a std::string
+    ///
+    std::string to_string();
```
xdg: Thanks.  That was left over from an earlier design before I decided we needed a standalone bsoncxx::decimal128 class.

src/bsoncxx/types/value.hpp
```diff
@@ -127,6 +127,11 @@ class BSONCXX_API value {
     explicit value(b_int64) noexcept;
 
     ///
+    /// Construct a value from a BSON Decimal128
```
xdg: Done

xdg:
Rebased to pick up private header suffix changes.  Added CXX-1027 on docs.

examples/bsoncxx/decimal128.cpp
```diff
@@ -0,0 +1,49 @@
+// Copyright 2015 MongoDB Inc.
```
acmorrow: We don't do that.

xdg: New file, so updated to 2016.

src/bsoncxx/builder/core.hpp
```diff
@@ -263,6 +268,11 @@ class BSONCXX_API core {
     void append(const oid& value);
 
     ///
+    /// Append a decimal128 object as a BSON Decimal128.
+    ///
+    void append(const decimal128& value);
```
xdg: For whatever reason *all* the appends are const ref.  Even b_int64, which is literally just an in64_t.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,98 @@
...
+   public:
+    ///
+    /// Constructs a BSON Decimal128 value representing zero.
+    ///
+    decimal128() = default;
+
+    ///
+    /// Constructs a BSON Decimal128 from high and low 64-bit big-endian parts.
+    ///
+    /// @param high are the high 64-bits.
```
xdg: I guess the rendering is more like a DT/DD pair.  So name followed by some sentence like thing is fine.  I personally tend to dislike extremely short sentence fragments with periods.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,98 @@
...
+    /// @param low are the low 64-bits.
+    ///
+    decimal128(uint64_t high, uint64_t low) noexcept : _high(high), _low(low) {
```
xdg: According to [this SO post](https://stackoverflow.com/questions/9180164/implicit-generated-members-and-noexcept), the generated default should be noexcept automatically.

src/bsoncxx/decimal128.hpp
```diff
@@ -0,0 +1,98 @@
...
+    /// @relates decimal128
+    ///
+    friend BSONCXX_API bool BSONCXX_CALL operator==(const decimal128& lhs, const decimal128& rhs);
+    friend BSONCXX_API bool BSONCXX_CALL operator!=(const decimal128& lhs, const decimal128& rhs);
+    ///
+    /// @}
+    ///
+
+    ///
+    /// Accessor for high 64 bits
```
xdg: done.

src/bsoncxx/decimal128.hpp
jrassi: Ah!  Yep, that's consistent with my further reading on cppreference.  I must have either forgotten that the implicitly-declared default constructor was considered to be `noexcept(true)`, or that explicitly-defaulted default constructors are considered to have the same exception specifier as their equivalent implicitly-declared default constructors.

src/bsoncxx/builder/core.hpp
jrassi: That's close to true, but not quite true.  All of the append() overloads taking a b_xxx type pass a const reference (probably just to be consistent with each other), though the append() overloads taking small raw types pass by value:

```
240     ///
241     /// Append a native boolean as a BSON boolean.
242     ///
243     void append(bool value);
244
245     ///
246     /// Append a native double as a BSON double.
247     ///
248     void append(double value);
249
250     ///
251     /// Append a native int32_t as a BSON 32-bit signed integer.
252     ///
253     void append(std::int32_t value);
254
255     ///
256     /// Append a native int64_t as a BSON 64-bit signed integer.
257     ///
258     void append(std::int64_t value);
```

xdg: Right!  I thought you were talking about the b_ version.  I'll amend the plain decimal128 one.  (This is an example of the problem too many types with similar names!)

acmorrow:
You beat me by about 2 minutes with the merge. I had just pulled this up. I took a quick pass through and this looks good to me as well. Sorry for the late review.

PR #550, opened 2016-10-01
CXX-999 Support read-only views

jrassi:
@acmorrow @xdg, PTAL.

I modeled the design of database::create_view() off of the approach taken for [JAVA-2286](https://jira.mongodb.org/browse/JAVA-2286).  Note that I couldn't find record of any other driver having a collMod helper, so I just went ahead and added the view-related options to options::modify_collection() (though I also filed [CXX-1058](https://jira.mongodb.org/browse/CXX-1058) to have us consider deprecating database::modify_collection() outright).

Follow-up work:
- Collation support for views will be addressed in [CXX-971](https://jira.mongodb.org/browse/CXX-971).
- Updating the 3.3 development series documentation link will be addressed in [CXX-1061](https://jira.mongodb.org/browse/CXX-1061).

### DIFF
--- a/src/mongocxx/options/create_view.cpp
+++ b/src/mongocxx/options/create_view.cpp
@@ -1,0 +1,43 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mongocxx/options/create_view.hpp>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/types.hpp>
+
+#include <mongocxx/config/private/prelude.hh>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+namespace options {
+
+create_view& create_view::pipeline(class pipeline pipeline) {
+    _pipeline = std::move(pipeline);
+    return *this;
+}
+
+bsoncxx::document::value create_view::to_document() const {
+    auto doc = bsoncxx::builder::stream::document{};
+
+    if (_pipeline) {
+        doc << "pipeline" << _pipeline->view_array();
+    }
+
+    return doc.extract();
+}
+
+}  // namespace options
+MONGOCXX_INLINE_NAMESPACE_END
+}  // namespace mongocxx
--- a/src/mongocxx/pipeline.hpp
+++ b/src/mongocxx/pipeline.hpp
@@ -18,6 +18,7 @@
 #include <string>
 #include <memory>
 
+#include <bsoncxx/array/view.hpp>
 #include <bsoncxx/document/view.hpp>
 #include <bsoncxx/document/view_or_value.hpp>
 
@@ -174,9 +175,16 @@ class MONGOCXX_API pipeline {
     pipeline& unwind(std::string field_name);
 
     ///
+    /// @return A view of the underlying BSON array this pipeline represents.
+    ///
+    bsoncxx::array::view view_array() const;
+
+    ///
     /// @return A view of the underlying BSON document this pipeline represents.
     ///
-    bsoncxx::document::view view() const;
+    /// @deprecated The view_array() method should be used instead of this method.
+    ///
+    MONGOCXX_DEPRECATED bsoncxx::document::view view() const;
 
    private:
     friend class collection;
--- a/src/mongocxx/test/database.cpp
+++ b/src/mongocxx/test/database.cpp
@@ -33,6 +33,18 @@ using bsoncxx::builder::stream::document;
 using bsoncxx::builder::stream::finalize;
 using bsoncxx::builder::stream::open_document;
 
+namespace {
+
+int get_max_wire_version(const client& mongo_client) {
+    auto reply = mongo_client["admin"].run_command(document{} << "isMaster" << 1 << finalize);
+    auto max_wire_version = reply.view()["maxWireVersion"];
+    REQUIRE(max_wire_version);
+    REQUIRE(max_wire_version.type() == bsoncxx::type::k_int32);
+    return max_wire_version.get_int32().value;
+}
+
+}  // namespace
+
 TEST_CASE("A default constructed database is false-ish", "[database]") {
     instance::current();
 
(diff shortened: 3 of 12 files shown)


jrassi:
The AppVeyor and Travis failures on this pull request are legit.  I will create fixup commits on Monday or Tuesday and push them to this branch.

jrassi:
Fixup commits pushed to address the AppVeyor and Travis failures.

I am still getting to the bottom of why the MSVC compile was failing, but the story so far is that declaring a `create_view` class with the `__declspec(dllexport)` attribute and a `boost::optional<pipeline>` member caused the compiler to attempt to reference `pipeline`'s deleted copy assignment operator.  Explicitly deleting `create_view`'s copy constructor and copy assignment operators resolves the issue, and I'm still investigating as to why.

xdg:
My initial inclination is that adding `create_view` and a corresponding options class is the wrong direction.  The more we deviate from the server command API, the more at risk we are of an impedance mismatch should it change.  I'm going to talk to Jeff about the rationale for the Java driver implementing it the way they did.

I also think if we're going to deprecate `modify_collection`, which I think should happen, then we shouldn't change it to work with views.

I'm going to hold off further review until I discuss with Jeff.

src/mongocxx/test/database.cpp
```diff
@@ -33,6 +34,18 @@ using bsoncxx::builder::stream::document;
 using bsoncxx::builder::stream::finalize;
 using bsoncxx::builder::stream::open_document;
 
+namespace {
+
+int get_max_wire_version(const client& mongo_client) {
```
xdg: This seems like the kind of thing that should go into the helper header or into a private .cpp file.  This is likely to be useful elsewhere over time.

src/mongocxx/test/database.cpp
```diff
@@ -370,4 +383,50 @@ TEST_CASE("Database integration tests", "[database]") {
...
+            // The server supports views.
+            REQUIRE(view.count(bsoncxx::document::view{}) == 1);
+        } else {
+            // The server doesn't support views. On these versions of the server, view creation
+            // requests are treated as ordinary collection creation requests.
+            REQUIRE(view.count(bsoncxx::document::view{}) == 0);
+        }
+    }
+
+    SECTION("A view can be modified on a database", "[database]") {
```
xdg: can drop this as we drop modify support.

src/mongocxx/test/options/create_view.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+TEST_CASE("create_view", "[create_view]") {
+    instance::current();
+
+    options::create_view cv;
+
+    SECTION("Can be exported to a document") {
+        cv.pipeline(std::move(pipeline{}.limit(1)));
+
+        auto doc = cv.to_document();
```
xdg: as discussed, this shouldn't be public or unit tested

xdg: With this removed, we still need a section should test that pipeline can be set/retrieved.

src/mongocxx/test/options/modify_collection.cpp
```diff
@@ -19,6 +19,7 @@
 #include <bsoncxx/document/value.hpp>
 #include <bsoncxx/document/view.hpp>
 #include <bsoncxx/json.hpp>
+#include <bsoncxx/stdx/string_view.hpp>
```
xdg: all the modify changes should be dropped

src/mongocxx/options/modify_collection.hpp
```diff
@@ -18,6 +18,8 @@
```
jrassi: Done.

src/mongocxx/pipeline.hpp
```diff
@@ -174,9 +175,16 @@ class MONGOCXX_API pipeline {
...
+    ///
+    bsoncxx::array::view view_array() const;
+
+    ///
     /// @return A view of the underlying BSON document this pipeline represents.
     ///
-    bsoncxx::document::view view() const;
+    /// @deprecated The view_array() method should be used instead of this method.
+    ///
+    MONGOCXX_DEPRECATED bsoncxx::document::view view() const;
```
jrassi:
```
commit a75b38f15319e53f2a611d63613c30eac4f3b63f
Author: J. Rassi <rassi@10gen.com>
Date:   Sat Oct 1 15:49:56 2016 -0400

    CXX-999 Add pipeline::view_array(), deprecate pipeline::view()

    This commit provides a mechanism to get the underlying array out of a
    pipeline object, which logically wraps an array builder.  This is
    needed for database commands which take a pipeline option as a BSON
    array type.

    The pipeline::view() method exposes a document wrapper around this
    array, which is strictly less useful.
```

In other words: pipelines are arrays, and we should expose them as arrays.  Note that an `array::view` can be converted to a `document::view`, but a `document::view` can't be converted to an `array::view`.  It's a design error that `pipeline::view()` didn't return an array originally; hence the addition of the new method and deprecation of the existing one (there's no point in maintaining both).

src/mongocxx/test/database.cpp
jrassi: Agree, though the test infrastructure is in a somewhat messy state right now, and consequently I don't know where the actual best home for this is yet.  Mind I defer this suggestion to the next time we need to do a wire version check, or when we refactor some of the test infrastructure?

src/mongocxx/test/database.cpp
jrassi: Done.

src/mongocxx/test/options/modify_collection.cpp
jrassi: Done.

src/mongocxx/test/options/create_view.cpp
jrassi: See reply above.

src/mongocxx/options/create_view.cpp
```diff
@@ -0,0 +1,43 @@
...
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+namespace options {
+
+create_view& create_view::pipeline(class pipeline pipeline) {
+    _pipeline = std::move(pipeline);
+    return *this;
+}
+
+bsoncxx::document::value create_view::to_document() const {
```
xdg: Let's leave this as is and expand CXX-1053, but let's change that to 3.1-required.  That way we can close this PR/ticket but won't lose track.  Unlike the server, we have much more direct control of what we ship versus what we drop (as you've seen as we've re-triaged the list a couple times).

src/mongocxx/options/modify_collection.cpp
```diff
@@ -41,11 +41,21 @@ modify_collection& modify_collection::index(bsoncxx::document::view_or_value ind
     return *this;
 }
 
+modify_collection& modify_collection::pipeline(class pipeline pipeline) {
```
xdg: That's fine.

src/mongocxx/pipeline.hpp
xdg: Right.  I missed that the pipeline builder was `bsoncxx::builder::stream::array` in the first place`

src/mongocxx/test/database.cpp
xdg: Agreed.

xdg:
LGTM. I'm fine deferring the changes you suggested deferring.

PR #551, opened 2016-10-02
CXX-1062 read_concern::acknowledge_string("") sets level to default

jrassi:
Fixes a bug where passing the empty string to
read_concern::acknowledge_string() would incorrectly set the read
concern level to read_concern::level::k_unknown, instead of the
correct value of read_concern::level::k_server_default.


jrassi:
@acmorrow @xdg, PTAL.

xdg:
LGTM

PR #553, opened 2016-10-04
CXX-1033 Re-enable decimal128 in libbson compilation for CI

xdg:
libmongoc master has updated its libbson submodule to include
the fix that should allow us to remove `--disable-decimal-bid`.


xdg:
@jrassi PTAL

PR #554, opened 2016-10-04
CXX-1049 Exclude private headers based on .hh suffix

xdg:
The prior wild card match is too greedy, matching parent
directories.  This restricts the exclusions to subdirectories.


xdg:
@acmorrow @jrassi PTAL.

src/bsoncxx/CMakeLists.txt
```diff
@@ -160,8 +160,8 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     FILES_MATCHING
       PATTERN "*.h"
       PATTERN "*.hpp"
-      REGEX ".+/private" EXCLUDE
-      REGEX ".+/test" EXCLUDE
+      REGEX "${CMAKE_CURRENT_SOURCE_DIR}/.*private" EXCLUDE
```
jrassi: Will `${CMAKE_CURRENT_SOURCE_DIR}` will be interpreted as part of a regex?  If so, that doesn't seem quite correct to me, in the case where the filename contains any special characters.  Note also that you removed the slash preceding `private`, which I think should be restored.

I read the relevant CMake docs and can't think of a bulletproof counter-proposal, though.  Can you use `\Q` and `\E` in these regular expressions?  That won't work if the filename includes `\E`, but it's probably good enough for our purposes.

xdg: I'll try `\Q` and see what happens.

I remove the slash before private because we want to catch both top-level and sublevel private.  Possibly we need two regexes for those cases:

```
REGEX "${CMAKE_CURRENT_SOURCE_DIR}/private" EXCLUDE
REGEX "${CMAKE_CURRENT_SOURCE_DIR}/.*/private" EXCLUDE
```

jrassi: Ah, I had forgotten that we have sub-level private directories.

Suggest also adding a trailing slash.

xdg:
@jrassi @acmorrow PTAL.  Updated to the new .hh exclusion approach.

xdg:
Whoops.  Pushed with a new fix.

src/mongocxx/CMakeLists.txt
```diff
@@ -165,11 +165,7 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     DESTINATION ${MONGOCXX_HEADER_INSTALL_DIR}
     COMPONENT dev
     FILES_MATCHING
-        PATTERN "*.h"
```
jrassi: Same question as above.

src/bsoncxx/CMakeLists.txt
```diff
@@ -158,10 +158,7 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     DESTINATION ${BSONCXX_HEADER_INSTALL_DIR}
     COMPONENT dev
     FILES_MATCHING
-      PATTERN "*.h"
```
jrassi: Are you sure we don't want to export any C headers?  Haven't looked, but thought I'd ask.

xdg: I checked and the only .h header was a private one.  From Drews comment yesterday, it sounds like we want to limit ourselves to .hpp.

xdg:
Regarding getting rid of private directories, the config/private directories have files that do nothing.  Unless we expect to need them in the future, they could be dropped.  I'm not sure about the others -- possibly having both "private" and ".hh" is a double warning to readers that they are private headers.  I.e. the directory name tells them, even though we have .hh for cmake's benefit.

PR #555, opened 2016-10-04
CXX-1068 Fix incomplete iterator API

rkarasev:
When I use bsoncxx::document::view inside Google tests (https://github.com/google/googletest) I faced with this error:

`error: conversion from ‘bsoncxx::v_noabi::document::view::iterator’ to non-scalar type ‘bsoncxx::v_noabi::document::view::const_iterator’ requested
   for (typename C::const_iterator it = container.begin();`

I opened Google's code and found this code:
###### 

// Used to print an STL-style container when the user doesn't define
// a PrintTo() for it.
template <typename C>
void DefaultPrintTo(IsContainer, false_type, const C& container, ::std::ostream\* os) {
  // ...
  for (typename C::const_iterator it = container.begin();  it != container.end(); ++it, ++count) {
###### 

STL-style containers return const_iterator using methods begin/end for const objects, but your implementation returns non const iterators for const objects. 

I fixed this problem in my commit.

### DIFF
--- a/src/bsoncxx/document/view.cpp
+++ b/src/bsoncxx/document/view.cpp
@@ -148,7 +148,15 @@ view::const_iterator view::cend() const {
     return const_iterator();
 }
 
-view::iterator view::begin() const {
+view::const_iterator view::begin() const {
+    return cbegin();
+}
+
+view::const_iterator view::end() const {
+    return const_iterator();
+}
+
+view::iterator view::begin() {
     bson_t b;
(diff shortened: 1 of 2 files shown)


xdg:
Thanks!  This has been assigned ticket [CXX-1068](https://jira.mongodb.org/browse/CXX-1068) for tracking.  I believe we'll need to make a similar change to the array iterator, but that's straightforward.

PR #556, opened 2016-10-04
CXX-976 Allow users to set a limit on acceptable staleness

jrassi:
@acmorrow @xdg, PTAL.


xdg:
LGTM.

This is another example where I don't find the mocks to do much, given that we already have tested actual round-tripping through the implementation.

PR #557, opened 2016-10-06
CXX-1072 create_collection::to_document(): serialize _no_padding fix

jrassi:
Previously, options::create_collection::_no_padding was always
serialized incorrectly.


jrassi:
@xdg, PTAL.

This fixes the database integration test suite, which has been failing against nightly master of the server since the work for [SERVER-25459](https://jira.mongodb.org/browse/SERVER-25459) was pushed.

PR #560, opened 2016-10-13
CXX-1052 Pull find::convert_all_modifiers() into new class in private/

jrassi:
@xdg: per your offline request, I've merged #544 and opened this new review for a second consideration of my suggested improvements to the legacy modifiers transformation logic.


src/mongocxx/options/private/rewriter.cpp
```diff
@@ -198,12 +198,7 @@ find rewriter::rewrite_find_modifiers(const find& options) {
         }
     }
 
-    // Ideally, we'd like to return an options::find object with the 'modifiers' option not engaged,
-    // but the current options::find API does not provide a mechanism to clear an existing option.
-    // While not ideal, it's sufficient for our purposes to set 'modifiers' to an empty view here
-    // (the alternative would be to build up a new options object from scratch, but this would be
-    // verbose and error-prone).
-    converted_options.modifiers(bsoncxx::document::view{});
```
jrassi: > Why do you think that's better than disengaging the optional?

The advantage of this approach is that knowledge of the legacy modifiers conversion logic is moved out of the `options::find` class, which is left to be a simple struct-like object consistent with its original design.

Disengaging the optional here (required due to a current limitation of the options API, which we should totally also discuss separately) isn't great, but I didn't consider it a strong con at the time.  However, I see your concern have posted a fix-up commit with a new `MONGOCXX_PRIVATE` method on `options::find`, which eliminates the need to set `modifiers` to an empty view here..  I could also instead have the `options::find` class friend the `options::rewriter::rewrite_find_modifiers` method; let me know if that seems preferable to you.

jrassi:
Note that, as of a61870e, this review is entirely code movement.

xdg:
LGTM

PR #562, opened 2016-10-13
CXX-1088 RHEL 5 Evergreen variant should use v3.2-latest server binary

jrassi:
The server no longer builds on RHEL 5, so the source server binary
link for the RHEL 5.5 Evergreen variant is out of date.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -10,7 +10,10 @@ variables:
     ## source variables
     source:
         linux_64_source: &linux_64_source https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-latest.tgz
-        rhel55_source: &rhel55_source http://downloads.mongodb.org/linux/mongodb-linux-x86_64-rhel55-latest.tgz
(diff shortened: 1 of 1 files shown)


jrassi:
Note that CXX-999 depends on this, which introduces (in tests) the assumption that a max_wire_version == 5 connection implies that the server supports views.  This assumption is currently violated on the RHEL 5 builder, which is currently using a binary that hasn't been updated since 3.3.8.

bjori:
Oh. I didn't realize the `*-rhel55-latest.tgz` was no longer updated.
I'd just remove the entry and replace `rhel55_source` with `linux_64_source` in the buildvariants

jrassi:
Updated, PTAL.

jrassi:
PTAL again.  I didn't realize that recent builds of the server dropped support for glibc 2.5 entirely, so I've restored the RHEL 5.5-specific server binary, and changed the link to download v3.2-latest instead of -latest.

jrassi:
Actually, no need to take a look again, this is just the same thing we discussed in C++ driver chat.  I'll go ahead and merge.  Thanks for your help on this!

PR #563, opened 2016-10-13
CXX-1088 Fix ARM variant to use latest server binary available

jrassi:
@bjori, PTAL.


jrassi:
> Actually. Maybe do s/aarch64/arm64/g for the whole file? Otherwise it might be confusing later on if a variable was called aarch64 or arm64 ...

Done and merged.

PR #566, opened 2016-10-18
CXX-971 (part 8) Add client helpers to test_util

jrassi:
@xdg, PTAL.

This is groundwork for the collation integration tests.

### DIFF
--- a/src/mongocxx/exception/error_code.hpp
+++ b/src/mongocxx/exception/error_code.hpp
@@ -33,6 +33,7 @@ enum class error_code : std::int32_t {
     k_ssl_not_supported,
     k_unknown_read_concern,
     k_unknown_write_concern,
+    k_server_response_malformed,
     // Add new constant string message to error_code.cpp as well!
 };
 
--- a/src/mongocxx/test_util/client_helpers.cpp
+++ b/src/mongocxx/test_util/client_helpers.cpp
@@ -1,0 +1,41 @@
+// Copyright 2016 MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mongocxx/test_util/client_helpers.hh>
+
+#include <bsoncxx/builder/stream/document.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/exception/error_code.hpp>
+#include <mongocxx/exception/operation_exception.hpp>
+
+#include <mongocxx/config/private/prelude.hh>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+namespace test_util {
+
+std::int32_t get_max_wire_version(const client& client) {
+    auto reply =
+        client["admin"].run_command(bsoncxx::builder::stream::document{}
+                                    << "isMaster" << 1 << bsoncxx::builder::stream::finalize);
+    auto max_wire_version = reply.view()["maxWireVersion"];
+    if (!max_wire_version || max_wire_version.type() != bsoncxx::type::k_int32) {
+        throw operation_exception{error_code::k_server_response_malformed};
+    }
+    return max_wire_version.get_int32().value;
+}
+
+}  // namespace test_util
+MONGOCXX_INLINE_NAMESPACE_END
+}  // namespace mongocxx
--- a/src/mongocxx/test_util/client_helpers.cpp
+++ b/src/mongocxx/test_util/client_helpers.cpp
@@ -30,7 +30,11 @@ std::int32_t get_max_wire_version(const client& client) {
         client["admin"].run_command(bsoncxx::builder::stream::document{}
                                     << "isMaster" << 1 << bsoncxx::builder::stream::finalize);
     auto max_wire_version = reply.view()["maxWireVersion"];
-    if (!max_wire_version || max_wire_version.type() != bsoncxx::type::k_int32) {
+    if (!max_wire_version) {
+        // If wire version is not available (i.e. server version too old), it is assumed to be zero.
+        return 0;
+    }
+    if (max_wire_version.type() != bsoncxx::type::k_int32) {
         throw operation_exception{error_code::k_server_response_malformed};
     }
     return max_wire_version.get_int32().value;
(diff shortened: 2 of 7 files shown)


src/mongocxx/exception/error_code.hpp
```diff
@@ -33,6 +33,7 @@ enum class error_code : std::int32_t {
     k_ssl_not_supported,
     k_unknown_read_concern,
     k_unknown_write_concern,
+    k_server_response_malformed,
```
xdg: Please update error_code.cpp with a string message as well.
(While you're at it, please sort the messages in the same order as the .hpp file)

I just added a minor comment on the 3.1 dev branch about this. Hopefully that won't cause a merge conflict.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -0,0 +1,49 @@
...
+    if (max_wire_version.type() != bsoncxx::type::k_int32) {
+        throw operation_exception{error_code::k_server_response_malformed};
+    }
+    return max_wire_version.get_int32().value;
+}
+
+bool supports_collation(const client& client) {
+    return get_max_wire_version(client) >= 5;
+}
+
```
xdg: yay! we have a place to hang all the other conditionals we'll need for integration testing.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <bsoncxx/builder/stream/document.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/exception/error_code.hpp>
+#include <mongocxx/exception/operation_exception.hpp>
+
+#include <mongocxx/config/private/prelude.hh>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
```
xdg: I don't think we need this, as this is only a test file that isn't installed, right?  I think this macro is only for public ABI versioning.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+    return max_wire_version.get_int32().value;
+}
+
+bool supports_collation(const client& client) {
+    return get_max_wire_version(client) >= 5;
+}
+
+}  // namespace test_util
+MONGOCXX_INLINE_NAMESPACE_END
```
xdg: Ditto not needed?

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -0,0 +1,58 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <cstdint>
+
+#include <mongocxx/config/private/prelude.hh>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
```
xdg: Ditto not needed?

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -0,0 +1,58 @@
...
+
+class client;
```
xdg: I think even without the inline namespace this should work as mongocxx::client should resolve to the full version with the inline ABI version namespace.

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -0,0 +1,58 @@
...
+
+namespace test_util {
+
+///
```
xdg: Since these aren't being installed for end users, then we probably shouldn't have doxygen comments for them.

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -0,0 +1,58 @@
...
+///   The client to introspect.
+///
+/// @return Whether or not the client supports collation.
+///
+/// @throws mongocxx::operation_exception if the operation fails, or the server reply is malformed.
+///
+bool supports_collation(const client& client);
+
+}  // namespace test_util
+MONGOCXX_INLINE_NAMESPACE_END
```
xdg: Ditto not needed?

src/mongocxx/test_util/mock.hh
```diff
@@ -25,7 +25,10 @@
 #include <unordered_map>
 #include <vector>
 
+#include <mongocxx/config/private/prelude.hh>
+
 namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
```
xdg: Ditto not needed?

src/mongocxx/test_util/mock.hh
```diff
@@ -183,4 +186,7 @@ class mock<R (*)(Args...)> {
 };
 
 }  // namespace test_util
-}  // namespace mongo
+MONGOCXX_INLINE_NAMESPACE_END
```
xdg: Ditto not needed?

src/mongocxx/test_util/client_helpers.cpp
jrassi: Yay indeed!

src/mongocxx/test_util/client_helpers.cpp
jrassi: See reply above.

src/mongocxx/test_util/client_helpers.hh
jrassi: See reply above.

src/mongocxx/test_util/client_helpers.hh
jrassi: See reply above.

src/mongocxx/test_util/client_helpers.hh
jrassi: See reply above.

src/mongocxx/test_util/mock.hh
jrassi: See reply above.

src/mongocxx/test_util/mock.hh
jrassi: See reply above.

src/mongocxx/exception/error_code.hpp
jrassi: Done.

src/mongocxx/test_util/client_helpers.cpp
jrassi: See the comment message I wrote.

```
commit e250db1588b700d2e9fb4054923b6e48ef529837
Author: J. Rassi <rassi@10gen.com>
Date:   Tue Oct 18 15:49:07 2016 -0400

    CXX-971 Move test_util namespace into ABI namespace

    While test_util is not part of the ABI, this makes the mock class
    consistent with other private code that is declared within the ABI
    namespace (such as mongocxx::libbson).
```

src/mongocxx/test_util/client_helpers.hh
jrassi: Hmm.  I investigated, and confirmed indeed that other header comments for private code don't use `@param` etc, but they do use doxygen-style triple-comments.  Changed in fixup commits.

It'd be worth thinking at some point about a way to make the header-comments-for-private-code situation make more sense going forward.

acmorrow: The intention was that private members (and private classes!) where to explicitly *not* get doxygen /// comments. Just plain old // comments.

PR #569, opened 2016-10-18
CXX-971 (part 11) Add collation support for non-find reads

jrassi:
@xdg, PTAL.

On top of #568.

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -741,6 +741,24 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
         assert_contains_one("bar");
         assert_contains_one("quux");
     }
+
+    SECTION("distinct with collation", "[collection]") {
+        auto doc = document{} << "x"
+                              << "foo" << finalize;
+
+        auto predicate = document{} << "x"
+                                    << "FOO" << finalize;
+
+        auto distinct_opts = options::distinct{}.collation(case_insensitive_collation.view());
+
+        auto distinct_results = coll.distinct("x", predicate.view(), distinct_opts);
+        if (test_util::supports_collation(mongodb_client)) {
+            REQUIRE(std::distance(distinct_results.begin(), distinct_results.end()) == 1);
+        } else {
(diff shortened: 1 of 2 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -706,6 +741,24 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
...
+    SECTION("distinct with collation", "[collection]") {
+        auto doc = document{} << "x"
+                              << "foo" << finalize;
+
+        auto predicate = document{} << "x"
+                                    << "FOO" << finalize;
+
+        auto distinct_opts = options::distinct{}.collation(case_insensitive_collation.view());
+
+        auto distinct_results = coll.distinct("foo", predicate.view(), distinct_opts);
```
xdg: Should this be `coll.distinct("x",predicate.view(), distinct_opts);` since the field is `x`?

src/mongocxx/test/collection.cpp
```diff
@@ -678,9 +713,9 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
         bulk.append(model::insert_one{std::move(doc3)});
         bulk.append(model::insert_one{std::move(doc4)});
 
-        distinct_coll.bulk_write(bulk);
+        coll.bulk_write(bulk);
```
xdg: Why are we using a different collection?  The commit message "minor cleanup" doesn't explain enough.

jrassi: All of the sub-tests in the "CRUD functionality" test case use `coll`, except for tests that explicitly require multiple collections.  This change makes the "distinct test" consistent with the rest of the test cases.

src/mongocxx/test/collection.cpp
jrassi: Yep, you're right, thanks for spotting this.  Pushed a fixup commit addressing this.  This doesn't change correctness of the test (the distinct operation now returns one result with "foo", instead of one result with null), but it clarifies the intent.

PR #570, opened 2016-10-19
CXX-971 (part 12) Add collation support for basic writes

jrassi:
@xdg, PTAL.

On top of #568.

### DIFF
--- a/src/mongocxx/collection.cpp
+++ b/src/mongocxx/collection.cpp
@@ -506,6 +506,9 @@ stdx::optional<result::update> collection::update_one(view_or_value filter, view
     class bulk_write bulk_op(bulk_opts);
 
     model::update_one update_op(filter, update);
+    if (options.collation()) {
+        update_op.collation(*options.collation());
+    }
     if (options.upsert()) {
         update_op.upsert(*options.upsert());
     }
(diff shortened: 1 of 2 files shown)


src/mongocxx/collection.cpp
```diff
@@ -416,7 +416,8 @@ stdx::optional<result::replace_one> collection::replace_one(view_or_value filter
     class bulk_write bulk_op(bulk_opts);
 
     model::replace_one replace_op(filter, replacement);
-    if (options.upsert()) replace_op.upsert(options.upsert().value());
+    if (options.collation()) replace_op.collation(*options.collation());
+    if (options.upsert()) replace_op.upsert(*options.upsert());
```
xdg: Almost all options applications in this file use a block form of `if`.  See https://gist.github.com/xdg/3870ef40ceabf1b9b0c9c592034ff740

I'd rather move towards that common style instead of repeating the uncommon style.

Why are we changing the upsert line?

jrassi: Changed all one-line conditionals in this file to use braces.  See new commit and fixup commits posted.  I also filed [CXX-1095](https://jira.mongodb.org/browse/CXX-1095) to track clang-tidy integration; see the ticket description for more details.

I changed the upsert line because the use of `optional::operator*()` is much more common than `optional::value()` in this file, and in the rest of the codebase.

PR #571, opened 2016-10-19
CXX-971 (part 13) Add collation support for find_one_and_* writes

jrassi:
@xdg, PTAL.

On top of #568.


src/mongocxx/collection.cpp
```diff
@@ -520,23 +520,28 @@ stdx::optional<bsoncxx::document::value> collection::find_one_and_replace(
     auto opts = libmongoc::find_and_modify_opts_new();
     auto opts_cleanup = make_guard([&opts] { libmongoc::find_and_modify_opts_destroy(opts); });
     int flags = ::MONGOC_FIND_AND_MODIFY_NONE;
-    scoped_bson_t bson_replacement{replacement};
-    scoped_bson_t bson_sort{options.sort()};
-    scoped_bson_t bson_projection{options.projection()};
```
xdg: For future reference, I find this sort of code churn distracting when it's not related to the purpose of the CR.  I have no problem with this going into a "minor: Avoid creating variables when not needed" commit that doesn't need CR.  Otherwise, I'm left puzzling out the significance of the move. In this case, It also got me wondering about lifetime, reading the C driver code, and I've just filed a CDRIVER ticket about how the option copy behavior is undocumented.

jrassi: Hmm, I'm having trouble understanding your position here.  If glancing at this change makes you uncertain about its correctness, then to me it doesn't seem appropriate at all to commit it as an unreviewed minor change.  Note also that this change is committed separately in this review (8c213c335a2549452e56e4171709ad4e83374870).  Perhaps let's discuss further in person.

PR #575, opened 2016-10-21
CXX-1051 Deprecate options::find::modifiers()

jrassi:
@xdg, PTAL.

As part of the resolution of CXX-1056, we will go through all of the `@deprecated`-tagged method declarations and add MONGOCXX_DEPRECATED specifiers, including these two.


jrassi:
Per our previous discussions, let me know also if you'd like comment-only changes like these to be unreviewed, going forward.

PR #579, opened 2016-12-14
CXX-915 Add CMake examples for package search mechanisms

acmorrow:
This adds examples for how to find the mongocxx and bsoncxx libraries and link against them using the two primary CMake package search mechanisms:

https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html
https://cmake.org/cmake/help/v3.0/command/find_package.html

A follow up task should be to automatically run these in the CI loop so they don't go stale.


xdg:
I added a commit and rebased to master.

PR #581, opened 2017-02-09
CXX-1187 Handle null return values from libmongoc for uri's

aseering:
This is a proposed fix for [CXX-1187](https://jira.mongodb.org/browse/CXX-1187), as discussed on that ticket.

### DIFF
--- a/src/mongocxx/test/uri.cpp
+++ b/src/mongocxx/test/uri.cpp
@@ -67,7 +67,7 @@ TEST_CASE("URI", "[uri]") {
         REQUIRE(u.hosts().size() == 1);        
         REQUIRE(u.hosts()[0].name == "localhost");
         REQUIRE(u.hosts()[0].port == 27017);
-        REQUIRE(u.hosts()[0].family == 0);
+        // Don't check 'u.hosts()[0].family'.  Value is platform-dependent.
         REQUIRE(u.to_string() == "mongodb://localhost:27017");
         REQUIRE(u.write_concern().journal() == false);
         REQUIRE(u.write_concern().majority() == false);
--- a/src/mongocxx/test/uri.cpp
+++ b/src/mongocxx/test/uri.cpp
@@ -20,14 +20,40 @@
 #include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/uri.hpp>
 
-#include <bsoncxx/json.hpp>
-
 TEST_CASE("URI", "[uri]") {
     SECTION("Default URI") {
         REQUIRE_NOTHROW(mongocxx::uri{});
         REQUIRE_NOTHROW(mongocxx::uri{mongocxx::uri::k_default_uri});
         REQUIRE(mongocxx::uri{}.to_string() ==
                 mongocxx::uri{mongocxx::uri::k_default_uri}.to_string());
+
+        mongocxx::uri u{};
+
+        // Values that should be empty with a blank URI.
+        REQUIRE(u.auth_mechanism() == "");
+        REQUIRE(u.auth_source() == "admin");
+        REQUIRE(u.database() == "");
+        REQUIRE(u.hosts().size() == 1);
+        REQUIRE(u.hosts()[0].name == "localhost");
+        // Don't check 'u.hosts()[0].family'.  Value is platform-dependent.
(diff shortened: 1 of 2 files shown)


src/mongocxx/test/uri.cpp
```diff
@@ -41,4 +43,36 @@ TEST_CASE("URI", "[uri]") {
...
+        REQUIRE(u.password() == "");
+        REQUIRE(u.read_concern().acknowledge_string().empty());
+        REQUIRE(!u.read_preference().tags());
+        REQUIRE(!u.read_preference().max_staleness());
+        REQUIRE(u.replica_set() == "");
+        REQUIRE(u.username() == "");
+        REQUIRE(!u.write_concern().nodes());
+        
+        // Values with defaults with a blank URI
+        REQUIRE(u.ssl() == false);
```
aseering: I don't know that these values are necessarily correct. I have simply verified that (1) they aren't brazenly unreasonable to my untrained eye, and (2) after the fix below, none of these accessor functions segfault :-)

src/mongocxx/uri.cpp
```diff
@@ -26,6 +26,19 @@
...
+
+std::string to_string_null_safe(const char* str)
+{
+    if (str == nullptr) {
+        return std::string();
+    }
+    return str;
+}
+
+}
```
aseering: Stylistically, this is not the most elegant approach ever.  But it works.  I'm guessing that this function may be useful for other wrapper classes as well, but I didn't see a more central place to put it.

src/mongocxx/test/uri.cpp
```diff
@@ -67,7 +67,7 @@ TEST_CASE("URI", "[uri]") {
         REQUIRE(u.hosts().size() == 1);        
         REQUIRE(u.hosts()[0].name == "localhost");
         REQUIRE(u.hosts()[0].port == 27017);
-        REQUIRE(u.hosts()[0].family == 0);
```
aseering: There's presumably a #define'd constant for this in socket.h on Linux/Mac (AF_INET?), but it's elsewhere on Windows...

src/mongocxx/test/uri.cpp
```diff
@@ -41,4 +43,36 @@ TEST_CASE("URI", "[uri]") {
...
+        // Values with defaults with a blank URI
+        REQUIRE(u.ssl() == false);
+        REQUIRE(u.read_concern().acknowledge_level() == mongocxx::read_concern::level::k_server_default);
+        REQUIRE(u.read_preference().mode() == mongocxx::read_preference::read_mode::k_primary);
+        REQUIRE(u.auth_source() == "admin");
+        REQUIRE(u.hosts().size() == 1);        
+        REQUIRE(u.hosts()[0].name == "localhost");
+        REQUIRE(u.hosts()[0].port == 27017);
+        // Don't check 'u.hosts()[0].family'.  Value is platform-dependent.
+        REQUIRE(u.to_string() == "mongodb://localhost:27017");
```
jrassi: Suggest replacing the right-hand side with `mongocxx::uri::k_default_uri`.

src/mongocxx/uri.cpp
jrassi: Oops, one last nit: per code style, please replace this line with:

```
}  // namespace
```

src/mongocxx/test/uri.cpp
```diff
@@ -20,6 +20,8 @@
 #include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/uri.hpp>
 
+#include <bsoncxx/json.hpp>
```
aseering: One of the changes suggested in one of your comments below drops the need for this header.  So I think I'll just delete this line, if that's ok.

aseering:
Thanks!  I've updated the code per your suggestions.  Let me know if I missed anything, or if you see anything else that would benefit from a clean-up.

jrassi:
Thanks for addressing my feedback.  This LGTM.  We very much appreciate your report of this issue and your submission of another quality pull request!

Merged in [c045cea](https://github.com/mongodb/mongo-cxx-driver/commit/c045cea1e66b3bebcb96c838efdf99dcb40c6086).

PR #582, opened 2017-02-26
CXX-1231 When patching include paths to use 'mnmlstc', avoid filename collisions

aseering:
By default, perl's "-i" option tells it to:
* Open the specified file.
* Unlink the file.
* Create a new file with the same name.
* Copy the old file to the new file, transforming it per the regex.

This is not valid on Windows because you can't have two files with the same name, even if one of them has been unlinked.

The quick fix is to tell Perl to instead create a backup file, so it moves the original file rather than unlinking it.


aseering:
Hi again!  I hit this while compiling with `mnmlstc` on Windows.  I realize this may not be a common use case because VC++ apparently doesn't support mnmlstc yet.  But I'm using gcc 6 via [WSL](https://msdn.microsoft.com/en-us/commandline/wsl/faq), on a pre-release version of Windows 10.

It would arguably be cleaner to go remove perl's backup files after perl runs.  On the other hand, disk space is cheap, these files are small, and someone might want to go debug this regex someday, in which case having both versions of the file around could be convenient.

xdg:
Assigned [CXX-1231](https://jira.mongodb.org/browse/CXX-1231) for tracking.

xdg:
Amended to delete the .bak files (so they aren't installed) and rebased as 4bfcf67b667c259d7230157b91784875710be270

Thank you!

PR #583, opened 2017-04-23
minor: fix typos in tutorial

ksuarz:
Hey team, this is just a few minor typos in the tutorial. I'm not familiar with the C++ driver contribution process, so please let me know if I should file a JIRA ticket/edit the commit messages/etc.

In addition: I'm not familiar with formatting, so I don't exactly know how to fix this, but it seems that`h5` doesn't seem to have styling on [the tutorial page](https://mongodb.github.io/mongo-cxx-driver/mongocxx-v3/tutorial/). For example, see the "Get A Single Document That Matches a Filter" section; it's formatted the same as plain text.

### DIFF
--- a/docs/content/mongocxx-v3/tutorial.md
+++ b/docs/content/mongocxx-v3/tutorial.md
@@ -124,7 +124,7 @@ mongocxx::collection coll = db["test"];
 
 ### Create a Document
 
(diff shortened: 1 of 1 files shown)


saghm:
LGTM. Thanks for the PR!

PR #586, opened 2017-07-26
CXX-1413 fix view_or_value's description

dusmart:
When we pass the argument to a document::view_or_value parameter, it uses r-value reference if and only if the argument's type is document::value


saghm:
This has been assigned [CXX-1413](https://jira.mongodb.org/browse/CXX-1413) for tracking

saghm:
Committed in https://github.com/mongodb/mongo-cxx-driver/commit/c6e6c4bfbc2020b08c3917d6a8aca8d1d6c9f8f1. Thanks for the fix!

PR #589, opened 2017-09-21
CXX-1241: Implement Extended JSON spec

derickr:
Preliminary implementation. I chose to *add* two functions, just like in PHP, to prevent BC from being broken.

### DIFF
--- a/src/bsoncxx/json.cpp
+++ b/src/bsoncxx/json.cpp
@@ -41,14 +41,13 @@ void bson_free_deleter(std::uint8_t* ptr) {
     bson_free(ptr);
 }
 
-}  // namespace
-
-std::string BSONCXX_CALL to_json(document::view view) {
+std::string to_json_helper(document::view view, decltype(bson_as_json) converter) {
     bson_t bson;
     bson_init_static(&bson, view.data(), view.length());
 
     size_t size;
-    auto result = bson_as_json(&bson, &size);
+    auto result = converter(&bson, &size);
+
     if (!result)
         throw exception(error_code::k_failed_converting_bson_to_json);
 
@@ -56,6 +55,21 @@ std::string BSONCXX_CALL to_json(document::view view) {
     const std::unique_ptr<char[], decltype(deleter)> cleanup(result, deleter);
 
     return {result, size};
+}
+
+}  // namespace
+
+std::string BSONCXX_CALL to_json(document::view view, ExtendedJsonMode mode) {
+    switch (mode) {
+        case ExtendedJsonMode::k_legacy:
+            return to_json_helper(view, bson_as_json);
+
+        case ExtendedJsonMode::k_relaxed:
+            return to_json_helper(view, bson_as_relaxed_extended_json);
+
+        case ExtendedJsonMode::k_canonical:
+            return to_json_helper(view, bson_as_canonical_extended_json);
+    }
 }
 
 document::value BSONCXX_CALL from_json(stdx::string_view json) {
--- a/src/bsoncxx/json.hpp
+++ b/src/bsoncxx/json.hpp
@@ -34,16 +34,25 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 extern const bool k_silence_doxygen;
 
 ///
-/// Converts a BSON document to a JSON string.
+/// An enumeration of the types of Extended JSON that the to_json function accepts
+enum class ExtendedJsonMode : std::uint8_t {
+    k_legacy,    ///< to produce Legacy Extended JSON
+    k_canonical, ///< to produce Canonical Extended JSON
+    k_relaxed,   ///< to produce Relaxed Extended JSON
+};
+
+///
+/// Converts a BSON document to a JSON string, in extended format.
 ///
 /// @param view
 ///   A valid BSON document.
 ///
 /// @throws bsoncxx::exception with error details if the conversion failed.
 ///
-/// @returns A JSON string.
+/// @returns An extended JSON string.
 ///
-BSONCXX_API std::string BSONCXX_CALL to_json(document::view view);
+BSONCXX_API std::string BSONCXX_CALL
+to_json(document::view view, ExtendedJsonMode mode = ExtendedJsonMode::k_legacy);
 
 ///
 /// Constructs a new document::value from the provided JSON text
--- a/src/bsoncxx/test/json.cpp
+++ b/src/bsoncxx/test/json.cpp
@@ -70,4 +70,41 @@ TEST_CASE("CXX-941 is resolved") {
     REQUIRE(output ==
             "{ \"obj_name\" : \"{\\\"id1\\\":\\\"val1\\\", \\\"id2\\\":\\\"val2\\\"}\" }");
 }
+
+TEST_CASE("CXX-1246: Legacy Extended JSON (Implicit)") {
+    using namespace bsoncxx;
+    types::b_binary bin_val{binary_sub_type::k_uuid, 8, reinterpret_cast<const uint8_t*>("deadbeef")};
+    auto doc = make_document(kvp("number", 42), kvp("bin", bin_val));
+    auto output = bsoncxx::to_json(doc.view());
+    REQUIRE(output ==
+            R"({ "number" : 42, "bin" : { "$binary" : "ZGVhZGJlZWY=", "$type" : "04" } })");
+}
+
+TEST_CASE("CXX-1246: Legacy Extended JSON (Explicit)") {
+    using namespace bsoncxx;
+    types::b_binary bin_val{binary_sub_type::k_uuid, 8, reinterpret_cast<const uint8_t*>("deadbeef")};
+    auto doc = make_document(kvp("number", 42), kvp("bin", bin_val));
+    auto output = to_json(doc.view(), ExtendedJsonMode::k_legacy);
+    REQUIRE(output ==
+            R"({ "number" : 42, "bin" : { "$binary" : "ZGVhZGJlZWY=", "$type" : "04" } })");
+}
+
+TEST_CASE("CXX-1246: Relaxed Extended JSON") {
+    using namespace bsoncxx;
+    types::b_binary bin_val{binary_sub_type::k_uuid, 8, reinterpret_cast<const uint8_t*>("deadbeef")};
+    auto doc = make_document(kvp("number", 42), kvp("bin", bin_val));
+    auto output = to_json(doc.view(), ExtendedJsonMode::k_relaxed);
+    REQUIRE(output ==
+            R"({ "number" : 42, "bin" : { "$binary" : { "base64": "ZGVhZGJlZWY=", "subType" : "04" } } })");
+}
+
+TEST_CASE("CXX-1246: Canonical Extended JSON") {
+    using namespace bsoncxx;
+    types::b_binary bin_val{binary_sub_type::k_uuid, 8, reinterpret_cast<const uint8_t*>("deadbeef")};
+    auto doc = make_document(kvp("number", 42), kvp("bin", bin_val));
+    auto output = to_json(doc.view(), ExtendedJsonMode::k_canonical);
+    REQUIRE(
+        output ==
+        R"({ "number" : { "$numberInt" : "42" }, "bin" : { "$binary" : { "base64": "ZGVhZGJlZWY=", "subType" : "04" } } })");
+}
 }  // namespace
--- a/docs/content/mongocxx-v3/installation.md
+++ b/docs/content/mongocxx-v3/installation.md
@@ -37,6 +37,7 @@ have problems, please file a bug report via
 
 The mongocxx driver builds on top of the MongoDB C driver.
 
+* For mongocxx-3.3.x, libmongoc 1.7.0 or later is required.
 * For mongocxx-3.2.x, libmongoc 1.7.0 or later is required.
 * For mongocxx-3.1.x, libmongoc 1.5.0 or later is required.
 * For mongocxx-3.0.x, we recommend the last 1.4.x version of libmongoc


src/bsoncxx/json.cpp
```diff
@@ -43,12 +43,28 @@ void bson_free_deleter(std::uint8_t* ptr) {
 
 }  // namespace
 
-std::string BSONCXX_CALL to_json(document::view view) {
+enum class jsonMode { LEGACY, CANONICAL, RELAXED };
```
acmorrow: - If this is only used in this file, it belongs in the unnamed namespace
- Types start with capital letters
- Enum constants start with k, so kLegacy, etc.

src/bsoncxx/json.cpp
```diff
@@ -43,12 +43,28 @@ void bson_free_deleter(std::uint8_t* ptr) {
...
+
+std::string _to_json_helper(document::view view, jsonMode mode)
```
acmorrow: There is no need to use leading _ in C++. Just move it up into the unnamed namespace, which accomplishes the same thing better. You are flirting with undefined behavior here for no good reason.

src/bsoncxx/json.cpp
```diff
@@ -43,12 +43,28 @@ void bson_free_deleter(std::uint8_t* ptr) {
...
+{
     bson_t bson;
     bson_init_static(&bson, view.data(), view.length());
 
     size_t size;
-    auto result = bson_as_json(&bson, &size);
+    char  *result;
```
acmorrow: The * looks in the wrong place to me. What does clang format say?

src/bsoncxx/json.cpp
```diff
@@ -43,12 +43,28 @@ void bson_free_deleter(std::uint8_t* ptr) {
...
+
+    switch (mode) {
```
acmorrow: Since all these variations call functions with the same signature, why don't you just pass the function you want to call in here?

```
std::string to_json_helper(document::view view, decltype(&bson_as_json) converter) {
   bson_t bson;
   bson_init_static(...);
   size_t size;
   auto result = converter(&bson, &size);
   if (!result)
       throw...
   return {result, size};
}
```

Then `to_canonical_json` is just:

```
... {
    return to_json_helper(view, &bson_as_canonical_extended_json);
}
```

And you don't need the enum at all.

src/bsoncxx/test/json.cpp
```diff
@@ -70,4 +70,28 @@ TEST_CASE("CXX-941 is resolved") {
     REQUIRE(output ==
             "{ \"obj_name\" : \"{\\\"id1\\\":\\\"val1\\\", \\\"id2\\\":\\\"val2\\\"}\" }");
 }
+
+TEST_CASE("CXX-1246: Legacy Extended JSON") {
+    bsoncxx::types::b_decimal128 decimal_val("-1234E+999");
+    auto doc = make_document(kvp("number", 42), kvp("dec128", decimal_val));
+    std::string output = bsoncxx::to_json(doc.view());
```
acmorrow: pretty reasonable to use auto to capture the result of to_json in this test and below.

src/bsoncxx/json.hpp
```diff
@@ -34,18 +34,42 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+/// Converts a BSON document to a JSON string, in canonical extended format.
+///
+/// @param view
+///   A valid BSON document.
+///
+/// @throws bsoncxx::exception with error details if the conversion failed.
+///
+/// @returns A canonical extended JSON string.
+///
+BSONCXX_API std::string BSONCXX_CALL to_canonical_extended_json(document::view view);
```
mbroadst: Not that the modes are likely to change very much in the future, but I would generally opt for a single method with public enumeration:

```
BSONCXX_API std::string BSONCXX_CALL to_extended_json(document::view view, ExtendedJsonMode mode);
```

Is there an obvious reason to prefer three explicit methods that I'm missing? Seems like more symbols to maintain, and not very extensible from an ABI perspective.

src/bsoncxx/test/json.cpp
mbroadst: +1

src/bsoncxx/json.cpp
mbroadst: Likewise, if we opt for a single method that takes an enum argument we could reduce the code dupe.

src/bsoncxx/json.cpp
mbroadst: @acmorrow generally I agree. Out of morbid curiosity could you elaborate on the comment about undefined behavior? Are you just referring to potential conflicts when mangling?

acmorrow: https://stackoverflow.com/a/228797

Basically what is here is legal (I think?) because _to_json isn't at global scope, it is at namespace scope, but that is too close for comfort.

- __to_json would be definitely illegal
- _To_json would be definitely illegal.

Why get that close to the line when just putting it in the unnamed namespace achieves the same effect. Leading underscores in any context are ill-advised. Best to avoid them.

src/bsoncxx/json.hpp
acmorrow: I think this is the most critical part of the review, since it is the user facing API. Lets sort this one out. Arguments for three methods, rather than public enum and one method?

src/bsoncxx/json.cpp
acmorrow: - ALL_CAPS_IS_SHOUTY_AND_UNECESSARY_IN_CXX: https://accu.org/index.php/journals/1923
- The k prefix comes from the Google Style Guide and is consistently followed throughout the codebase. Taking another look, it should really be `k_legacy`, etc. But presumably we aren't keeping them.

src/bsoncxx/json.hpp
acmorrow: We aren't currently enforcing ABI stability. So it would still be possible to add a new enum parameter that took a defaulted value. That would still get you API stability. Had we declared ABI stability, I'd agree with you. But since we haven't, I don't think we should use it to decide API design decisions.

src/bsoncxx/json.cpp
acmorrow: Correct.

src/bsoncxx/json.hpp
mbroadst: I might also add that for `to_json` we are returning the "JSON" representation, with `to_extended_json` we are returning the "Extended JSON" representation, those are the two things we should provide API for.  Returning "relaxed" or "canonical" Extended JSON is an _optional_ variant of Extended JSON, which is why I see it as an enum passed to a single function.  Originally we had a "strict" mode for Extended JSON as well, so you can see there is potential for growth in this enum.   So thats my stylistic argument, beyond my "less exported symbols is better" argument :smile:

mbroadst: Also, from the spec:
```
A generator MUST allow users to produce strings in either the Canonical Extended JSON format or the Relaxed Extended JSON format. If generators provide a default format, the default SHOULD be the Relaxed Extended JSON format.
```

src/bsoncxx/json.hpp
```diff
@@ -34,18 +34,39 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 extern const bool k_silence_doxygen;
 
 ///
-/// Converts a BSON document to a JSON string.
+/// An enumeration of the types of Extended JSON that the to_extended_json function accepts
+///
+/// The values are:
+///   k_canonical: to produce Canonical Extended JSON
+///   k_relaxed: to produce Relaxed Extended JSON
+enum class ExtendedJsonMode { k_canonical, k_relaxed };
```
acmorrow: IIRC, it was guidance for public enums that they always declare a size class so that adding new enumerators with explicit values could never change the size leading to an ABI break. It appears that that rule has not been rigorously followed.

src/bsoncxx/json.hpp
```diff
@@ -34,18 +34,39 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+///
+/// Converts a BSON document to a JSON string, in legacy extended format.
 ///
 /// @param view
 ///   A valid BSON document.
 ///
 /// @throws bsoncxx::exception with error details if the conversion failed.
 ///
-/// @returns A JSON string.
+/// @returns A legacy extended JSON string.
```
acmorrow: Should their also be an enumerator value that causes this function to be called? If the old format was 'extended', as the comment suggests, then surely it should be possible to obtain that behavior via the new `to_extended_json` API as well.

And, if you did that, you could mark this function as deprecated.

src/bsoncxx/json.hpp
```diff
@@ -34,18 +34,39 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+///
+/// @param view
+///   A valid BSON document.
+///
+/// @throws bsoncxx::exception with error details if the conversion failed.
+///
+/// @returns A extended JSON string.
+///
+BSONCXX_API std::string BSONCXX_CALL
+to_extended_json(document::view view, ExtendedJsonMode mode = ExtendedJsonMode::k_relaxed);
```
acmorrow: I wouldn't add a default. They bind statically in the caller. It made sense to add a default when we were considering adding the enum to `to_json` as an API non-break, but as this is a new function I don't see an upside.

Users can easily write their own wrapper function to default in a value if they choose.

src/bsoncxx/json.hpp
derickr: I'm happy to add it, but as you also said before (paraphrased) "we don't handle ABI breaks".

src/bsoncxx/json.hpp
derickr: For future applications, "legacy" extended shouldn't really be used at all, so I am happy to mark this as "deprecated", but wouldn't want to add it to the enum regardless.

src/bsoncxx/json.hpp
derickr: Sorry, nm. I misread that. I'll remove the default.

src/bsoncxx/json.hpp
acmorrow: This entire project has been undertaken in an effort to set an ABI. The fact that we haven't managed to do it yet isn't an argument against doing the engineering right so that we *can* do it in the future, and have a shot at maintaining an ABI once we do.

src/bsoncxx/json.hpp
acmorrow: By doing this, you are forcing people to change both what function they use, and what format they get, at the same time. Giving them a nicer transitional path is better. I can move to the non-deprecated function but retain the format I want/expect. Then later I can upgrade to the new format when my application or other consumers are ready to deal. Why force both changes at once?

src/bsoncxx/json.hpp
```diff
@@ -34,16 +34,26 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
 
 ///
-/// Converts a BSON document to a JSON string.
+/// An enumeration of the types of Extended JSON that the to_extended_json function accepts
+///
+/// The values are:
+///   k_legacy: to produce Legacy Extended JSON
+///   k_canonical: to produce Canonical Extended JSON
+///   k_relaxed: to produce Relaxed Extended JSON
+enum class ExtendedJsonMode : unsigned int { k_legacy, k_canonical, k_relaxed };
```
acmorrow: unsigned int isn't a fixed width type. Use std::uint8_t or similar.

src/bsoncxx/json.hpp
```diff
@@ -34,16 +34,26 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+///
+/// Converts a BSON document to a JSON string, in extended format.
 ///
 /// @param view
 ///   A valid BSON document.
 ///
 /// @throws bsoncxx::exception with error details if the conversion failed.
 ///
-/// @returns A JSON string.
+/// @returns A extended JSON string.
```
acmorrow: s/A/An

src/bsoncxx/json.hpp
```diff
@@ -34,16 +34,26 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 extern const bool k_silence_doxygen;
 
 ///
-/// Converts a BSON document to a JSON string.
+/// An enumeration of the types of Extended JSON that the to_extended_json function accepts
+///
+/// The values are:
+///   k_legacy: to produce Legacy Extended JSON
```
acmorrow: I'll bet doxygen has a way to explicitly doc comment enumeration values, if so we should use that.

src/bsoncxx/json.cpp
```diff
@@ -58,6 +57,21 @@ std::string BSONCXX_CALL to_json(document::view view) {
     return {result, size};
 }
 
+}  // namespace
+
+std::string BSONCXX_CALL to_json(document::view view, ExtendedJsonMode mode) {
+	switch (mode) {
```
acmorrow: This looks like clang-format wasn't invoked.

src/bsoncxx/json.hpp
```diff
@@ -34,16 +34,25 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 extern const bool k_silence_doxygen;
 
 ///
-/// Converts a BSON document to a JSON string.
+/// An enumeration of the types of Extended JSON that the to_extended_json function accepts
```
acmorrow: Nit: the function to_extended_json no longer exists. And can you doxygen mention it somehow?

src/bsoncxx/json.hpp
```diff
@@ -34,16 +34,25 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+enum class ExtendedJsonMode : std::uint8_t {
+    k_legacy,    ///< to produce Legacy Extended JSON
```
acmorrow: Instead of "to produce", how about "Selects"

src/bsoncxx/test/json.cpp
```diff
@@ -70,4 +70,40 @@ TEST_CASE("CXX-941 is resolved") {
...
+            R"({ "number" : 42, "dec128" : { "$numberDecimal" : "-1.234E+1002" } })");
+}
+
+TEST_CASE("CXX-1246: Relaxed Extended JSON") {
+    using namespace bsoncxx;
+    types::b_decimal128 decimal_val("-1234E+999");
+    auto doc = make_document(kvp("number", 42), kvp("dec128", decimal_val));
+    auto output = to_json(doc.view(), ExtendedJsonMode::k_relaxed);
+    REQUIRE(output ==
+            R"({ "number" : 42, "dec128" : { "$numberDecimal" : "-1.234E+1002" } })");
```
acmorrow: Maybe I'm just having trouble seeing it, but what is the difference of the output here on line 97 from that on line 88?

If they don't actually differ, then I'd suggest finding a test case for which they do, assuming there is one.

mbroadst: @derickr perhaps a better example of Legacy vs Relaxed might be `$regex`, as that is called out explicitly in the  [Extended JSON spec](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst)

PR #590, opened 2017-09-26
Fix Travis builds.

derickr:


mbroadst:
I actually think we should be able to fix travis builds in a far more complete way with new travis features.  For instance all of the "apt" management going on here, can be moved to a new `addons` section:
```
env:
  - CXX=g++-4.8
addons:
  apt:
    sources:
      - ubuntu-toolchain-r-test
    packages:
      - g++-4.8
```

There is a lot more detail on this page:
  - https://docs.travis-ci.com/user/installing-dependencies/

derickr:
Oh sure, this is not nearly ready to get merged. Right now, I am waiting for the Travis build to test this change though. I had hoped it would have been faster.

mbroadst:
Ah, okay.  An added benefit to switching over to `addons` is that you can opt-in to containerized builds (by specifying `sudo: false` in `.travis.yml`) to speed up the tests runs a lot.

derickr:
Yeah, exactly. Wanted to "just make it work" first though.

derickr:
Closing, so that I can do quick iterations of testing through my fork.

PR #591, opened 2017-10-03
CXX-1441, CXX-1443: Remove deprecated API, and pin libmongoc to 1.8.0 for Evergreen

derickr:


acmorrow:
I'm confused. I'm fairly certain that in the past the mongocxx build built and tested itself against both the current stable release of libmongoc/libbson, and master. That way we new when we had been broken by changes in master when they happened, rather than all at once the next time we upgraded. If we are no longer going to do both, I'd rather we tracked master for that same reason.

mbroadst:
@acmorrow we need to have some constants in our life here. The latest version of the c driver pushed upstream is 1.7.0, so that's what users would expect to build against when using the system installed version of the driver - I say that we standardize upon that. Depending on the master version of the driver needlessly introduced a build error that prevented the 3.2.0-rc0 release a month ago when the driver was handed off.  I'm absolutely fine basing the 3.3 (3.6 compat) version of this driver on the latest version of libmongoc, but moving forward I'm not okay simply tracking a moving target while we try to keep up locally.

acmorrow:
If you don't track C driver master in at least one builder (and I'd argue, all), then you are just hiding future work from yourself, and, more importantly, removing a very useful signal for the C driver team. We have on more than one occasion discovered that the C driver had broken something important or subtle on master because the C++ driver was actively tracking it. Additionally, our users build this way: clone the repos, check out master, and build. Our master should build against C driver master at all times. If it breaks we should fix it.

mbroadst:
We are indeed tracking master, on our master branch - the signal is there.  This commit is against the 3.2 branch and thus tracks our guarantees for this release.  We will keep the pinned version here at 1.7.0, which is the expectation for this version of the driver, and continue to track master on master for future development.

acmorrow:
That makes more sense, thanks for clarifying.

acmorrow:
@derickr - FYI:
- the commit message style for the C++ driver is server style: `CXX-ABC Description`. No leading "Fix" or other verb, no colon after the ticket ID. If there is more than one ticket `CXX-ABC CXX-123 Description`.
- If there is no ticket, commit message says "minor: do a little thing".

PR #592, opened 2017-10-10
CXX-1439, CXX-1447, CXX-1453, CXX-1454 Travis and Evergreen improvements/fixes, reduce use of deprecated methods in tests and headers

derickr:


acmorrow:
Hi @derickr  - This isn't really actionable as a code review. It does too many different things, and keeps changing so it is a moving target. Please close it, and then stage out a series of incremental reviews on a per-ticket basis.

derickr:
@acmorrow - It has [WIP] prefixed for a reason - Work In Progress. It's just so that @mbroadst and I can have a discussion about it easily.

acmorrow:
@derickr - When it is a PR, every watcher of the repo gets a notification when you push a new commit to it.

.travis.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+mongo --quiet -p 27017 --eval 'db.version();'
```
mbroadst: what's the purpose of running `db.version()` here?

derickr: It helps with debugging - in case Travis decides to update and the tests start failing for some unexplained reason.

.travis.yml
```diff
@@ -1,79 +1,63 @@
...
-    - sudo apt-get install cmake
-
     # Install Mongo C Driver
     - pushd mongo-c-driver
 
-    # TODO: Update this to our real minimum for the C++11 driver 3.2 release, once known.
-    - git checkout master
-
-    - ./autogen.sh --enable-tests=no --enable-examples=no --with-libbson=bundled; make; sudo make install
+    - git checkout 1.8.1
```
mbroadst: I wonder, if we already have that travis script above, whether we should put this into a new script under a `.travis` folder (similar to the `.evergreen` folder)?

.travis.yml
```diff
@@ -90,10 +74,10 @@ script:
     - ./src/mongocxx/test/test_instance
 
     # Install headers and libs for the examples
-    - make install
+    #- make install
```
mbroadst: why are these all commented out?

src/bsoncxx/builder/core.cpp
```diff
@@ -214,7 +214,7 @@ class core::impl {
             }
         }
 
-        ~frame() {
+        ~frame() noexcept(false) {
```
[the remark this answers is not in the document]
derickr: I agree, but the ~frame already throws the exception. I don't think it should do that, but also am unsure whether we can "just remove it". Hence, for now I thought it'd be better to add the right ``noexcept()`` so that things at least compile

acmorrow:
@derickr - So, this is no longer a WIP review? If so, I reiterate my prior objection. One ticket, one PR, please.

src/bsoncxx/builder/core.cpp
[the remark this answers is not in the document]
derickr: I agree, but the dtor currently *does* throw, and by marking it as such, it at least does not thrown any warnings that prevent compilation with -Werror. I am happy to make a ticket to fix this, however, I am not sure how yet and not break the "API".

mbroadst:
@acmorrow I'm comfortable that these fixes are related enough, generally do not overlap, and are organized by ticket on each of the 5 commits.  While we can discuss the finer details of the PR policy moving forward, we desperately need to move forward to the 3.2 release so let's please focus on that now.

src/mongocxx/hint.hpp
```diff
@@ -100,6 +100,8 @@ class MONGOCXX_API hint {
    private:
     stdx::optional<bsoncxx::document::view_or_value> _index_doc;
     stdx::optional<bsoncxx::string::view_or_value> _index_string;
+
+    bsoncxx::document::value create_document() const;
```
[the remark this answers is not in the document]
acmorrow: A better approach would be to create a new public inline function called `to_document_no_deprecated` that is the same as `to_document` but doesn't have the deprecated tag. Then have `to_document` invoke `to_document_no_deprevated`. It also gives users a way out of the same bind. Then when you actually remove it, you take out both.

Also, methods before members, always, and private functions begin with a leading `_`.

src/mongocxx/hint.hpp
```diff
@@ -155,10 +157,6 @@ MONGOCXX_INLINE hint::operator bsoncxx::types::value() const {
     return to_value();
 }
 
-MONGOCXX_INLINE hint::operator bsoncxx::document::value() const {
```
acmorrow: You de-inlined the definition of this function, but you didn't remove the `MONGOCXX_INLINE` from its declaration.

src/mongocxx/validation_criteria.hpp
```diff
@@ -139,17 +139,15 @@ class MONGOCXX_API validation_criteria {
     stdx::optional<bsoncxx::document::view_or_value> _rule;
     stdx::optional<validation_level> _level;
     stdx::optional<validation_action> _action;
+
+    bsoncxx::document::value create_document() const;
```
acmorrow: Ditto re ordering, `_`, and making a public inline named `no_deprecated` instead.

src/mongocxx/validation_criteria.hpp
```diff
@@ -139,17 +139,15 @@ class MONGOCXX_API validation_criteria {
...
 };
 
 MONGOCXX_API bool MONGOCXX_CALL operator==(const validation_criteria& lhs,
                                            const validation_criteria& rhs);
 MONGOCXX_API bool MONGOCXX_CALL operator!=(const validation_criteria& lhs,
                                            const validation_criteria& rhs);
 
-MONGOCXX_INLINE validation_criteria::operator bsoncxx::document::value() const {
```
acmorrow: Ditto - you de-inlined this but didn't take `MONGOCXX_INLINE` off the declaration.

src/mongocxx/options/create_view.hpp
```diff
@@ -130,11 +130,9 @@ class MONGOCXX_API create_view {
 
     friend MONGOCXX_API bool MONGOCXX_CALL operator==(const create_view&, const create_view&);
     friend MONGOCXX_API bool MONGOCXX_CALL operator!=(const create_view&, const create_view&);
-};
 
-MONGOCXX_INLINE create_view::operator bsoncxx::document::value() const {
```
acmorrow: Ditto re `MONGOCXX_INLINE` still on declaration.

src/mongocxx/options/create_view.hpp
```diff
@@ -130,11 +130,9 @@ class MONGOCXX_API create_view {
...
-    return to_document();
-}
+    bsoncxx::document::value create_document() const;
```
acmorrow: Again, ordering, `_`, and the `no_deprecated` approach.

src/mongocxx/options/create_collection.hpp
```diff
@@ -260,11 +260,9 @@ class MONGOCXX_API create_collection {
                                                       const create_collection&);
     friend MONGOCXX_API bool MONGOCXX_CALL operator!=(const create_collection&,
                                                       const create_collection&);
-};
 
-MONGOCXX_INLINE create_collection::operator bsoncxx::document::value() const {
-    return to_document();
-}
+    bsoncxx::document::value create_document() const;
```
acmorrow: Here too: `MONGOCXX_INLINE`, ordering, `_`, `no_deprecated`.

src/mongocxx/options/create_collection.cpp
```diff
@@ -133,12 +134,22 @@ bsoncxx::document::value create_collection::to_document() const {
     }
 
     if (_validation) {
+        BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_BEGIN;
```
acmorrow: Do you still need to do this if you updated the validation class?

src/bsoncxx/builder/core.cpp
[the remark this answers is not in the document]
acmorrow: Additionally, this seems independent of the goal of addressing deprecations in the associated commit. It should probably have its own ticket and commit.

src/mongocxx/test/database.cpp
```diff
@@ -42,6 +42,12 @@ using bsoncxx::builder::basic::make_document;
 
 bool check_for_collections(cursor cursor, std::set<std::string> expected_colls) {
     for (auto&& coll : cursor) {
+        // Skip system collections
```
acmorrow: Comment says what you are doing, but not why.

src/mongocxx/test/database.cpp
```diff
@@ -42,6 +42,12 @@ using bsoncxx::builder::basic::make_document;
...
+        auto pos = coll["name"].get_utf8().value.to_string().find("system.");
```
acmorrow: This exact subexpression is reiterated on line 51. Maybe pull it into a variable?

src/bsoncxx/builder/core.cpp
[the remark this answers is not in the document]
hanumantmk: Originally that destructor didn't throw (it silently ate errors).  That was probably a mistake, but a minor one (because the only failure mode for array/doc_end is if you've exceeded the max buffer size).

The "correct" solution is probably to provide a separate close() instead of ~frame().  failure of close() could either throw or be acted upon to throw.  ~frame() would become a noop.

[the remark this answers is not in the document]
derickr: I've created a ticket for this: https://jira.mongodb.org/browse/CXX-1458

src/mongocxx/hint.hpp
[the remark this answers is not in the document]
derickr: To the first paragraph:
- That seems like a similar idea to the changes in this PR, except that your ``to_document_no_deprecated`` is called ``create_document``, and that it is a private method instead of a "public inline function". Can you explain why your suggestion is better?

To the second paragraph:
- Is there any *technical* reason why you would want methods before members? Or is it just a convention?
- I thought that in the past you said that you never wanted a leading ``_`` for anything‽

src/mongocxx/test/database.cpp
derickr: Fixed in a separate commit which I'll squash before merge.

src/mongocxx/test/database.cpp
derickr: Fixed in a separate commit which I'll squash before merge.

.mci.yml
```diff
@@ -28,7 +28,7 @@ variables:
 
     ## cmake flag variables
     cmake_flags:
-        linux_cmake_flags: &linux_cmake_flags -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-missing-field-initializers" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wconversion -Wnarrowing -Wno-expansion-to-defined -pedantic -Werror -Wno-missing-field-initializers"
+        linux_cmake_flags: &linux_cmake_flags -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-missing-field-initializers" -DCMAKE_CXX_FLAGS="-Wall -Wextra -Wconversion -Wnarrowing -pedantic -Werror -Wno-missing-field-initializers"
```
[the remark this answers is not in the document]
sgolemon: Seems like that flag should be compiler dependent then, not platform specific.  clang exists on linux as well, after all.

Also, I'm hesitant to believe clang doesn't *ever* need this.  clang-5 has added a number of new warning notices...

src/mongocxx/hint.hpp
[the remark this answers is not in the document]
acmorrow: - The `_no_deprecated` convention was the planned convention for doing this sort of forwarding function. It clearly expresses the intention, and keeps the word "deprecated" at the call site, making it clearer that you are calling a deprecated function. You could make the decoration `no_deprecated_warning` if you wanted. I'm willing to compromise on that, but the adornment for the "does not issue a deprecation warning" shim should definitely contain the word deprecated so that it can be grepped for.

- There is no technical reason, it is a convention, and one that we have followed everywhere else in the codebase. Please continue to do so.

- The rules are tricky, but a leading underscore followed by [a-z] is OK at class scope, it is only forbidden at global scope. Please see http://en.cppreference.com/w/cpp/language/identifiers. Leading underscore for private members and private functions is the convention in the codebase, unless it would lead to using a reserved identifier.

.mci.yml
```diff
@@ -283,7 +283,7 @@ tasks:
         - func: "start_mongod"
         - func: "install_c_driver"
           vars:
-              MONGOC_VERSION: 1.7.0
+              MONGOC_VERSION: 1.8.1
```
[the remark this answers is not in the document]
acmorrow: If you let people build against 1.8.0, they *will* build the tests, and they will complain that they don't build. You should also consistently keep the enforced minimum in the build system and your CI system 1:1, its just good practice. If you let them drift, I promise strange things will go wrong over time.

src/mongocxx/options/create_collection.cpp
acmorrow: Right, and here is where you could call the `no_deprecated` or `no_deprecated_warning` version of the function.

src/mongocxx/test/database.cpp
```diff
@@ -42,13 +42,15 @@ using bsoncxx::builder::basic::make_document;
 
 bool check_for_collections(cursor cursor, std::set<std::string> expected_colls) {
     for (auto&& coll : cursor) {
-        // Skip system collections
-        auto pos = coll["name"].get_utf8().value.to_string().find("system.");
+        // Skip system collections which the MMAPv1 storage engine returns,
+        // while WiredTiger does not.
+        auto nameString = coll["name"].get_utf8().value.to_string();
```
acmorrow: We don't use camelCase in this codebase, we use snake_case.

src/bsoncxx/builder/core.cpp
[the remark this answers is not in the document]
acmorrow: @hanumantmk - Wouldn't adding a `close` method defeat the purpose of frame, which is basically an RAII type? Should we just be only doing the throw if `std::uncaught_exception` returns false?

PR #594, opened 2017-11-04
CXX-1453 Suppress ignored attributes in test mocks due to GCC bug

acmorrow:
This works around the GCC bug where it is interpreting the export attributes on the mongo-c-driver functions while constructing the mocks. Just suppress the warning in that narrow context with a diagnostic pragma.

### DIFF
--- a/src/mongocxx/private/libmongoc.cpp
+++ b/src/mongocxx/private/libmongoc.cpp
@@ -21,11 +21,23 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace libmongoc {
 
 #ifdef MONGOCXX_TESTING
+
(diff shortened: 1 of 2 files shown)


acmorrow:
@derickr - Per comments elsewhere, I'm not holding this for travis. As far as whether it is a bug or not, did you read https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81605? It does not make sense for visibility attributes to count as part of the type of the function.

hanumantmk:
lgtm.  There's no difference between having the function pointer in hand and having the function with visibility.  If you've got it, you can call it the same (i.e. not part of the type).

We should go back and mention that on the gcc bug report

PR #595, opened 2017-11-04
CXX-1460 Small fixups

acmorrow:
These two commits don't address the central issue of CXX-1460, which is that GCC thinks there is an ambiguity between the scoped_bson_t constructors in C++17 mode, but clang requires both overloads. I'm not sure which compiler is in the right, yet.

In the meantime, this adds a clearly missing header, and adds some missing calls to `std::move` that were observed while looking into the implementation of `scoped_bson_t`.

### DIFF
--- a/src/mongocxx/private/libbson.cpp
+++ b/src/mongocxx/private/libbson.cpp
@@ -40,14 +40,14 @@ scoped_bson_t::scoped_bson_t(bsoncxx::document::view_or_value doc)
 scoped_bson_t::scoped_bson_t(bsoncxx::stdx::optional<bsoncxx::document::view_or_value> doc)
     : _is_initialized{doc} {
     if (doc) {
-        _doc = doc;
+        _doc = std::move(doc);
         doc_to_bson_t(*_doc, &_bson);
     }
 }
 
 void scoped_bson_t::init_from_static(bsoncxx::document::view_or_value doc) {
     _is_initialized = true;
-    _doc = doc;
+    _doc = std::move(doc);
     doc_to_bson_t(*_doc, &_bson);
 }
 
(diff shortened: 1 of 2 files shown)


src/mongocxx/private/libbson.cpp
```diff
@@ -40,22 +40,22 @@ scoped_bson_t::scoped_bson_t(bsoncxx::document::view_or_value doc)
...
+    _doc = std::move(doc);
     doc_to_bson_t(*_doc, &_bson);
 }
 
 void scoped_bson_t::init_from_static(
     bsoncxx::stdx::optional<bsoncxx::document::view_or_value> doc) {
     if (doc) {
         _is_initialized = true;
-        _doc = doc;
+        _doc = std::move(doc);
```
derickr: Could this be related to https://jira.mongodb.org/browse/CXX-1388 ?

acmorrow: Don't think so, no. This was purely a missed optimization.

acmorrow:
@derickr - I'm not holding this based on the state of travis, per my comments in #596

src/mongocxx/private/libbson.hh
```diff
@@ -65,7 +65,7 @@ class MONGOCXX_TEST_API scoped_bson_t {
     //
     // The internal bson_t is initialized if the optional is populated.
     //
-    scoped_bson_t(bsoncxx::stdx::optional<bsoncxx::document::view_or_value> doc);
+    explicit scoped_bson_t(bsoncxx::stdx::optional<bsoncxx::document::view_or_value> doc);
```
hanumantmk: This seems to do more than you mentioned in the pr and definitely changes the way scoped_bson_t's will be constructed from arguments.  Not saying if it's right or wrong, but probably wrong for this pr

hanumantmk:
lgtm % adding explicit to scoped_bson_t optional overload

PR #596, opened 2017-11-04
CXX-1461 Add missing unreachable to exhaustive switch statement

acmorrow:


acmorrow:
@derickr - As far as travis goes, I strongly disagree. Travis is already broken on master, and it isn't our build system of record: evergreen is. Now, you can tell me that evergreen is red too, and that is true, but the fix for that is this PR! I'd much prefer to merge this, get evergreen green, then use that to patch queue my other PRs, and only once we have evergreen green on all branches, then worry about travis.

PR #598, opened 2017-11-04
CXX-1463 SFINAE out copy ctor competitor for b_regex and b_codewscope

acmorrow:


src/bsoncxx/types.hpp
```diff
@@ -370,7 +370,7 @@ struct BSONCXX_API b_regex {
     /// @param options
     ///   The regex options
     ///
-    template <typename T, typename U = stdx::string_view>
+    template <typename T, typename U = stdx::string_view, typename std::enable_if<!std::is_same<b_regex, typename std::decay<T>::type>::value, int>::type = 0>
```
derickr: I'd be very curious to hear what this does.

mbroadst: https://eli.thegreenplace.net/2014/sfinae-and-enable_if/ is a nice quick description of SFINAE for this purpose

acmorrow: @derickr - Please see all of the other constructor templates in this file. They all follow this pattern, except for `b_regex` and `b_codewscope`.

We got away with this because they were two argument functions, but when @saghm added a default for `b_regex` it now needs the SFINAE work. I did `b_codewscope` too since if someone later added a default for its `scope` parameter, it would have the same problem.

hanumantmk:
lgtm

PR #599, opened 2017-11-04
CXX-1447 Deprecate things the right way

acmorrow:
This re-structures the deprecations such that all deprecated functions and constructors have an associated `_deprecated` variant, or, for the case of constructors, a `deprecated` tag type. Then, the deprecated versions are implemented in terms of the non-deprecated one.

For us, this gives us an easy way to continue to call deprecated functions without warnings. It does the same thing for our users, except for them they are still clearly calling a deprecated function since the word deprecated is in the method, or on the line of code.

### DIFF
--- a/src/bsoncxx/oid.cpp
+++ b/src/bsoncxx/oid.cpp
@@ -34,7 +34,11 @@ oid::oid() {
 
 const oid::init_tag_t oid::init_tag{};
 
-oid::oid(init_tag_t) : oid::oid() {}
+oid::oid(init_tag_t) : oid::oid(init_tag_deprecated) {}
+
+const oid::init_tag_t_deprecated oid::init_tag_deprecated{};
+
+oid::oid(init_tag_t_deprecated) : oid::oid() {}
 
 oid::oid(const bsoncxx::stdx::string_view& str) {
     if (!bson_oid_is_valid(str.data(), str.size())) {
@@ -63,6 +67,10 @@ std::string oid::to_string() const {
 }
 
 oid::operator bool() const {
+    return operator_bool_deprecated();
+}
+
+bool oid::operator_bool_deprecated() const {
     return true;
 }
 
--- a/src/mongocxx/read_preference.hpp
+++ b/src/mongocxx/read_preference.hpp
@@ -99,6 +99,8 @@ class MONGOCXX_API read_preference {
     ///
     read_preference();
 
+    struct deprecated_tag{};
+
     ///
     /// Constructs a new read_preference.
     ///
@@ -108,6 +110,7 @@ class MONGOCXX_API read_preference {
     /// @deprecated The constructor with no arguments and the method mode() should be used.
     ///
     MONGOCXX_DEPRECATED read_preference(read_mode mode);
+    read_preference(read_mode mode, deprecated_tag);
 
     ///
     /// Constructs a new read_preference with tags.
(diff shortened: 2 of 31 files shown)


src/bsoncxx/oid.cpp
```diff
@@ -34,7 +34,11 @@ oid::oid() {
 
 const oid::init_tag_t oid::init_tag{};
 
-oid::oid(init_tag_t) : oid::oid() {}
+oid::oid(init_tag_t) : oid::oid(init_tag_deprecated) {}
+
+const oid::init_tag_t_deprecated oid::init_tag_deprecated{};
+
+oid::oid(init_tag_t_deprecated) : oid::oid() {}
```
derickr: Is this a new deprecation?

src/mongocxx/read_preference.hpp
```diff
@@ -108,6 +110,7 @@ class MONGOCXX_API read_preference {
     /// @deprecated The constructor with no arguments and the method mode() should be used.
     ///
     MONGOCXX_DEPRECATED read_preference(read_mode mode);
+    read_preference(read_mode mode, deprecated_tag);
```
derickr: Do I understand correctly that you added an overloaded method with an explicit new argument to force this version over the MONGOCXX_DEPRECATED variant above?

src/bsoncxx/oid.cpp
acmorrow: No - it lets you invoke the deprecated constructor without triggering the deprecation warning.

src/mongocxx/read_preference.hpp
acmorrow: This is a constructor, not a method, and I'm not forcing anything. I'm providing a way that users can continue to call the deprecated method without getting deprecation warnings/errors, but they need to explicitly write out 'deprecated_tag', so it is clear at the call site that they are using somethign deprecated. I can't use the same trick I used elsewhere of just appending `_deprecated` to the method name, because the name of a constructor must match the class name.

acmorrow:
@derickr - Per other comments, I'm not holding this for travis.

acmorrow:
Ping on this one @mbroadst.

I think the technique here is pretty sound - it will let us continue to use the deprecated entry points in tests, etc. until we remove them without needing to spam in macros everywhere, and it will let users continue to access the deprecated APIs even if they use -Wall -Werror, by calling the versions uglified with `_deprecated`.

I did have one possible way of doing this slightly better, which would be to actually make them even a little more ugly, and for a deprecated symbol `foo` in libbsoncxx, we would name the deprecation-warning-clean function `foo_bsoncxx_deprecated`, and for a deprecated symbol `bar` in libmongocxx, we would name it `bar_mongocxx_deprecated`. That would have the advantage that it would:

- Be consistently greppable by including the library name
- Make clear in which library the deprecation occured
- Map 1:1 with the `{BSON,MONGO}CXX_DEPRECATED` macros

What do you think?

mbroadst:
@acmorrow I think this current change suffices (no need to go the extra step an include library name), as hard as it is to swallow adding all this ugly API 😄   The names are still sufficiently greppable, and I'm not _quite_ sure how useful it is even to know whether something is deprecated in bsoncxx vs mongocxx, as you are likely (as a user) to be using them as a unit rather than in pieces.

LGTM

acmorrow:
@mbroadst - Yeah its definitely ugly, but it is good for our users. We should err on the side of hard/distasteful for us, easy for them.

PR #602, opened 2017-11-09
CXX-1464 Support using the C++17 std::string_view and std::optional

acmorrow:
This also refactors the polyfill selection to be a little saner. 

The `string_view::to_string` method that we were using off the various polyfills doesn't actually exist in C++17. To work around, I needed to introduce a new free function `bsoncxx::string::to_string` such that it will work in both C++1{1,4} and C++17.

CC @mbroadst @hanumantmk


mbroadst:
👏 thank you for this!  LGTM on first glance, but I will go over it more thoroughly after lunch today.

acmorrow:
@mbroadst - Squashed and rebased so our repaired travis build can have a go at it.

src/bsoncxx/string/to_string.hpp
```diff
@@ -0,0 +1,40 @@
+// Copyright 2015 MongoDB Inc.
```
hanumantmk: 17?

PR #603, opened 2017-11-10
CXX-1439 Get travis green again

acmorrow:
Apply CXX-1439 on the v3.0 branch.


acmorrow:
@mbroadst - No need to review this, just applying the CXX-1439 changes from master to the older branches, but doing it as a PR to verify.

PR #604, opened 2017-11-10
CXX-1439 Get travis green again

acmorrow:


acmorrow:
@mbroadst This is the v3.1 branch analogue of #603 - no need to review.

PR #605, opened 2017-11-12
CXX-1458 Don't bother closing bson subobjects in stack dtor

acmorrow:
@hanumantmk PTAL. I think this does what we need.


acmorrow:
@hanumantmk - Ping?

hanumantmk:
lgtm

PR #607, opened 2017-12-01
fix compare key name error

kevin1018:
If a document has a "de" key and a "desc" key, when searching for the "de" key, it may match the "desc" key.
So compare the key length to make sure match the right key.

### DIFF
--- a/src/bsoncxx/document/view.cpp
+++ b/src/bsoncxx/document/view.cpp
@@ -125,7 +125,7 @@ view::const_iterator view::find(stdx::string_view key) const {
 
     while (bson_iter_next(&iter)) {
         const char* ikey = bson_iter_key(&iter);
-        if (0 == strncmp(key.data(), ikey, key.size())) {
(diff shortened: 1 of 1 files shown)


acmorrow:
Could you please add a regression test for this? A test that would fail to pass if this change were not applied, but that does pass if it is? I'd recommend adding it to `src/bsoncxx/test/bson_get_values.cpp`.

acmorrow:
I've actually gone ahead and written the regression tests - I'll fold them into this commit when I merge it.

acmorrow:
I've pulled your commit in to a new PR that adds a unit test, so I'm closing this in favor of #608. Thanks for contributing to the MongoDB C++ Driver!

acmorrow:
@kevin1018 FYI your commit was merged in c80ce3e8638975baa0f82aa54e28ce416896b219. Thank you for help in testing the release candidate and resolving this issue, which was also reported here: https://jira.mongodb.org/browse/CXX-1476

PR #608, opened 2017-12-04
CXX-1476 Fix broken key lookup and add unit tests

acmorrow:
Subsumes PR #607

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2336,7 +2336,7 @@ TEST_CASE("create_index tests", "[collection]") {
 
         bool unique = options.unique().value();
         auto validate = [unique](bsoncxx::document::view index) {
-            auto expire_after = index["expireAfter"];
+            auto expire_after = index["expireAfterSeconds"];
             REQUIRE(expire_after);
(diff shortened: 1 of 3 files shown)


acmorrow:
@hanumantmk PTAL. Might want to see CXX-992 as well for background.

acmorrow:
@xdg @mbroadst @hanumantmk - This is a blocker for 3.2.0-rc1. Could one of you please take a look at it ASAP?

xdg:
@saghm PTAL

src/mongocxx/test/collection.cpp
```diff
@@ -2336,7 +2336,7 @@ TEST_CASE("create_index tests", "[collection]") {
 
         bool unique = options.unique().value();
         auto validate = [unique](bsoncxx::document::view index) {
-            auto expire_after = index["expireAfter"];
+            auto expire_after = index["expireAfterSeconds"];
```
saghm: Did the previous version only work due to the bug being fixed above, or is this change unrelated?

acmorrow: Only worked due to the bug.

acmorrow:
@saghm ping?

PR #609, opened 2018-01-13
CXX-1493 and CXX-1446

acmorrow:
- Upgrades us to C driver 1.9.2 required
- Suppresses warnings (to errors with -Werror) for calling newly deprecated C driver functions
- Removes an unused call to the GLE code that was deprecated in C driver 1.8


acmorrow:
@mbroadst and @sgolemon - PTAL. This is the mechanical work needed to get us moved to C driver 1.9.x without having deprecation warnings break our -Werror builds on evergreen and travis.

src/mongocxx/private/libmongoc_symbols.hh
```diff
@@ -29,7 +31,9 @@ MONGOCXX_LIBMONGOC_SYMBOL(bulk_operation_update_many_with_opts)
 MONGOCXX_LIBMONGOC_SYMBOL(bulk_operation_update_one_with_opts)
 MONGOCXX_LIBMONGOC_SYMBOL(cleanup)
 MONGOCXX_LIBMONGOC_SYMBOL(client_destroy)
+BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_BEGIN
 MONGOCXX_LIBMONGOC_SYMBOL(client_find_databases)
```
mbroadst: this is just a nitpick, but since there are just a few of these would it make more sense just switch over to the `with_opts` variants now?

acmorrow: I considered it, and in fact started down that path. However, the error handling model for the `with_opts` is different than that for the old variants - they aren't just drop-in replacements. It seemed better to do that for 3.3.0.

Also, this fix may be incomplete. On older compilers, we still seem to get deprecation warnings, which get promoted to errors since we build with -Werror. I think for some reason older GCC is evaluating the call through the function pointer to the deprecated function as being a usage of a deprecated thing.

https://evergreen.mongodb.com/task_log_raw/cxx_driver_ubuntu1404_debug_compile_and_test_with_shared_libs_patch_fc9a44325e03bb8bb166e814caeb23413b4ba7af_5a5a51e7e3c3316075002589_18_01_13_18_37_41/0?type=T#L1312

I'll probably just disable -Werror on that builder, but expect another update to this PR soon.

acmorrow:
@mbroadst - Had to plumb in some additional deprecation suppressions to make older GCC happy. PTAL?

PR #610, opened 2018-01-23
CXX-1488 Coverity fixes

acmorrow:
This is a grab bag of small fixes related to things that coverity flagged. LKM on slack if you want a link the report.

I'm going to drop some explanatory comments on each chunk of change.

PTAL @sgolemon, @hanumantmk

### DIFF
--- a/src/bsoncxx/array/element.hpp
+++ b/src/bsoncxx/array/element.hpp
@@ -37,8 +37,6 @@ class BSONCXX_API element : private document::element {
    public:
     element();
 
-    explicit element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t offset);
-
     using document::element::operator bool;
 
     using document::element::type;
@@ -72,6 +70,13 @@ class BSONCXX_API element : private document::element {
     using document::element::raw;
     using document::element::length;
     using document::element::offset;
+
+   private:
+    friend class view;
+
+    BSONCXX_PRIVATE explicit element(const std::uint8_t* raw,
+                                     std::uint32_t length,
+                                     std::uint32_t offset);
 };
 
 }  // namespace array
--- a/src/bsoncxx/array/view.cpp
+++ b/src/bsoncxx/array/view.cpp
@@ -59,7 +59,7 @@ view::const_iterator& view::const_iterator::operator++() {
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element = element{nullptr, 0, 0};
+        _element = element{};
     } else {
         _element = element{i.raw, i.len, i.off};
     }
--- a/src/bsoncxx/array/view.cpp
+++ b/src/bsoncxx/array/view.cpp
@@ -30,7 +30,7 @@ namespace array {
 
 namespace {
 bson_iter_t to_bson_iter_t(element e) {
-    bson_iter_t i;
+    bson_iter_t i{};
     i.raw = e.raw();
     i.len = e.length();
     i.next_off = e.offset();
--- a/src/bsoncxx/builder/core.cpp
+++ b/src/bsoncxx/builder/core.cpp
@@ -622,7 +622,9 @@ core& core::open_array() {
 core& core::concatenate(const bsoncxx::document::view& view) {
     if (_impl->is_array()) {
         bson_iter_t iter;
-        bson_iter_init_from_data(&iter, view.data(), view.length());
+        if (!bson_iter_init_from_data(&iter, view.data(), view.length())) {
+            throw bsoncxx::exception{error_code::k_cannot_append_document};
+        }
 
         while (bson_iter_next(&iter)) {
             stdx::string_view key = _impl->next_key();
--- a/src/bsoncxx/document/element.cpp
+++ b/src/bsoncxx/document/element.cpp
@@ -44,7 +44,7 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace document {
 
-element::element() : _raw(nullptr), _length(0), _offset(0) {}
+element::element() : element(nullptr, 0, 0) {}
 
 element::element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t offset)
     : _raw(raw), _length(length), _offset(offset) {}
@@ -52,22 +52,12 @@ element::element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t of
 const std::uint8_t* element::raw() const {
     return _raw;
 }
-void element::raw(const std::uint8_t* raw) {
-    _raw = raw;
-}
 
 std::uint32_t element::length() const {
     return _length;
 }
-void element::length(std::uint32_t length) {
-    _length = length;
-}
-
 std::uint32_t element::offset() const {
     return _offset;
-}
-void element::offset(std::uint32_t offset) {
-    _offset = offset;
 }
 
 bsoncxx::type element::type() const {
--- a/src/bsoncxx/document/element.hpp
+++ b/src/bsoncxx/document/element.hpp
@@ -79,20 +79,6 @@ class BSONCXX_API element {
     element();
 
     ///
-    /// Construct an element as an offset into a buffer of bson bytes.
-    ///
-    /// @param raw
-    ///   A pointer to the raw bson bytes.
-    ///
-    /// @param length
-    ///   The size of the bson buffer.
-    ///
-    /// @param offset
-    ///   The element's offset into the buffer.
-    ///
-    explicit element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t offset);
-
-    ///
     /// Conversion operator to bool which is true for valid elements.
     ///
     explicit operator bool() const;
@@ -105,14 +91,6 @@ class BSONCXX_API element {
     const std::uint8_t* raw() const;
 
     ///
-    /// Setter for the raw bson bytes the element points to.
-    ///
-    /// @param raw
-    ///   The bytes this element should point to.
-    ///
-    void raw(const std::uint8_t* raw);
-
-    ///
     /// Getter for length of the raw bson bytes the element points to.
     ///
     /// @return a pointer to the length of the raw bson bytes.
@@ -120,27 +98,11 @@ class BSONCXX_API element {
     std::uint32_t length() const;
 
     ///
-    /// Setter for length of the raw bson bytes the element points to.
-    ///
-    /// @param length
-    ///   The length of the bytes this element points to.
-    ///
-    void length(std::uint32_t length);
-
-    ///
     /// Getter for the offset into the raw bson bytes the element points to.
     ///
     /// @return the offset into the raw bson bytes.
     ///
     std::uint32_t offset() const;
-
-    ///
-    /// Setter for the offset into the raw bson bytes the element points to.
-    ///
-    /// @param offset
-    ///   The offset into the bytes this element points to.
-    ///
-    void offset(std::uint32_t offset);
 
     ///
     /// Getter for the type of the element.
@@ -390,6 +352,25 @@ class BSONCXX_API element {
     array::element operator[](std::uint32_t i) const;
 
    private:
+    ///
+    /// Construct an element as an offset into a buffer of bson bytes.
+    ///
+    /// @param raw
+    ///   A pointer to the raw bson bytes.
+    ///
+    /// @param length
+    ///   The size of the bson buffer.
+    ///
+    /// @param offset
+    ///   The element's offset into the buffer.
+    ///
+    BSONCXX_PRIVATE explicit element(const std::uint8_t* raw,
+                                     std::uint32_t length,
+                                     std::uint32_t offset);
+
+    friend class view;
+    friend class array::element;
+
     const std::uint8_t* _raw;
     std::uint32_t _length;
     std::uint32_t _offset;
--- a/src/bsoncxx/document/view.cpp
+++ b/src/bsoncxx/document/view.cpp
@@ -58,7 +58,7 @@ view::const_iterator& view::const_iterator::operator++() {
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element = element{nullptr, 0, 0};
+        _element = element{};
     } else {
         _element = element{i.raw, i.len, i.off};
     }
--- a/src/bsoncxx/document/view.cpp
+++ b/src/bsoncxx/document/view.cpp
@@ -29,7 +29,7 @@ namespace document {
 
 namespace {
 bson_iter_t to_bson_iter_t(element e) {
-    bson_iter_t i;
+    bson_iter_t i{};
     i.raw = e.raw();
     i.len = e.length();
     i.next_off = e.offset();
--- a/src/bsoncxx/private/itoa.cpp
+++ b/src/bsoncxx/private/itoa.cpp
@@ -20,7 +20,7 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace {
 
-const char* kIndexTable =
+constexpr char kIndexTable[] =
     "0\0"
     "1\0"
     "2\0"
@@ -1024,10 +1024,16 @@ const char* kIndexTable =
 }  // namespace
 
 itoa::itoa(uint32_t val) : _val(val) {
-    init();
+    _init();
 }
 
-void itoa::init() {
+itoa& itoa::operator=(uint32_t new_val) {
+    _val = new_val;
+    _init();
+    return *this;
+}
+
+void itoa::_init() {
     if (_val < 10) {
         _str = kIndexTable + (2 * _val);
         _len = 1;
@@ -1052,30 +1058,6 @@ void itoa::init() {
         _str = _buf + i;
         _len = static_cast<std::uint8_t>(size - i);
     }
-}
-
-itoa::itoa() : itoa(0) {}
-
-itoa::itoa(const itoa& rhs) : itoa(rhs._val) {}
-
-itoa& itoa::operator=(const itoa& rhs) {
-    _val = rhs._val;
-    init();
-    return *this;
-}
-
-itoa& itoa::operator=(uint32_t i) {
-    _val = i;
-    init();
-    return *this;
-}
-
-const char* itoa::c_str() const {
-    return _str;
-}
-
-std::size_t itoa::length() const {
-    return _len;
 }
 
 BSONCXX_INLINE_NAMESPACE_END
--- a/src/bsoncxx/private/itoa.hh
+++ b/src/bsoncxx/private/itoa.hh
@@ -26,23 +26,37 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 
 class BSONCXX_TEST_API itoa {
    public:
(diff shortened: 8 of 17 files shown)


src/bsoncxx/array/element.hpp
```diff
@@ -37,8 +37,6 @@ class BSONCXX_API element : private document::element {
    public:
     element();
 
-    explicit element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t offset);
```
acmorrow: This was unsafe to have as a public constructor, and not needed. I made it private and friended in the places that needed it. Noticed while looking into the coverity report about the uninit'd field in bson_iter_t.

src/bsoncxx/array/element.hpp
```diff
@@ -72,6 +70,13 @@ class BSONCXX_API element : private document::element {
     using document::element::raw;
     using document::element::length;
     using document::element::offset;
+
+   private:
+    friend class view;
```
acmorrow: Need to friend in view now that constructor is private.

src/bsoncxx/array/view.cpp
```diff
@@ -30,7 +30,7 @@ namespace array {
 
 namespace {
 bson_iter_t to_bson_iter_t(element e) {
-    bson_iter_t i;
+    bson_iter_t i{};
```
acmorrow: This should silence coverity, but more work is required, see https://jira.mongodb.org/browse/CXX-1499

src/bsoncxx/array/view.cpp
```diff
@@ -59,7 +59,7 @@ view::const_iterator& view::const_iterator::operator++() {
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element = element{nullptr, 0, 0};
+        _element = element{};
```
acmorrow: Just an obvious cleanup.

src/bsoncxx/builder/core.cpp
```diff
@@ -622,7 +622,9 @@ core& core::open_array() {
 core& core::concatenate(const bsoncxx::document::view& view) {
     if (_impl->is_array()) {
         bson_iter_t iter;
-        bson_iter_init_from_data(&iter, view.data(), view.length());
+        if (!bson_iter_init_from_data(&iter, view.data(), view.length())) {
```
acmorrow: Coverity was pointing out that we weren't checking the return value here, so I added the logic that seemed reasonable.

src/bsoncxx/document/element.cpp
```diff
@@ -44,31 +44,21 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace document {
 
-element::element() : _raw(nullptr), _length(0), _offset(0) {}
+element::element() : element(nullptr, 0, 0) {}
```
acmorrow: Noticed while doing constructor changes - no need to re-iterate the field names, just delegate.

src/bsoncxx/document/element.cpp
```diff
@@ -44,31 +44,21 @@ namespace bsoncxx {
...
 
 element::element(const std::uint8_t* raw, std::uint32_t length, std::uint32_t offset)
     : _raw(raw), _length(length), _offset(offset) {}
 
 const std::uint8_t* element::raw() const {
     return _raw;
 }
-void element::raw(const std::uint8_t* raw) {
```
acmorrow: These three removed because removed in header, see that for details.

src/bsoncxx/document/element.hpp
```diff
@@ -78,20 +78,6 @@ class BSONCXX_API element {
     ///
     element();
 
-    ///
```
acmorrow: This constructor was unsafe to have public. Made it private and friended in the necessary things.

src/bsoncxx/document/element.hpp
```diff
@@ -104,44 +90,20 @@ class BSONCXX_API element {
     ///
     const std::uint8_t* raw() const;
 
-    ///
```
acmorrow: This and the next two functions were unsafe and unused. Too easy to tweak an element into a bad state. Element is supposed to be an immutable view anyway. What good would these do you?

src/bsoncxx/document/element.hpp
```diff
@@ -390,6 +352,25 @@ class BSONCXX_API element {
...
+    /// @param raw
+    ///   A pointer to the raw bson bytes.
+    ///
+    /// @param length
+    ///   The size of the bson buffer.
+    ///
+    /// @param offset
+    ///   The element's offset into the buffer.
+    ///
+    BSONCXX_PRIVATE explicit element(const std::uint8_t* raw,
```
acmorrow: Unsafe ctor moved here and needed friends added.

src/bsoncxx/document/view.cpp
```diff
@@ -29,7 +29,7 @@ namespace document {
 
 namespace {
 bson_iter_t to_bson_iter_t(element e) {
-    bson_iter_t i;
+    bson_iter_t i{};
```
acmorrow: As in array.cpp

src/bsoncxx/document/view.cpp
```diff
@@ -58,7 +58,7 @@ view::const_iterator& view::const_iterator::operator++() {
     bson_iter_next(&i);
 
     if (!bson_iter_next(&i)) {
-        _element = element{nullptr, 0, 0};
+        _element = element{};
```
acmorrow: ditto

src/bsoncxx/private/itoa.cpp
```diff
@@ -20,7 +20,7 @@ namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 namespace {
 
-const char* kIndexTable =
+constexpr char kIndexTable[] =
```
acmorrow: Always should have been constexpr.

src/bsoncxx/private/itoa.cpp
```diff
@@ -1024,10 +1024,16 @@ const char* kIndexTable =
 }  // namespace
 
 itoa::itoa(uint32_t val) : _val(val) {
-    init();
+    _init();
```
acmorrow: init was missing a leading underscore because it was private.

src/bsoncxx/private/itoa.cpp
```diff
@@ -1024,10 +1024,16 @@ const char* kIndexTable =
...
 }
 
-void itoa::init() {
+itoa& itoa::operator=(uint32_t new_val) {
```
acmorrow: This is just git being bad at diff - moved up from 1061 in old.

src/bsoncxx/private/itoa.cpp
```diff
@@ -1054,29 +1060,5 @@ void itoa::init() {
...
-    return *this;
-}
-
-itoa& itoa::operator=(uint32_t i) {
-    _val = i;
-    init();
-    return *this;
-}
-
-const char* itoa::c_str() const {
```
acmorrow: I inlined the accessors. Silly to call a function to get these, and itoa is package private.

src/bsoncxx/private/itoa.cpp
```diff
@@ -1054,29 +1060,5 @@ void itoa::init() {
     }
 }
 
-itoa::itoa() : itoa(0) {}
-
-itoa::itoa(const itoa& rhs) : itoa(rhs._val) {}
-
-itoa& itoa::operator=(const itoa& rhs) {
```
acmorrow: I removed the copy and assignment operators since coverity was warning about not checking for self-assign and I didn't want to deal with it. The only case we actually need to support is assigning an integer to it.

src/bsoncxx/private/itoa.hh
```diff
@@ -26,24 +26,38 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
 
 class BSONCXX_TEST_API itoa {
    public:
-    itoa();
-    itoa(uint32_t i);
-    itoa(const itoa& rhs);
-    itoa& operator=(const itoa& rhs);
-    itoa& operator=(uint32_t i);
+    explicit itoa(uint32_t i = 0);
```
acmorrow: Consolidate the ctors and make them explicit.

src/bsoncxx/private/itoa.hh
```diff
@@ -26,24 +26,38 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
 
+    itoa(const itoa& rhs) = delete;
```
acmorrow: Copying one of these doesn't make much sense.

src/bsoncxx/private/itoa.hh
```diff
@@ -26,24 +26,38 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+    itoa& operator=(const itoa&) = delete;
+
+    itoa& operator=(uint32_t new_value);
```
acmorrow: But assigning it a new value does.

src/bsoncxx/private/itoa.hh
```diff
@@ -26,24 +26,38 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
-    BSONCXX_PRIVATE void init();
+    BSONCXX_PRIVATE void _init();
 
     uint32_t _val;
     const char* _str;
     uint8_t _len;
     char _buf[11];
 };
 
+inline uint32_t itoa::val() const {
```
acmorrow: The accessors seem sensible to inline.

src/mongocxx/collection.hpp
```diff
@@ -19,6 +19,7 @@
 #include <memory>
 #include <string>
 
+#include <bsoncxx/builder/basic/array.hpp>
```
acmorrow: Fixing up the headers for index_view revealed that this header was missing from collection.hpp

src/mongocxx/collection.hpp
```diff
@@ -717,7 +718,7 @@ class MONGOCXX_API collection {
 
     ///
     /// Gets an index_view to the collection.
-    class index_view indexes();
+    index_view indexes();
```
acmorrow: No need for an elaborated type specifier here, was probably cargo-cult'ed in over the summer.

src/mongocxx/index_view.hpp
```diff
@@ -17,17 +17,9 @@
 #include <string>
 #include <vector>
 
-#include <bsoncxx/builder/basic/array.hpp>
```
acmorrow: I have no idea why all these includes where here but they aren't needed.

src/mongocxx/index_view.cpp
```diff
@@ -28,10 +28,10 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 index_view::index_view(void* coll)
     : _impl{stdx::make_unique<impl>(static_cast<mongoc_collection_t*>(coll))} {}
 
+index_view::index_view(index_view&&) noexcept = default;
```
acmorrow: Index view had a move constructor but no move assign, which made coverity sad.

src/mongocxx/index_view.hpp
```diff
@@ -38,14 +30,11 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 
 class MONGOCXX_API index_view {
    public:
-    index_view() = delete;
-
-    index_view(const index_view&) = delete;
+    index_view(index_view&&) noexcept;
+    index_view& operator=(index_view&&) noexcept;
```
acmorrow: Give index_view both move ctor and move assign, and make them properly noexcept.

src/mongocxx/options/client.cpp
```diff
@@ -21,7 +21,7 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
 client& client::ssl_opts(ssl ssl_opts) {
-    _ssl_opts = ssl_opts;
+    _ssl_opts = std::move(ssl_opts);
```
acmorrow: Coverity pointed out that this was a large parameter by value, which is true. Usually I don't care though because we are moving from the argument. In this case we weren't, so I fixed it.

src/mongocxx/private/collection.hh
```diff
@@ -42,11 +42,13 @@ class collection::impl {
           client_impl{i.client_impl} {}
 
     impl& operator=(const impl& i) {
-        libmongoc::collection_destroy(collection_t);
-        collection_t = libmongoc::collection_copy(i.collection_t);
+        if (this != &i) {
```
acmorrow: Coverity complained that this type wasn't protected against self assignment. It actually shouldn't be possible to ever self assign these, since they are always owned by unique_ptr, but just in case that rule ever changed, it seemed pretty harmless to add it. The code was definitely wrong if you did manage a self assignment.

src/mongocxx/private/database.hh
```diff
@@ -36,10 +36,12 @@ class database::impl {
           name{i.name} {}
 
     impl& operator=(const impl& i) {
-        libmongoc::database_destroy(database_t);
-        database_t = libmongoc::database_copy(i.database_t);
-        client_impl = i.client_impl;
-        name = i.name;
+        if (this != &i) {
```
acmorrow: Ditto re coverity and self assignment.

src/mongocxx/private/index_view.hh
```diff
@@ -16,9 +16,14 @@
 
 #include <vector>
 
+#include <bsoncxx/builder/basic/array.hpp>
```
acmorrow: After removing the needless headers from the public index_view header, more were required in the private impl.

src/mongocxx/test/collection.cpp
```diff
@@ -366,8 +366,8 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
                     // copyIds[idx] doesn't exist, but ids[idx] does.
                     REQUIRE(rhsIdVal != rhs.end());
 
-                    const auto &lhsVal = lhsIdVal.second;
-                    const auto &rhsVal = rhsIdVal->second;
+                    const auto& lhsVal = lhsIdVal.second;
```
acmorrow: Lint fixup from prior commit.

src/mongocxx/test/gridfs/bucket.cpp
```diff
@@ -502,7 +502,7 @@ TEST_CASE("mongocxx::gridfs::downloader::read with arbitrary sizes", "[gridfs::d
 
     std::int64_t file_length = 100;
     std::int32_t chunk_size = 9;
-    std::int32_t read_size;
+    std::int32_t read_size = 0;
```
acmorrow: Coverity seemed not to be able to see that we always assigned this in one of the sections. Fair enough. Init it to zero, and then require that it be non-zero before we start the test body.

src/bsoncxx/array/element.hpp
```diff
@@ -72,6 +70,13 @@ class BSONCXX_API element : private document::element {
...
+
+    BSONCXX_PRIVATE explicit element(const std::uint8_t* raw,
```
mathbunnyru: I think "explicit" is not needed here.

src/bsoncxx/array/view.cpp
mathbunnyru: Maybe it is even possible to write element = {}; ?

src/bsoncxx/document/element.hpp
mathbunnyru: "explicit" is not needed

src/bsoncxx/document/view.cpp
mathbunnyru: maybe _element = {}; is possible, I'm not sure

src/bsoncxx/private/itoa.cpp
mathbunnyru: Nice one

acmorrow:
+ @kevinAlbs and @mbroadst

src/bsoncxx/array/view.cpp
mbroadst: weird, they should be equivalent?

src/mongocxx/options/client.cpp
mbroadst: shouldn't this be a const ref?

src/mongocxx/private/collection.hh
mbroadst: why not just be opinionated about this and delete the operator outright?

src/mongocxx/test/collection.cpp
mbroadst: /me wonders if we should be linting as part of the travis job

mbroadst:
LGTM, just a few clarifying questions in there

src/bsoncxx/array/view.cpp
acmorrow: In C++ yes, but bson_iter_t is a C struct, so a POD. Fields are not default initialized, so we get whatever garbage is on the stack here in any fields that we don't immediately set. This at least zero-inits any we don't explicitly set. That is still probably wrong, but less so, hence CXX-1499

src/bsoncxx/array/view.cpp
acmorrow: Maybe. I'm not too inclined to change it though.

src/bsoncxx/document/element.hpp
acmorrow: Interestingly, in C++11, explicit does have semantics on multi-argument ctors: it inhibits copy initialization from an initializer list:

https://stackoverflow.com/questions/39122101/explicit-constructor-taking-multiple-arguments/39122237

I'll need to think about what that means for the project internal implementation, but as the function is now private it can't negatively affect the user facing API. Good looking out though. I'll leave it for now until I make a decision about how I feel about the C++11 semantics.

src/bsoncxx/document/view.cpp
acmorrow: Probably, not worth changing IMHO.

src/bsoncxx/private/itoa.cpp
acmorrow: thx.

src/mongocxx/options/client.cpp
acmorrow: No, it doesn't need to be. We are taking the argument by value because no matter what we are making at least one copy. If we took it by const-ref, we would copy assign into `_ssl_opts`. As written, we get one copy (into the parameter) and one move (into `_ssl_opts`). However, if at the call site the `ssl` object is a temporary, then we actually get two moves and no copies.

src/mongocxx/private/collection.hh
acmorrow: Ah, because it does in fact get get invoked. Its just that I know that it will never get invoked as a self-assignment.

src/mongocxx/test/collection.cpp
acmorrow: Yeah... its weird. I really thought we used to?

src/bsoncxx/array/element.hpp
acmorrow: Same comment as above.

acmorrow:
Thanks @mbroadst.

src/bsoncxx/document/element.hpp
mathbunnyru: Thank you for the explanation.

PR #612, opened 2018-03-26
cmake: honor LIB_SUFFIX

timn:
On some systems, e.g., Fedora, libraries are installed to /usr/lib64. A typical pattern is to honor a variable named LIB_SUFFIX. It is common and specified by default in the Fedora build system if necessary.

This patch extends the installation paths from "lib" to "lib${LIB_SUFFIX" in order to allow for proper packaging and installation.

The suffix can be set as "-DLIB_SUFFIX=64" and is ignored if not set.

### DIFF
--- a/src/bsoncxx/CMakeLists.txt
+++ b/src/bsoncxx/CMakeLists.txt
@@ -216,12 +216,12 @@ install(FILES
 install(TARGETS
     bsoncxx
     RUNTIME DESTINATION bin COMPONENT runtime
-    LIBRARY DESTINATION lib COMPONENT runtime
-    ARCHIVE DESTINATION lib COMPONENT dev
+    LIBRARY DESTINATION lib${LIB_SUFFIX} COMPONENT runtime
+    ARCHIVE DESTINATION lib${LIB_SUFFIX} COMPONENT dev
 )
 
 set(PACKAGE_INCLUDE_INSTALL_DIRS ${BSONCXX_HEADER_INSTALL_DIR})
-set(PACKAGE_LIBRARY_INSTALL_DIRS lib)
+set(PACKAGE_LIBRARY_INSTALL_DIRS lib${LIB_SUFFIX})
 set(PACKAGE_LIBRARIES bsoncxx)
 
 include(CMakePackageConfigHelpers)
@@ -234,7 +234,7 @@ endif()
 
 configure_package_config_file(
   cmake/${PKG}-config.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/${PKG}-config.cmake
-  INSTALL_DESTINATION lib/cmake/${PKG}-${BSONCXX_VERSION}
+  INSTALL_DESTINATION lib${LIB_SUFFIX}/cmake/${PKG}-${BSONCXX_VERSION}
   PATH_VARS PACKAGE_INCLUDE_INSTALL_DIRS PACKAGE_LIBRARY_INSTALL_DIRS
 )
 
@@ -246,7 +246,7 @@ write_basic_package_version_file(
 
 install(
   FILES ${CMAKE_CURRENT_BINARY_DIR}/${PKG}-config.cmake ${CMAKE_CURRENT_BINARY_DIR}/${PKG}-config-version.cmake
-  DESTINATION lib/cmake/${PKG}-${BSONCXX_VERSION}
+  DESTINATION lib${LIB_SUFFIX}/cmake/${PKG}-${BSONCXX_VERSION}
 )
 
 add_subdirectory(test)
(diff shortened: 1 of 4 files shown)


acmorrow:
@timn - Thanks for the PR.

I'm actually a little surprised to find that CMake doesn't already have a way to do this, but it appears that it does not. It looks like at least LLVM and all other projects do this their own way too.

I think I'd like to namespace these variables a little harder though so that we don't collide with other CMake projects if/when embedded. Maybe `BSONCXX_LIBDIR_SUFFIX` and `MONGOCXX_LIBDIR_SUFFIX`, one for each library? On the other hand, they will never vary independently in practice. Perhaps `MONGO_CXX_DRIVER_LIBDIR_SUFFIX` if it is going to sit at the top level, above `mongocxx` and `bsoncxx`? We don't currently have a convention for top level defines, and the `MONGO_CXX_DRIVER` prefix seems plausible.

timn:
I chose the variable name that I already knew from other packages and the Fedora build system (cf. https://src.fedoraproject.org/rpms/cmake/blob/master/f/macros.cmake).

So, even if you were to choose custom names, I would still set them based on LIB_SUFFIX. Just because it is so prevalent. Just to give on other example cf. [OpenRAVE](https://github.com/rdiankov/openrave/blob/master/CMakeLists.txt). There, they made it an explicit set in the top file. Is that something that would be giving it enough visibility to go with the de-facto standard instead of mongocxx-specific macros?

Given this prevalence, even if you want to avoid clashes internally in mongo-cxx, it would probably still lead to unexpected results as LIB_SUFFIX is, in fact, set from the outside, as in the example of the Fedora build system. Therefore, I would see using something different even as more "dangerous".

acmorrow:
@timn - Interesting points.

Regarding packaging mongo-cxx-driver for Fedora, this has been something on our long term roadmap. However, we have not yet started to pursue it as we do not consider the mongocxx driver to have reached ABI stability yet. You may notice that the ABI version for the install is currently `v_noabi`. Something to consider.

timn:
I have a working package for Fedora. You can find it at [timn/mongo-cxx-driver copr](https://copr.fedorainfracloud.org/coprs/timn/mongo-cxx-driver/). It's working in the sense that I successfully ran "make test" after building it against a local MongoDB 3.4 instance. I haven't actually updated all of our code in [Fawkes](https://www.fawkesrobotics.org), yet, to test this more extensively.

I think it's important to have it out there today as more and more code currently get's written for the legacy driver and it's ever harder to eventually replace it. The copr is optional, so people can start playing with it, as we will. Hopefully killing a particularly nasty spurious uncaught exception we experience with the old driver.

acmorrow:
@timn - I wonder: should we instead be using https://cmake.org/cmake/help/git-master/module/GNUInstallDirs.html

timn:
That loos like a viable alternative. Since it's just about the install paths, patching in LIB_SUFFIX is the typical thing I end up doing as a minimal patch when packaging. That being said, I welcome a responsive upstream and I'm happy to test a patch to see if it yields the desired results on Fedora. 

It is worth noting that I am not the maintainer of the package in Fedora (but of a package that depends on it). Once I had a chance to test my package in more detail I'd send a heads-up to propose an upgrade of the core package. Since this is a breaking change (unless keeping a compat legacy package around) it will probably take some time.

acmorrow:
@timn - I've filed https://jira.mongodb.org/browse/CXX-1540 for us to evaluate GNUInstallDirs, and I'll let you know when we have something you can test out. I think it is the right way to handle the issue.

acmorrow:
@timn - Can you give this branch a try: https://github.com/acmorrow/mongo-cxx-driver/tree/CXX-1540

It plumbs in GNUInstallDirs, and you can set the installation path via `-DCMAKE_INSTALL_LIBDIR=lib64` while running CMake. 

Can you see if it fixes the issue for you? We will also do some testing on our side. In particular, I want to know how it works on windows.

acmorrow:
@timn - Circling back to this. Were you able to try out the branch using `GNUInstallDirs`? Did it work for you? Would it be sufficient to address your packaging issues? I'm extra curious about this because if this proves to be the right thing, I want to encourage the MongoDB C driver CMake build to migrate to it as well.

timn:
@acmorrow - sorry, didn't get to this, yet. I'll try it later today. Isn't the C driver using autotools?

timn:
@acmorrow -  I have checked the branch and it compiles and installs properly without my patch. With GNUInstallDirs being part of the default cmake install that seems to be the way to go. I'll update my copr repo once you release 3.3. Thanks for looking into this!

acmorrow:
Thanks @timn for checking it out. I'll work on getting in for 3.3. Regarding the C driver, it currently has both autotools and CMake build systems integrated. However, the autotools build will be removed in favor of CMake in an upcoming release.

acmorrow:
Fixed in 20cdd0f5269ca06b3c215ccd679fbb4c98f9414b

PR #615, opened 2018-04-13
OpenSSL 1.1.x compilation fixes for "legacy" branch

kedazo:
Hello there,

Yes I am aware about that this branch shouldn't be used anylonger, however i made some small mods which makes this compilable with OpenSSL 1.1.x.
These are very little mods, maybe someone is interested.

Br,
David


acmorrow:
Merging this would just encourage people to continue to use the legacy driver as they find their OpenSSL versions upgraded, so I don't think we should merge it. Still, I appreciate your mentioning it on the associated ticket, as that should get people here.

On the other hand, I'd rather not leave PRs open forever, and I'm not merging it because, as you note, the driver is EOL. So, I think it is best to close it, but I think you have left enough breadcrumbs that doing so won't remove its value. Let me know if you strongly disagree.

acmorrow:
Closing, driver is EOL.

Climax777:
Thanks for this. Resurrected an old project and this came in handy

PR #616, opened 2018-04-16
CXX-1553 Fix wrong call to b_timestamp ctor

AsT4re:
In BSON document element class, the get_timestamp() accessor create a new b_timestamp object but with parameters timestamp and increment in the wrong order.


acmorrow:
LGTM. @ajdavis could you PTAL?

acmorrow:
@ajdavis  - Yeah, there is some discussion on the ticket. I think we should make a new type for it that declares the fields as private, in the correct order, and then takes the timestamp part as a `std::chrono::seconds` type.

ajdavis:
Thanks, I've opened CXX-1555 to improve this.

PR #624, opened 2018-08-13
Update CMakeLists.txt

cos-public:
Cl compiler reports __cplusplus == 199711L for any /std option (https://developercommunity.visualstudio.com/content/problem/120156/-cplusplus-macro-still-defined-as-pre-c11-value.html). To report correct standard version /Zc:__cplusplus should be set, otherwise \src\bsoncxx\stdx\make_unique.hpp (at least) fails to compile, as it relies on version in __cplusplus.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,12 @@ endif()
 
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Force MS compiler to report correct version in __cplusplus define
+# to be able to compile without boost polyfill
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:__cplusplus")
+endif()
+
 # Include the required modules
 include(GenerateExportHeader)
 include(InstallRequiredSystemLibraries)


acmorrow:
Do compilers older than VS 2017 honor `/Zc:__cplusplus`? If not, I think this would break those builds. Additionally, this isn't sufficient, as application of `/Zc:__plusplus` is also needed in any consumer of libbsoncxx in these same cases, so I think this isn't sufficient either.

Still, it seems like an edge case that should be fixed, so I'm open to ideas.

cos-public:
As for the first point: I don't think it makes sense to define /Zc:__cplusplus for older versions of VS, so may be it could be ifdef'ed in CMakeLists for VS >= 2017 (I'm not that familiar with cmake to define it by myself using your coding standards).
As for the second point: I compiled and used mongocxx successfully in my test project without defining /Zc:__cplusplus. Probably I didn't require stuff from stdx\make_unique.hpp, and I don't know whether it's supposed to be used in applications. May be you could do some additional check in make_unique.hpp? Or even leave it as it is, just add some hints in error message about /Zc:__cplusplus for those who would want to use this header.

kevinAlbs:
Hi @cos-public, thank you for the pull request! We've addressed this in https://jira.mongodb.org/browse/CXX-1613. We now document [here](http://mongocxx.org/mongocxx-v3/installation/#step-4-configure-the-driver):

> For building with Visual Studio 2017 (without a C++17 polyfill), it is necessary to configure with an additional option, /Zc:__cplusplus to opt into the correct definition of __cplusplus

I'll close this because I don't think there's any outstanding action items.

Best,
Kevin

PR #633, opened 2018-10-05
fix allow_invalid_certificates documentation

gchebanov:
```c++
    ///
    /// If false, the driver will not verify the server's CA file.
    ///
    /// @param allow_invalid_certificates
    ///   Whether or not to check the server's CA file.
    ///
    /// @return
    ///   A reference to the object on which this member function is being called.  This facilitates
    ///   method chaining.
    ///
    ssl& allow_invalid_certificates(bool allow_invalid_certificates);
```
Should be "If true, the driver will not verify the server's CA file."


kevinAlbs:
Thanks for the contribution @gchebanov

PR #638, opened 2018-11-07
CXX-1688 Fix building with GCC 8.x

tamaroth:
When returning `optional<T>` from within the `transaction::impl` methods, the constructor that takes the implementation of the concern/preference must be public.


kevinAlbs:
This looks great! Can you run our clang-format script and update? 
```
python ./etc/clang-format.py format
```

EHadoux:
Hello, are you planning to make a new release containing this commit soon? Just so I know if I'd rather commit to using the git repo or if I should wait a bit.
Thanks!

PR #639, opened 2018-11-07
CXX-1689 Fix string_view to string conversion.

tamaroth:


kevinAlbs:
Thank you for the contribution @tamaroth!

PR #640, opened 2019-01-05
CXX-1711 Add CMake flags to build both shared and static libraries

bcaimano:


CMakeLists.txt
```diff
@@ -75,6 +75,56 @@ include(InstallRequiredSystemLibraries)
...
+else()
+    # Give a fatal error if we have a non-sensical combination
+    if(BUILD_SHARED_AND_STATIC_LIBS)
+        message(FATAL_ERROR
+                "BUILD_SHARED_LIBS is OFF but BUILD_SHARED_AND_STATIC_LIBS is ON. \
+To build static libraries only, set both BUILD_SHARED_LIBS and BUILD_SHARED_AND_STATIC_LIBS to OFF"
+        )
+    endif()
+
+    set(BSONCXX_BUILD_SHARED OFF CACHE BOOL "")
```
bcaimano: I should really change these ones to INTERNAL as well.

CMakeLists.txt
```diff
@@ -75,6 +75,56 @@ include(InstallRequiredSystemLibraries)
...
+
+# Allow the user to decide whether to use shared libraries or static libraries
+# for the mongo-c-driver
+option(BUILD_SHARED_LIBS_WITH_STATIC_MONGOC
+       "Use static mongo-c-driver libraries with shared mongo-cxx-driver libraries"
+        OFF
+)
+
+# Disambiguate our options into clear flags
+if(BUILD_SHARED_LIBS)
```
bcaimano: So I added two external options. However, I wanted to make sure I maintained existing behavior. So I also added these six INTERNAL variables to explicitly turn on and off features. If this is too messy, I can definitely hide it in a cmake module.

src/bsoncxx/CMakeLists.txt
```diff
@@ -117,85 +117,72 @@ if (BSONCXX_POLY_USE_BOOST)
...
-)
-add_library(bsoncxx_testing
-    ${bsoncxx_sources}
-)
-target_compile_definitions(bsoncxx_testing PUBLIC BSONCXX_TESTING)
-
-set_target_properties(bsoncxx_testing PROPERTIES
-    OUTPUT_NAME bsoncxx-testing
-)
+function(bsoncxx_add_library TARGET OUTPUT_NAME LINK_TYPE)
```
bcaimano: Since nearly the same target is created two or three times, I wrapped the creation in a CMake function.

src/bsoncxx/CMakeLists.txt
```diff
@@ -210,40 +197,46 @@ install(FILES
...
 set(PACKAGE_LIBRARIES bsoncxx)
 
 include(CMakePackageConfigHelpers)
 
-if (BUILD_SHARED_LIBS)
-    set(PKG "libbsoncxx")
-else()
-    set(PKG "libbsoncxx-static")
-endif()
+function(bsoncxx_install TARGET)
```
bcaimano: Like with `bsoncxx_add_library`, here I'm making sure we do the same pattern for both installs via a CMake function

CMakeLists.txt
kevinAlbs: I like it – the six internal variables clarify the logic.

CMakeLists.txt
kevinAlbs: Please do.

src/bsoncxx/CMakeLists.txt
kevinAlbs: Nice!

CMakeLists.txt
```diff
@@ -75,6 +75,56 @@ include(InstallRequiredSystemLibraries)
...
+    if(BUILD_SHARED_AND_STATIC_LIBS)
+        set(BSONCXX_BUILD_STATIC ON CACHE INTERNAL "")
+        set(MONGOCXX_BUILD_STATIC ON CACHE INTERNAL "")
+    else()
+        set(BSONCXX_BUILD_STATIC OFF CACHE INTERNAL "")
+        set(MONGOCXX_BUILD_STATIC OFF CACHE INTERNAL "")
+    endif()
+
+    if(BUILD_SHARED_LIBS_WITH_STATIC_MONGOC)
+        set(BSONCXX_LINK_STATIC ON CACHE INTERNAL "")
```
kevinAlbs: Since this only applies to linking with libmongoc/libbson, can we rename this internal variable to:
`BSONCXX_LINK_WITH_STATIC_MONGOC`? And similarly `MONGOCXX_LINK_WITH_STATIC_MONGOC`?

src/mongocxx/CMakeLists.txt
```diff
@@ -135,78 +135,71 @@ include_directories(
...
-    target_compile_definitions(mongocxx PUBLIC MONGOCXX_STATIC)
-    target_compile_definitions(mongocxx_mocked PUBLIC MONGOCXX_STATIC)
+    endif()
 
-    set_target_properties(mongocxx PROPERTIES
-        OUTPUT_NAME mongocxx-static
-    )
-endif()
+    if(LINK_TYPE STREQUAL "STATIC")
+        # Nothing
```
kevinAlbs: Do we need `target_compile_definitions(${TARGET} PUBLIC MONGOCXX_STATIC)`?

CMakeLists.txt
bcaimano: Sure, happy to do so.

src/mongocxx/CMakeLists.txt
bcaimano: So we don't actually use that symbol internally as far as I can tell, we're handing that to other projects that include our headers via pkg-config or cmake config. That said, it really doesn't hurt to do it. (I think I missed it when I did my original conversion.)

bcaimano:
So I noticed that that section of docs could be a bit easier to follow, I've add a couple of commits that rephrase it with a lot of hopefully drop-in-terminal shell code snippets. Apologies for the extra stuff to look over, I can't quite help myself when it comes to technical prose.

docs/content/mongocxx-v3/installation.md
```diff
@@ -118,92 +118,128 @@ obtain.
...
 ```
-cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local ..
+cmake ..                                \
+    -DCMAKE_BUILD_TYPE=Release          \
+    -DCMAKE_INSTALL_PREFIX=/usr/local
 ```
 
-`/usr/local` may be replaced with the directory that `libmongoc` was
-installed to.
+> ***NOTE*** The trailing `..` above is important!  Don't omit it.
```
rtimmons: this is now a leading `..` 😸

bcaimano: Well, it trails its line. The entire note is somewhat patronizing, you won't be able to build without some understanding of CMake.

src/bsoncxx/CMakeLists.txt
```diff
@@ -117,85 +117,72 @@ if (BSONCXX_POLY_USE_BOOST)
...
-
-    set_target_properties(bsoncxx PROPERTIES
-        OUTPUT_NAME bsoncxx-static
+    endif()
+
+    if(LINK_TYPE STREQUAL "STATIC")
+        target_compile_definitions(bsoncxx_static PUBLIC BSONCXX_STATIC)
+    endif()
+
+    if (BSONCXX_POLY_USE_MNMLSTC AND NOT BSONCXX_POLY_USE_SYSTEM_MNMLSTC)
```
rcsanchez97: Remove space between `if` and `(`, for consistency.

src/bsoncxx/CMakeLists.txt
```diff
@@ -117,85 +117,72 @@ if (BSONCXX_POLY_USE_BOOST)
...
+        add_dependencies(${TARGET} EP_mnmlstc_core)
+        ExternalProject_Get_Property(EP_mnmlstc_core source_dir)
+        target_include_directories(${TARGET} PUBLIC ${source_dir}/include)
+    elseif (BSONCXX_POLY_USE_BOOST)
```
rcsanchez97: Remove space between `elseif` and `(`, for consistency.

src/bsoncxx/config/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ install(FILES
     COMPONENT dev
 )
 
-if (BUILD_SHARED_LIBS)
+if (BSONCXX_BUILD_SHARED)
```
rcsanchez97: Remove space between `if` and `(`, for consistency.

src/mongocxx/CMakeLists.txt
```diff
@@ -135,78 +135,71 @@ include_directories(
...
 else()
-    target_link_libraries(mongocxx bsoncxx ${libmongoc_libraries})
-    target_link_libraries(mongocxx_mocked bsoncxx ${libmongoc_libraries})
+    set(MONGOCXX_LIBRARY_FOR_EXAMPLES "mongocxx_static" CACHE INTERNAL "")
+    mongocxx_add_library(mongocxx_mocked "mongocxx-mocked" STATIC)
+endif()
+target_link_libraries(mongocxx_mocked PUBLIC bsoncxx_testing)
+target_compile_definitions(mongocxx_mocked PUBLIC MONGOCXX_TESTING)
+
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
rcsanchez97: Remove space between `if` and `(`, for consistency.

src/bsoncxx/CMakeLists.txt
rcsanchez97: This function, along with `bsoncxx_install` should be factored out into a separate file, perhaps `cmake/BsoncxxUtil.cmake`, which can then be included with `include(BsoncxxUtil)` near the top of this `CMakeLists.txt`.

src/bsoncxx/CMakeLists.txt
rcsanchez97: Same comment as with `bsoncxx_add_library`.

src/mongocxx/CMakeLists.txt
```diff
@@ -135,78 +135,71 @@ include_directories(
...
-target_compile_definitions(mongocxx_mocked PUBLIC MONGOCXX_TESTING)
-
-if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    target_compile_options(mongocxx_mocked PRIVATE /bigobj)
-endif()
-
-set_target_properties(mongocxx_mocked PROPERTIES
-    OUTPUT_NAME mongocxx-mocked
-)
+function(mongocxx_add_library TARGET OUTPUT_NAME LINK_TYPE)
```
rcsanchez97: Apply a similar pattern here, as I suggested with `bsoncxx_add_library` and `bsoncxx_install`.

src/mongocxx/CMakeLists.txt
```diff
@@ -220,41 +213,47 @@ install(FILES
...
 set(PACKAGE_LIBRARIES mongocxx)
 
 include(CMakePackageConfigHelpers)
 
-if (BUILD_SHARED_LIBS)
-    set(PKG "libmongocxx")
-else()
-    set(PKG "libmongocxx-static")
-endif()
+function(mongocxx_install TARGET)
```
rcsanchez97: Apply a similar pattern here, as I suggested with `bsoncxx_add_library` and `bsoncxx_install`.

src/bsoncxx/CMakeLists.txt
bcaimano: @rcsanchez97, sure, happy to do so.

bcaimano:
Okay, @rcsanchez97, @kevinAlbs, I think I've addressed the last round of comments.

docs/content/mongocxx-v3/installation.md
```diff
@@ -118,92 +118,126 @@ obtain.
...
+    -DCMAKE_BUILD_TYPE=Release                      \
+    -DCMAKE_PREFIX_PATH=/opt/mongo-c-driver         \
+    -DCMAKE_INSTALL_PREFIX=/opt/mongo-cxx-driver
 ```
 
-`/usr/local` may be replaced with the directory that `libmongoc` was
-installed to.
+> *Note* If you need multiple paths in a CMake PATH variable, separate them with a semicolon like
+> this:
+> `-DCMAKE_PREFIX_PATH="/your/cdriver/prefix;/some/other/path"`
```
kevinAlbs: Oh, that's a nice reminder.

docs/content/mongocxx-v3/installation.md
```diff
@@ -118,92 +118,126 @@ obtain.
...
+Instead of the `-DCMAKE_PREFIX_PATH` option, users must specify the `libmongoc` installation
+directory by using the `-DLIBMONGOC_DIR` and `-DLIBBSON_DIR` options:
 
 ```sh
-cmake -DCMAKE_BUILD_TYPE=Release -DBSONCXX_POLY_USE_BOOST=1 \
-    -DCMAKE_INSTALL_PREFIX=/usr/local ..
+cmake ..                                            \
+    -DCMAKE_BUILD_TYPE=Release                      \
+    -DLIBMONGOC_DIR=/opt/mongo-c-driver             \
+    -DLIBBSON_DIR=/opt/mongo-c-dirver               \
```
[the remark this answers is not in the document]
bcaimano: Ohhhh man, that's funny. Yeah, should be fixed.

kevinAlbs:
Woo, thanks!

docs/content/mongocxx-v3/installation.md
rcsanchez97: I am a bit late on this comment, but I believe that semi-colon is Windows-specific.  For Unix-like systems a colon would be used instead.

kevinAlbs: Actually this seems to work for me (on macOS at least). https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html says it's a "Semicolon-separated list".

rcsanchez97: Interesting...I seemed to recall that it used the platform-specific separator and so use colon on Linux.  Of course, given the very cross-platform nature of CMake it is not all that surprising to find that it can intelligently use either and convert as a appropriate.  Feel free to disregard my earlier comment.

bcaimano: So it's actually the list separator in CMake. IIRC, it works like this:
```
set(A B C D)
message(FATAL_ERROR "${A}")
// Prints B;C;D
```

rcsanchez97:
That makes sense.  When I made my initial comment I missed that the
string is wrapped in quotes.  I just saw the semicolon and thought that
would confuse a Unix-y shell into thinking it was a command separator.

PR #641, opened 2019-01-30
Cast maxAwaitTimeMS value to int64_t to avoid ambiguity issues

bazineta:
Resolves CXX-1720.


kevinAlbs:
This line exceeds our 100 character limit specified in .clang-format. Can you run clang-format, we have a helper script to do so:
```
python ./etc/clang-format.py format
```

bazineta:
Inasmuch as there's nothing in the contribution guidelines regarding clang-format, and running it appears to perform a download and nothing else, if you'd explain what you want here, in simple terms, as if speaking to a small child, I'll attempt to comply.

kevinAlbs:
Thanks for the contribution!

> perform a download and nothing else
clang-format.py downloads clang-format 3.8 if you don't have it.

> there's nothing in the contribution guidelines regarding clang-format
Apologies, I'll add that now.

PR #645, opened 2019-06-21
Provide correct SOVERSION

russkel:
Currently some odd named libraries are being created: `/usr/lib/libmongocxx.so._noabi` which is causing issues with the Alpine Linux packaging system and its SO dependency tracing, and likely other distros. By using the numeric ABI 0.0.0 as SOVERSION this SO file disappears.

### DIFF
--- a/cmake/BsoncxxUtil.cmake
+++ b/cmake/BsoncxxUtil.cmake
@@ -20,7 +20,7 @@ function(bsoncxx_add_library TARGET OUTPUT_NAME LINK_TYPE)
         set_target_properties(${TARGET} PROPERTIES
             CXX_VISIBILITY_PRESET hidden
             VISIBILITY_INLINES_HIDDEN ON
-            SOVERSION ${BSONCXX_ABI_VERSION}
+            SOVERSION ${BSONCXX_VERSION_NO_EXTRA}
         )
     endif()
 
--- a/cmake/MongocxxUtil.cmake
+++ b/cmake/MongocxxUtil.cmake
@@ -20,7 +20,7 @@ function(mongocxx_add_library TARGET OUTPUT_NAME LINK_TYPE)
         set_target_properties (${TARGET} PROPERTIES
             CXX_VISIBILITY_PRESET hidden
             VISIBILITY_INLINES_HIDDEN ON
-            SOVERSION ${MONGOCXX_ABI_VERSION}
+            SOVERSION ${MONGOCXX_VERSION_NO_EXTRA}
         )
     endif()
 


acmorrow:
Hi @russkel - The use of the `noabi` tag is deliberate. When we were building the early versions of the C++ driver prototype, we wanted to set ourselves up to be able to ship libraries with a stable ABI. So we plumbed ABI information through the entire driver and build system, including things like its inline namespace, header install path, and filenames.

However, we weren't, at that time, ready to commit to maintaining a stable ABI.

We considered putting in `0` for the ABI tag as you are suggesting here. That would have resulted in a `SONAME` of `libmongocxx.so.0`, which would probably work fine for you.

The problem is that we felt that even a `0` still looked too much like declaring a stable ABI. So we opted instead to use the string `_noabi` to clearly indicate to anyone who looked at the file that our soname symlink was not declaring a stable ABI.

We do have plans to actually set our first stable ABI, hopefully in the near future. At that point, we will update to `1` and begin honoring ABI.

Meanwhile, I'm not sure exactly why having `libmongocxx.so_noabi` for the `SONAME` would break anything for Alpine. If you could provide some background on that it would be helpful. As far as I'm aware the ELF spec doesn't mandate the use of numeric values anywhere in `DT_SONAME`. It merely is recording the name under which the linker should record `DT_NEEDED` references when linking. Using a string-y value should be fine. If the Alpine linux packaging system has made an assumption that these values must be numeric, I'd suggest that that assumption should be revisited.

Finally, the change you have suggested here doesn't quite do what you want. If you want to alter the ABI number of the library, you should change `BSONCXX_ABI_VERSION` and `MONGOCXX_ABI_VERSION`. These values are properly interpolated through the build system so that things like the file include path, and, more importantly, the inline namespace, match the ABI version and `SONAME`.

russkel:
Hi @acmorrow, thanks for the detailed reply.

> We considered putting in 0 for the ABI tag as you are suggesting here. That would have resulted in a SONAME of libmongocxx.so.0, which would probably work fine for you.

Oddly I was getting a libmongocxx.so.0 lib as well as the .so._noabi lib, and then the Alpine Linux build script complained about not being able to find .so._noabi. I will have a look again at what is happening when I build and attempt to package without this patch. Initially this is what made me think something was wrong.

> Meanwhile, I'm not sure exactly why having libmongocxx.so_noabi for the SONAME would break anything for Alpine. If you could provide some background on that it would be helpful. As far as I'm aware the ELF spec doesn't mandate the use of numeric values anywhere in DT_SONAME. It merely is recording the name under which the linker should record DT_NEEDED references when linking. Using a string-y value should be fine. If the Alpine linux packaging system has made an assumption that these values must be numeric, I'd suggest that that assumption should be revisited.

I probably made the incorrect assumption that a non-numeric SOVERSION was invalid in my haste in getting the driver compiled so I could do some work ;).

> Finally, the change you have suggested here doesn't quite do what you want. If you want to alter the ABI number of the library

All I wanted was for a non-numeric SOVERSION, and this seemed to fix it, but as you say maybe the issue lies elsewhere. I will raise this with the Alpine Linux people.

Cogitri:
> The problem is that we felt that even a 0 still looked too much like declaring a stable ABI. So we opted instead to use the string _noabi to clearly indicate to anyone who looked at the file that our soname symlink was not declaring a stable ABI.

That's exactly what `0` is for though: Indicating that there's no stable ABI&API yet and that anything may change and break at any point and `here be dragons`.

> Meanwhile, I'm not sure exactly why having libmongocxx.so_noabi for the SONAME would break anything for Alpine. 

Because we detect ABI breakages by checking if the SONAME has changed. If it didn't => ABI remained the same. The `noabi` suffix doesn't tell us anything about this, making it super annoying to deal with (we'd just need to manually test it with something like ABIChecker).

acmorrow:
> > The problem is that we felt that even a 0 still looked too much like declaring a stable ABI. So we opted instead to use the string _noabi to clearly indicate to anyone who looked at the file that our soname symlink was not declaring a stable ABI.
> 
> That's exactly what `0` is for though: Indicating that there's no stable ABI&API yet and that anything may change and break at any point and `here be dragons`.
> 

If you can point me to the relevant section in the ELF standard that states that, I'll be happy to consider it. On my system, for instance, I have:

` 0x000000000000000e (SONAME)             Library soname: [libpthread.so.0]`

Should I conclude that pthreads does not offer a stable ABI?


> > Meanwhile, I'm not sure exactly why having libmongocxx.so_noabi for the SONAME would break anything for Alpine.
> 
> Because we detect ABI breakages by checking if the SONAME has changed. If it didn't => ABI remained the same (and AFAICS about all other distros so the same, like Void Linux, Gentoo, Exherbo ...). The `noabi` suffix doesn't tell us anything about this, making it super annoying to deal with (we'd just need to manually test it with something like ABIChecker).

Which you should be doing anyway (or use libabigail's utilities, etc.), because developers fail to bump the SONAME when they should, basically all the time.

My recommendation would be that you simply not package this for alpine. It doesn't have a stable ABI yet, so packaging it doesn't make sense.

rcsanchez97:
On Mon, Jul 15, 2019 at 07:32:23AM -0700, Andrew C. Morrow wrote:
> 
>    My recommendation would be that you simply not package this for alpine. It
>    doesn't have a stable ABI yet, so packaging it doesn't make sense.
> 
Something else to consider is some sort of experimental packaging or
marker.  For example, when I packaged the C++ driver for Debian I
specifically packaged it for the experimental distribution precisely
because of the lack of stable ABI.  Is something like that available or
feasible for Alpine?

Cogitri:
> Is something like that available or feasible for Alpine?

Yes, the package would first go into `edge`, the rolling release flavour of Alpine. It'd first go into a stable release with 3.11, which is due in 6 months.

> because developers fail to bump the SONAME when they should, basically all the time.

Hm, I haven't experienced that in the years I've been working on various distros, but alright. I do agree that using a proper ABI scanner _would_ be a good idea and I'll look into that in a bit though.

> If you can point me to the relevant section in the ELF standard that states that, I'll be happy to consider it. 

It doesn't necessarily say only numbers are allowed, but that the soname must be bumped upon ABI changes to make sure that no consumers of the lib use a version of the lib with an incompatible ABI [1] [2] [3]. I just really don't see the reason for the `noabi` suffix - if you want to tell devs that the ABI&API isn't stable, then make only 0.x releases, add a disclaimer to the README and to the docs, those are way better pointers than the soname. 

[1]: https://autotools.io/libtool/version.html
[2]: http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html

> The soname has the prefix ``lib'', the name of the library, the phrase ``.so'', followed by a period and a version number that is incremented whenever the interface changes

[3]: https://docs.oracle.com/cd/E19683-01/817-3677/chapter4-2-rason/index.html

Thanks

acmorrow:
> > Is something like that available or feasible for Alpine?
> 
> Yes, the package would first go into `edge`, the rolling release flavour of Alpine. It'd first go into a stable release with 3.11, which is due in 6 months.
> 
> > because developers fail to bump the SONAME when they should, basically all the time.
> 
> Hm, I haven't experienced that in the years I've been working on various distros, but alright. I do agree that using a proper ABI scanner _would_ be a good idea and I'll look into that in a bit though.

Yeah, the issue is that maintaining a correct ABI for C++ is much harder than it is for C. All sorts of small things can break it. Our goal for when we *do* set ABI for the C++ driver is to directly integrate with libabigail in our CI system, so that we know when we do break ABI.

And we expect that we *will* change ABI, fairly often. That will result in new ABI versions, hopefully as a part of our development process, without human intervention.



> 
> > If you can point me to the relevant section in the ELF standard that states that, I'll be happy to consider it.
> 
> It doesn't necessarily say only numbers are allowed, but that the soname must be bumped upon ABI changes to make sure that no consumers of the lib use a version of the lib with an incompatible ABI [1](https://autotools.io/libtool/version.html) [2](http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html) [3](https://docs.oracle.com/cd/E19683-01/817-3677/chapter4-2-rason/index.html). I just really don't see the reason for the `noabi` suffix - if you want to tell devs that the ABI&API isn't stable, then make only 0.x releases, add a disclaimer to the README and to the docs, those are way better pointers than the soname.
> 
> As of [2](http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html) :

Sure, but that is just guidance. I think, for instance, that it would be reasonable, and potentially interesting, to give your library a SONAME that was the git commit hash that set the current ABI. That would let you tag ABI changepoints in your commit history.

There also isn't anything that says that ABI values in SONAMES need to be incrementing. In fact, there might be good reasons not to do that. Seeing an ABI of 3 might suggest that it was a superset of 2, but that may in fact not even be true.


In the end, we have very deliberately done something that is admittedly a little odd here, but we did it with good intentions: we want to very clearly indicate to consumers that there is no ABI. The fact that this got on the Alpine Linux radar suggests that it had our intended effect.

Once we do set an ABI, we intend to follow the best practices here. But we aren't ready to commit to that just yet.

kevinAlbs:
Hello all, I'm closing this PR as it appears the discussion has led to the decision that we will not merge this PR, and will keep the `noabi` suffix until we set ABI. We do intend to set ABI as part of https://jira.mongodb.org/browse/CXX-804. As @acmorrow mentioned, that depends on setting up automated ABI change detection.

russkel:
Thanks @kevinAlbs. This patch will likely remain in the Alpine Linux mongocxx testing package until an ABI is set.

PR #648, opened 2019-09-12
Cmake boost include

fechanti:
When trying to compile mongocxx driver against boost 1.71 I had to update boost linking. Note that using recent cmake (e.g 3.15) seems to fix the issue by introducing backward compatibility, but I think using Boost::boost is cleaner.


hungbui20384:
Ok

PR #651, opened 2020-02-19
Enabling Travis builds for s390x

cnnaik:
Adding support for s390x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
       compiler: clang
       env: CONFIG=Debug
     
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Hi @cnnaik, thank you for the contribution! The C++ driver is already tested on s390x in Evergreen, our continuous integration testing platform: https://evergreen.mongodb.com/waterfall/cxx-driver

We generally only use our testing matrix for Travis for quick test feedback on pull-requests, and rely on Evergreen for comprehensive testing.

PR #655, opened 2020-05-29
CXX-1976 Add RetryableWriteError error labels to retryable transaction tests

bazile-clyde:


bazile-clyde:
Last evergreen results: https://evergreen.mongodb.com/version/5ed155160ae60631eb2e9484

.gitignore
```diff
@@ -1,5 +1,6 @@
 # Directories that are generated
 /tags
+cmake-build*
```
bazile-clyde, line 3: We needed this to avoid having every build file show up when we run `git status`. This is already done in libmongoc

data/transactions/test_files.txt
```diff
@@ -16,10 +18,13 @@ pin-mongos.json
...
 read-pref.json
 reads.json
+retryable-abort-errorLabels.json
 retryable-abort.json
+retryable-commit-errorLabels.json
 retryable-commit.json
 retryable-writes.json
 run-command.json
 transaction-options.json
+transaction-options-repl.json
```
kevinAlbs, line 28: Ah, nice catch! Looks like this test was missed from before.

One day we should remove these test_files.txt, and just glob the entire directory.

bazile-clyde: I agree. We have a ticket for it here: https://jira.mongodb.org/browse/CXX-1917

PR #657, opened 2020-06-01
CXX-2028 fix writeConcern population in find_and_modify

MohammadGhazanfar:
Mongo C++ Driver doesn’t populate the _writeConcern_ doc properly while sending a _find_one_and_update_ request to the mongo server. This was silently happening up until mongo 4.2 introduced stricter checks on the received payload. After which, a failure is seen.

This commit fixes _find_and_modify_ to populate _writeConcern_ in accordance with: https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/

More details: https://jira.mongodb.org/browse/CXX-2028


MohammadGhazanfar:
Thanks @bazile-clyde !

PR #658, opened 2020-06-01
CXX-1937 dos2unix: command not found, CXX-2025 changeStream err

bazile-clyde:


bazile-clyde:
Each commit is a separate fix. Both are tiny.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -303,6 +303,11 @@ bool matches(types::bson_value::view main,
     }
 
     if (main.type() == type::k_document) {
+        // the value '42' acts as placeholders for "any value"
+        if (pattern.type() == type::k_utf8 && 0 == pattern.get_utf8().value.compare("42") ) {
+                return true;
+        }
```
kevinAlbs, line 309: Run clang-format on this to fix the "lint" task on evergreen. For the C++ driver, there's a separate script:
```
python  ./etc/clang_format.py format
```

PR #659, opened 2020-06-02
CXX-2028 Add regression test for write concern

kevinAlbs:
Test explicitly passing a write concern to a collection find_one_and_* function is applied correctly.

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -32,6 +32,8 @@
 #include <mongocxx/exception/write_exception.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/pipeline.hpp>
+#include <mongocxx/private/libbson.hh>
+#include <mongocxx/private/libmongoc.hh>
 #include <mongocxx/read_concern.hpp>
 #include <mongocxx/test_util/client_helpers.hh>
 #include <mongocxx/write_concern.hpp>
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -2665,4 +2667,56 @@ TEST_CASE("bulk_write with container", "[collection]") {
...
+
+/* Regression test for CXX-2028. */
+TEST_CASE("find_and_x operations append write concern correctly", "[collection]") {
+    instance::current();
+    mongocxx::client client{uri{}};
+    mongocxx::write_concern wc;
+    wc.acknowledge_level(mongocxx::write_concern::level::k_acknowledged);
+
+    mongocxx::collection collection = client["fam_wc"]["collection"];
+    collection.drop();
```
bazile-clyde, line 2679: Just FYI, there's an uncaught `mongocxx::operation_exception` here and in a lot of the methods below. I'm not exactly sure if you'd want to handle that or leave it as is. Most of the code looks like it ignores exceptions with a few commands passed to `REQUIRE_NOTHROW`

kevinAlbs: Oh good point. I thought that might be a bug, since we should be ignoring "ns not found" errors. But the driver will not throw for "ns not found": https://github.com/mongodb/mongo-cxx-driver/blob/060b6dd5fc23eade27710a30e6a12eeecb84e38a/src/mongocxx/collection.cpp#L1196-L1203

If an exception was thrown, that would be unexpected and should fail the test. So I think this is ok.

src/mongocxx/test/collection.cpp
```diff
@@ -2665,4 +2667,56 @@ TEST_CASE("bulk_write with container", "[collection]") {
...
+    collection.insert_one(make_document(kvp("x", 1)));
+
+    stdx::optional<bsoncxx::document::value> doc;
+    /* 4.4. servers will reply with an error, causing an exception. */
+    /* find_one_and_update */
+    mongocxx::options::find_one_and_update find_one_and_update_opts;
+    find_one_and_update_opts.write_concern(wc);
+    doc = collection.find_one_and_update(make_document(),
```
samantharitter, line 2687: you should be able to just pass {} as the first parameter here as an empty filter, rather than having to call make_document()

`collection.find_one_and_replace({}, ...);`

src/mongocxx/test/collection.cpp
```diff
@@ -2665,4 +2667,56 @@ TEST_CASE("bulk_write with container", "[collection]") {
...
+    REQUIRE(doc);
+
+    /* < 4.4 servers will not return an error for unexpected fields. Add a visitor function to check
+     * manually. */
+    bool called = false;
+    auto visitor = libmongoc::find_and_modify_opts_append.create_instance();
+    visitor->visit([&](mongoc_find_and_modify_opts_t*, const bson_t* extra) {
+        bson_iter_t iter;
+        called = true;
+        REQUIRE(bson_iter_init_find(&iter, extra, "writeConcern"));
```
samantharitter, line 2712: How about adding a further check for the fields we expect to see inside of this sub-document?

src/mongocxx/test/collection.cpp
```diff
@@ -2665,4 +2667,56 @@ TEST_CASE("bulk_write with container", "[collection]") {
...
 }
+
+/* Regression test for CXX-2028. */
+TEST_CASE("find_and_x operations append write concern correctly", "[collection]") {
+    instance::current();
+    mongocxx::client client{uri{}};
+    mongocxx::write_concern wc;
+    wc.acknowledge_level(mongocxx::write_concern::level::k_acknowledged);
+
+    mongocxx::collection collection = client["fam_wc"]["collection"];
```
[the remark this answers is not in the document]
kevinAlbs: Done.

src/mongocxx/test/collection.cpp
kevinAlbs: Oh right, done.

src/mongocxx/test/collection.cpp
kevinAlbs: Done.

PR #661, opened 2020-06-02
Fix replica set tests

kevinAlbs:
This fixes the replica set task. Windows is still a mystery, but I suspect it is an issue with the hosts.


data/transactions/test_files.txt
```diff
@@ -5,7 +5,6 @@ commit.json
 count.json
 delete.json
 error-labels.json
-errors-client.json
```
kevinAlbs, line 8: This was missing from the test_files.txt list before, so we weren't running it. After it was added, it revealed that this test file was failing. I think the file was prematurely synced. It will be synced as part of https://jira.mongodb.org/browse/CXX-1799, which requires implementing the `assertSessionTransactionState` test operation. (It isn't done yet for libmongoc's test runner either).

.mci.yml
```diff
@@ -88,6 +88,7 @@ functions:
                  rm -rf "mongo-cxx-driver"
                  rm -fr "mongo-c-driver"
                  rm -fr mongod
+                 rm -fr drivers-evergreen-tools
```
kevinAlbs, line 91: I'm not sure why Windows variants only started failing now, but I saw a related failure here:

https://evergreen.mongodb.com/task/cxx_driver_windows_2k8_release_compile_and_test_with_shared_libs_060b6dd5fc23eade27710a30e6a12eeecb84e38a_20_06_02_15_58_04

src/mongocxx/test/spec/crud.cpp
```diff
@@ -30,6 +30,19 @@ using namespace mongocxx::spec;
 using bsoncxx::builder::basic::kvp;
 using bsoncxx::builder::basic::make_document;
 
+bool is_unsupported(bsoncxx::stdx::string_view description) {
+    std::set<std::string> unsupported_tests = {"Deprecated count with empty collection",
+                                               "Deprecated count with collation",
+                                               "Deprecated count without a filter",
+                                               "Deprecated count with a filter",
+                                               "Deprecated count with skip and limit"};
```
kevinAlbs, line 38: CXX-1931 removed the deprecated count helper (in favor of the non-deprecated countDocuments and estimatedDocumentCount). We cannot run these tests.

.mci.yml
```diff
@@ -293,6 +295,9 @@ functions:
                       ${test_params} ./src/mongocxx/test/test_command_monitoring_specs
                       ${test_params} ./src/mongocxx/test/test_instance
                       ${test_params} ./src/mongocxx/test/test_transactions_specs
+                      ${test_params} ./src/mongocxx/test/test_logging
+                      ${test_params} ./src/mongocxx/test/test_retryable_reads_specs
```
kevinAlbs, line 299: The danger of having many test runners is that easy miss some, either on evergreen or locally. When we have time, I'd like to simplify this to one test runner (or have one script that runs them all), and remove the `*_TESTS_PATH` environment variables.

data/transactions/test_files.txt
samantharitter: Nice detective work!

.mci.yml
samantharitter: Yes, I've seen the failure for when the directory already exists.  Seems worthwhile to add this here!  On one of my branches, I've also changed the line where we clone drivers-evergreen-tools to check if the directory exists first.  With this cleanup added, that check is probably overkill.

src/mongocxx/test/spec/crud.cpp
samantharitter: Acknowledged.

.mci.yml
samantharitter: +1 for cleaning up the test runners!

PR #662, opened 2020-06-03
Backport for CXX-2028

kevinAlbs:
Backporting CXX-2028 required some manual fix-ups. The logic around find_and_modify changed in 3.3 in https://jira.mongodb.org/browse/CXX-1565.

Evergreen: https://evergreen.mongodb.com/version/5ed7ba569ccd4e0ac28b8ab3


.evergreen/find_cmake.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
kevinAlbs, line 1: This is a copy from master in order to build the latest C driver. All C++ branches going back to 3.3 test against the latest libmongoc. But the 3.2 branch was building 1.9.2. There were test failures due to a bug that was fixed in 1.11:

https://github.com/mongodb/mongo-c-driver/commit/806cddb2acde03f5462460491f38fa21332b7d97#diff-5853ab613d7df98e940febe8c1553fc6R370

Specifically, a default write concern could get included in a command. Here's an example of one failure:
https://evergreen.mongodb.com/task/cxx_driver_v3.2_ubuntu1604_debug_asan_compile_and_test_with_shared_libs_patch_c5a6e0a1cc2876eed025c01dc61a935d8d554c53_5ed70c362a60ed42a93e81e3_20_06_03_02_34_39

IMO since we test against latest C driver on all other branches I think it makes sense to do so for the 3.2 branch too. It's always safe for users to upgrade libmongoc, and this permits us to test older C++ branches against the latest server.

.evergreen/install.sh
```diff
@@ -15,28 +15,13 @@ usage() {
 [ $# -lt 1 ] && { usage; exit 2; }
 
 VERSION=${1}
-PREFIX=${PREFIX:-$(pwd)"/deps-install"}
```
kevinAlbs, line 18: Same with this file, a straight copy to build the latest C driver (the 1.9 version of libmongoc still used autotools to build)

etc/memcheck.suppressions
```diff
@@ -8,3 +8,66 @@
    obj:*libsasl2.so*
    fun:sasl_client_init
 }
+{
```
kevinAlbs, line 11: Valgrind suppressions copied from master.

src/mongocxx/collection.cpp
```diff
@@ -91,7 +91,7 @@ mongocxx::stdx::optional<bsoncxx::document::value> find_and_modify(
         if (!wc->is_acknowledged() && opts["collation"]) {
             throw mongocxx::logic_error{mongocxx::error_code::k_invalid_parameter};
         }
-        opts_builder.append(concatenate(wc->to_document()));
+        opts_builder.append(kvp("writeConcern", wc->to_document()));
```
kevinAlbs, line 94: This is the fix. This differs from https://github.com/mongodb/mongo-cxx-driver/commit/cf4faedfa96c2b63e846a0ec169d458dfc84dbd4 since this code changed in 3.3.

.evergreen/find_cmake.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
```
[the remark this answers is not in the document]
kevinAlbs: Hah good point. I copied this file from the master branch without looking at it. I don't see it used anywhere so I'll remove it here, and on master.

.evergreen/find_cmake.sh
```diff
@@ -0,0 +1,33 @@
...
+
+find_cmake ()
+{
+  if [ ! -z "$CMAKE" ]; then
+    return 0
+  elif [ -f "/Applications/cmake-3.2.2-Darwin-x86_64/CMake.app/Contents/bin/cmake" ]; then
```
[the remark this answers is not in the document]
kevinAlbs: I'm not sure what the original intent of OS was. Since this logic works as-is I'll leave it.

kevinAlbs:
Closing and pushing to fix up commits locally.

PR #664, opened 2020-06-05
CXX-1994 add remaining URI getters

kevinAlbs:
Adds and tests getters for URI options missing from `mongocxx::uri`.


src/mongocxx/uri.cpp
```diff
@@ -129,5 +130,160 @@ class write_concern uri::write_concern() const {
...
+
+    value = libmongoc::uri_get_option_as_utf8(uri, opt_name.c_str(), NULL);
+    if (!value) {
+        return {};
+    }
+
+    return stdx::string_view{value};
+}
+
+static stdx::optional<std::int32_t> _int32_option(mongoc_uri_t* uri, std::string opt_name) {
```
kevinAlbs, line 144: libmongoc provides a `mongoc_uri_get_option_as_int32`, but I chose not to use it.

Like `mongoc_uri_get_option_as_utf8` it accepts a fallback argument to return if the option is unset. But it treats an option in a URI set to 0 as unset (e.g. `mongodb://host/?connectTimeoutMS=0`). But some options allow 0 (e.g. `localThresholdMS`, `connectTimeoutMS`).

http://mongoc.org/libmongoc/current/mongoc_uri_get_option_as_int32.html

src/mongocxx/uri.cpp
```diff
@@ -129,5 +130,160 @@ class write_concern uri::write_concern() const {
...
+    }
+    bson_iter_document(&iter, &len, &data);
+    return bsoncxx::document::view(data, len);
+}
+
+stdx::optional<stdx::string_view> uri::appname() const {
+    return _string_option(_impl->uri_t, "appname");
+}
+
+/* Special case. authMechanismProperties are stored as part of libmongoc's credentials. */
```
kevinAlbs, line 185: Ah, C style comments. I'll update.

src/mongocxx/uri.hpp
```diff
@@ -32,10 +32,13 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
 ///
 /// Class representing a MongoDB connection string URI.
 ///
-/// @todo return const char* instead or stringview
-/// @todo harmonize with C library (options, credentials, etc...)
+/// @todo Use optionals for all option getters. Otherwise users cannot distinguish between a case of
+/// an empty option explicitly present in the URI and the option being omitted. Also change getters
+/// that return a std::string to return a string_view since a copy is not necessary.
```
kevinAlbs, line 37: I synced up with the URI options spec: https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst

We already had some option getters, and some URI options are implied through other getters. E.g. `w` and `wTimeoutMS` are returned via `write_concern()`.

Looking at Java's API further convinced me to return optionals:
https://mongodb.github.io/mongo-java-driver/4.1/apidocs/mongodb-driver-core/com/mongodb/ConnectionString.html

I created https://jira.mongodb.org/browse/CXX-2038 and will paste link into this comment in the next commit.

samantharitter: Yes, I agree with both of these future changes.

src/mongocxx/uri.hpp
```diff
@@ -170,6 +181,168 @@ class MONGOCXX_API uri {
     /// @return A write_concern that represents what was specified in the uri.
     ///
     class write_concern write_concern() const;
+    ///
```
samantharitter, line 184: style nit, newline above this comment block

src/mongocxx/uri.cpp
```diff
@@ -129,5 +130,160 @@ class write_concern uri::write_concern() const {
...
 
+static stdx::optional<stdx::string_view> _string_option(mongoc_uri_t* uri, std::string opt_name) {
+    const char* value;
+
+    value = libmongoc::uri_get_option_as_utf8(uri, opt_name.c_str(), NULL);
+    if (!value) {
+        return {};
+    }
+
+    return stdx::string_view{value};
```
samantharitter, line 141: I'm surprised this doesn't need to be `return {stdx::string_view{value}};`

kevinAlbs: Huh, yeah I wasn't sure either. But I think it corresponds to constructor 8 on cppreference
https://en.cppreference.com/w/cpp/utility/optional/optional

```
template < class U = value_type >
constexpr optional( U&& value );
```

src/mongocxx/uri.hpp
kevinAlbs: Done.

src/mongocxx/test/uri.cpp
```diff
@@ -69,5 +73,273 @@ TEST_CASE("URI", "[uri]") {
...
+    uri = mongocxx::uri{"mongodb://host/?" + optname + "=false"};
+    REQUIRE(getter_fn(uri));
+    REQUIRE(false == *getter_fn(uri));
+
+    /* Present but empty. libmongoc considers it unset. */
+    uri = mongocxx::uri{"mongodb://host/?" + optname + "="};
+    REQUIRE(!getter_fn(uri));
+
+    /* Present, but numeric. */
+    uri = mongocxx::uri{"mongodb://host/?" + optname + "=1"};
```
bazile-clyde, line 132: All numeric values would result in `true` right?

src/mongocxx/uri.hpp
```diff
@@ -48,11 +51,15 @@ class MONGOCXX_API uri {
     static const std::string k_default_uri;
 
     ///
-    /// Constructs a uri from an optional MongoDB uri string. If no uri string is specified,
-    /// uses the default uri string, 'mongodb://localhost:27017'.
+    /// Constructs a uri from an optional MongoDB URI string. If no URI string is specified,
```
bazile-clyde, line 54:
```suggestion
    /// Constructs a URI from an optional MongoDB URI string. If no URI string is specified,
```

src/mongocxx/uri.hpp
```diff
@@ -171,6 +182,169 @@ class MONGOCXX_API uri {
...
+    ///
+    /// @return An optional bsoncxx::document::view
+    ///
+    stdx::optional<bsoncxx::document::view> auth_mechanism_properties() const;
+
+    ///
+    /// Returns the list of compressors present in the uri or an empty list if "compressors" was not
+    /// present or contained no valid compressors.
+    ///
+    /// @return An optional bsoncxx::document::view
```
bazile-clyde, line 203:
```suggestion
    /// @return An optional stdx::string_view
```

src/mongocxx/test/uri.cpp
kevinAlbs: I actually wasn't sure off hand. libmongoc actually only interprets a few numbers as boolean values:
https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-uri.c#L1097-L1111

Apparently `-1` is considered false...

I added a test for good measure.

src/mongocxx/uri.hpp
kevinAlbs: I took the view that `uri` refers to the class, which is lowercase, and `URI` is the concept.

src/mongocxx/uri.hpp
kevinAlbs: Oh good catch! Fixed.

PR #665, opened 2020-06-10
CXX-2040 do not upload working directory

kevinAlbs:
Uploading all binaries built for testing produces a large upload.
Windows hosts appear to timeout during upload and repeat the task
from the start.

See the linked BUILD ticket to CXX-2040 for more context.


rcsanchez97:
@kevinAlbs Thanks for fixing this!  This problem was affecting some patch builds I was testing.  At first I thought there was a defect in my modifications to the Evergreen helper scripts.  Glad to see there is a simple work around.

PR #666, opened 2020-06-11
Skip FLE prose test on pre 4.2 servers

kevinAlbs:
The tasks on the `ppc64le Ubuntu 16.04 (MongoDB Latest)` variant appear to be failing due to other reasons.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -595,7 +595,7 @@ buildvariants:
     #######################################
     - matrix_name: "integration"
       matrix_spec: {os: "*", mongodb_version: "*"}
-      display_name: "${os} ${mongodb_version}"
+      display_name: "${os} (MongoDB ${mongodb_version})"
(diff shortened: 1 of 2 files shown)


.mci.yml
```diff
@@ -617,7 +617,7 @@ buildvariants:
     #         Linux Buildvariants         #
     #######################################
     - name: ubuntu1604-release
-      display_name: "Ubuntu 16.04 Release"
+      display_name: "Ubuntu 16.04 Release (MongoDB Latest)"
```
kevinAlbs, line 620: For convenience. Each variant specifies the server version as a variant expansion, each variant runs exactly one server version.

samantharitter: Acknowledged.

kevinAlbs:
Hah no worries. It did not occur to me in review either.

PR #667, opened 2020-06-11
CXX-2032 simplify Evergreen scripts

rcsanchez97:
Full Evergreen patch build: https://evergreen.mongodb.com/version/5ee191a87742ae18980fa466

### DIFF
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -52,10 +52,10 @@ esac
 
 cd build
 "$CMAKE" "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" -DCMAKE_VERBOSE_MAKEFILE=ON -DMONGOCXX_ENABLE_SLOW_TESTS=ON -DENABLE_UNINSTALL=ON "$@" ..
-"$CMAKE" --build . -- $CMAKE_BUILD_OPTS
-"$CMAKE" --build . --target install
-"$CMAKE" --build . --target $CMAKE_EXAMPLES_TARGET
+"$CMAKE" --build . --config $BUILD_TYPE -- $CMAKE_BUILD_OPTS
+"$CMAKE" --build . --config $BUILD_TYPE --target install
+"$CMAKE" --build . --config $BUILD_TYPE --target $CMAKE_EXAMPLES_TARGET
 
 if [ "$_RUN_DISTCHECK" ]; then
-   DISTCHECK_BUILD_OPTS="-j$CONCURRENCY" "$CMAKE" --build . --target distcheck
+   DISTCHECK_BUILD_OPTS="-j$CONCURRENCY" "$CMAKE" --build . --config $BUILD_TYPE --target distcheck
 fi
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -19,33 +19,43 @@ fi
 
 OS=$(uname -s | tr '[:upper:]' '[:lower:]')
 
-cd build
-cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" -DMONGOCXX_ENABLE_SLOW_TESTS=ON -DENABLE_UNINSTALL=ON "$@" ..
+if [ -f /proc/cpuinfo ]; then
+    CONCURRENCY=$(grep -c ^processor /proc/cpuinfo)
+elif which sysctl; then
+    CONCURRENCY=$(sysctl -n hw.logicalcpu)
+else
+    echo "$0: can't figure out what value of -j to pass to 'make'" >&2
+    exit 1
+fi
 
 case "$OS" in
-    darwin|linux)
-        if [ -f /proc/cpuinfo ]; then
-            CONCURRENCY=$(grep -c ^processor /proc/cpuinfo)
-        elif which sysctl; then
-            CONCURRENCY=$(sysctl -n hw.logicalcpu)
-        else
-            echo "$0: can't figure out what value of -j to pass to 'make'" >&2
-            exit 1
-        fi
-        make "-j$CONCURRENCY" all VERBOSE=1
-        make install VERBOSE=1
-        make "-j$CONCURRENCY" examples VERBOSE=1
-        if [ "$RUN_DISTCHECK" ]; then
-                make DISTCHECK_BUILD_OPTS="-j$CONCURRENCY" distcheck VERBOSE=1
-        fi
-        ;;
-    cygwin*)
-        MSBuild.exe /p:Configuration=${BUILD_TYPE} /m ALL_BUILD.vcxproj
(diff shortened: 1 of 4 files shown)


.evergreen/compile.sh
```diff
@@ -19,33 +19,43 @@ fi
...
+      CMAKE_BUILD_OPTS="-j $CONCURRENCY"
+      CMAKE_EXAMPLES_TARGET=examples
+      if [ "$RUN_DISTCHECK" ]; then
+         _RUN_DISTCHECK=$RUN_DISTCHECK
+      fi
+      ;;
+
+   cygwin*)
+      CMAKE_BUILD_OPTS="/maxcpucount:$CONCURRENCY"
+      CMAKE_EXAMPLES_TARGET=examples/examples
```
kevinAlbs, line 42: I didn't know that the target name for the examples was different on Windows. Do you know off-hand if it just the examples target that differs?

.evergreen/compile.sh
```diff
@@ -19,33 +19,43 @@ fi
...
+. .evergreen/find_cmake.sh
+
+cd build
+"$CMAKE" "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" -DCMAKE_VERBOSE_MAKEFILE=ON -DMONGOCXX_ENABLE_SLOW_TESTS=ON -DENABLE_UNINSTALL=ON "$@" ..
+"$CMAKE" --build . -- $CMAKE_BUILD_OPTS
+"$CMAKE" --build . --target install
+"$CMAKE" --build . --target $CMAKE_EXAMPLES_TARGET
+
+if [ "$_RUN_DISTCHECK" ]; then
+   DISTCHECK_BUILD_OPTS="-j$CONCURRENCY" "$CMAKE" --build . --target distcheck
```
kevinAlbs, line 60: Is it safe to pass `-j$CONCURRENCY` to Windows here?

Though, based on the previous `case`, it looks like `RUN_DISTCHECK` was not checked on Cygwin. So perhaps this introduces additional coverage of running distcheck on Windows as well.

.evergreen/uninstall_check_windows.cmd
```diff
@@ -7,11 +7,8 @@ echo > %INSTALL_DIR%\lib\canary.txt
 
 dir %INSTALL_DIR%\share\mongo-cxx-driver
 
-set PATH=%PATH%;"c:\Program Files (x86)\MSBuild\14.0\Bin"
-set PATH=%PATH%;"c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin"
-
 pushd build
-MSBuild.exe /m generate_uninstall\uninstall.vcxproj
+%CMAKE% --build . --target uninstall
```
kevinAlbs, line 11: [This Windows task](https://evergreen.mongodb.com/task/cxx_driver_windows_2k8_release_uninstall_check_windows_patch_7a06a355915f2ac105349e7925aed43d460df5b6_5ee191e2850e615ee6ea37d2_20_06_11_02_07_30) appears to expand $CMAKE to the empty string:
```
[2020/06/11 02:36:07.091] C:\data\mci\598ae302535af5d363ca095dfc713e06\mongo-cxx-driver\build>--build . --target uninstall
[2020/06/11 02:36:07.091] '--build' is not recognized as an internal or external command,
[2020/06/11 02:36:07.091] operable program or batch file.
```

I think this just needs a call to `. .evergreen/find_cmake.sh` above.

.evergreen/install_c_driver.sh
```diff
@@ -112,6 +72,30 @@ case "$OS" in
         ;;
 esac
 
+# build libbson
+mkdir cmake_build
+cd cmake_build
+"$CMAKE" -G "$GENERATOR" -DENABLE_MONGOC=OFF $CMAKE_ARGS ..
+"$CMAKE" --build . -- $CMAKE_BUILD_OPTS
```
kevinAlbs, line 79: The previous Windows cmake commands also had `--config "Debug"`. I believe Debug may be the default. If we're not sure whether it's safe to omit, I'd lean towards being explicit. In the past, I've seen user reports about Windows bugs linking to both the debug/release forms of the C runtime, which usually appear as non-obvious runtime crashes (here's an example: https://jira.mongodb.org/browse/CXX-1961).

Testing now on my Mac, passing `--config` seems to be ignored. It might be safe to always specify `--config Debug` here, and for the `--build` commands below.

.evergreen/compile.sh
```diff
@@ -19,33 +19,43 @@ fi
...
+   darwin|linux)
+      CMAKE_BUILD_OPTS="-j $CONCURRENCY"
+      CMAKE_EXAMPLES_TARGET=examples
+      if [ "$RUN_DISTCHECK" ]; then
+         _RUN_DISTCHECK=$RUN_DISTCHECK
+      fi
+      ;;
+
+   cygwin*)
+      CMAKE_BUILD_OPTS="/maxcpucount:$CONCURRENCY"
```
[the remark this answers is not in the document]
rcsanchez97: Good point.  Using the `--config` option for all builds seems like the best way, so I've done that.

.evergreen/compile.sh
rcsanchez97: That is the only target with that peculiarity.

.evergreen/compile.sh
rcsanchez97: The original idea behind the `distcheck` target is that it would only ever be executed on a Unix-like system.  Before the changes in this PR, the `distcheck` invocation followed the other cmake/make invocations inside the `case`.  However, since we now use the `case` as a way to setup what is needed for each platform before invoking common cmake commands, I have taken this particular approach.  It isn't perfectly clean, as it is a little platform-specific, but it seemed like the simplest way to make the necessary change.

.evergreen/uninstall_check_windows.cmd
rcsanchez97: Good catch.  However, `. .evergreen/find_cmake.sh` won't work in a batch file.  I've simply defined the variable based on the standard location of the CMake binary on Evergreen hosts.

.evergreen/install_c_driver.sh
rcsanchez97: Done.

rcsanchez97:
@kevinAlbs I have just force-pushed an update that rebases two commits on to `master`.  The first commit is the original from this PR and the second is the changes that address the comments you made.

kevinAlbs:
There was an unexpected task failure on macOS: https://evergreen.mongodb.com/task/cxx_driver_macos_1014_compile_and_test_with_static_libs_patch_c9ab2c2aee54ebb43f9bb80ffcf50d958ee4bccf_5ee57a2a2fbabe0572abc698_20_06_14_01_15_23/0
It looked unrelated though. I reran it, and it passed the second time.

rcsanchez97:
This change was merged in commit 796c11cf0

PR #668, opened 2020-06-15
CXX-1965 Ensure that 'errInfo' is propagated

bazile-clyde:


src/mongocxx/test/collection.cpp
```diff
@@ -2721,4 +2721,51 @@ TEST_CASE("find_and_x operations append write concern correctly", "[collection]"
...
+
+    bsoncxx::document::element result;
+    try {
+        coll.insert_one(doc.view());
+    } catch (const operation_exception& e) {
+        auto error = e.raw_server_error()->view();
+        result = error["writeConcernErrors"][0]["errInfo"];
+    }
+
+    REQUIRE(result);
```
kevinAlbs, line 2768: It's probably not strictly necessary since the C++ driver does not parse write concern errors or distinguish them, but the spec also says to match `result` against the `errInfo` document.

bazile-clyde: Done

src/mongocxx/test/collection.cpp
```diff
@@ -2721,4 +2721,53 @@ TEST_CASE("find_and_x operations append write concern correctly", "[collection]"
...
+
+    coll.drop();
+    auto doc = make_document(kvp("x", types::b_int32{1}));
+
+    bsoncxx::document::element result;
+    try {
+        coll.insert_one(doc.view());
+    } catch (const operation_exception& e) {
+        auto error = e.raw_server_error()->view();
+        result = error["writeConcernErrors"][0]["errInfo"];
```
kevinAlbs, line 2767: I'm not 100% sure, but the cause of the failing Valgrind tasks may be that once `e` goes out of scope, the view to `errInfo` is no longer valid. Perhaps copying out "errInfo" to a `bsoncxx::document::value` will fix that.

FWIW tests passed on the last commit, so I suspect it's this. I don't see why the other changes would be causing that.

src/mongocxx/test/collection.cpp
```diff
@@ -17,7 +17,6 @@
 
 #include <bsoncxx/builder/basic/document.hpp>
 #include <bsoncxx/json.hpp>
-#include <bsoncxx/private/suppress_deprecation_warnings.hh>
```
bazile-clyde, line 20: This was also unused.

src/mongocxx/test/collection.cpp
```diff
@@ -2721,4 +2721,53 @@ TEST_CASE("find_and_x operations append write concern correctly", "[collection]"
...
+        sub_doc.append(kvp("provenance", "clientSupplied"));
+    }));
+
+    auto fail_point = builder::basic::document{};
+    fail_point.append(kvp("configureFailPoint", "failCommand"));
+
+    using bsoncxx::builder::basic::sub_array;
+    fail_point.append(kvp("data", [&err_info](sub_document sub_doc) {
+        sub_doc.append(
+            kvp("failCommands", [&err_info](sub_array sub_arr) { sub_arr.append("insert"); }));
```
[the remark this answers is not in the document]
bazile-clyde: Removed.

src/mongocxx/test/collection.cpp
bazile-clyde: I J\just did the testing inside the catch to avoid having to copy out err info.

kevinAlbs: Oh nice, even simpler.

bazile-clyde:
> Run lint before merging. Otherwise LGTM!

Done. Closing and merging.

PR #671, opened 2020-06-16
CXX-1868 Collection and index creation in multi-doc txns

bazile-clyde:


src/mongocxx/test/spec/transactions.cpp
```diff
@@ -378,7 +378,7 @@ void run_transactions_tests_in_file(const std::string& test_path) {
         };
 
         if (test["expectations"]) {
-            apm_checker.compare(test["expectations"].get_array().value, false, visitor);
+            apm_checker.compare(test["expectations"].get_array().value, true, visitor);
```
bazile-clyde, line 381: `has_collection` calls `list_collection` under the hood. This gives us two extra `command_started_event`s: one for `assertCollectionExists` and one for `assertCollectionNotExists`. Not sure if there's a to avoid this though. Open to suggestions.

kevinAlbs: Can we run `assertCollectionExists`, `assertCollectionNotExists`, `assertIndexExists`, and `assertIndexNotExists` on a separate mongocxx::client? There is some precedent for that. `targetedFailPoint` is meant to be run from a different client for that very reason:

> Tests that use the "targetedFailPoint" operation do not include
``configureFailPoint`` commands in their command expectations. Drivers MUST
ensure that ``configureFailPoint`` commands do not appear in the list of logged
commands, either by manually filtering it from the list of observed commands or
by using a different MongoClient to execute ``configureFailPoint``.

src/mongocxx/test/spec/operation.cpp
```diff
@@ -1160,6 +1160,18 @@ document::value operation_runner::_run_configure_fail_point(bsoncxx::document::v
     return result.extract();
 }
 
+document::value operation_runner::_create_index(const document::view& operation) {
+    auto arguments = operation["arguments"];
+    auto session = _lookup_session(arguments.get_document().value);
```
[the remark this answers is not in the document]
bazile-clyde: Done. Used REQUIRE.

src/mongocxx/test/spec/operation.cpp
```diff
@@ -1273,6 +1289,42 @@ document::value operation_runner::run(document::view operation) {
         auto bucket = _db->gridfs_bucket();
         bucket.download_to_stream(operation["arguments"]["id"].get_value(), &null_stream);
 
+        return empty_document;
+    } else if (key.compare("createCollection") == 0) {
+        auto collection_name = operation["arguments"]["collection"].get_utf8().value;
+        auto session = _lookup_session(operation["arguments"].get_document().value);
```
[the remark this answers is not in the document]
bazile-clyde: Done

src/mongocxx/test/spec/transactions.cpp
bazile-clyde: Good call. That worked!

kevinAlbs: Nice!

PR #673, opened 2020-06-17
CXX-2042 Add 4.4 tests to evergreen

kevinAlbs:


kevinAlbs:
Replaced by https://github.com/mongodb/mongo-cxx-driver/pull/676

PR #674, opened 2020-06-17
add codecov to travis-ci

bazile-clyde:


bazile-clyde:
@kevinAlbs updated to ignore `third_party` libs

PR #675, opened 2020-06-18
CXX-1973 Bump minimum required C driver

kevinAlbs:
1.17.0 is required for 4.4 support and Client-Side Field Level
Encryption.

Add a test to verify we can compile and test against the oldest
version of the C driver supported.

As a note: once this is merged, if the version of libmongoc you have installed locally is older than 1.17.0 (or reports as older) you may need to reinstall the C driver from the 1.17.0-beta2 release (or from master).


.mci.yml
```diff
@@ -168,7 +168,7 @@ functions:
                         PREFIX=$(cygpath -m "$PREFIX")
                     fi
                     export GENERATOR="${cdriver_cmake_generator}"
-                    ./.evergreen/install_c_driver.sh ${MONGOC_VERSION|master}
+                    ./.evergreen/install_c_driver.sh ${mongoc_version|master}
```
kevinAlbs, line 171: I changed the case for consistency with other expansions. I've been bitten before when treating an expansion as a bash variable and vice-versa.

.evergreen/build_example_projects.sh
```diff
@@ -9,6 +9,9 @@ else
     DIR=shared
 fi
 
+. .evergreen/find_cmake.sh
+export CMAKE="$CMAKE"
```
kevinAlbs, line 13: The Ubuntu 18.04 hosts do not appear to have `cmake` in the path. `compile.sh` and `install_c_driver.sh` both use this script to discover the path to cmake, so this follows suit.

.mci.yml
```diff
@@ -744,6 +734,22 @@ buildvariants:
...
+    # Test that we actually support building against the minium required C version claimed.
+    - name: ubuntu1804-min-libmongoc
+      display_name: "Ubuntu 18.04 with minimum libmongoc (MongoDB 4.4)"
+      batchtime: 1440 # 1 day
+      expansions:
+          build_type: "Debug"
+          tar_options: *linux_tar_options
+          cmake_flags: *linux_cmake_flags
+          mongodb_version: *version_latest
+          mongoc_version: "1.17.0-beta2"
```
kevinAlbs, line 746: With this, the intent is if we introduce API into libmongoc, and start relying on that API in C++, we'll be alerted with a test failure if we do not bump the minimum C driver requirement.

src/mongocxx/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ message ("mongocxx version: ${MONGOCXX_VERSION}")
 set(MONGOCXX_INLINE_NAMESPACE "v${MONGOCXX_ABI_VERSION}")
 set(MONGOCXX_HEADER_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/mongocxx/${MONGOCXX_INLINE_NAMESPACE}" CACHE INTERNAL "")
 
-set(LIBMONGOC_REQUIRED_VERSION 1.15.0)
+set(LIBMONGOC_REQUIRED_VERSION 1.17.0)
```
kevinAlbs, line 32: This is the only non-test change.

.mci.yml
samantharitter: Acknowledged

.evergreen/build_example_projects.sh
samantharitter: Acknowledged

src/mongocxx/CMakeLists.txt
samantharitter: I actually already needed to update my C driver installation, so seems good to make it "official" through the cmake

examples/projects/bsoncxx/cmake/static/build.sh
```diff
@@ -3,20 +3,16 @@ set -o errexit
...
 set -o xtrace
 
-if [ -z "$BUILD_TYPE" ]; then
-    BUILD_TYPE=Release
-fi
-
-if [ -z "$CXX_STANDARD" ]; then
-    CXX_STANDARD=11
-fi
+BUILD_TYPE=${BUILD_TYPE:-Release}
```
samantharitter, line 6: TIL what the :- operator is!

PR #676, opened 2020-06-18
CXX-2042 Add 4.4 tasks to evergreen

kevinAlbs:
- Upgrade Windows in matrix tasks to run "latest" and "4.4" as part of
matrix.
- Fix a test helper for lowercasing error messages.
- Skip FLE tests we cannot run on unsupported operations.
- Build on VS 2017 with /EHsc to prevent tests from failing due to
destructors not being called when exceptions are thrown. (CXX-2054)

### DIFF
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -30,6 +30,7 @@ fi
 
 case "$OS" in
    darwin|linux)
+      GENERATOR=${GENERATOR:-"Unix Makefiles"}
       CMAKE_BUILD_OPTS="-j $CONCURRENCY"
       CMAKE_EXAMPLES_TARGET=examples
       if [ "$RUN_DISTCHECK" ]; then
@@ -38,6 +39,7 @@ case "$OS" in
       ;;
 
    cygwin*)
+      GENERATOR=${GENERATOR:-"Visual Studio 14 2015 Win64"}
       CMAKE_BUILD_OPTS="/maxcpucount:$CONCURRENCY"
       CMAKE_EXAMPLES_TARGET=examples/examples
       ;;
@@ -51,7 +53,7 @@ esac
 . .evergreen/find_cmake.sh
 
 cd build
-"$CMAKE" "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" -DCMAKE_VERBOSE_MAKEFILE=ON -DMONGOCXX_ENABLE_SLOW_TESTS=ON -DENABLE_UNINSTALL=ON "$@" ..
+"$CMAKE" -G "$GENERATOR" "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" -DCMAKE_VERBOSE_MAKEFILE=ON -DMONGOCXX_ENABLE_SLOW_TESTS=ON -DENABLE_UNINSTALL=ON "$@" ..
 "$CMAKE" --build . --config $BUILD_TYPE -- $CMAKE_BUILD_OPTS
 "$CMAKE" --build . --config $BUILD_TYPE --target install
 "$CMAKE" --build . --config $BUILD_TYPE --target $CMAKE_EXAMPLES_TARGET
--- a/.mci.yml
+++ b/.mci.yml
@@ -9,6 +9,7 @@ variables:
 
     mongodb_version:
         version_latest: &version_latest latest
+        version_44: &version_44 4.4
         version_42: &version_42 4.2
         version_40: &version_40 4.0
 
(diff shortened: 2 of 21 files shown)


.evergreen/compile.sh
```diff
@@ -38,6 +39,7 @@ case "$OS" in
       ;;
 
    cygwin*)
+      GENERATOR=${GENERATOR:-"Visual Studio 14 2015 Win64"}
```
kevinAlbs, line 42: Since we have two options for generators on Windows, I made this more consistent with `install_c_driver.sh` and made the generator an optional environment variable for this script.

.mci.yml
```diff
@@ -29,12 +30,16 @@ variables:
...
-        msvc2015_cmake_flags: &msvc2015_cmake_flags -G "Visual Studio 14 2015 Win64" -DBOOST_ROOT=c:/local/boost_1_60_0
+        msvc2015_cmake_flags: &msvc2015_cmake_flags -DBOOST_ROOT=c:/local/boost_1_60_0
+        msvc2015_generator: &msvc2015_generator Visual Studio 14 2015 Win64
         # The VS option /Zc:__cplusplus is necessary to build with VS2017's
         # native C++17 support. __cplusplus should indicate the standard of C++
         # the compiler supports. VS2017 erroneously reports it as 199711L
         # (instead of 201703L) without this option.
-        msvc2017_cmake_flags: &msvc2017_cmake_flags -G "Visual Studio 15 2017 Win64" -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_FLAGS="/Zc:__cplusplus"
+        # The VS option /EHsc is necessary to opt into C++ standard exception unwinding.
+        # Without this option, some destructors may not be called in tests (CXX-2054).
```
kevinAlbs, line 40: The warning I saw compiling tests with Visual Studio 2017 is:
```
C:\code\mongo-cxx-driver\src\mongocxx\test\model\insert_one.cpp(33,9): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc [C:\code\mongo-cxx-driver\cmake-build\src\mongocxx\test\test_driver.vc
xproj]
```

Based on the [description of the warning](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4530?view=vs-2019) and the [docs for Windows C++ exception handling](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=vs-2019) this seems like the exception handling we want. What I'm unsure about is whether this is an option we should be building with by default, or if we should note this in the installation docs. I created https://jira.mongodb.org/browse/CXX-2055 to answer that.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -380,9 +380,8 @@ bool matches(document::view doc, document::view pattern, match_visitor visitor_f
 
 std::string tolowercase(stdx::string_view view) {
     std::string out;
-    out.reserve(view.size());
     for (size_t i = 0; i < view.length(); i++) {
-        out[i] = static_cast<char>(::tolower(view[i]));
+        out += static_cast<char>(::tolower(view[i]));
```
kevinAlbs, line 384: This was causing an issue with error text comparison when running locally. I do not believe the `reserve` guarantees we can assign to individual characters. cppreference around the operator overload says:

>  No bounds checking is performed. If pos > size(), the behavior is undefined.

https://en.cppreference.com/w/cpp/string/basic_string/operator_at

.evergreen/compile.sh
samantharitter: Acknowledged

.mci.yml
```diff
@@ -29,12 +30,16 @@ variables:
...
         asan_cmake_flags: &asan_cmake_flags -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-missing-field-initializers" -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0 -fsanitize=address -O1 -g -fno-omit-frame-pointer -Wall -Wextra -Wconversion -Wnarrowing -pedantic -Werror"
         ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-missing-field-initializers" -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0 -fsanitize=undefined -fsanitize-blacklist=$(pwd)/etc/ubsan.blacklist -fno-sanitize-recover=undefined -O1 -g -fno-omit-frame-pointer -Wall -Wextra -Wconversion -Wnarrowing -pedantic -Werror"
-        msvc2015_cmake_flags: &msvc2015_cmake_flags -G "Visual Studio 14 2015 Win64" -DBOOST_ROOT=c:/local/boost_1_60_0
+        msvc2015_cmake_flags: &msvc2015_cmake_flags -DBOOST_ROOT=c:/local/boost_1_60_0
+        msvc2015_generator: &msvc2015_generator Visual Studio 14 2015 Win64
```
samantharitter, line 34: This could just be left to be the default, right?  I mean it seems fine if we specify it anyway, in case we change the default later.

.mci.yml
samantharitter: Interesting, I'd love to learn more about this.

src/mongocxx/test_util/client_helpers.cpp
samantharitter: The idea behind calling reserve, with any container type, is to allocate its space ahead of time in one go, so that it's not hitting multiple growths as you append to it.  It's meant to be more efficient.  But, I don't think the `reserve`d space is initialized and ready for use, it's more like it's adjacent to the container.  I found this amusing analogy on stack overflow:

"Calling .reserve() is like blowing up mountains to give you some free land. The amount of free land is the .capacity(). The land is there but that doesn't mean you can live there. You have to build houses in order to move in. The number of houses is the .size() (= .length())."

Let's keep the call to `reserve` in there, but still change the inside of the loop to append with += instead of assigning character by character.

.mci.yml
kevinAlbs: I'm not sure off-hand if it's required to be explicit if multiple versions of Visual Studio are available. On my local machine, which has only Visual Studio 2019, it isn't required. So, that might be ok. But being explicit might also be beneficial. We intend to test that compilation on these compilers works. If the default on the hosts were to change (however unlikely), we may lose that test coverage.

.mci.yml
kevinAlbs: From #cxx-discuss, it seems like enabling the flag by default is the way to go. The server also does this.

src/mongocxx/test_util/client_helpers.cpp
kevinAlbs: I like that analogy. Done.

PR #678, opened 2020-06-25
CXX-1948, CXX-2058

bazile-clyde:


src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -41,79 +35,96 @@ using namespace bsoncxx;
...
+    if (test["description"].get_utf8().value.compare(
+            "Change Stream should error when an invalid aggregation stage is passed in") == 0) {
+        UNSCOPED_INFO(
+            "Skipping test with invalid pipeline stages. The C++ driver cannot test them.");
+        return true;
+    }
+    return should_skip_spec_test({uri{}}, test.get_document().value);
+}
+
+void test_setup(document::view test, document::view test_spec) {
```
bazile-clyde, line 52: Just pulled this logic out like in `transactions.cpp` and outsourced some of the work to the function in `util.cpp`

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -41,79 +35,96 @@ using namespace bsoncxx;
...
-            continue;
-        }
-
-        options::change_stream cs_opts{};
-        if (test["changeStreamOptions"]) {
-            auto options = test["changeStreamOptions"].get_document().value;
-            if (options["batchSize"]) {
-                cs_opts.batch_size(options["batchSize"].get_int32().value);
+        auto description = test["description"].get_utf8().value;
+        SECTION(to_string(description)) {
```
bazile-clyde, line 83: Wrapping this in a `SECTION` shows the description under `TEST_CASE` ("Change stream spec tests") for failed tests. Which gives output like:
```
Change stream spec tests
  Executing a watch helper on a Collection results in notifications for changes
  to the specified collection
``` 
The descriptions of additional test cases that failed are displayed, indented, under the `TEST_CASE` as well.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -41,79 +35,96 @@ using namespace bsoncxx;
...
+            auto cs = [&]() {
+                pipeline pipeline{};
+                if (test["changeStreamPipeline"]) {
+                    pipeline = build_pipeline(test["changeStreamPipeline"].get_array().value);
+                }
 
-        // "Using `globalClient`, run every operation in operations in serial against the server."
-        if (test["operations"]) {
+                std::string db_name = to_string(test_spec["database_name"].get_utf8().value);
+                std::string coll_name = to_string(test_spec["collection_name"].get_utf8().value);
```
bazile-clyde, line 114: I just moved these names into this lambda since it was the only place they were used. The rest of the additions is from indentation.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -122,47 +133,50 @@ void run_change_stream_tests_in_file(const std::string& test_path) {
...
+
+            // "Assert that the changes received from changeStream MATCH the results in
+            // results.success"
+            if (!had_error) {
+                REQUIRE(expected_result["success"]);
+                for (auto&& expected_change : expected_result["success"].get_array().value) {
+                    REQUIRE(std::count_if(
+                        changes.begin(), changes.end(), [&](const document::value& res) {
+                            return matches(res.view(), expected_change.get_document().value);
+                        }));
```
bazile-clyde, line 164: Same logic as before but did a bit of refactoring.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -122,47 +133,50 @@ void run_change_stream_tests_in_file(const std::string& test_path) {
...
                 }
+            }
 
-                REQUIRE(found);
+            // Disable the failpoint.
+            if (test["failPoint"]) {
+                bsoncxx::v_noabi::stdx::string_view fail_point =
+                    (get_max_wire_version(client) >= 9) ? "failGetMoreAfterCursorCheckout"
+                                                        : "failPoint";
+                disable_fail_point(client, fail_point);
```
bazile-clyde, line 173: Added this since we have to handle failPoints now.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -172,11 +186,11 @@ TEST_CASE("Change stream spec tests", "[change_stream_spec]") {
 
     client client{uri{}};
     if (!test_util::is_replica_set(client)) {
-        WARN("Skipping - not a replica set");
+        UNSCOPED_INFO("Skipping - not a replica set");
```
bazile-clyde, line 189: I tried to reduce the number of warnings. Displaying everything as a warning made it difficult to debug.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -28,63 +28,39 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
-    using bsoncxx::to_json;
+    auto is_kill_cursor = [](bsoncxx::document::value v) {
+        return v.view()["command_started_event"]["command"]["killCursors"];
+    };
 
     auto events_iter = _events.begin();
+    auto events_end = (_skip_kill_cursors)
+                          ? std::remove_if(_events.begin(), _events.end(), is_kill_cursor)
+                          : _events.end();
+
```
bazile-clyde, line 44: "Removing" the `killCursors` early on simplified most of this logic. At first, `killCursor` commands were interwoven in the list of events. Some of the new tests would try to match one of these `killCursor` commands with an `expectation` in the loop below. I think it's simpler to just move these commands out of the way if we don't want to use them rather than adding checks in our loops for them.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -28,63 +28,39 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
-                    "more expectations than events captured\nExpectations:\n\n%s\n\nCaptured:\n\n",
-                    to_json(expectations).c_str());
-            this->print_all();
-            REQUIRE(false);
-        }
-        REQUIRE(events_iter != _events.end());
+        REQUIRE(events_iter != events_end);
 
         auto expected = expectation.get_document().view();
+        CAPTURE(to_json(*events_iter), expectation);
```
bazile-clyde, line 49: `fprintf` printed on a different buffer from the `catch2` statements so output was often out of order.
This `CAPTURE` gives output like: 
```
mongo-cxx-driver/src/mongocxx/test/spec/monitoring.cpp:50: FAILED:
  to_json(*events_iter) := "{ "command_started_event" ...
  expectation := { ...
```
If the command below fails.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -28,63 +28,39 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
-
-        if (events_iter != _events.end()) {
-            fprintf(stderr, "Error: extra event captured: %s\n", to_json(*events_iter).c_str());
-            events_iter++;
-        }
-
-        REQUIRE(false);
-    }
+
+    REQUIRE((allow_extra || events_iter == events_end));
```
bazile-clyde, line 54: We no longer have to worry about kill cursors here. So we either allow extra events or events_iter is at the "end" (which may be the start of kill cursor commands if we skipped them). If this fails we know we have too many events and the line number will be printed.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -28,63 +28,39 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
-            fprintf(stderr,
-                    "Error: event not found in APM checker.\n Expected: %s\n\nActual:",
-                    to_json(expected).c_str());
-            print_all();
-            REQUIRE(false);
-        }
+        CAPTURE(to_json(expected).c_str());
+        REQUIRE(std::count_if(_events.begin(), _events.end(), [&](bsoncxx::document::view doc) {
+            return test_util::matches(doc, expected);
+        }));
```
bazile-clyde, line 63: Removes `fprintf` in favor of capture.

src/mongocxx/test/spec/monitoring.hh
```diff
@@ -43,7 +43,7 @@ class apm_checker {
 
    private:
     std::vector<bsoncxx::document::value> _events;
-    bool _skip_kill_cursors;
+    bool _skip_kill_cursors = false;
```
bazile-clyde, line 46: I was getting warnings since we were using an initialized variable.

src/mongocxx/test/spec/util.cpp
```diff
@@ -130,12 +130,12 @@ void configure_fail_point(const client& client, document::view test) {
     }
 }
 
-void disable_fail_point(const client& client, stdx::string_view failpoint) {
+void disable_fail_point(const client& client, stdx::string_view fail_point) {
```
bazile-clyde, line 133: Kept this consistent with `disable_fail_point`. Also, my idea raises warnings for misspellings.

src/mongocxx/test/spec/monitoring.cpp
kevinAlbs: As a sidenote, with the removal of kill cursors events, we might be able to remove `allow_extra`. Though other spec tests set it too, so I'm not sure off-hand.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -28,63 +28,39 @@ MONGOCXX_INLINE_NAMESPACE_BEGIN
...
-                return test_util::matches(doc, expected);
-            }) == _events.end()) {
-            fprintf(stderr,
-                    "Error: event not found in APM checker.\n Expected: %s\n\nActual:",
-                    to_json(expected).c_str());
-            print_all();
-            REQUIRE(false);
-        }
+        CAPTURE(to_json(expected).c_str());
+        REQUIRE(std::count_if(_events.begin(), _events.end(), [&](bsoncxx::document::view doc) {
```
kevinAlbs, line 61: Same comment, would `find_if` be more appropriate?

src/mongocxx/test/spec/change_stream.cpp
kevinAlbs: Actually, I checked this out locally. Running the test runner with `-s` shows the output of `INFO` and `UNSCOPED_INFO` logs.

src/mongocxx/test/change_streams.cpp
```diff
@@ -677,6 +675,7 @@ TEST_CASE("Watch a Collection", "[min36]") {
         SECTION("A range-based for loop iterates twice") {
             int count = 0;
             for (const auto& v : x) {
+                (void)v;
```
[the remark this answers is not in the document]
bazile-clyde: Yep!

src/mongocxx/test/change_streams.cpp
```diff
@@ -451,7 +449,7 @@ TEST_CASE("Documentation Examples", "[min36]") {
         // Start Changestream Example 3
         stdx::optional<bsoncxx::document::view> resume_token;
         change_stream stream = inventory.watch();
-        for (auto& event : stream) {
+        for (auto it = stream.begin(); it != stream.end(); it++) {
```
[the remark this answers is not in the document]
bazile-clyde: I think the iterator should be avoided when possible. The _only_ reason I'm using it here is because `event` would be an unused variable otherwise. As a side note, I used the construct `(void)var;` because that test explicitly tested against a range-based loop so I had to keep it. It didn't seem as appropriate here but I'm open to suggestions.

src/mongocxx/test/index_view.cpp
```diff
@@ -344,9 +344,9 @@ TEST_CASE("drop_all", "[index_view]") {
         bsoncxx::document::view result_view = result.view();
 
         auto cursor1 = indexes.list();
-        REQUIRE(std::distance(cursor1.begin(), cursor1.end()) == models.size() + 1);
-        REQUIRE((result_view["numIndexesAfter"].get_int32() -
-                 result_view["numIndexesBefore"].get_int32()) == models.size());
+        REQUIRE((unsigned)std::distance(cursor1.begin(), cursor1.end()) == models.size() + 1);
+        REQUIRE((unsigned)(result_view["numIndexesAfter"].get_int32() -
```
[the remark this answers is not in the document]
bazile-clyde: I'm actually not sure why this wasn't caught earlier. This is a comparison of a signed an unsigned int, which throws a warning, which we mark as an error.

src/mongocxx/test/spec/change_stream.cpp
bazile-clyde: I guess using `find_if` would save us from having to iterate through the entire list. Done

src/mongocxx/test/spec/monitoring.cpp
bazile-clyde: Good idea. Done

src/mongocxx/test/spec/monitoring.cpp
bazile-clyde: Just from trying it locally, it causes tests to break though I'm not sure why. I'll investigate

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -41,79 +35,96 @@ using namespace bsoncxx;
...
-            auto target = std::string(test["target"].get_utf8().value);
-            if (target == "collection") {
-                return client[db1_name][coll1_name].watch(pipeline, cs_opts);
-            } else if (target == "database") {
-                return client[db1_name].watch(pipeline, cs_opts);
-            } else {
-                return client.watch(pipeline);
-            }
-        }();
+            // "Using client, create a changeStream changeStream against the specified target"
```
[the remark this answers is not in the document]
bazile-clyde: Done

src/mongocxx/test/spec/monitoring.cpp
kevinAlbs: Thanks for checking. There's no need to investigate further IMO. It's out of scope for this change.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -122,47 +133,50 @@ void run_change_stream_tests_in_file(const std::string& test_path) {
...
+                        }));
                 }
+            }
 
-                REQUIRE(found);
+            // Disable the failpoint.
+            if (test["failPoint"]) {
+                bsoncxx::v_noabi::stdx::string_view fail_point =
+                    (get_max_wire_version(client) >= 9) ? "failGetMoreAfterCursorCheckout"
+                                                        : "failPoint";
```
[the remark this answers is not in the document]
bazile-clyde: Done

src/mongocxx/test/spec/monitoring.cpp
bazile-clyde: Done

bazile-clyde:
> Oh nice. Will UNSCOPED_INFO show up on high verbosity? I would want to verify when developing if tests are being skipped when I expect them to run.

Ah, it actually didn't show no matter the verbosity since there's no assertion. I changed it back to `WARN`.

src/mongocxx/test/index_view.cpp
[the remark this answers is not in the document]
kevinAlbs: Huh, it may be that we're not using a new enough version of g++ in evergreen. The tests on Ubuntu 16.04 use g++ 5.4.0. This is an improvement regardless.

PR #679, opened 2020-06-29
CXX-2021 implement client side encryption prose tests

samantharitter:
Evergreen patch run: https://evergreen.mongodb.com/version/5efa0a1a32f4175dd509669b


src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
+
+void _round_trip(class client_encryption* client_encryption,
+                 bsoncxx::types::bson_value::view datakey) {
+    auto to_encrypt_doc = document{} << "v"
+                                     << "test" << finalize;
+    auto to_encrypt = to_encrypt_doc.view()["v"].get_value();
+
+    options::encrypt encrypt_opts;
+    encrypt_opts.algorithm(options::encrypt::encryption_algorithm::k_deterministic);
+    encrypt_opts.key_id(datakey.get_binary());
```
kevinAlbs, line 669: Actually, related to my comment about ownership above, IIUC this could also be problematic. The underlying bytes that b_binary points to are unowned. I imagine this is also UB:

```
{
   bsoncxx::types::bson_value::value data_key = client_encryption->create_data_key("aws", std::move(opts));
   encrypt_opts.key_id (data_key.get_binary());
}
// encrypt_opts is holding onto binary data that has been destroyed.
```

I think `encrypt_opts` would need to make a copy of the underlying data.

This seems out of scope of this ticket. But if you agree that this seems like a bug, I think we should address this before releasing. Perhaps a straightforward solution is to have the argument of `key_id` be a `bsoncxx::types::bson_value::view_or_value`? It also may be an API improvement in that you can directly pass the return value of `create_data_key` as the option for the key id.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
+    auto decrypted_val = client_encryption->decrypt(encrypted_val);
+
+    REQUIRE(decrypted_val == to_encrypt);
+}
+
+void _run_endpoint_test(class client_encryption* client_encryption,
+                        bsoncxx::document::view masterkey,
+                        stdx::optional<std::string> error_msg = stdx::nullopt) {
+    options::data_key opts;
+    opts.master_key(masterkey);
```
kevinAlbs, line 680: Can you add `INFO (masterkey)`? If this fails, it'll be easier to see which case failed.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -251,6 +275,8 @@ void run_datakey_and_double_encryption(Callable create_data_key,
 }
 
 TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
+    instance::current();
```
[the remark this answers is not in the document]
samantharitter: Yes, I believe the first call to this method initializes libmongoc, so if a test runs first and does not call this, we have a problem!

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
+
+    std::vector<bsoncxx::document::view> docs;
+    docs.push_back(std::move(over_2mib_1));
+    docs.push_back(std::move(over_2mib_2));
+
+    // Expect the bulk write to succeed and split after first doc (i.e. two inserts occur).
+    // Note: C++ driver does not verify using command monitoring, as splitting logic is handled by
+    // the C driver.
+    auto res = coll.insert_many(docs);
+    REQUIRE(res->inserted_count() == 2);
```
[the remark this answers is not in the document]
samantharitter: I think that, as per the comment on L564, and knowing that libmongoc is handling all the splitting logic, it's not worth testing that here through command monitoring.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
+    options::client client_encrypted_opts;
+    _add_client_encrypted_opts(&client_encrypted_opts, limits_schema.view(), std::move(kms_doc));
+    class client client_encrypted {
+        uri{}, std::move(client_encrypted_opts)
+    };
+
+    // Using client_encrypted perform the following operations:
+    // Insert { "_id": "over_2mib_under_16mib",
+    //          "unencrypted": <the string "a" repeated 2097152 times> }.
+    std::string over_2mib_under_16mib(static_cast<size_t>(2097152), 'a');
```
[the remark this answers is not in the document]
samantharitter: Yeah it's a handy constructor!  Ah, yes, the static_cast is probably a hold-over from when I was trying to use brace initialization, and the compiler kept trying to cast that number to a char instead of calling this constructor.  Removed.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
 TEST_CASE("Corpus", "[client_side_encryption]") {
-    // TODO CXX-2021
+    // TODO CXX-2022
+}
+
+void _round_trip(class client_encryption* client_encryption,
+                 bsoncxx::types::bson_value::view datakey) {
+    auto to_encrypt_doc = document{} << "v"
+                                     << "test" << finalize;
+    auto to_encrypt = to_encrypt_doc.view()["v"].get_value();
```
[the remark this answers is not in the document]
samantharitter: So, this line
`auto to_encrypt = to_encrypt_doc.view()["v"].get_value();`

 is actually dealing with four data types: the owning `document::value` `to_encrypt_doc`, the temporary `document::view` `to_encrypt_doc.view()`, a temporary `document::element` that is created when we use `["v"]`, and the final captured `bson_value::view` `to_encrypt`.

I see no problem with this flow; even though the `document::view` and the `document::element` in the chain are temporary objects.  Both of those temporaries are views into the owning `to_encrypt_doc`'s buffer, so there's no lifetime issues when these temporaries go away.  In fact, I think this is/should be one of our "recommended" flows, since it utilizes the view-ness of views in order to avoid copies and moves.

That being said, I rather hate the construction `doc_value.view()["field"]`, and I'm excited to make it better as part of [CXX-2046](https://jira.mongodb.org/browse/CXX-2046).

This flow uses a similar "chain of temporaries" and is how we recommend users access bson fields:
```
auto doc_value = document{} << "key" << "value" << finalize; //1
auto doc_view = doc_value.view(); //2
auto string_val = doc_view["key"].get_utf8().value; //3
```
Line `//3` that pulls out the string is creating two temporary non-owning view types, the flow is `document::view` -> `document::element` -> `bsoncxx::types::b_utf8` -> `stdx::string_view`.  Here, yes, the `document::view doc_view` is not a temporary in this one-liner, as in the line of code in question, but it's still not owning any of the data.  Pulling it out into a named variable on line `//2` is really just a convenience that makes the code a little more readable, IMO.

I would very much like to discourage users from creating the b_type structs manually.  The b_type structs are inherently view-y and have no owning counterpart, so if users create their own manually they must manage the lifetime of all the component parts correctly.  To me this sounds like a nice foot-gun.  While, yes, this code that you've written will be correct:
```
std::string str = "test";
bsoncxx::bson_value::view to_encrypt {bsoncxx::types::b_utf8 {str}};
```
I think it's much more likely that users will write this and then have a problem:
`bsoncxx::bson_value::view to_encrypt {bsoncxx::types::b_utf8 {"test"}};`

I would rather encourage users to work with `document` as the starting point whenever possible, even though it's a little roundabout, since it will make the ownership semantics easier to reason about and to do correctly:
```
auto doc = make_document(kvp("v", "test")); // or, auto doc = document{} << "v" << "test" << finalize;
auto doc_view = doc.view(); // optional intermediate step
auto to_encrypt = doc_view["v"].get_value(); // this is a bson_value::view
```

If they want to construct an owning bson_value type, they have two options:
```
// 1: use bson_value::value ctor to convert to an owning bson_value
bson_value::value to_encrypt_owning{to_enrypt};
// 2: call get_owned_value() instead of get_value(), when accessing the bson_value in the first place
auto to_encrypt_owning = doc_view["v"].get_owning_value();
```

src/mongocxx/test/client_side_encryption.cpp
samantharitter: Ah good catch!  This is definitely a concern.  I agree, we should change this method to take a view_or_value.  I opened [CXX-2063](https://jira.mongodb.org/browse/CXX-2063) to do this work before the release.

src/mongocxx/test/client_side_encryption.cpp
samantharitter: Done.

src/mongocxx/test/client_side_encryption.cpp
kevinAlbs: Thanks!

src/mongocxx/test/client_side_encryption.cpp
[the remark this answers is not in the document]
samantharitter: I like that helper idea!  I created a ticket for it: [CXX-2066](https://jira.mongodb.org/browse/CXX-2066).  I agree, it is roundabout, and I think that's a shortcoming of our bson value/document type system hierarchy.  It would be great to fix this in the future, so owning `bson_value`s can be easily created without going through the document builder.

As an improvement after the release, I'd like to expose constructors for `bson_value::value` outright.  I see these constructors as taking the component parts of each `b_type` struct, but not actually using the `b_type` structs.  To me the `b_type` structs are problematic because they are inherently non-owning, and it requires a copy to make owning `bson_value`'s from them.  I have created a ticket for this as well: [CXX-2067](https://jira.mongodb.org/browse/CXX-2067).

However, I think that in a real world use case it's likely the user would already have a document containing a value they wanted to encrypt.  I'm imagining that explicit encryption will mostly be used by non-enterprise users, who will write conversion methods to take documents on their way to being inserted, parse a sensitive field out, encrypt that field, then reconstruct the document and insert it.  But who knows, maybe I'm totally wrong about this!  And that use case has its thorny bits, too; we don't have a "replace" facility to drop-in replacement bson values.  In fact, I'm not sure how we would implement that, either...

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -322,28 +360,569 @@ TEST_CASE("Datakey and double encryption", "[client_side_encryption]") {
...
+    // Create a MongoClient configured with auto encryption (referred to as client_encrypted),
+    // with local KMS provider as follows:
+    //     { "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }
+    // and with the keyVaultNamespace set to keyvault.datakeys.
+    char key_storage[96];
+    memcpy(&(key_storage[0]), kLocalMasterKey, 96);
+    bsoncxx::types::b_binary local_master_key{
+        bsoncxx::binary_sub_type::k_binary, 96, (const uint8_t*)&key_storage};
+
+    auto kms_doc = document{} << "local" << open_document << "key" << local_master_key
```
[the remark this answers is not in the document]
samantharitter: Yes, I don't see why not!  Done.

src/mongocxx/test/client_side_encryption.cpp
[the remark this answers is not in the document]
samantharitter: Done!

samantharitter:
Hey @kevinAlbs I believe I've fixed the windows file issue, I lost a couple of crucial line changes to .mci.yml in one of my rebases.

.mci.yml
```diff
@@ -278,6 +278,9 @@ functions:
                   set -o xtrace
 
                   if [ "Windows_NT" == "$OS" ]; then
+                      CRUD_TESTS_PATH=$(cygpath -m $CRUD_TESTS_PATH)
+                      ENCRYPTION_TESTS_PATH=$(cygpath -m $ENCRYPTION_TESTS_PATH)
```
kevinAlbs, line 282: I don't think these paths are necessary. The `CRUD_TESTS_PATH` was here before, but I believe it was redundant. These tests are run through the RUN_TESTS target, and AFAICT they use the paths that are defined in src/mongocxx/test/CMakeLists.txt. Other evidence of that is that none of the other test paths are specified here (e.g. `TRANSACTIONS_TESTS_PATH`). Can you remove this?

src/mongocxx/test/client_side_encryption.cpp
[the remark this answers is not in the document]
kevinAlbs: Thanks for creating the tickets! I put some thoughts in CXX-2067.

.mci.yml
samantharitter: These are the lines that fixed my windows failure...

samantharitter: Perhaps the CRUD path is unnecessary?  The ENCRYPTION redefine with cygpath here seems to be necessary to help windows find the right files in the test.

kevinAlbs: Ah sorry, I don't think I fully understand how paths are getting through. If this is necessary for Evergreen hosts, then I'm fine with it. I would think the CRUD path is unnecessary, but if we remove the need for passing paths through environment variables in https://jira.mongodb.org/browse/CXX-2036, this will go away in the future anyway.

kevinAlbs:
I missed this before, but I believe Travis is failing due to running the prose tests when environment variables are not set. I think rebasing on master and adding calls to `should_run_client_side_encryption_test()` in the test should resolve that.

.mci.yml
samantharitter: Yes, I'm not 100% sure why it's necessary to call cygpath here, but not for all the test environment variables.  And since we are planning to remove these soon, I don't want to go too far down that rabbit hole!

PR #682, opened 2020-07-01
CXX-2048 add back tests

kevinAlbs:
Adds tests back after upgrading the PowerPC variant to Ubuntu 18.04.

Code coverage reports a decrease in client_side_encryption.cpp. I'm not sure why. Code coverage is only uploaded in Travis AFAIK, and Travis was never running CSFLE tests (since AWS credentials are not passed through).

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -764,8 +764,8 @@ buildvariants:
           - name: compile_and_test_with_shared_libs
           - name: compile_and_test_with_static_libs
 
-    - name: power8-ubuntu1604
-      display_name: "ppc64le Ubuntu 16.04 (MongoDB Latest)"
+    - name: power8-ubuntu1804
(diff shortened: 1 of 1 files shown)


bazile-clyde:
LGTM. Ignore the code coverage report. I have a branch that uploads code coverage files from all of our Linux and macOS evergreen builds. That raised the code coverage percentage to 88%, but until it's pushed all PRs will, of course, be missing those files and show an artificially low code coverage percentage.

kevinAlbs:
> LGTM. Ignore the code coverage report. I have a branch that uploads code coverage files from all of our Linux and macOS evergreen builds. That raised the code coverage percentage to 88%, but until it's pushed all PRs will, of course, be missing those files and show an artificially low code coverage percentage.

Oh nice. Thanks!

PR #683, opened 2020-07-01
CXX-1875 add read/write concern tests

kevinAlbs:
- Reuse transaction spec runner for read/write concern integration tests.
- Allow empty read/write concern in spec tests.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -268,6 +268,7 @@ functions:
                   export TRANSACTIONS_TESTS_PATH="$(pwd)/../data/transactions"
                   export WITH_TRANSACTION_TESTS_PATH="$(pwd)/../data/with_transaction"
                   export RETRYABLE_READS_TESTS_PATH="$(pwd)/../data/retryable-reads"
+                  export READ_WRITE_CONCERN_OPERATION_TESTS_PATH="$(pwd)/../data/read-write-concern/operation"
 
                   # export environment variables for encryption tests (and redact them)
                   set +o xtrace
@@ -300,6 +301,7 @@ functions:
                       ${test_params} ./src/mongocxx/test/test_transactions_specs
                       ${test_params} ./src/mongocxx/test/test_logging
                       ${test_params} ./src/mongocxx/test/test_retryable_reads_specs
+                      ${test_params} ./src/mongocxx/test/test_read_write_concern_specs
 
 
                       # Some platforms like OS X don't support the /mode syntax to the -perm option
--- a/src/mongocxx/test/CMakeLists.txt
+++ b/src/mongocxx/test/CMakeLists.txt
@@ -146,6 +146,13 @@ add_executable(test_retryable_reads_specs
     ${spec_test_common}
 )
 
+add_executable(test_read_write_concern_specs
+    ${CMAKE_SOURCE_DIR}/src/mongocxx/test_util/client_helpers.cpp
+    ${THIRD_PARTY_SOURCE_DIR}/catch/main.cpp
+    spec/read_write_concern.cpp
+    ${spec_test_common}
+)
+
 target_link_libraries(test_driver mongocxx_mocked ${libmongoc_target})
 target_link_libraries(test_logging mongocxx_mocked ${libmongoc_target})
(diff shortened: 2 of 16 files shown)


src/mongocxx/test/sdam-monitoring.cpp
```diff
@@ -265,7 +265,6 @@ TEST_CASE("SDAM Monitoring", "[sdam_monitoring]") {
 TEST_CASE("Heartbeat failed event", "[sdam_monitoring]") {
     instance::current();
     options::apm apm_opts;
-    stdx::optional<oid> topology_id;
```
kevinAlbs, line 268: Unused.

src/mongocxx/test/spec/transactions.cpp
```diff
@@ -408,5 +412,7 @@ TEST_CASE("Transactions spec automated tests", "[transactions_spec]") {
         "TRANSACTIONS_TESTS_PATH", &run_transactions_tests_in_file, unsupported_transaction_tests);
 
     run_tests_in_suite("WITH_TRANSACTION_TESTS_PATH", &run_transactions_tests_in_file);
+
+    run_tests_in_suite("READ_WRITE_CONCERN_OPERATION_TESTS_PATH", &run_transactions_tests_in_file);
```
kevinAlbs, line 416: I admit reusing the transactions test runner for this seems a little strange. But the format is the same. A unified test runner is in the near future (https://jira.mongodb.org/browse/DRIVERS-709). I think that will be the time to consolidate them all.

src/mongocxx/test/spec/util.cpp
```diff
@@ -56,7 +57,21 @@ uint32_t error_code_from_name(string_view name) {
     return 0;
 }
 
+/* Called with the entire test file and individual tests. */
 bool should_skip_spec_test(const client& client, document::view test) {
+    std::set<std::string> unsupported_tests = {
+        "CreateIndex and dropIndex omits default write concern",
```
kevinAlbs, line 63: As described in https://jira.mongodb.org/browse/CXX-1351, there is a `create_index()` method, but not a  `drop_index()` on the collection class.

.mci.yml
```diff
@@ -261,6 +261,7 @@ functions:
                   export TRANSACTIONS_TESTS_PATH="$(pwd)/../data/transactions"
                   export WITH_TRANSACTION_TESTS_PATH="$(pwd)/../data/with_transaction"
                   export RETRYABLE_READS_TESTS_PATH="$(pwd)/../data/retryable-reads"
+                  export READ_WRITE_CONCERN_OPERATION_TESTS_PATH="$(pwd)/../data/read-write-concern/operation"
```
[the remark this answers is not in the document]
kevinAlbs: Haha so am I

src/mongocxx/test/spec/util.cpp
kevinAlbs: Agreed, it makes syncing tests a lot simpler. The C driver has a similar list for tests of unsupported operations.

src/mongocxx/test/CMakeLists.txt
```diff
@@ -205,6 +205,8 @@ set_tests_properties(transactions_specs PROPERTIES
   ENVIRONMENT "TRANSACTIONS_TESTS_PATH=${CMAKE_SOURCE_DIR}/data/transactions")
 set_property(TEST transactions_specs
   APPEND PROPERTY ENVIRONMENT "WITH_TRANSACTION_TESTS_PATH=${CMAKE_SOURCE_DIR}/data/with_transaction")
+set_property(TEST transactions_specs
```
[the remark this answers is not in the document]
kevinAlbs: Yeah, looking at this again, splitting this off into a separate runner is not much additional work. I made a separate test runner for the read and write concern tests, and moved the transaction test runner to `spec/util.cpp` to be shared between them. When we do consolidate them, it's one less exceptional case to think about.

src/mongocxx/test/spec/transactions.cpp
kevinAlbs: I split this into a separate test runner.

PR #677, opened 2020-06-22
CXX-2041 fix travis compile

kevinAlbs:
- Remove non-essential getter with optional b_binary, which gives
an error with MNMLSTC and clang 3.6
- Remove redundant friend declaration
- Add explicit static casts for constructing a bson_value::view from
bson_value::value and std::string from string_view
- Fix session lambda to have a void return

PR #684, opened 2020-07-02
CXX-2041 Add necessary include

kevinAlbs:


src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -37,6 +37,7 @@
 #include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/exception/operation_exception.hpp>
 #include <mongocxx/stdx.hpp>
+#include <mongocxx/private/libmongoc.hh>
```
kevinAlbs, line 40: I added `should_run_client_side_encryption_test` in this file as part of https://github.com/mongodb/mongo-cxx-driver/pull/677

That relies on a preprocessor define in the mongoc header mongoc-config.h. I mistakenly did not include the the mongoc headers, so the check always skipped. I noticed this when looking at test logs.

The client side encryption prose tests will also need to call `should_run_client_side_encryption_test` before running. Travis does not build against libmongoc with client side encryption enabled.

samantharitter:
Looks like it might need a clang-format to pass lint, otherwise LGTM

PR #685, opened 2020-07-07
CXX-1738 implement debian-package-build task

rcsanchez97:
Patch build for the new task: https://spruce.mongodb.com/version/5f03c5613066153d099fba2e/tasks

### DIFF
--- a/.evergreen/debian_package_build.sh
+++ b/.evergreen/debian_package_build.sh
@@ -1,0 +1,71 @@
+#!/bin/sh
+
+#
+# Test mongocxx's Debian packaging scripts.
+#
+# Supported/used environment variables:
+#   IS_PATCH    If "true", this is an Evergreen patch build.
+
+set -o xtrace
+set -o errexit
+
+on_exit () {
+  if [ -e ./unstable-chroot/debootstrap/debootstrap.log ]; then
+    echo "Dumping debootstrap.log"
+    cat ./unstable-chroot/debootstrap/debootstrap.log
+  fi
+}
+trap on_exit EXIT
+
+if [ "${IS_PATCH}" = "true" ]; then
+  git diff HEAD -- . ':!debian' > ../upstream.patch
+  git diff HEAD -- debian > ../debian.patch
+  git clean -fdx
+  git reset --hard HEAD
+  if [ -s ../upstream.patch ]; then
+    [ -d debian/patches ] || mkdir debian/patches
+    mv ../upstream.patch debian/patches/
+    echo upstream.patch >> debian/patches/series
+    git add debian/patches/*
+    git commit -m 'Evergreen patch build - upstream changes'
+    git log -n1 -p
+  fi
+  if [ -s ../debian.patch ]; then
+    git apply --index ../debian.patch
+    git commit -m 'Evergreen patch build - Debian packaging changes'
+    git log -n1 -p
+  fi
+fi
+
+cd ..
+
+git clone https://salsa.debian.org/installer-team/debootstrap.git debootstrap.git
+export DEBOOTSTRAP_DIR=`pwd`/debootstrap.git
+sudo -E ./debootstrap.git/debootstrap unstable ./unstable-chroot/ http://cdn-aws.deb.debian.org/debian
+cp -a mongo-cxx-driver ./unstable-chroot/tmp/
+sudo chroot ./unstable-chroot /bin/bash -c "(set -o xtrace && \
+  apt-get install -y build-essential git-buildpackage fakeroot debhelper cmake curl ca-certificates libboost-dev libsasl2-dev libicu-dev doxygen && \
+  mkdir /tmp/mongo-c-driver && \
+  curl -o deb.tar.gz -L https://s3.amazonaws.com/mciuploads/mongo-c-driver/master/mongo-c-driver-debian-packages-latest.tar.gz && \
+  tar zxvf deb.tar.gz && \
+  apt-get install -y ./*.deb && \
+  cd /tmp/mongo-cxx-driver && \
+  git clean -fdx && \
+  git reset --hard HEAD && \
+  python3 etc/calc_release_version.py > build/VERSION_CURRENT && \
+  git add --force build/VERSION_CURRENT && \
+  git commit build/VERSION_CURRENT -m 'Set current version' && \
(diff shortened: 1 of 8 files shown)


rcsanchez97:
@kevinAlbs FYI, for the moment this change uses a set of C driver packages downloaded from Evergreen at a particular commit.  However, if we have a static URL for the location of the Debian packages built from the latest C driver commit, I could switch to using that instead.

.evergreen/debian_package_build.sh
```diff
@@ -0,0 +1,71 @@
...
+cd ..
+
+git clone https://salsa.debian.org/installer-team/debootstrap.git debootstrap.git
+export DEBOOTSTRAP_DIR=`pwd`/debootstrap.git
+sudo -E ./debootstrap.git/debootstrap unstable ./unstable-chroot/ http://cdn-aws.deb.debian.org/debian
+cp -a mongo-cxx-driver ./unstable-chroot/tmp/
+sudo chroot ./unstable-chroot /bin/bash -c "(set -o xtrace && \
+  apt-get install -y build-essential git-buildpackage fakeroot debhelper cmake curl ca-certificates libboost-dev libsasl2-dev libicu-dev doxygen && \
+  mkdir /tmp/mongo-c-driver && \
+  curl -o deb.tar.gz -L https://s3.amazonaws.com/mciuploads/mongo-c-driver/master/mongo-c-driver-debian-packages-1.17.0-pre-mongo_c_driver_44906882c3f2bbd2a3605e359814e74e8fe40ec9.tar.gz && \
```
kevinAlbs, line 49: The C driver's debian-package-build task uploads to a fixed latest URL. For example [this task](https://evergreen.mongodb.com/task/mongo_c_driver_packaging_debian_package_build_49d5590b0164e629c6adc1ed1b88e6852f62bba9_20_07_07_16_23_59) has this attachment:
https://s3.amazonaws.com/mciuploads/mongo-c-driver/master/mongo-c-driver-debian-packages-latest.tar.gz

.evergreen/debian_package_build.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+cd ..
+
+git clone https://salsa.debian.org/installer-team/debootstrap.git debootstrap.git
+export DEBOOTSTRAP_DIR=`pwd`/debootstrap.git
+sudo -E ./debootstrap.git/debootstrap unstable ./unstable-chroot/ http://cdn-aws.deb.debian.org/debian
+cp -a mongo-cxx-driver ./unstable-chroot/tmp/
+sudo chroot ./unstable-chroot /bin/bash -c "(set -o xtrace && \
+  apt-get install -y build-essential git-buildpackage fakeroot debhelper cmake curl ca-certificates libboost-dev libsasl2-dev libicu-dev doxygen && \
```
kevinAlbs, line 47: To check, libicu and libsasl are libmongoc dependencies. But we need those because we are building the C driver from the debian package, right?

debian/rules
```diff
@@ -24,7 +24,8 @@ export DEB_LDFLAGS_MAINT_APPEND := -lpthread
 override_dh_auto_configure:
 	dh_auto_configure -B$(CURDIR)/build -- \
 	-DBUILD_VERSION=$(DEB_VERSION_UPSTREAM) \
-	-DBSONCXX_POLY_USE_BOOST=1
+	-DBSONCXX_POLY_USE_BOOST=1 \
+	-DENABLE_UNINSTALL=OFF
```
kevinAlbs, line 28: Was disabling uninstall necessary?

.mci.yml
```diff
@@ -531,6 +531,35 @@ tasks:
             export INSTALL_VERSION=${mongocxx_version_finder_sh}
             cmd.exe /c .\\.evergreen\\uninstall_check_windows.cmd
 
+    - name: debian-package-build
+      commands:
+      - func: "setup"
+      - command: expansions.update
```
kevinAlbs, line 537: Is the `mongocxx_version_finder_sh` expansion unused in this task?

debian/build_snapshot.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
+
+#
+# build_snapshot.sh
+#
+
+#
+# Copyright 2018 MongoDB, Inc.
```
kevinAlbs, line 8: 2020

debian/gbp.conf
```diff
@@ -19,6 +19,8 @@ prebuild = bash -c "
     # Create upstream tarball from reference, exclude items that do not belong
     pushd $GBP_GIT_DIR/.. &&
     ls -1 &&
+    # TODO: remove this line and the next after the first time the change is pushed
+    echo '/.* !export-ignore' >$GBP_GIT_DIR/info/attributes &&
```
kevinAlbs, line 23: I assume we cannot do this as a separate commit in this PR since Evergreen would consider all commits part of one patch.

.evergreen/debian_package_build.sh
rcsanchez97: Thanks for the pointer! Done.

.evergreen/debian_package_build.sh
rcsanchez97: We have libicu and libsasl listed as dependencies for building the C++ driver in the `debian/control` file.  I'd have to try the build after removing those to see if we really need them. In short, though, the `debian_package_build.sh` script needs to use the build dependencies specified in `debian/control`.

debian/rules
rcsanchez97: Yes.  Otherwise, we generate and install a file (`uninstall.sh`) which the packages don't install.  The build scripts will notice that a file was left behind and then fail the package build.

.mci.yml
rcsanchez97: Yes, though I failed to notice this until after I pushed the PR.  I have an unpushed commit that removes this.  When I update the PR, this will be removed.

debian/gbp.conf
rcsanchez97: True.  The alternative is to remove this and then live with the first Evergreen waterfall build containing this change failing.  All subsequent changes should then work as expected.  I am happy to go either with it.

kevinAlbs: I'm fine with doing a follow up.

PR #686, opened 2020-07-08
CXX-1111 Remove std::iterator inheritance

gspatace:
Directly specify std::iterator_traits fields in classes


kevinAlbs:
@gspatace Thank you for the contribution, this change LGTM!

PR #687, opened 2020-07-08
CXX-2063 options::encrypt::key_id() takes a bson_value::view_or_value

samantharitter:
- implement bson_value::view_or_value
- change options::encrypt::key_id() to take a bson_value::view_or_value

In our original discussion of this, we talked about changing this method to take a bson_value::value, to avoid creating tricky lifetime semantics.  However, most of our options that take documents take a document::view_or_value.  This allows the user to decide whether to give the options class ownership (by std::move-ing in a document::value) or to intentionally pass in a view and manage the lifetime themselves.  I think this is the most consistent and flexible approach for bson_value, too.

### DIFF
--- a/src/bsoncxx/types/bson_value/value.hpp
+++ b/src/bsoncxx/types/bson_value/value.hpp
@@ -49,7 +49,7 @@ class BSONCXX_API value {
     ///
     /// Create an owning copy of a bson_value::view.
     ///
-    value(const view&);
+    explicit value(const view&);
 
     ///
     /// Get a view over the bson_value owned by this object.
--- a/src/mongocxx/options/encrypt.cpp
+++ b/src/mongocxx/options/encrypt.cpp
@@ -26,10 +26,8 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
-encrypt& encrypt::key_id(bsoncxx::types::b_binary key_id) {
-    // MNMLSTC on clang 3.6.2 reports an error when assigning to an optional of a struct.
+encrypt& encrypt::key_id(bsoncxx::types::bson_value::view_or_value key_id) {
     _key_id = std::move(key_id);
-    _key_id_set = true;
     return *this;
 }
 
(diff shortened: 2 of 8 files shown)


src/mongocxx/options/encrypt.cpp
```diff
@@ -26,10 +26,8 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
-encrypt& encrypt::key_id(bsoncxx::types::b_binary key_id) {
-    // MNMLSTC on clang 3.6.2 reports an error when assigning to an optional of a struct.
```
kevinAlbs, line 30: Yay for my hacky workaround going away!

src/mongocxx/options/encrypt.cpp
```diff
@@ -51,24 +49,35 @@ const stdx::optional<encrypt::encryption_algorithm>& encrypt::algorithm() const
     return _algorithm;
 }
 
+const stdx::optional<bsoncxx::types::bson_value::view_or_value>& encrypt::key_id() const {
+    return _key_id;
+}
```
kevinAlbs, line 54: Awesome! This resolved https://jira.mongodb.org/browse/CXX-2059 as well. Feel free to self-assign and close when this is merged.

src/mongocxx/options/encrypt.cpp
```diff
@@ -51,24 +49,35 @@ const stdx::optional<encrypt::encryption_algorithm>& encrypt::algorithm() const
...
+
 void* encrypt::convert() const {
     using libbson::scoped_bson_t;
 
     mongoc_client_encryption_encrypt_opts_t* opts = libmongoc::client_encryption_encrypt_opts_new();
 
-    if (_key_id_set && _key_alt_name) {
+    if (_key_id && _key_alt_name) {
```
kevinAlbs, line 61: Do you recall if this empty if block was intentional? If it's just for the purpose of documentation I'd suggest removing the if and expanding the comment to say "libmongoc will error if both id and alt name are set".

src/bsoncxx/types/bson_value/value.hpp
```diff
@@ -49,7 +49,7 @@ class BSONCXX_API value {
     ///
     /// Create an owning copy of a bson_value::view.
     ///
-    value(const view&);
+    explicit value(const view&);
```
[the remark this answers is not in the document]
samantharitter: Yes!  view_or_value did not like that the non-explicit constructor allowed conversions from value to view and vice versa.  The error was caused by the use of a ternary operator in view_or_value:
`_view(_value ? *_value : std::move(other._view))`

With a non-explicit constructor, the compiler couldn't decide whether to convert the first return to a view, or convert the second return to a value in order to make the types match.  With the constructor explicit, the ambiguity goes away.  I didn't see a downside to having the constructor be explicit, as `bson_value::value` has a `view()` operator, so it's easy enough to call that if needed.

src/mongocxx/options/encrypt.cpp
samantharitter: haha!

src/mongocxx/options/encrypt.cpp
samantharitter: Oh right, I forgot we made a ticket for that, that's good.

src/mongocxx/options/encrypt.cpp
samantharitter: I think it was just a readability thing, happy to remove it.

PR #689, opened 2020-07-08
CXX-1911 GridFS index checking should support indexes created in the shell

bazile-clyde:


src/mongocxx/test/gridfs/bucket.cpp
```diff
@@ -1059,4 +1056,57 @@ TEST_CASE("gridfs download large file", "[gridfs::bucket]") {
...
+
+    auto downloader = bucket.open_download_stream(result.id());
+    auto bytes_read = downloader.read(to_read.data(), file_size);
+    REQUIRE(bytes_read == file_size);
+
+    SECTION("default indexes created") {
+        auto bucket_collection_name = GENERATE(values({"fs.chunks", "fs.files"}));
+        auto indexes = db[bucket_collection_name].list_indexes();
+        REQUIRE(std::distance(indexes.begin(), indexes.end()) == 2);
+    }
```
bazile-clyde, line 1086: So after uploading arbitrary data the default indexes are 
```
  { "v" : 2, "key" : { "_id" : 1 }, "name" : "_id_" }
  { "v" : 2, "unique" : true, "key" : { "files_id" : 1, "n" : 1 }, "name" : "files_id_1_n_1" }
```
This looks ok to me but weirdly, on line 1092, changing `_id` to `files_id` fails since the DB command doesn't match that index (`files_id`) with the one created here (`_id`) and doesn't throw an error. However, in C ([here](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-gridfs-bucket.c#L183)) we test this using `files_id`. I just wanted to point this out because I'm not sure why this works in C but not here.

src/mongocxx/test/gridfs/bucket.cpp
```diff
@@ -1059,4 +1056,57 @@ TEST_CASE("gridfs download large file", "[gridfs::bucket]") {
...
+    SECTION("fs.chunks 'files_id' and 'n' compares by numeric value") {
+        auto create_index_cmd =
+            make_document(kvp("createIndexes", "fs.chunks"), kvp("indexes", [&](sub_array sub_arr) {
+                              sub_arr.append(make_document(
+                                  kvp("key", make_document(kvp("_id", 1.0), kvp("n", 1.0))),
+                                  kvp("name", "files_id_1_n_1"),
+                                  kvp("unique", true)));
+                          }));
+        REQUIRE_THROWS_WITH(db.run_command(create_index_cmd.view()),
+                            Catch::Contains("must have unique name"));
```
bazile-clyde, line 1097: I'm expecting both indexes (`fs.chunks` and `fs.files`) to _not_ get created since they already exist. That should confirm that we're correctly matching the floating-point values here with the ints created by default. 

These do throw slightly different errors though and I wanted to point that out in case I'm doing anything wrong.

bazile-clyde:
FYI, [here](https://evergreen.mongodb.com/version/5f072e4157e85a4a43f82c7f) is the latest patch build to address the evergreen failures. I did two things: 
1) added yet another substring the `db.run_command` could throw... 
```
        REQUIRE_THROWS_WITH(db.run_command(create_index_cmd.view()),
                            Catch::Contains("must have unique name") ||
                                Catch::Contains("An existing index has the same name"));
```
2) included the numeric library for `iota` for windows.

I'm still not sure why we get so many different but similar error messages. Will investigate

src/mongocxx/test/gridfs/bucket.cpp
kevinAlbs: I think the difference in server errors has to do with how server is comparing existing indexes.

As you noted, the gridfs bucket should have created this index on the chunks collection:
`{ "v" : 2, "unique" : true, "key" : { "files_id" : 1, "n" : 1 }, "name" : "files_id_1_n_1" }`

Since this is creating an index with a different key ` { _id: 1.0, n: 1.0 }`, but the same name. I think the server interprets that as an attempt to create a different index, since `_id` does not match `file_id`.

The test below attempts to create `{ filename: 1.0, uploadDate: 1.0 }`. The server may ignore the numeric type difference, and consider that an attempt to create the same exact index, which triggers the different error.

That's my guess. Checking in the shell seems to indicates the same. Running this twice:
```
db.runCommand({createIndexes: "coll", indexes: [{key: {x: 1}, name: "abc"}]})
```
Results in `all indexes already exist`

But creating a different index with the same name:
```
db.runCommand({createIndexes: "coll", indexes: [{key: {y: 1}, name: "index"}]})
```
Gets back `An existing index has the same name as the requested index`

src/mongocxx/test/gridfs/bucket.cpp
```diff
@@ -1059,4 +1056,57 @@ TEST_CASE("gridfs download large file", "[gridfs::bucket]") {
...
+    auto bytes_read = downloader.read(to_read.data(), file_size);
+    REQUIRE(bytes_read == file_size);
+
+    SECTION("default indexes created") {
+        auto bucket_collection_name = GENERATE(values({"fs.chunks", "fs.files"}));
+        auto indexes = db[bucket_collection_name].list_indexes();
+        REQUIRE(std::distance(indexes.begin(), indexes.end()) == 2);
+    }
+
+    SECTION("fs.chunks 'files_id' and 'n' compares by numeric value") {
```
kevinAlbs, line 1088: Per discussion, instead of attempting to create conflicting indexes after the gridfs bucket, create the indexes first (with varying the value between `1` and `1.0`, similar to the libmongoc test). Then assert that GridFS did not create additional indexes. 

We want to test the index comparison logic in GridFS. Creating the indexes first, then checking that GridFS did not create additional indexes, verifies that GridFS correctly detected that the indexes exist.

Note, looking at `create_indexes_if_nonexistent`, IIUC GridFS isn't checking for indexes at all (just that the collections are empty). So I expect tests would fail, and that `create_indexes_if_nonexistent` will need to be updated.

bazile-clyde: Doing it this way worked!! Now I create the index for `chunks` and `files` manually before uploading and downloading arbitrary data and ensure that no more indexes were created. I also updated the `TEST_CASE` and `SECTION` description to explicitly state that this is being tested. Also, TIL, that this code works...
```
TEST_CASE (...) {
    int i = 0;
    REQUIRE(i==0);
    SECTION(...) {
        i++;
        REQUIRE(i==1);
    }
    SECTION(...) {
        i++;
        REQUIRE(i==1);
    }
    REQUIRE(i==1);
}
```
This is what I modeled my code after.

kevinAlbs: Nice!

Re: my note about `create_indexes_if_nonexistent`, it really does not check whether indexes exist before attempting to create them. But I think that's ok, because the server will not return an error when attempting to create the same index twice.

Here's what happens creating the same index twice:
```
db> db.runCommand({"createIndexes": "coll", indexes: [{"name": "test", key: {test: 1}}]})
db> db.runCommand({"createIndexes": "coll", indexes: [{"name": "test", key: {test: 1}}]})
{
	"numIndexesBefore" : 4,
	"numIndexesAfter" : 4,
	"note" : "all indexes already exist",
	"ok" : 1,
	"$clusterTime" : {
		"clusterTime" : Timestamp(1594661966, 1),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1594661966, 1)
}
```

The second reply is not an error ("ok": 1). That seems ok to me. The GridFS spec does not say how to check for indexes, pymongo seems to rely on the same behavior, and even if we did check for indexes explicitly it'd be the same number of commands.

PR #691, opened 2020-07-10
CXX-1998 'CommitQuorum' option support for 'createIndexes’ command on MongoDB 4.4

bazile-clyde:


bazile-clyde:
@kevinAlbs I see you closed [CDRIVER-3627](https://jira.mongodb.org/browse/CDRIVER-3627). From the spec

> Q: Why does the driver manually throw errors if the commitQuorum option is specified against a pre 4.4 server?
Starting in 3.4, the server validates all options passed to the createIndexes command, but due to a bug in versions 4.2.0-4.2.5 of the server ([SERVER-47193](https://jira.mongodb.org/browse/SERVER-47193)), specifying commitQuorum does not result in an error. The option is used interally by the server on those versions, and its value could have adverse effects on index builds. To prevent users from mistakenly specifying this option, drivers manually verify it is only sent to 4.4+ servers.

I'm not sure if this was missed in C or it was decided that we shouldn't do it. This implementation doesn't currently throw an error for server version less than 4.4. If we decided against it in C then I'll add it here but if it was left out we should probably throw an error from libmongoc.

kevinAlbs:
> I'm not sure if this was missed in C or it was decided that we shouldn't do it. This implementation doesn't currently throw an error for server version less than 4.4. If we decided against it in C then I'll add it here but if it was left out we should probably throw an error from libmongoc.

It was not applicable for libmongoc. The only index creation helpers in libmongoc are deprecated (and not used by the C++ driver), and the generic command helpers should not be inspecting the command options.

To do the wire version check in the C++ driver, the only way I see is to use libmongoc's `mongoc_client_select_server` and inspect the wire version of the primary's server description using `mongoc_server_description_ismaster`. Neither function is used by the C++ driver, so you'll need to add it to `libmongoc_symbols.hh`. But... there's more. `mongoc_client_select_server` requires a `mongoc_client_t*`, but `mongocxx::index_view` only has a `mongoc_collection_t*`. I think `mongocxx::collection::indexes()` will need to additionally pass that through, which it can get from `_get_impl()->client_impl->client_t`.

TBH it seems like a very roundabout way for this check, but it also might be the best option we have right now. This is an unfortunate consequence of libmongoc not providing index helpers, it pushes more work onto wrapping drivers :/ I think this would be an argument for adding index helpers to libmongoc I do not know the rationale for why they were deprecated originally.

src/mongocxx/private/index_view.hh
```diff
@@ -134,6 +134,10 @@ class index_view::impl {
             opts_doc.append(bsoncxx::builder::concatenate_doc{session->_get_impl().to_document()});
         }
 
+        if (options.commit_quorum()) {
```
jmikola, line 137: I'd suggest a comment above this to explain why you're passing this command-specific option via `opts_bson`, which is typically reserved for generic command options. A reference to [CDRIVER-3748](https://jira.mongodb.org/browse/CDRIVER-3748) would work, like:

> libmongoc validates commitQuorum via command opts (CDRIVER-3748)

src/mongocxx/options/index_view.hpp
```diff
@@ -84,9 +84,63 @@ class MONGOCXX_API index_view {
...
 
+    ///
+    /// Sets the commit quorum for this operation.
+    ///
+    /// This option may only be used with MongoDB version 4.4 or later.
+    ///
+    /// @param commit_quorum
+    ///   Integer representing the minimum number of data-bearing voting replica set members (i.e.
+    ///   commit quorum), including the primary, that must report a successful index build before
+    ///   the primary marks the indexes as ready. A value of @c 0 disables quorum-voting behavior.
```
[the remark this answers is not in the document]
kevinAlbs: Ah gotcha. That's fine with me.

src/mongocxx/private/index_view.hh
```diff
@@ -134,6 +133,25 @@ class index_view::impl {
...
+            auto server_description = libmongoc::client_select_server(
```
kevinAlbs, line 137: `server_description` may be NULL. If so, `error` is set.

If `NULL` is returned, you can use `mongoc_error.hh`'s `throw_exception`, to throw an exception directly from the `bson_error_t`.

Similar to my comment below, perhaps use `write_exception`:

```
throw_exception<write_exception>(error);
```

Secondly, the returned server description needs to be manually destroyed with `mongoc_server_description_destroy`.

src/mongocxx/private/index_view.hh
```diff
@@ -134,6 +133,25 @@ class index_view::impl {
...
+                _client, true /* for_writes */, nullptr /* read_prefs */, &error);
+            auto is_master = libmongoc::server_description_ismaster(server_description);
+
+            bson_iter_t iter;
+            bson_iter_init_find(&iter, is_master, "maxWireVersion");
```
kevinAlbs, line 142: This could return false if the ismaster reply does not contain `maxWireVersion`. I don't think that would realistically happen (if there was no primary available, then `mongoc_client_select_server` would have returned NULL). But I think it is worth checking. If no `maxWireVersion` is available, throw an error.

src/mongocxx/private/index_view.hh
```diff
@@ -134,6 +133,25 @@ class index_view::impl {
...
+            int32_t max_wire_version = bson_iter_int32(&iter);
+
+            if (max_wire_version < 9) {
+                throw logic_error{
```
kevinAlbs, line 146: Looking at the documentation around the exception classes, `logic_error` is for API misuse. Since this is not an error from a server. Perhaps, in the absence of a more apt error class `write_exception` is more applicable?

kevinAlbs, line 146: The server description is leaked if this throws, this can be destroyed by calling the C function directly:

```
mongoc_server_description_destroy(server_description);
```

Or, I think the more idiomatic C++ solution would be wrap the C type in something that calls `mongoc_server_description_destroy` in the destructor (similar to how `scoped_bson_t` wraps a `bson_t`). If it seems reasonable, you could add a private helper class in this file for that:

```
class scoped_server_description {
public:
   scoped_server_description (mongoc_server_description_t *sd) : sd (sd) {}
   ~scoped_server_description () { mongoc_server_description_destroy (sd); }
   mongoc_server_description_t* sd;
};
```

Then, changing the assignment above should prevent any leaks on early returns or throws:

```
scoped_server_description scoped_sd(
    libmongoc::client_select_server(_client, true /* for_writes */, nullptr /* read_prefs */, &error)
);
```

src/mongocxx/private/libmongoc_symbols.hh
```diff
@@ -154,6 +154,7 @@ MONGOCXX_LIBMONGOC_SYMBOL(client_pool_push)
 MONGOCXX_LIBMONGOC_SYMBOL(client_pool_set_apm_callbacks)
 MONGOCXX_LIBMONGOC_SYMBOL(client_pool_try_pop)
 MONGOCXX_LIBMONGOC_SYMBOL(client_reset)
+MONGOCXX_LIBMONGOC_SYMBOL(client_select_server);
```
kevinAlbs, line 157: The trailing ; is not needed, and I think the cause of some compilation warnings on Evergreen.

src/mongocxx/private/index_view.hh
bazile-clyde: Done.

src/mongocxx/private/index_view.hh
bazile-clyde: Done

src/mongocxx/private/index_view.hh
bazile-clyde: Done

bazile-clyde: Thanks for the code snippet. Done!

src/mongocxx/private/libmongoc_symbols.hh
bazile-clyde: Done

src/mongocxx/test/index_view.cpp
```diff
@@ -172,6 +172,52 @@ TEST_CASE("create_one", "[index_view]") {
...
+
+        auto key = make_document(kvp("a", 1));
+        index_model model(key.view());
+        options::index_view options;
+
+        auto commit_quorum_regex =
+            Catch::Matches("(.*)commit( )?quorum(.*)", Catch::CaseSensitive::No);
+
+        using namespace test_util;
+        bool is_supported = compare_versions(get_server_version(mongodb_client), "4.4") >= 0 &&
```
kevinAlbs, line 190: The test failure on s390x seems related to the version of MongoDB running.

The logs of:
https://evergreen.mongodb.com/task/cxx_driver_ubuntu1604_zseries_compile_and_test_with_shared_libs_patch_510d2b0c04d1103bd1252088cda78522346b1f25_5f0f5f7f7742ae3af8c73ca8_20_07_15_19_56_48

say:
```
db version v4.3.0-2027-gbe045fe
```

That test is running against single, so `is_supported` would be false, and an exception is expected. But since that server version is likely reporting wire version 9, your wire version check is not getting hit. I think the right thing to do is change this comparison to check the wire version instead:

```
bool is_supported =  get_max_wire_version (mongodb_client) >= 9 && get_topology (mongoc_client) != "single";
```

src/mongocxx/private/index_view.hh
```diff
@@ -219,6 +239,16 @@ class index_view::impl {
     }
 
     mongoc_collection_t* _coll;
+    mongoc_client_t* _client;
+
+    class scoped_server_description {
+       public:
+        explicit scoped_server_description(mongoc_server_description_t* sd) : sd(sd) {}
```
kevinAlbs, line 246: Oh nice, I forgot about marking that explicit.

src/mongocxx/options/index_view.hpp
```diff
@@ -84,9 +84,63 @@ class MONGOCXX_API index_view {
...
+    ///
+    /// @see
+    ///   https://docs.mongodb.com/master/reference/command/createIndexes
+    ///
+    index_view& commit_quorum(int commit_quorum);
+
+    ///
+    /// Sets the commit quorum for this operation.
+    ///
+    /// The server-side default is "votingMembers".
```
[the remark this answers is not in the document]
kevinAlbs: IIRC `dochub.mongodb...` links are permanent links so are slightly preferred, but the all but one of the links are to `docs.mongodb...` so I think that is fine to use for consistency. We can change them en-masse at some point in the future if need be.

src/mongocxx/test/index_view.cpp
```diff
@@ -172,6 +172,53 @@ TEST_CASE("create_one", "[index_view]") {
...
+                REQUIRE_THROWS_WITH(indexes.create_one(model, options), commit_quorum_regex);
+            }
+        }
+
+        SECTION("fails with invalid int") {
+            options.commit_quorum(-1);
+            REQUIRE_THROWS_WITH(indexes.create_one(model, options), commit_quorum_regex);
+        }
+
+        SECTION("works with string") {
```
kevinAlbs, line 208: This section, and the section below, appear to be the ones failing.

As the capturing of `is_supported` shows, `is_supported` is false, an an exception is expected.

IIUC an exception is expected because the test runs against a single topology, but "majority" should only be supported on a replica set. And the section below expects a server to always reject the value "bad_str"

I still believe this is due to the intermediate 4.3.0 server version this is running against on zSeries. The server may have implemented validation after 4.3.0, so the reported wire version is still 9, but invalid commitQuorum values are still accepted. https://jira.mongodb.org/browse/SERVER-41846 in particular (which has fixVersion 4.3.1) seems related. 

We could investigate further and check with the server team. But IMO, it is less important that we test server side validation of these values, and more important that we test that `commitQuroum` works when supported, and that the driver returns a client side error if the wire version is < 9.

I'd argue for removing the "fails with invalid string" test below, and only testing the "is_supported" case here.

bazile-clyde: Per our discussion, I skipped single typologies and got rid of the `invalid with...` tests.

src/mongocxx/test/index_view.cpp
```diff
@@ -174,6 +174,9 @@ TEST_CASE("create_one", "[index_view]") {
     }
 
     SECTION("commitQuorum option") {
+        if (test_util::get_topology(mongodb_client) == "single")
```
kevinAlbs, line 177: Can you add a warning for consistency with other skips? `WARN("skip: commitQuorum option requires a replica set");`

bazile-clyde: Done.

PR #692, opened 2020-07-10
CXX-2069 Treat CursorNotFound as a resumable change stream error

bazile-clyde:


bazile-clyde:
Handled by C. Latest evg: https://evergreen.mongodb.com/version/5f08e0090305b922a456eb84

PR #694, opened 2020-07-17
CXX-1853, CXX-1947 pass hint to update and delete

bazile-clyde:


src/mongocxx/test/spec/crud.cpp
```diff
@@ -59,107 +59,110 @@ void run_crud_tests_in_file(std::string test_path) {
...
-
-        if (is_unsupported(test["description"].get_utf8().value)) {
-            continue;
-        }
-
-        if (should_skip_spec_test(client, test.get_document())) {
-            continue;
-        }
+        auto description = test["description"].get_utf8().value;
+        SECTION(to_string(description)) {
```
bazile-clyde, line 63: Everything below this (excluding line 91, which I changed) stayed the same. It's just indented.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -389,42 +389,24 @@ std::string tolowercase(stdx::string_view view) {
...
-        REQUIRE_BSON_MATCHES(doc, next_expected);
-    }
+    auto actual = coll->find({}, options);
+    std::equal(std::begin(actual),
+               std::end(actual),
+               std::begin(expected["data"].get_array().value),
+               [](document::view doc, bsoncxx::array::element ele) {
+                   REQUIRE_BSON_MATCHES(doc, ele.get_document().value);
+                   return true;
+               });
```
bazile-clyde, line 407: During my debugging efforts, I had a difficult time trying to parse what was happening here. I realized all this did was assert that two lists contained the same elements and had the same length. `REQUIRE_BSON_MATCHES` will show mismatches in the debugging output.

bazile-clyde: Updated this in the last commit. Since this checks from `begin(range1)` to `end(range1)` and iterates `range2` this would've returned `true` if `range1` was empty. I swapped the ranges so we check at least the length of `expected_data`. Afterward actual should be empty since it's a cursor if they're the same length.

bazile-clyde: Also, I removed the `write_concern` swap logic and ran it evergreen ([here](https://evergreen.mongodb.com/version/5f11c0d1850e61351dc5975f)). All of the tests pass so I'm not sure if we still need it.

src/mongocxx/collection.hpp
```diff
@@ -15,10 +15,6 @@
 #pragma once
 
 #include <algorithm>
-#include <cstdint>
-#include <memory>
-#include <string>
```
kevinAlbs, line 20: Nice. I'm curious, did you use a separate tool to determine these headers were unused? Or did this come from clang-tidy / compiler warnings?

src/mongocxx/test/collection_mocked.cpp
```diff
@@ -598,6 +597,10 @@ TEST_CASE("Collection", "[collection]") {
                     }
                 }
 
+                if (expected_hint) {
+                    REQUIRE(options_view["hint"].get_utf8() == expected_hint->get_utf8());
+                }
```
kevinAlbs, line 602: Here, and in the other expectations, it might be worth adding:

```
else {
    REQUIRE (!options_view["hint"]);
}
```

To ensure we're not sending an empty hint document when none is given.

src/mongocxx/bulk_write.cpp
```diff
@@ -61,6 +59,9 @@ bulk_write& bulk_write::append(const model::write& operation) {
             if (operation.get_update_one().collation()) {
                 options_builder.append(kvp("collation", *operation.get_update_one().collation()));
             }
+            if (operation.get_update_one().hint()) {
+                options_builder.append(kvp("hint", operation.get_update_one().hint()->to_value()));
```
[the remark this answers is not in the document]
bazile-clyde: Done

src/mongocxx/collection.hpp
bazile-clyde: We probably _should_ get a tool but CLION just highlights it for me.

src/mongocxx/test/collection_mocked.cpp
bazile-clyde: Done

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -389,42 +389,26 @@ std::string tolowercase(stdx::string_view view) {
...
-            FAIL("found more documents than expected.");
-        }
-
-        auto next_expected = *it;
-        it++;
-
-        REQUIRE_BSON_MATCHES(doc, next_expected);
-    }
-
+    options.sort(make_document(kvp("_id", 1)));
```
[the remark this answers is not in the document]
bazile-clyde: Oh, this was actually you. I just made it one line. I closed that ticket and linked the one it was done in. Side note, I ran these tests without the `read_concern` logic and it worked fine. Not really sure if we still need it.

src/mongocxx/test_util/client_helpers.cpp
bazile-clyde: Ah, nevermind. After investigating CDRIVER-2901 it seems better to leave it in.

kevinAlbs: This looks a lot cleaner. And I agree, using local read concern seems necessary.

PR #698, opened 2020-07-21
add code cov for all linux tests

bazile-clyde:


CMakeLists.txt
```diff
@@ -128,7 +131,7 @@ endif()
 
 
 # If the user did not customize the install prefix,
-# set it to live under build so we don't inadverently pollute /usr/local
+# set it to live under build so we don't inadvertently pollute /usr/local
```
bazile-clyde, line 134: misspelling

bazile-clyde:
This adds code coverage for build variants compiled with g++ or clang. It does not add code coverage for Windows, Mac, or any variants that already have ASAN, UBSAN or Valgrind enabled.

CMakeLists.txt
rcsanchez97: Lol.  That has been like that for 5 1/2 years.  I don't even know how many times I've read through this file and not noticed the misspelling.

.mci.yml
```diff
@@ -601,6 +618,7 @@ axes:
                 cmake_flags: *linux_cmake_flags
                 build_type: "Debug"
                 tar_options: *linux_tar_options
+                code_coverage_cmake_flags: *code_coverage_cmake_flags
```
kevinAlbs, line 621: Awesome! Looking at the runtime difference, it seems like tests only take one more minute with code coverage enabled. That seems like a worthwhile trade.

We'd likely get even more accurate coverage if we run the replica set tests with code coverage as well (e.g. transactions tests only run against a replica set). Currently there's only one replica set task that's run on the release variant.

If we add `compile_and_test_with_shared_libs_replica_set` to the "integration" matrix below, that may do the trick. Though there may be some unexpected failures / additional setup.

No need to do this now if it's non-trivial. This is an overall improvement as-is.

bazile-clyde: Ah, good call. I was able to test against a replica set for Linux only. Windows fails so I left out of this commit.

PR #701, opened 2020-07-23
CXX-1619 add new prose tests for oid

samantharitter:


src/bsoncxx/test/oid.cpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    SECTION("overflows predictably") {
+        /* The C++ driver does not have the ability to set a context for oids, so we
+           rely on the C driver's testing of this behaviour */
+    }
+
+    SECTION("oid", "uses a different machine/process number after fork is called") {
+        /* The C++ driver does not have the ability to set a context for oids, so we
+           rely on the C driver's testing of this behaviour */
```
kevinAlbs, line 68: If this uses the default behavior, then I think we could test this. The default `bson_context_t` has the PID cache disabled: http://mongoc.org/libbson/current/bson_context_get_default.html

samantharitter: Unfortunately the catch test framework does not support forking *sigh*

kevinAlbs: Ah gotcha.

kevinAlbs (approved):

PR #703, opened 2020-07-27
CXX-2007 hidden indexes support

bazile-clyde:


src/mongocxx/options/index_view.cpp
```diff
@@ -25,7 +25,7 @@ namespace mongocxx {
 MONGOCXX_INLINE_NAMESPACE_BEGIN
 namespace options {
 
-index_view::index_view() : _max_time(), _write_concern() {}
+index_view::index_view() : _max_time(), _write_concern(), _commit_quorum() {}
```
bazile-clyde, line 28: Forgot to do this during the commit_quorum support PR

src/mongocxx/test/collection.cpp
```diff
@@ -2406,13 +2406,15 @@ TEST_CASE("create_index tests", "[collection]") {
...
+        options.hidden(true);
         options.expire_after(std::chrono::seconds(500));
         options.name(index_name);
 
         REQUIRE_NOTHROW(coll.create_index(keys.view(), options));
 
         bool unique = options.unique().value();
-        auto validate = [unique](bsoncxx::document::view index) {
+        bool hidden = options.hidden().value();
+        auto validate = [=](bsoncxx::document::view index) {
```
samantharitter, line 2417: So, the `[=]` copy-capture will capture all current variables by copy.  `[&]`, reference-capture, is a better choice here, because there aren't any lifetime issues with these variables that would require making a copy for the lambda.  The fewer copies we make the better!

src/mongocxx/options/index_view.cpp
samantharitter: Acknowledged.

src/mongocxx/test/collection.cpp
bazile-clyde: Done

src/mongocxx/test/collection.cpp
```diff
@@ -2406,13 +2406,13 @@ TEST_CASE("create_index tests", "[collection]") {
 
         options::index options{};
         options.unique(true);
+        if (test_util::newer_than(mongodb_client, "4.4"))
```
bazile-clyde, line 2409: Wrapping the `hidden` option with this check (and below on line 2426) fixed the test failures.

PR #704, opened 2020-07-27
CXX-1920 Remove zSeries ubuntu 16.04

bazile-clyde:


.mci.yml
```diff
@@ -106,6 +106,7 @@ functions:
                  g++ --version || true
                  clang --version || true
                  cmake --version || true
+                 openssl version || true
```
bazile-clyde, line 109: This helped debug the incompatible cmake and openssl version issue

.mci.yml
```diff
@@ -745,16 +748,17 @@ buildvariants:
...
           - name: compile_and_test_with_static_libs
 
-    - name: ubuntu1604-zseries
-      display_name: "s390x Ubuntu 16.04 (MongoDB Latest)"
+    - name: ubuntu1804-zseries
+      display_name: "s390x Ubuntu 18.04 (MongoDB Latest)"
       batchtime: 1440 # 1 day
       expansions:
           build_type: "Release"
+          cmake: "cmake"
```
bazile-clyde, line 756: Looking for `cmake` in `/opt/cmake/bin/cmake` caused the test failures [here](https://evergreen.mongodb.com/task_log_raw/cxx_driver_ubuntu1804_zseries_compile_and_test_with_shared_libs_patch_3fe4528b7d33b5230f52895b3051be07cbfb0032_5f1aff467742ae3ba1b0e81c_20_07_24_15_34_56/0?type=T#L357) since that version depended on openssl 1.1.1. This defaults to the one in `/usr/bin`

bazile-clyde:
Unfortunately, upgrading this to 18.04 didn't fix the test failures as we hoped ([context](https://jira.mongodb.org/browse/CXX-2079))

.mci.yml
samantharitter: Acknowledged

.mci.yml
```diff
@@ -172,6 +173,7 @@ functions:
                     if [ "Windows_NT" == "$OS" ]; then
                         PREFIX=$(cygpath -m "$PREFIX")
                     fi
+                    export CMAKE=${cmake}
```
samantharitter, line 176: Does this line have an effect?  Who consumes the new CMAKE variable?  Also, what is the output of ${cmake} ?  When I run that locally in my shell, the result is empty.

.mci.yml
```diff
@@ -211,6 +213,7 @@ functions:
                   if [ "Windows_NT" == "$OS" ]; then
                       MONGOC_PREFIX=$(cygpath -m "$MONGOC_PREFIX")
                   fi
+                  export CMAKE=${cmake}
```
samantharitter, line 216: Same questions as above?

.mci.yml
samantharitter: Acknowledged.

.mci.yml
bazile-clyde: It overrides the checks in `find_cmake.sh`. A lot of files use that script to find `cmake` iff the env var `CMAKE` isn't set. I set it here to avoid that.

.mci.yml
```diff
@@ -745,16 +749,17 @@ buildvariants:
           # signature.
           - name: compile_and_test_with_static_libs
 
-    - name: ubuntu1604-zseries
-      display_name: "s390x Ubuntu 16.04 (MongoDB Latest)"
+    - name: ubuntu1804-zseries
+      display_name: "s390x Ubuntu 18.04 (MongoDB 4.4)"
```
bazile-clyde, line 753: So "lastest" for this platform points to 4.3, not 4.4, which is what we want to test. Upgrading the server version to 4.4 fixed the tests ([latest evg build](https://evergreen.mongodb.com/version/5f20a1fbd1fe0740912aa8d4))

.mci.yml
bazile-clyde: Answered above.

bazile-clyde:
> Why don't we put a custom skip in the crud spec test runner for this failing hint test? Before we bump the zSeries ubuntu, I'd like to make sure it can have a green test run.

Got it. The last tests I ran are green after running against 4.4 and not 4.3 like we were doing previously.

.mci.yml
kevinAlbs: Hah, ironically the latest is older than the released.

I think we'll want to add a `latest` variant again later, when the latest download does update beyond 4.4. I created CXX-2085 to track that.

.mci.yml
kevinAlbs: Good find! Seems related to the issue I encountered attempting to upgrade the `power8-ubuntu1604` to Ubuntu 18.04: https://jira.mongodb.org/browse/CXX-2048. I'll likely use your new expansion to refer to the installation of cmake.

PR #705, opened 2020-08-06
CXX-1916 Only sync json files for spec tests

bazile-clyde:


bazile-clyde:
This just deletes the YML and RST files. Later the `test_files.txt` files will be deleted too and the only thing left in these folders will be the JSON files. We can then simply just run all files in this folder through the test runner.

samantharitter:
LGTM!

PR #707, opened 2020-08-17
CXX-2093 fix rpm-package-build task

rcsanchez97:
This fixes the failure of {{spec.patch}} to apply as a result of downstream RPM spec.  Note that the task continues to fail because of an unrelated transient issue in the mock environment provisioning.

Evergreen patch build: https://spruce.mongodb.com/version/5f3aeafe7742ae2a6ba05436/tasks


rcsanchez97:
@kevinAlbs I've merged this PR, but I will wait for another few days for the transient mock failure to resolve itself and for the task to turn green on the waterfall before I close the Jira ticket.

PR #708, opened 2020-08-19
update badges

bazile-clyde:
It isn't clear which links link to Doxygen or mongocxx.org. So, I added badges that link to our Doxygen and mongocxx pages. I also made the link to our evergreen build a badge so it's more uniform with the others. It didn't look clickable and the icon isn't as well know as we might like.


bazile-clyde:
Also added one for our license

PR #710, opened 2020-08-24
CXX-1530 remove unused and unmaintained examples

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/5f43f7b8c9ec4443c8b13db1/tasks


examples/mongocxx/aggregate.cpp
```diff
@@ -1,69 +0,0 @@
...
-    // respectively. Therefore, a mongocxx::instance must be created before using the driver and
-    // must remain alive for as long as the driver is in use.
-    mongocxx::instance inst{};
-    mongocxx::client conn{mongocxx::uri{}};
-
-    auto db = conn["test"];
-
-    // Group documents by field and calculate count.
-    {
-        // @begin: cpp-group-documents-by-a-field-and-calculate-count
```
kevinAlbs, line 39: Apologies if there was miscommunication. I think these examples may be valuable for users reading through the source. There may be some redundancy with the examples written in `examples/mongocxx/mongodb.com/` but some of these examples aren't covered there (like `view_or_value_variant.cpp`) and having them as independent examples is probably more digestible.

I think we only need to remove the comment annotations like this one, since they aren't used.

rcsanchez97: :facepalm:

PR #711, opened 2020-08-24
CID 113953/113952 Uninitialized pointer read

bazile-clyde:
This commit fixes two high impact coverity issues.

### DIFF
--- a/src/mongocxx/test/spec/operation.cpp
+++ b/src/mongocxx/test/spec/operation.cpp
@@ -1275,7 +1275,7 @@ document::value operation_runner::_create_index(const document::view& operation)
     return _coll->create_index(*session, keys, opts.extract());
 }
 
-operation_runner::operation_runner(collection* coll) : _coll(coll) {}
+operation_runner::operation_runner(collection* coll) : operation_runner(nullptr, coll) {}
 operation_runner::operation_runner(database* db,
                                    collection* coll,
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/spec/operation.cpp
```diff
@@ -1275,7 +1275,7 @@ document::value operation_runner::_create_index(const document::view& operation)
     return _coll->create_index(*session, keys, opts.extract());
 }
 
-operation_runner::operation_runner(collection* coll) : _coll(coll) {}
```
bazile-clyde, line 1278: Calling this constructor will leave `_db` uninitialized (and only `_db` since the other parameters default to `nullptr`). Since `_db` is a pointer, we'd be reading from an arbitrary memory adress.

src/mongocxx/test/spec/operation.cpp
```diff
@@ -1275,7 +1275,7 @@ document::value operation_runner::_create_index(const document::view& operation)
...
+operation_runner::operation_runner(collection* coll) : operation_runner(nullptr, coll) {}
```
bazile-clyde, line 1278: A [delegating constructor](https://docs.microsoft.com/en-us/cpp/cpp/delegating-constructors?view=vs-2019) that sets `_db` to `nullptr` fixes the issue.

PR #713, opened 2020-08-26
CXX-2094 Resync change stream spec tests

kevinAlbs:


data/change_stream/change-streams.json
```diff
@@ -6,7 +6,7 @@
   "tests": [
     {
       "description": "$changeStream must be the first stage in a change stream pipeline sent to the server",
-      "skipReason" : "CXX-2025",
+      "skipReason" : "CXX-2084",
```
kevinAlbs, line 9: This is the new ticket to track this skipped test.

PR #714, opened 2020-08-26
CXX-1095 Add clang-tidy integration and Evergreen task

bazile-clyde:
[Clang-tidy](https://clang.llvm.org/extra/clang-tidy/) has a ton of checks you can add. I added the bare minimum, least controversial ones in this PR:
```
cert- | Checks related to CERT Secure Coding Guidelines.
clang-analyzer- | Clang Static Analyzer checks.
cppcoreguidelines- | Checks related to C++ Core Guidelines.
```
This PR fails as expected. We have a lot of nits to clean up. In the future, we should probably add checks one-by-one.


.mci.yml
```diff
@@ -801,6 +820,18 @@ buildvariants:
           # signature.
           - name: compile_and_test_with_static_libs
 
+    - name: ubuntu1804-debug
```
bazile-clyde, line 823: This is to use the latest version of clang-tidy.

.mci.yml
```diff
@@ -848,7 +879,7 @@ buildvariants:
 
     # Test that we actually support building against the minium required C version claimed.
     - name: ubuntu1804-min-libmongoc
-      display_name: "Ubuntu 18.04 with minimum libmongoc (MongoDB 4.4)"
+      display_name: "Ubuntu 18.04 with minimum libmongoc (MongoDB Latest)"
```
bazile-clyde, line 882: I'm not sure why the description said 4.4 but it uses the latest. I'm not sure if we actually wanted to use the latest or 4.4 though.

etc/run-clang-tidy.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
kevinAlbs, line 1: To confirm, developers should be able to run this themselves, `./etc/run-clang-tidy.sh`, right? If so, I'd make a note in https://github.com/mongodb/mongo-cxx-driver/blob/master/CONTRIBUTING.md to run this script before submitting a PR.

kevinAlbs (changes requested):
This is great! I'm a little worried about adding a failing task that might be failing for some time until we resolve those checks. It's not clear to me how to address some of those warnings. The stats for those counts of those warnings are:
```
[('cppcoreguidelines-pro-type-union-access', 76),
 ('cppcoreguidelines-pro-bounds-array-to-pointer-decay', 62),
 ('cppcoreguidelines-pro-bounds-pointer-arithmetic', 53),
 ('cppcoreguidelines-pro-type-vararg', 50),
 ('cppcoreguidelines-pro-type-const-cast', 44),
 ('cppcoreguidelines-pro-type-member-init', 36),
 ('cppcoreguidelines-pro-type-reinterpret-cast', 32),
 ('cppcoreguidelines-pro-type-cstyle-cast', 15),
 ('cert-err60-cpp', 8),
 ('cppcoreguidelines-owning-memory', 8),
 ('cppcoreguidelines-special-member-functions', 7),
 ('cppcoreguidelines-pro-bounds-constant-array-index', 5),
 ('cert-err58-cpp', 5),
 ('cert-dcl21-cpp', 4),
 ('cppcoreguidelines-pro-type-static-cast-downcast', 2)]
```
(generated [with this script](https://gist.github.com/kevinAlbs/26cb33857365e42e6c9055bc97956c9a))

The casting warnings seem like a straightforward fix, but `cppcoreguidelines-pro-type-union-access` warnings are related to the bson_types structs being used in a union. And `cppcoreguidelines-pro-type-vararg` are produced for calls into C functions that take variadic args like `bson_set_error`. I'm not sure what the workaround is.

So, unless we know how to fix those now, is it possible to list the warnings we are currently getting to `run-clang-tidy.sh` to treat those as warnings? Otherwise, we'll likely be ignoring the output of this task for a while, and it will continue to make PR test results considered failures.

bazile-clyde:
Per our discussion, I remove "warnings-as-errors"

kevinAlbs (approved):
LGTM!

PR #715, opened 2020-08-27
CXX-2036 give default values for env vars

bazile-clyde:


src/mongocxx/test/spec/util.hh
```diff
@@ -97,8 +97,9 @@ uri get_uri(document::view test);
 // and run them with the provided callback function.
 //
 using test_runner = std::function<void(const std::string& file)>;
-void run_tests_in_suite(std::string ev, test_runner cb, std::set<std::string> unsupported_tests);
-void run_tests_in_suite(std::string ev, test_runner cb);
+void run_tests_in_suite(std::string ev,
+                        test_runner cb,
+                        std::set<std::string> unsupported_tests = {});
```
bazile-clyde, line 102: Just gave this a default value rather than overloading it.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -197,6 +197,9 @@ TEST_CASE("Change stream spec tests", "[change_stream_spec]") {
         return;
     }
 
-    run_tests_in_suite("CHANGE_STREAM_TESTS_PATH", &run_change_stream_tests_in_file);
+    auto tests_path = std::getenv("CHANGE_STREAM_TESTS_PATH")
+                          ? std::getenv("CHANGE_STREAM_TESTS_PATH")
+                          : std::string{MONGOCXX_SOURCE_DIR} + "/data/change_stream";
```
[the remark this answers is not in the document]
kevinAlbs: There are a few other missing `require (!path.empty())`. Instead of fixing them, perhaps it's best to have a utility function in `src/mongocxx/test/spec/util.hh` to return a path given the environment variable name? It doesn't save much code, but it might be nice to have all of the environment variable path handling in one spot. (At some point we could remove the environment variables entirely, or just have one `DATA_PATH`).

src/mongocxx/test/spec/util.cpp
```diff
@@ -434,10 +434,11 @@ uri get_uri(document::view test) {
...
 }
 
-void run_tests_in_suite(std::string ev, test_runner cb, std::set<std::string> unsupported_tests) {
-    char* tests_path = std::getenv(ev.c_str());
-    INFO("checking for path from environment variable: " << ev);
-    REQUIRE(tests_path);
+void run_tests_in_suite(std::string tests_path,
+                        test_runner cb,
+                        std::set<std::string> unsupported_tests) {
+    CAPTURE(tests_path);
```
kevinAlbs, line 440: Nice, that's helpful.

bazile-clyde:
I'll be holding off on this since, per the comment in [CDRIVER-3784](https://jira.mongodb.org/browse/CDRIVER-3784), `CMAKE_SOURCE_DIR` is getting removed.

PR #718, opened 2020-09-03
Fix build with libcxx

orivej:
Fixes error "implicit instantiation of undefined template 'std::string'"
requested in mongocxx/options/auto_encryption.hpp:256:29:
    stdx::optional<ns_pair> _key_vault_namespace;

and likewise in client_encryption.hpp.


kevinAlbs (approved):
Thanks for the contribution!

PR #719, opened 2020-09-05
Remove unused headers

bazile-clyde:


bazile-clyde:
This commit removes unused include directives. Unfortunately, this had to be done semi-manually. CLion allows you to analyze all source files for unused includes but I could not find the mythical "bulk fix" option. So I ended up using the list to get rid of as many of these issues as I could. This obviously isn't a scalable strategy but once we do find a reliable tool this will make the initial analysis that much easier.

On Include What You Use ([IWYU](https://github.com/include-what-you-use/include-what-you-use)): 
From their ReadMe:
> This is alpha quality software -- at best (as of July 2018). It was originally written to work specifically in the Google source tree, and may make assumptions, or have gaps, that are immediately and embarrassingly evident in other types of code.

In the meantime, I'll have to search for other tools.

src/mongocxx/cursor.cpp
```diff
@@ -14,17 +14,11 @@
...
-
 #include <bsoncxx/private/libbson.hh>
 #include <bsoncxx/stdx/make_unique.hpp>
 #include <mongocxx/exception/private/mongoc_error.hh>
 #include <mongocxx/exception/query_exception.hpp>
 #include <mongocxx/private/cursor.hh>
-#include <mongocxx/private/libbson.hh>
 #include <mongocxx/private/libmongoc.hh>
 
 #include <mongocxx/config/private/prelude.hh>
```
rayangler, line 24: I think I understand the general organization of includes for each file, but I'm just curious on why this particular include is at the very end, separated from other includes?

rayangler (approved):
LGTM!

src/mongocxx/cursor.cpp
bazile-clyde: While I'm not sure why this style choice was chosen originally, I'd guess because it's special. Pretty much all of our files need to include the prelude to push important macros. So we wouldn't want to risk this getting removed/modified. Separating it definitely helped _me_ avoid removing it in this commit.

kevinAlbs: CONTRIBUTING.md notes the general layout:
https://github.com/mongodb/mongo-cxx-driver/blob/master/CONTRIBUTING.md#headers
Though it is not enforced by clang format.

src/mongocxx/gridfs/bucket.cpp
```diff
@@ -31,8 +31,6 @@
 #include <mongocxx/options/index.hpp>
 #include <mongocxx/stdx.hpp>
 
-#include <mongocxx/config/private/prelude.hh>
```
kevinAlbs, line 34: I'd vote for leaving this in. Though CONTRIBUTING.md does not specify prelude as being necessary for .cpp files, the only cpp files that do not include prelude are tests/examples. It might be useful to have prelude included in all source files.

Additionally, I think MONGOCXX_INLINE_NAMESPACE_BEGIN is most directly included through prelude.hh => prelude.hpp. Though it'd pick it up from other headers, having the more direct include seems preferable.

src/mongocxx/options/encrypt.cpp
```diff
@@ -12,16 +12,14 @@
...
-
 #include <bsoncxx/types/private/convert.hh>
+#include <mongocxx/config/private/prelude.hh>
 #include <mongocxx/exception/error_code.hpp>
 #include <mongocxx/exception/logic_error.hpp>
+#include <mongocxx/options/encrypt.hpp>
 #include <mongocxx/private/libbson.hh>
 #include <mongocxx/private/libmongoc.hh>
 
-#include <mongocxx/config/private/prelude.hh>
```
kevinAlbs, line 23: I suggest leaving the prelude header include here per CONTRIBUTING.md's guidelines.

src/mongocxx/private/change_stream.hh
```diff
@@ -17,11 +17,12 @@
 #include <bsoncxx/document/view.hpp>
 #include <bsoncxx/stdx/optional.hpp>
 #include <mongocxx/change_stream.hpp>
+#include <mongocxx/config/private/prelude.hh>
+#include <mongocxx/exception/private/mongoc_error.hh>
+#include <mongocxx/exception/query_exception.hpp>
 #include <mongocxx/private/libbson.hh>
 #include <mongocxx/private/libmongoc.hh>
 
-#include <mongocxx/config/private/prelude.hh>
```
kevinAlbs, line 23: Ditto about moving the header back.

kevinAlbs (changes requested):
This is certainly a refreshing cleanup. Just a couple comments about the prelude header but otherwise looks great!

src/mongocxx/gridfs/bucket.cpp
bazile-clyde: Done

src/mongocxx/options/encrypt.cpp
bazile-clyde: Done

src/mongocxx/private/change_stream.hh
bazile-clyde: Done

kevinAlbs (approved):
LGTM!

PR #720, opened 2020-09-08
CXX-1643 Replace mongoc_uri_new for better exception message

rayangler:


src/mongocxx/exception/error_code.cpp
```diff
@@ -54,7 +54,7 @@ class error_category final : public std::error_category {
             case error_code::k_server_response_malformed:
                 return "the response from the server was malformed";
             case error_code::k_invalid_uri:
-                return "an invalid MongoDB URI was provided";
+                return "invalid URI schema, expecting 'mongodb://' or 'mongodb+srv://'";
```
bazile-clyde, line 57: Why this change?

src/mongocxx/uri.cpp
```diff
@@ -48,9 +48,12 @@ uri::uri(std::unique_ptr<impl>&& implementation) {
     _impl.reset(implementation.release());
 }
 
-uri::uri(bsoncxx::string::view_or_value uri_string)
-    : _impl(stdx::make_unique<impl>(libmongoc::uri_new(uri_string.terminated().data()))) {
-    if (_impl->uri_t == nullptr) {
+uri::uri(bsoncxx::string::view_or_value uri_string) {
+    bson_error_t error = {0};
```
bazile-clyde, line 52: I think this is causing the ASAN issues on evergreen. Since `bson_error_t` is a _C_ type and not a _C++_ type it doesn't have a constructor. In general, _C_ isn't an object-oriented language. That means no constructors among other things. You could probably just remove the assignment here since `libmongoc::uri_new_with_error` will handle initialization.

src/mongocxx/uri.cpp
```diff
@@ -48,9 +48,12 @@ uri::uri(std::unique_ptr<impl>&& implementation) {
...
+    _impl = stdx::make_unique<impl>(
+        libmongoc::uri_new_with_error(uri_string.terminated().data(), &error));
+
+    if (error.domain) {
```
bazile-clyde, line 56: Since [mongoc_uri_new_with_error](http://mongoc.org/libmongoc/1.16.2/mongoc_uri_new_with_error.html) will set `error` to `NULL` on success, this will fail if the function succeeds. What you can do instead is check the result of `mongoc_uri_new_with_error` instead. According to the docs, that function returns `NULL` if it fails. Here's an example from `index_view.hh:197-204`:
```
        bson_error_t error;

        bool result = libmongoc::collection_write_command_with_opts(
            _coll, command_bson.bson(), opts_bson.bson(), reply.bson_for_init(), &error);

        if (!result) {
            throw_exception<operation_exception>(error);
        }
```

bazile-clyde (changes requested):
Looking good so far! I left a couple of comments that might help with the failing tests on evergreen.

src/mongocxx/exception/error_code.cpp
rayangler: Based on the description of the ticket, I assumed that we wanted the warning generated by the C driver to be used as the error message when throwing the exception (as well as provides additional context to the "invalid URI" error). Should we leave it to how it was before?

bazile-clyde: Your assumption is right! The error message here before was for the general `k_invalid_uri`  error code. The text here now refers to a specific error code. Let's keep this general and add the error message (`error.message`) returned from the _C_ function as an argument to the `logic_error` constructor. So instead of:
```
        throw logic_error{error_code::k_invalid_uri};
```
We can just do this:
```
        throw logic_error{error_code::k_invalid_uri, error.message};
```

src/mongocxx/test/uri.cpp
```diff
@@ -71,6 +71,11 @@ TEST_CASE("URI", "[uri]") {
             mongocxx::uri{invalid};
         } catch (const mongocxx::logic_error& e) {
             REQUIRE(e.code() == mongocxx::error_code::k_invalid_uri);
+
+            std::string invalid_schema =
+                "Invalid URI Schema, expecting 'mongodb://' or 'mongodb+srv://': ";
+
+            REQUIRE(e.what() == invalid_schema + e.code().message());
```
rayangler, line 78: Not sure if this test is necessary but I added it in the moment to make sure we were successfully getting the correct error message.

bazile-clyde (approved):
LGTM! 
To merge:

1. Hit the _squash and merge_ button below.
2. Remove any commit messages from the body box (the one on bottom)
3. Ensure the commit message (in the top box) follows our standards ([here](http://mongocxx.org/contributing/)). Leave the postfix "(#720)" too. That'll link to the PR.
4. Click _Confirm squash and merge_

PR #724, opened 2020-09-16
CXX-2096 Add BSON UDL for from_json()

rayangler:


src/bsoncxx/json.cpp
```diff
@@ -89,5 +89,11 @@ document::value BSONCXX_CALL from_json(stdx::string_view json) {
     return document::value{buf, length, bson_free_deleter};
 }
 
+document::value BSONCXX_CALL operator"" _bson(const char* str, size_t len) {
+    // len isn't used, but it is required for the UDL to use a char*
```
samantharitter, line 93: I think you can get around this by changing the signature so it does not name the len variable:
`document::value BSONCXX_CALL operator"" _bson(const char* str, size_t) {...}`

But also, it looks like there should be a signature possible that just takes (const char*) [according to this page](https://en.cppreference.com/w/cpp/language/user_literal), did you have trouble implementing that?

src/bsoncxx/json.hpp
```diff
@@ -68,6 +68,22 @@ BSONCXX_API std::string BSONCXX_CALL to_json(document::view view,
 ///
 BSONCXX_API document::value BSONCXX_CALL from_json(stdx::string_view json);
 
+///
+/// Constructs a new document::value from the provided JSON text
```
samantharitter, line 72: Nit, please add a period at the end of the sentence

src/bsoncxx/test/json.cpp
```diff
@@ -112,4 +112,20 @@ TEST_CASE("CXX-1246: Canonical Extended JSON") {
         output ==
         R"({ "number" : { "$numberInt" : "42" }, "bin" : { "$binary" : { "base64": "ZGVhZGJlZWY=", "subType" : "04" } } })");
 }
+
+TEST_CASE("UDL _bson works like from_json()") {
```
samantharitter, line 116: For completeness' sake, please add a test for the empty document "{}"

samantharitter (changes requested):
These code changes look pretty good, just a few comments!

But: I have some questions about whether this is a useful thing to add to the driver.  Did a request for this feature come from the community?  Maybe @bazile-clyde knows?

Since this option can't take a JSON representation mode, it's a little less useful than the to_json() method without really being much quicker or shorter to type.  Also I don't think the UDL syntax is super widely known, so I'm not sure people would use this unless it was a specific community-requested feature.  I'm not sure it's actually worth us committing and then supporting this.

src/bsoncxx/json.cpp
rayangler: I'll try it with `document::value BSONCXX_CALL operator"" _bson(const char* str, size_t) {...}`! Thanks for letting me know you could do that.

I originally tried just using `(const char*)` but it's not an applicable literal operator. On that page, there is a list of accepted parameters and it lists `(const char*)` as being only used:
> as fallbacks for integer and floating-point user-defined literals

bazile-clyde:
> These code changes look pretty good, just a few comments!
> 
> But: I have some questions about whether this is a useful thing to add to the driver. Did a request for this feature come from the community? Maybe @bazile-clyde knows?
> 
> Since this option can't take a JSON representation mode, it's a little less useful than the to_json() method without really being much quicker or shorter to type. Also I don't think the UDL syntax is super widely known, so I'm not sure people would use this unless it was a specific community-requested feature. I'm not sure it's actually worth us committing and then supporting this.

This came out of the usability requests from the engineers that worked on genny as a nice to have. There are also a few tickets that point to nlohmann/JSON as a model for easily creating documents. One of the simpler ways to generate JSON in that library is also a UDL. As far as how widely known UDLs are, I'm not sure. Strings, dates, numeric types, etc. in the standard library all have them. So they aren't hidden. Is it worth committing and supporting? I think so. It's just a wrapper function. So about 3 lines of code for quick usability improvement. With all of the complaints we've received about how difficult it is to create BSON documents, I don't see why not.

samantharitter (approved):
LGTM!  @bazile-clyde thank you for clarifying the history of this ticket, sounds like people will use it!

src/bsoncxx/test/json.cpp
```diff
@@ -112,4 +112,25 @@ TEST_CASE("CXX-1246: Canonical Extended JSON") {
...
+    using namespace bsoncxx;
+
+    SECTION("_bson and from_json() return the same value") {
+        auto expected_value = from_json(k_valid_json);
+        auto actual_value = R"({ "a" : 1, "b" : 2.0 })"_bson;
+        REQUIRE(actual_value == expected_value);
```
kevinAlbs, line 122: Can you add a test case from the ticket as well: `REQUIRE ("[1,2,3]" == from_json("[1,2,3]"))`

src/bsoncxx/json.cpp
```diff
@@ -89,5 +89,9 @@ document::value BSONCXX_CALL from_json(stdx::string_view json) {
     return document::value{buf, length, bson_free_deleter};
 }
 
+document::value BSONCXX_CALL operator"" _bson(const char* str, size_t) {
+    return from_json(str);
```
kevinAlbs, line 93: `str` is a NULL terminated C string.
`from_json` takes a `string_view`.
Because `string_view` is constructed without a length, it'll end up computing the length of the string (constructor 4 here: https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view)
I suggest explicitly constructing a `string_view`, and using the `size_t` argument passed to call constructor 3 of `string_view`. That'll avoid an extra length calculation. It's a small optimization, but also a simple one to make.

kevinAlbs (changes requested):
Looks good! Just a request for another test case and removal of extra string length calculation.

kevinAlbs (approved):
LGTM! The failures on travis are unrelated.

PR #725, opened 2020-09-16
CXX-2067 Expose bson_value::value type constructors

bazile-clyde:
All of the tests are done and works. I'm still trying to figure out how to combine the constructors in `value.cpp` and `value.hpp`. As far as generic programming goes, a variadic template could work like:
```
value(const type id); // base case

template<typename T, typename ... Args>
value(const type id, T t, Args ...args);

```
But it'd need to be defined in the `value.hpp`. IIUC that would create a new function for every different combination of arguments used. Which would cause a lot of overhead (memory and compilation) since it'd be a fairly large function. 

There's also the open issue of how to handle errors. As of now I just throw an error but I'm also working on that too.

### DIFF
--- a/src/bsoncxx/test/bson_value.cpp
+++ b/src/bsoncxx/test/bson_value.cpp
@@ -21,7 +21,6 @@
 #include <bsoncxx/types/bson_value/make_value.hpp>
 #include <bsoncxx/types/bson_value/value.hpp>
 #include <bsoncxx/types/bson_value/view.hpp>
-#include <iostream>
 #include <vector>
 
 namespace {
@@ -58,16 +57,7 @@ void coverting_construction_test(T actual, U expected) {
 }  // namespace
 
 TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
-    auto doc_value = make_document(kvp("hello", "world"));
-    auto doc2_value = make_document(kvp("a", 1));
-
-    auto doc = doc_value.view();
-    auto doc2 = doc2_value.view();
-
-    auto elem = doc["hello"];
-    auto elem2 = doc2["a"];
-
-    SECTION("can be constructed from a bson_value::view") {
+    SECTION("can be constructed from") {
         SECTION("bool") {
             auto test_doc = bson_value::make_value(b_bool{true});
             value_construction_test(test_doc.view());
@@ -77,13 +67,15 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
         }
 
         SECTION("utf8") {
-            auto test_doc = bson_value::make_value("super duper");
+            std::string value = "super duper";
+
+            auto test_doc = bson_value::make_value(value);
             value_construction_test(test_doc.view());
 
-            coverting_construction_test("super duper", test_doc);
-            coverting_construction_test(b_utf8{"super duper"}, test_doc);
-            coverting_construction_test(stdx::string_view{"super duper"}, test_doc);
-            coverting_construction_test(std::string{"super duper"}, test_doc);
+            coverting_construction_test(value, test_doc);
+            coverting_construction_test(value.c_str(), test_doc);
+            coverting_construction_test(b_utf8{value}, test_doc);
+            coverting_construction_test(stdx::string_view{value}, test_doc);
 
             auto test_empty = bson_value::make_value("");
             value_construction_test(test_empty.view());
@@ -313,6 +305,15 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
     }
 
     SECTION("can be constructed by a document::element") {
+        auto doc_value = make_document(kvp("hello", "world"));
+        auto doc2_value = make_document(kvp("a", 1));
+
+        auto doc = doc_value.view();
+        auto doc2 = doc2_value.view();
+
+        auto elem = doc["hello"];
+        auto elem2 = doc2["a"];
+
         bson_value::value value = elem.get_owning_value();
 
         SECTION("can create new views") {
--- a/src/bsoncxx/test/bson_value.cpp
+++ b/src/bsoncxx/test/bson_value.cpp
@@ -77,7 +77,7 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
             coverting_construction_test(stdx::string_view{value}, test_doc);
 
             const char raw_data[]{'s', 'u', 'p', 'e', 'r', ' ', 'd', 'u', 'p', 'e', 'r'};
-            coverting_construction_test(raw_data, test_doc);
+            coverting_construction_test(stdx::string_view{raw_data, 11}, test_doc);
 
             auto test_empty = bson_value::make_value("");
             value_construction_test(test_empty.view());
--- a/src/bsoncxx/test/bson_value.cpp
+++ b/src/bsoncxx/test/bson_value.cpp
@@ -257,11 +257,15 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
             auto doc = make_document(kvp("a", 1));
             auto test_doc = bson_value::make_value(doc.view());
             value_construction_test(test_doc.view());
+
+            coverting_construction_test(b_document{doc.view()}, test_doc);
             coverting_construction_test(value(doc.view()), test_doc);
 
             // Empty document
             test_doc = bson_value::make_value(document::view{});
             value_construction_test(test_doc.view());
+
+            coverting_construction_test(value(b_document{}), test_doc);
             coverting_construction_test(value(document::view{}), test_doc);
         }
 
--- a/src/bsoncxx/test/bson_value.cpp
+++ b/src/bsoncxx/test/bson_value.cpp
@@ -166,7 +166,7 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
         }
 
         SECTION("regex") {
-            /* options are sorted and any duplicate or invalid options are ignored */
+            /* options are sorted and any duplicate or invalid options are removed */
             auto test_doc = bson_value::make_value(b_regex{"amy", "imsx"});
             value_construction_test(test_doc.view());
 
@@ -176,7 +176,7 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
             auto empty_regex = bson_value::make_value(b_regex{"", ""});
             value_construction_test(empty_regex.view());
 
-            coverting_construction_test(value(type::k_regex), empty_regex);
+            coverting_construction_test(b_regex{"", ""}, empty_regex);
             coverting_construction_test(value(type::k_regex, ""), empty_regex);
             coverting_construction_test(value(type::k_regex, "", ""), empty_regex);
         }
@@ -196,12 +196,12 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
         SECTION("code") {
             auto test_doc = bson_value::make_value(b_code{"look at me I'm some JS code"});
             value_construction_test(test_doc.view());
+            coverting_construction_test(b_code{"look at me I'm some JS code"}, test_doc);
             coverting_construction_test(value(type::k_code, "look at me I'm some JS code"),
                                         test_doc);
 
             auto empty_code = bson_value::make_value(b_code{""});
             value_construction_test(empty_code.view());
-            coverting_construction_test(value(type::k_code), empty_code);
             coverting_construction_test(value(type::k_code, ""), empty_code);
         }
 
@@ -210,12 +210,17 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
             auto test_doc =
                 bson_value::make_value(b_codewscope{"it's me, Code with Scope", doc.view()});
             value_construction_test(test_doc.view());
+
+            coverting_construction_test(value(b_codewscope{"it's me, Code with Scope", doc.view()}),
+                                        test_doc);
             coverting_construction_test(
                 value(type::k_codewscope, "it's me, Code with Scope", doc.view()), test_doc);
 
             auto empty_doc = make_document(kvp("a", ""));
             auto empty_code = bson_value::make_value(b_codewscope{"", empty_doc.view()});
             value_construction_test(empty_code.view());
+
+            coverting_construction_test(value(b_codewscope{"", empty_doc.view()}), empty_code);
             coverting_construction_test(value(type::k_codewscope, "", empty_doc.view()),
                                         empty_code);
         }
@@ -223,13 +228,17 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
         SECTION("minkey") {
             auto test_doc = bson_value::make_value(b_minkey{});
             value_construction_test(test_doc.view());
+
             coverting_construction_test(value(type::k_minkey), test_doc);
+            coverting_construction_test(value(b_minkey{}), test_doc);
         }
 
         SECTION("maxkey") {
             auto test_doc = bson_value::make_value(b_maxkey{});
             value_construction_test(test_doc.view());
+
             coverting_construction_test(value(type::k_maxkey), test_doc);
+            coverting_construction_test(value(b_maxkey{}), test_doc);
         }
 
         SECTION("binary") {
--- a/src/bsoncxx/test/bson_value.cpp
+++ b/src/bsoncxx/test/bson_value.cpp
@@ -123,12 +123,12 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
         }
 
         SECTION("oid") {
-            oid _id{"0123456789abcdefABCDEFFF"};
-            auto test_doc = bson_value::make_value(b_oid{_id});
+            oid id{"0123456789abcdefABCDEFFF"};
+            auto test_doc = bson_value::make_value(b_oid{id});
             value_construction_test(test_doc.view());
 
-            coverting_construction_test(_id, test_doc);
-            coverting_construction_test(b_oid{_id}, test_doc);
+            coverting_construction_test(id, test_doc);
+            coverting_construction_test(b_oid{id}, test_doc);
         }
 
         SECTION("decimal128") {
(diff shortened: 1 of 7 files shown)


bazile-clyde:
I wanted to get some feedback before moving forward. So far, I've only finished constructors for a boolean (both `b_bool` and `bool`). The other types would look almost identical to this. 

Some types have a natural C++ built-in type equivalent, and others don't. So I'm using the `append` overloads from `core.hpp` to figure out which types I can implicitly construct. I'll need to do this for the implicit BSON constructor.

src/bsoncxx/test/bson_value.cpp
```diff
@@ -54,6 +54,28 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
...
+                REQUIRE(copy_initialization == test_doc);
+
+                bson_value::value direct_initialization(types::b_bool{true});
+                REQUIRE(direct_initialization == test_doc);
+
+                bson_value::value _static_cast = (bson_value::value)types::b_bool{true};
+                REQUIRE(_static_cast == test_doc);
+            }
+
+            SECTION("with implicit BSON type") {
```
bazile-clyde, line 69: These implicit tests won't exist for types like `symbol`, `regex`, etc.

src/bsoncxx/test/bson_value.cpp
```diff
@@ -54,6 +54,28 @@ TEST_CASE("types::bson_value::value", "[bsoncxx::types::bson_value::value]") {
...
             value_construction_test(test_doc.view());
+
+            SECTION("with explicit BSON type") {
+                bson_value::value copy_initialization = types::b_bool{true};
+                REQUIRE(copy_initialization == test_doc);
+
+                bson_value::value direct_initialization(types::b_bool{true});
+                REQUIRE(direct_initialization == test_doc);
+
+                bson_value::value _static_cast = (bson_value::value)types::b_bool{true};
```
samantharitter, line 65: Technically, this is not a static cast, this is a c-style cast.  I would suggest either changing the name or changing the cast, as it's misleading the way it's written:

`bson_value::value cast = (bson_value::value)types::b_bool{true};`
`bson_value::vale _static_cast = static_cast<bson_value::value>(types::b_bool{true});`

src/bsoncxx/test/bson_value.cpp
samantharitter: Acknowledged.  I would suggest writing convenience methods to do the `implicit_construction_test()` and `explicit_construction_test()`, similar to the `value_construction_test()` method in this file, so we can run the same tests easily for every type.  Writing that method will involve some (fun!) template meta-programming, I'm happy to give further suggestions on that or pair program it with you some time.

src/bsoncxx/types/bson_value/value.hpp
```diff
@@ -51,6 +51,9 @@ class BSONCXX_API value {
     ///
     explicit value(const view&);
 
+    value(b_bool b);  // TODO: use BSONCXX_ENUM for all b_ types
```
samantharitter, line 54: As the ticket describes, the b_types are not things we really want to encourage users to use.  There's a few reasons it's better to hide these types, the biggest reason being that they mostly do not own their own memory.

For now, let's only provide the constructors from the base types themselves--so, here, provide only the constructor that takes a `bool` and not the one that takes `types::b_bool`.

Unfortunately, this probably means you can't use the slick `BSONCXX_ENUM` :( though you could probably do something fancy to strip the `b_` prefixes off, I'm not sure it's worth it.

bazile-clyde: How would we create values that have the same underlying type? For example, is `value(std::string str)` a utf8, code or symbol.

src/bsoncxx/test/bson_value.cpp
bazile-clyde: Done

src/bsoncxx/test/bson_value.cpp
bazile-clyde: Done

src/bsoncxx/types/bson_value/value.hpp
kevinAlbs: > As the ticket describes, the b_types are not things we really want to encourage users to use.

I generally agree that using the b_types structs is potentially dangerous because it is not clear that it does not own its data. I'm not sure what the long term goal would be to remove them (in favor of view/value types that clarify ownership?).

> How would we create values that have the same underlying type? For example, is value(std::string str) a utf8, code or symbol?

Good point. If we do want to avoid using the `types` structs in the API, we'd need to avoid ambiguity when there is the same underlying type.
If we're sure we want to avoid the b_types structs, I'd be in favor of explicit named static constructors, like `value::from_string(const std::string& str)` and `value::from_symbol (const std::string& str)`. That seems straightforward at the cost of some extra verbosity.

bazile-clyde: Sort of unrelated, but I think the Genny users also complained about our aversion to allocating memory:

> The library seems to go out of its way to avoid allocating memory. While that is admirable, it does not make it easy when you are building essentially tree-like structures (such as BSON), and it hardly matters when it must eventually be sent over the network anyway.

This could also be circumvented by just having a constructor like:
```
template <typename... Args>
value(type k_type = k_undefined, Args&&... args);
```

kevinAlbs: > Until we remove the b_types we should probably support conversions between a generic value and a specific value.

That is a fair point. Unless we have a complete solution of how to replace `b_types` structs, we'd have an asymmetric API, where users must use the `b_type` structs when accessing element BSON values, but cannot use them to construct BSON values.

E.g. currently users get `b_type` structs from iterating, and can use them to construct document values:
```
bsoncxx::types::b_regex regex = doc.view()["field"].get_regex();
bsoncxx::document::value doc = make_document(kvp("regex", regex));
```

But we'd be requiring users to split out the fields the `b_type` in order to construct a `bson_value::value`
```
bsoncxx::types::bson_value val {regex.regex, regex.options};
```

So, in chatting with Clyde, the proposal is that the constructor should support the same set of values that the `bsoncxx::builder::core:append()` accepts. @samantharitter do you agree?

samantharitter: Hm, the type ambiguity does make this more complicated, I hadn't thought of that.  I like both @kevinAlbs 's factory method solution and @bazile-clyde 's type argument method solution as workarounds for that.

I agree that omitting constructors that take the b_types would lead to an asymmetrical API.  But, I'm also not convinced that that alone is a reason for us to add them to the driver.  There are plenty of quirks in the current driver API, and this particular feature is one that I requested, it didn't come from the community.

I think it would help if we figured out what our ideal future BSON library looks like and work backwards from that design.  For example, if we would like to ultimately remove the `b_type`s from the public API, it doesn't make sense to introduce more methods that use them, as this will lead to more user code that ultimately needs to be changed when we deprecate and then remove those types.

kevinAlbs: Per discussion, let's keep the `b_type` constructors to give a consistent API. If we remove the `b_type` structs in a future design, it will be part of a larger change.

src/bsoncxx/types/bson_value/value.hpp
```diff
@@ -38,6 +43,33 @@ namespace bson_value {
...
+    value(bsoncxx::document::view v);
+    value(bsoncxx::array::view v);
+    value(std::vector<unsigned char> v, binary_sub_type sub_type = {});
+
+    value(const type id);
+    value(const type id, uint64_t a, uint64_t b);
+    value(const type id, stdx::string_view a, stdx::string_view b = {});
+    value(const type id, stdx::string_view a, oid b);
+    value(const type id, stdx::string_view a, bsoncxx::document::view_or_value b);
+    value(const type id, const binary_sub_type sub_id, uint32_t size, const uint8_t* data);
```
kevinAlbs, line 71: Does this cover all types? If so, I'm ok with having these listed out as you have now, instead of trying to collapse them into one template.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,170 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+
+value::value(const type id) : _impl{stdx::make_unique<impl>()} {
+    if (id == type::k_minkey) {
+        _impl->_value.value_type = BSON_TYPE_MINKEY;
+    } else if (id == type::k_maxkey) {
+        _impl->_value.value_type = BSON_TYPE_MAXKEY;
+    } else if (id == type::k_undefined) {
+        _impl->_value.value_type = BSON_TYPE_UNDEFINED;
+    } else {
+        throw std::logic_error{"Must be min/max key or undefined"};
```
kevinAlbs, line 62: > There's also the open issue of how to handle errors. As of now I just throw an error but I'm also working on that too

Is the open question whether or not to throw? There is precedent for throwing in some constructors. At least client comes to mind: https://github.com/mongodb/mongo-cxx-driver/blob/78831317d0dde2b36c32236303991d3737c23972/src/mongocxx/client.cpp#L84

kevinAlbs (commented):
I have not fully reviewed yet, but a couple comments regarding the open questions.

src/bsoncxx/types/bson_value/value.hpp
bazile-clyde: Yep, this is everything

src/bsoncxx/types/bson_value/value.cpp
bazile-clyde: Yeah, I guess I'll have to throw regardless, for the memcpy and allocations I think. I'll clean this up.

src/bsoncxx/types/bson_value/value.hpp
bazile-clyde: @samantharitter thinking about it a bit more, the only way I can see us removing the `b_types` is if we create a class for each of the BSON types that don't easily map to built-in C++ types. This would be similar to what's done for `decimal128` and `oid`. Of course, these new types would be owning. We'd then only need a class for each BSON type and this class... I _think_. 

I think the next steps though would be a design doc of some kind in the way future but wanted to run this idea by both of you.

bazile-clyde:
@kevinAlbs and @samantharitter done. I:

- added documentation
- removed the type ids for constructors, where possible
- added exceptions for invalid type ids

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,162 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+}
+
+value::value(b_bool v) : value(v.value) {}
+value::value(bool v) : _impl{stdx::make_unique<impl>()} {
+    _impl->_value.value_type = BSON_TYPE_BOOL;
+    _impl->_value.value.v_bool = v;
+}
+
+value::value(b_code v) : value(v.type_id, v) {}
+value::value(b_regex v) : value(v.type_id, std::string{v.regex}, std::string{v.options}) {}
```
kevinAlbs, line 81: `v.regex` and `v.options` are `stdx::string_view`. Can they be passed directly?

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,162 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+    _impl->_value.value.v_binary.data_len = (uint32_t)size;
+    _impl->_value.value.v_binary.data = (uint8_t*)bson_malloc(size);
+    std::memcpy(_impl->_value.value.v_binary.data, data, size);
+}
+
+value::value(b_document v) : value(v.view()) {}
+value::value(bsoncxx::document::view v) : _impl{stdx::make_unique<impl>()} {
+    _impl->_value.value_type = BSON_TYPE_DOCUMENT;
+    _impl->_value.value.v_doc.data_len = (uint32_t)v.length();
+    _impl->_value.value.v_doc.data = (uint8_t*)bson_malloc0(v.length());
```
kevinAlbs, line 174: Use `bson_new_from_data`.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,162 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+    std::memcpy(_impl->_value.value.v_doc.data, v.data(), v.length());
+}
+
+value::value(b_array v) : value(v.value) {}
+value::value(bsoncxx::array::view v) : _impl{stdx::make_unique<impl>()} {
+    _impl->_value.value_type = BSON_TYPE_ARRAY;
+    _impl->_value.value.v_doc.data_len = (uint32_t)v.length();
+    _impl->_value.value.v_doc.data = (uint8_t*)bson_malloc0(v.length());
```
kevinAlbs, line 182: Use `bson_new_from_data`.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,162 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+}
+
+value::value(b_codewscope v) : value(v.code, v.scope) {}
+value::value(stdx::string_view code, bsoncxx::document::view_or_value scope)
+    : _impl{stdx::make_unique<impl>()} {
+    _impl->_value.value_type = BSON_TYPE_CODEWSCOPE;
+    _impl->_value.value.v_codewscope.code = make_copy_for_libbson(code);
+    _impl->_value.value.v_codewscope.code_len = (uint32_t)code.length();
+    _impl->_value.value.v_codewscope.scope_len = (uint32_t)scope.view().length();
+    _impl->_value.value.v_codewscope.scope_data = (uint8_t*)bson_malloc0(scope.view().length());
```
kevinAlbs, line 153: I suggest using `bson_new_from_data` instead of malloc + memcpy.

src/bsoncxx/types/bson_value/value.hpp
```diff
@@ -38,6 +43,177 @@ namespace bson_value {
...
+    ///
+    /// Constructs a BSON array value.
+    ///
+    value(bsoncxx::array::view v);
+
+    ///
+    /// Constructs a BSON binary data value.
+    ///
+    /// @param v
+    ///     a steam of bytes
```
kevinAlbs, line 128:
```suggestion
    ///     a stream of bytes
```

kevinAlbs (changes requested):
Looking great! Just a few comments about splitting up the constructors and additional tests. Apologies for the delayed review.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -77,12 +77,28 @@ value::value(bool v) : _impl{stdx::make_unique<impl>()} {
...
+        case type::k_undefined:
+            _impl->_value.value_type = BSON_TYPE_UNDEFINED;
+            break;
+        default:
+            throw bsoncxx::exception(error_code::k_invalid_bson_type_id);
+    }
+}
+
+value::value(b_code v) : value(v.type_id, v) {}
+value::value(b_symbol v) : value(v.type_id, v) {}
```
kevinAlbs, line 100: I suggest splitting off code/symbol from regex for the same reason as splitting of the minkey/maxkey/undefined constructors (so `bson_value::value{type::k_symbol, "x", "y"}` does not compile and run and ignore the second argument).

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -150,9 +157,9 @@ value::value(stdx::string_view code, bsoncxx::document::view_or_value scope)
     _impl->_value.value.v_codewscope.code = make_copy_for_libbson(code);
     _impl->_value.value.v_codewscope.code_len = (uint32_t)code.length();
     _impl->_value.value.v_codewscope.scope_len = (uint32_t)scope.view().length();
-    _impl->_value.value.v_codewscope.scope_data = (uint8_t*)bson_malloc0(scope.view().length());
-    std::memcpy(
-        _impl->_value.value.v_codewscope.scope_data, scope.view().data(), scope.view().length());
+
+    bson_t* cpy = bson_new_from_data(scope.view().data(), scope.view().length());
+    _impl->_value.value.v_codewscope.scope_data = bson_destroy_with_steal(cpy, true, &cpy->len);
```
kevinAlbs, line 162: Sorry, when I initially suggested `bson_new_from_data` I forgot that `scope_data` is storing a pointer to the data directly, not a `bson_t`. I had not realized you'd have to destroy the outer `bson_t`. This seems about the same in terms of readability. I'm fine with either. If you do want to change it back, I only suggest using `bson_malloc` instead of `bson_malloc0`.

kevinAlbs (commented):
Suggest splitting off one more constructor, otherwise looks good!

src/bsoncxx/types/bson_value/value.cpp
bazile-clyde: Done

src/bsoncxx/types/bson_value/value.cpp
bazile-clyde: Done.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -25,6 +27,162 @@ BSONCXX_INLINE_NAMESPACE_BEGIN
...
+            _impl->_value.value_type = BSON_TYPE_CODE;
+            _impl->_value.value.v_code.code = make_copy_for_libbson(a);
+            _impl->_value.value.v_code.code_len = (uint32_t)a.length();
+            break;
+        case type::k_symbol:
+            _impl->_value.value_type = BSON_TYPE_SYMBOL;
+            _impl->_value.value.v_symbol.symbol = make_copy_for_libbson(a);
+            _impl->_value.value.v_symbol.len = (uint32_t)a.length();
+            break;
+        case type::k_minkey:
```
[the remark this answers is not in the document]
kevinAlbs: Ah, I missed this response on my last pass. If `b` was required, this could be:
```
value(const type id, stdx::string_view a, stdx::string_view b); // regex with options
value(const type id, stdx::string_view a);                      // regex without options, symbol, code
```

src/bsoncxx/test/bson_value.cpp
```diff
@@ -38,139 +41,289 @@ void value_construction_test(bson_value::view test_view) {
...
-            auto test_doc = bson_value::make_value("super duper");
+            std::string value = "super duper";
+
+            auto test_doc = bson_value::make_value(value);
             value_construction_test(test_doc.view());
 
+            coverting_construction_test(value, test_doc);
+            coverting_construction_test(value.c_str(), test_doc);
+            coverting_construction_test(b_utf8{value}, test_doc);
+            coverting_construction_test(stdx::string_view{value}, test_doc);
```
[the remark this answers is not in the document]
kevinAlbs: I think this does need to be wrapped in a string view to pass the explicit length. Otherwise, the conversion will assume the string is null terminated to determine the length. I believe this is the cause of test failures on Windows: https://evergreen.mongodb.com/task_log_raw/cxx_driver_integration__os~windows_2k8_mongodb_version~4.0_compile_and_test_with_shared_libs_patch_11a863102c8549f82698ebe710394a0d3dd10c8a_5f90884332f417101242e124_20_10_21_19_13_08/0?type=T#L7212

kevinAlbs (approved):
LGTM with a test fix for Windows!

PR #726, opened 2020-09-24
CXX-2097 Create BSON serializer functions for user-defined objects

rayangler:
This PR will be opened as a draft while we make decisions on the implementation of serialization for user-defined objects. Please refer to the associated Google doc for further information.

Edit: Also ignore the random comments or irrelevant functions. I was testing out some stuff

### DIFF
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -24,6 +24,10 @@
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
 
+namespace array {
+class view;
+}
+
 namespace builder {
 namespace basic {
 class document;
@@ -84,7 +88,8 @@ class BSONCXX_API value {
     value& operator=(value&&) noexcept = default;
 
     // Serializer functions
-    template <typename T>
+    template <typename T,
+              typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
     explicit value(const T& user_object)
         : _data(new std::uint8_t[static_cast<std::size_t>(5)],
                 [](std::uint8_t* ptr) { delete[] ptr; }) {
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -28,12 +28,6 @@ namespace array {
 class view;
 }
 
-namespace builder {
-namespace basic {
-class document;
-}
-}
-
 namespace document {
 
 ///
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -201,7 +201,7 @@ class BSONCXX_API value {
     /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
     ///
     /// @param t
-    ///   A partially constructed object. The contents of the document object will be deserialized
+    ///   The object to construct. The contents of the document object will be deserialized
     ///   into t.
     ///
     template <typename T>
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -23,6 +23,13 @@
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
+
+namespace builder {
+namespace basic {
+class document;
+}
+}
+
 namespace document {
 
 ///
@@ -75,6 +82,11 @@ class BSONCXX_API value {
 
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
+
+//    template<typename T>
+//    explicit value(T& user_object) {
+//        to_bson(user_object, *this);
+//    }
 
     ///
     /// @returns A const_iterator to the first element of the document.
@@ -161,6 +173,14 @@ class BSONCXX_API value {
     /// @return A view over the value.
     ///
     BSONCXX_INLINE operator document::view() const noexcept;
+
+    // Function to help with serialization
+    template<typename T>
+    T get() {
+        T temp_object{};
+        from_bson(temp_object, this->view());
+        return temp_object;
+    }
 
     ///
     /// Transfer ownership of the underlying buffer to the caller.
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -81,12 +81,17 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
-    // Serializer functions
+    ///
+    /// Constructor used for serialization of user objects. This uses argument-dependent lookup
+    /// to find the function declaration `void to_bson(T& t, bsoncxx::document::value doc)`.
+    ///
+    /// @param user_object
+    ///   A custom user object to serialize. This object must be default-constructible for
+    ///   deserialization.
+    ///
     template <typename T,
               typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
-    explicit value(const T& user_object)
-        : _data(new std::uint8_t[static_cast<std::size_t>(5)],
-                [](std::uint8_t* ptr) { delete[] ptr; }) {
+    explicit value(const T& user_object) : value({}) {
         to_bson(user_object, *this);
     }
     template <typename T>
@@ -181,7 +186,13 @@ class BSONCXX_API value {
     ///
     BSONCXX_INLINE operator document::view() const noexcept;
 
-    // Function to help with serialization
+    ///
+    /// Constructs an object of type T from this document object. This method uses
+    /// argument-dependent lookup to find the function declaration
+    /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
+    ///
+    /// @note The class being deserialized into must be default-constructible.
+    ///
     template <typename T>
     T get() {
         T temp_object{};
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -186,14 +186,27 @@ class BSONCXX_API value {
     /// argument-dependent lookup to find the function declaration
     /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
     ///
-    /// @param t
-    ///   An optional partially constructed object. If no argument is passed, the function will
-    ///   create an object with its default constructor.
+    /// @note Type T must be default-constructible. Otherwise, use `void get(T& t)`.
     ///
     template <typename T>
-    T get(T t = {}) {
+    T get() {
+        T temp{};
+        from_bson(temp, this->view());
+        return temp;
+    }
+
+    ///
+    /// Constructs an object of type T from this document object. This method uses
+    /// argument-dependent lookup to find the function declaration
+    /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
+    ///
+    /// @param t
+    ///   A partially constructed object. The contents of the document object will be deserialized
+    ///   into t.
+    ///
+    template <typename T>
+    void get(T& t) {
         from_bson(t, this->view());
-        return t;
     }
 
     ///
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -24,11 +24,6 @@
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
-
-namespace array {
-class view;
-}
-
 namespace document {
 
 ///
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -17,6 +17,7 @@
 #include <cstdlib>
 #include <memory>
 
+#include <bsoncxx/array/view.hpp>
 #include <bsoncxx/document/view.hpp>
 
 #include <bsoncxx/config/prelude.hpp>
@@ -85,18 +86,17 @@ class BSONCXX_API value {
     /// Constructor used for serialization of user objects. This uses argument-dependent lookup
     /// to find the function declaration `void to_bson(T& t, bsoncxx::document::value doc)`.
     ///
-    /// @param user_object
-    ///   A custom user object to serialize. This object must be default-constructible for
-    ///   deserialization.
+    /// @param t
+    ///   A user-defined object to serialize into a BSON object.
     ///
     template <typename T,
               typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
-    explicit value(const T& user_object) : value({}) {
-        to_bson(user_object, *this);
+    explicit value(const T& t) : value({}) {
+        to_bson(t, *this);
     }
     template <typename T>
-    value& operator=(const T& user_object) {
-        *this = value{user_object};
+    value& operator=(const T& t) {
+        *this = value{t};
         return *this;
     }
 
@@ -191,13 +191,14 @@ class BSONCXX_API value {
     /// argument-dependent lookup to find the function declaration
     /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
     ///
-    /// @note The class being deserialized into must be default-constructible.
+    /// @param t
+    ///   An optional partially constructed object. If no argument is passed, the function will
+    ///   create an object with its default constructor.
     ///
     template <typename T>
-    T get() {
-        T temp_object{};
-        from_bson(temp_object, this->view());
-        return temp_object;
+    T get(T t = {}) {
+        from_bson(t, this->view());
+        return t;
     }
 
     ///
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -83,10 +83,10 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
-//    template<typename T>
-//    explicit value(T& user_object) {
-//        to_bson(user_object, *this);
-//    }
+    //    template<typename T>
+    //    explicit value(T& user_object) {
+    //        to_bson(user_object, *this);
+    //    }
 
     ///
     /// @returns A const_iterator to the first element of the document.
@@ -175,7 +175,7 @@ class BSONCXX_API value {
     BSONCXX_INLINE operator document::view() const noexcept;
 
     // Function to help with serialization
-    template<typename T>
+    template <typename T>
     T get() {
         T temp_object{};
         from_bson(temp_object, this->view());
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -83,10 +83,18 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
-    //    template<typename T>
-    //    explicit value(T& user_object) {
-    //        to_bson(user_object, *this);
-    //    }
+    // Serializer functions
+    template <typename T>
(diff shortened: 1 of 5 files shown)


src/bsoncxx/document/value.hpp
```diff
@@ -88,6 +107,14 @@ class BSONCXX_API value {
     ///
     BSONCXX_INLINE operator document::view() const noexcept;
 
+    // Function to help with serialization
+    template <typename T>
+    T get() {
```
samantharitter, line 112: When/where is this used?

src/bsoncxx/document/value.hpp
```diff
@@ -76,6 +81,20 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
+    // Serializer functions
+    template <typename T,
+              typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
+    explicit value(const T& user_object)
+        : _data(new std::uint8_t[static_cast<std::size_t>(5)],
+                [](std::uint8_t* ptr) { delete[] ptr; }) {
+        to_bson(user_object, *this);
```
samantharitter, line 90: So, the idea is that we would provide a to_bson(U, document::value*) method for every type U that we want to be able to serialize into a document?

src/bsoncxx/types/bson_value/view.cpp
```diff
@@ -184,6 +185,48 @@ bool operator!=(const view& lhs, const view& rhs) {
     return !(lhs == rhs);
 }
 
+// Serializer functions
+void view::_to_field(std::string& object_field) const {
```
samantharitter, line 189: It is considered a bit of an anti-pattern to pass a non-const reference as a method argument.  I would suggest rewriting these to be pointers, so that it's clear these fields will be modified:

```
void view::_to_field(std::string* object_field);
```

samantharitter (commented):
This is a cool idea!  I left some comments about style and about where to put the access point for to_field.  I like the overall design, though!

src/bsoncxx/document/value.hpp
rayangler: This would be used when the user tries to deserialize a BSON value into their custom user object like so:
```
test::Person test_person = bson_value.get<test::Person>();
```

I should change the comment above the function to have it say "deserialization" instead of "serialization", and make it clear that the above example is how the function would be used.

src/bsoncxx/document/value.hpp
rayangler: Yup! A `to_bson` serializer function should be written by the user within their object's namespace to take advantage of this

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,322 @@
...
+            kvp("_decimal128", types::b_decimal128{user_object.m_decimal128}),
+            kvp("_regex", types::b_regex{user_object.m_regex_string, user_object.m_regex_options}),
+            kvp("_timestamp",
+                types::b_timestamp{user_object.m_timestamp_increment,
+                                   user_object.m_timestamp_timestamp}));
+    }
+
+    friend void from_bson(Types_Test_Object& user_object,
+                          const bsoncxx::document::view& bson_object) {
+        bson_object["_double"].get_value().to_field(user_object.m_double);
```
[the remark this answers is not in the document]
rayangler: Yup, the user can choose to assign the BSON object's values like your example with `m_double`. I think having the new syntax in `document::view` instead would make it more convenient for the user.

However, we might choose to remove the `to_field` function from this PR as the main scope of the ticket was to focus on giving the user the ability to implicitly serialize their objects with the use of user-created `to_bson` and `from_bson` functions.

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    friend void to_bson(const Car& car, bsoncxx::document::value& bson_object) {
+        auto builder = bsoncxx::builder::stream::document{};
+        bson_object = builder << "manufacturer" << car._manufacturer << "model" << car._model
+                              << bsoncxx::builder::stream::finalize;
+    }
+
+    friend void from_bson(Car& car, const bsoncxx::document::view& bson_object) {
+        bson_object["manufacturer"].get_value().to_field(car._manufacturer);
```
bazile-clyde, line 75: Per our discussion, I'm not sure what we'd gain by allowing the user to do this:
```
bson_object["manufacturer"].get_value().to_field(car._manufacturer);
```
instead of 
```
car._manufacturer = bson_object["manufacturer"].get_string().value.to_string();
```
They'd still have to consider the type of the BSON value and field to be set since `to_field` will attempt to assign to the object's field based on it's argument type. That last part of the process is now obfuscated by the method call. Consider this case:
```
int i = 32;
bson_object["manufacturer"].get_value().to_field(i);
``` 
This will compile and set `i` to `"Tesla"` which results in a garbage value. So there isn't additional error checking. Had the user assigned a value like so:
```
int i = 32;
i = bson_object["manufacturer"].get_string().value.to_string();
```
They would've gotten a compile-time error. The `to_field` method would also have to handle rvalues, e.g., neither of these code snippets compile:
```
bson_object["manufacturer"].get_value().to_field("hello");
...
bson_object["manufacturer"].get_value().to_field(32);
```
I think we can remove the `to_field` function for now.

rayangler: I agree that the way `to_field` is implemented right now leaves very little to gain when the user can simply just assign the values from the BSON object to their own object themselves and it'd be best to remove it. I would like to address your examples though.

For the example in which we use `to_field(i)` where `i` is an `int`, we could change the `to_field` overloads to check for type first before assigning:
```
_to_field(int32_t& object_field) const {
    if (_type == bsoncxx::type::k_int32) {
        object_field = _b_int32.value;
    }
}
```
This would prevent the user's field from getting the incorrect value (i would remain 32 in this case). However, the user would definitely still need to pay attention to the types that they are using and may offer less flexibility, especially when it comes to implicit conversions. (For example, having an `int64_t i` while trying to obtain the value of a `b_int32` element would be impossible using `to_field` unless we make the function check that type as well.)

I am curious about your thoughts about the rvalues examples. Don't we want the code snippets to fail to compile when passing rvalues to the function since the assumed precondition is that we are using `to_field` to assign to an lvalue (member field)? The user shouldn't be doing something like `"hello" = bson_object...` to begin with.

bazile-clyde: Ah yes! Ignore the rvalue comments. It should fail in that case. Regarding the type checking, it's _possible_ but it would require a lot of work on our part to function correctly. It also isn't very idiomatic of C++ to provide functions solely for assignment. At least I haven't seen anything like it in the standard library. Usually, if you want to assign to a variable you have the implicit and explicit constructors, copy constructors, and move constructors, converting constructors, etc (see [this](https://en.cppreference.com/w/cpp/language/converting_constructor) for more examples).

src/bsoncxx/document/value.hpp
```diff
@@ -76,6 +81,20 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
+    // Serializer functions
+    template <typename T,
+              typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
+    explicit value(const T& user_object)
+        : _data(new std::uint8_t[static_cast<std::size_t>(5)],
```
bazile-clyde, line 88: This could be replaced with a delegating constructor that takes an empty document. So:
```
explicit value(const T& user_object) : value({}) {
```
That way we don't duplicate the logic of the constructor inline 76.

src/bsoncxx/document/value.hpp
```diff
@@ -76,6 +81,20 @@ class BSONCXX_API value {
...
+                [](std::uint8_t* ptr) { delete[] ptr; }) {
+        to_bson(user_object, *this);
+    }
+    template <typename T>
+    value& operator=(const T& user_object) {
```
bazile-clyde, line 93: `.hpp` files are usually just for method declarations. Not definitions. We could move these method definitions to `value.cpp` as is done for the other methods.

src/bsoncxx/document/value.hpp
```diff
@@ -88,6 +107,14 @@ class BSONCXX_API value {
     ///
     BSONCXX_INLINE operator document::view() const noexcept;
 
+    // Function to help with serialization
```
bazile-clyde, line 110: We use [doxygen](https://www.doxygen.nl/manual/docblocks.html) for our documentation. Doxygen needs comments to be formatted a specific way in order to parse and display them correctly. Just like the other comment blocks in this file, there needs to be 3 forward slashes before and after any comment blocks. The comments themselves also need to be preceded by 3 forward slashes.

src/bsoncxx/document/value.hpp
bazile-clyde: Since this will be user-facing code, maybe a more descriptive comment is needed here? For example:
```
///
/// Constructs an object of type T from this object. This method uses argument-dependent lookup to find the function declaration `void from_bson(T& t, const bsoncxx::document::view& doc)`.
///
```

src/bsoncxx/document/value.hpp
```diff
@@ -88,6 +107,14 @@ class BSONCXX_API value {
...
+    template <typename T>
+    T get() {
+        T temp_object{};
```
bazile-clyde, line 113: So this is awesome. It works just as intended. If you have time you could also allow partial construction of an object. In other words, you could have this method take in an optional value of type T and pass that to `from_bson`. So this method would become:
```
    template <typename T>
    T get(T o={}) {
        from_bson(o, this->view());
        return o;
    }
```
and allow the user to do this:
```
test::Person x;
test::Person y = expected_doc.get(x);
```
Note that the brackets (`<>`) aren't needed in this example since we can deduce the type T based on its argument. Feel free to hold off on this though. If you do decide to implement partial construction too you'd probably want to add a test like the code snippet.

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,322 @@
...
+                                kvp("age", person.age));
+}
+
+void from_bson(Person& person, const bsoncxx::document::view& bson_object) {
+    bson_object["first_name"].get_value().to_field(person.first_name);
+    bson_object["last_name"].get_value().to_field(person.last_name);
+    bson_object["age"].get_value().to_field(person.age);
+}
+
+class Car {
```
bazile-clyde, line 54: I think the `Person` class alone provides enough evidence that this feature works as intended. I don't think much is gained from testing with additional classes/structs, aside from more code to maintain. Since this feature will just check that some object of type T has the function _declarations_ above reachable from the call sites in `value.hpp` we shouldn't need to worry about how the _definitions_ of those function will look.

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,322 @@
...
+        bson_object["model"].get_value().to_field(car._model);
+    }
+
+   private:
+    std::string _manufacturer;
+    std::string _model;
+};
+
+// Class that uses most BSON Types
+class Types_Test_Object {
```
bazile-clyde, line 85: Echoing the comment above, I think the only class we need here is `Person` and its corresponding tests.

bazile-clyde (changes requested):

src/bsoncxx/document/value.hpp
rayangler: Separating the declaration and the definition of template functions would cause a linker error, iirc. [Here](https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl) is an example explanation for why this was happening.

Examples like `builder/basic/impl.hpp` and `builder/core.hpp` also adopt this practice of implementing a template function within the header file.

src/bsoncxx/test/bson_serialization.cpp
rayangler: Yup, good point! I'll remove them!

The original intention of creating separate classes/structs was to make sure the serialization could be done with either a class or a struct, but this was probably more for my own personal testing. It was also intended that the other classes would later be used as nested objects for future testing (but this is irrelevant now).

src/bsoncxx/document/value.hpp
rayangler: Is this a normal pattern and/or use case? It seems like doing this will result in both `Person x` and `Person y` having the same deserialized values from the `expected_doc`. It's useful for if the user would like to have 2 Persons with the same value, but I'm just not used to seeing this for assignment.

src/bsoncxx/document/value.hpp
bazile-clyde: Ah yes! Disregard.

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace test {
+
+struct Person {
+    std::string first_name;
+    std::string last_name;
+    int age;
+};
+
+// to_bson serializer function that can take a document::value&
```
bazile-clyde, line 39: I think the function declaration states this pretty well. I'd remove this comment.

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+    bsoncxx::document::value expected_doc =
+        make_document(kvp("first_name", expected_person.first_name),
+                      kvp("last_name", expected_person.last_name),
+                      kvp("age", expected_person.age));
+    auto expected_view = expected_doc.view();
+
+    SECTION("Conversion from BSON object to Person struct works") {
+        // BSON object -> Person
```
bazile-clyde, line 65: This comment and the one on line 74 can be removed since the description in `SECTION` already states this.

src/bsoncxx/document/value.hpp
```diff
@@ -23,6 +23,11 @@
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
+
+namespace array {
+class view;
+}
```
bazile-clyde, line 29: Could this be an include?
```
#include <bsoncxx/array/view.hpp>
```

src/bsoncxx/document/value.hpp
```diff
@@ -76,6 +81,25 @@ class BSONCXX_API value {
...
+    ///   A custom user object to serialize. This object must be default-constructible for
+    ///   deserialization.
+    ///
+    template <typename T,
+              typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
+    explicit value(const T& user_object) : value({}) {
+        to_bson(user_object, *this);
+    }
+    template <typename T>
+    value& operator=(const T& user_object) {
```
bazile-clyde, line 98: nit: You could remove the "user" in "user_object" here and above. It doesn't matter who owns or create the object. We usually just call this parameter _t_ or _v_ or even _value_ depending on the context.

bazile-clyde (changes requested):

src/bsoncxx/document/value.hpp
rayangler: OH, that makes sense. Thank you!

bazile-clyde (approved):
LGTM! Great work!

src/bsoncxx/document/value.hpp
bazile-clyde: Since you have the include now you should be able to delete this.

src/bsoncxx/document/value.hpp
```diff
@@ -76,6 +87,20 @@ class BSONCXX_API value {
     value(value&&) noexcept = default;
     value& operator=(value&&) noexcept = default;
 
+    // Serializer functions
+    template <typename T,
+              typename std::enable_if<!std::is_same<T, typename array::view>::value, int>::type = 0>
```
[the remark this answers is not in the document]
bazile-clyde: An alternative way to do this would be:
```
template <typename T>
explicit value(const T& t) : value({}) {
        // do stuff
}

template <typename T>
explicit value(
        typename std::enable_if<std::is_same<T, array::view>::value, array::view>::type t)
        : value((document::view)t) {}
```

src/bsoncxx/document/value.hpp
```diff
@@ -100,6 +119,21 @@ class BSONCXX_API value {
...
+    /// Constructs an object of type T from this document object. This method uses
+    /// argument-dependent lookup to find the function declaration
+    /// `void from_bson(T& t, const bsoncxx::document::view& doc)`.
+    ///
+    /// @param t
+    ///   An optional partially constructed object. If no argument is passed, the function will
+    ///   create an object with its default constructor.
+    ///
+    template <typename T>
+    T get(T t = {}) {
```
kevinAlbs, line 132: By accepting an argument of type `T`, this lets users convert to type `T` without explicitly specifying the type. E.g. they can write:

```
Person p;
doc.get(p)
```

But, I think we want the argument to be a reference `T& t`. Otherwise, the example above does not modify `p`, so users still need to store the return value. As it stands, users need to do one of the following:
```
Person p;
p = doc.get(p);
// or
Person p;
p = doc.get<Person>();
```

I'm assuming this corresponds to the `get_to` method in nlohmann/json: https://nlohmann.github.io/json/doxygen/classnlohmann_1_1basic__json.html

src/bsoncxx/test/bson_serialization.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    SECTION("Conversion from BSON object to Person struct works") {
+        test::Person test_person = expected_doc.get<test::Person>();
+        REQUIRE(test_person == expected_person);
+    }
+
+    SECTION("Conversion from BSON object to Person using partially constructed object") {
+        test::Person x;
+        test::Person y = expected_doc.get(x);
+        REQUIRE(y != x);
```
kevinAlbs, line 80: Similar to my comment above, it seems unnecessary to pass `x`. I think this should modify `x`.

src/bsoncxx/document/value.hpp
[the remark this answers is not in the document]
kevinAlbs: Oh nice use of SFINAE! Either approach seems fine to me.

src/bsoncxx/document/value.hpp
kevinAlbs: Ah, I see this was brought up in prior comment: https://github.com/mongodb/mongo-cxx-driver/pull/726#discussion_r501231115. @bazile-clyde does taking a `T&` seem right?

kevinAlbs (commented):
Looking great! I just have a question about one of the `get` overloads.

src/bsoncxx/document/value.hpp
bazile-clyde: Ah yes! A reference does seem better.

src/bsoncxx/test/bson_serialization.cpp
bazile-clyde: Yeah once it takes a reference this can be rewritten.

src/bsoncxx/document/value.hpp
bazile-clyde: Flagging this comment again.

rayangler: The forward declaration should already be removed

kevinAlbs (commented):
Fabulous work, LGTM.

kevinAlbs (approved):

bazile-clyde:
Feel free to merge this. Great work!

bazile-clyde (approved):

PR #728, opened 2020-09-30
fix std::put_time error on travis

samantharitter:


src/bsoncxx/test/oid.cpp
```diff
@@ -48,15 +51,21 @@ parsed_oid parse_oid(const oid& oid) {
...
 #endif
 
     return parsed;
 }
 
 void compare_string(const std::time_t& t, std::string time) {
-    std::stringstream time_stream;
-    time_stream << std::put_time(std::gmtime(&t), "%b %e, %Y %H:%M:%S UTC");
-    REQUIRE(time_stream.str() == time);
+    // std::stringstream time_stream;
```
kevinAlbs, line 61: Remove comments.

src/bsoncxx/test/oid.cpp
```diff
@@ -48,15 +51,21 @@ parsed_oid parse_oid(const oid& oid) {
...
+    char time_str[48];
+    if (0 == (strftime(time_str, sizeof(time_str), "%b %e, %Y %H:%M:%S UTC", std::gmtime(&t)))) {
```
kevinAlbs, line 63: Perhaps `REQUIRE (0 != strftime(...))`?

kevinAlbs (approved):
Thanks for looking into this! LGTM with some minor cleanup.

src/bsoncxx/test/oid.cpp
samantharitter: Yes this would be better!  Done.

src/bsoncxx/test/oid.cpp
samantharitter: Done.

kevinAlbs (approved):

PR #729, opened 2020-09-30
CXX-1587 update site generation for modern hugo

rcsanchez97:
I generated the site locally with hugo 0.54 and visually inspected the changed pages, comparing them to those on mongocxx.org for reference.


kevinAlbs (approved):
Woo, thank you for fixing this! I verified with Hugo v0.75.1

PR #732, opened 2020-10-08
CXX-1842 Make values have a superset of the view API

rayangler:
[The ticket](https://jira.mongodb.org/browse/CXX-1842) asked to let values be implicitly converted to views, but it seems to already be able to do that. This PR focuses on the second half, which is to let values use view's methods.

### DIFF
--- a/src/bsoncxx/document/value.cpp
+++ b/src/bsoncxx/document/value.cpp
@@ -57,11 +57,11 @@ document::view::const_iterator value::cend() const {
 }
 
 document::view::const_iterator value::begin() const {
-    return this->view().begin();
+    return cbegin();
 }
 
 document::view::const_iterator value::end() const {
-    return this->view().end();
+    return cend();
 }
 
 document::view::const_iterator value::find(stdx::string_view key) const {
@@ -74,7 +74,7 @@ element value::operator[](stdx::string_view key) const {
 }
 
 const std::uint8_t* value::data() const {
-    return static_cast<uint8_t*>(_data.get());
+    return _data.get();
 }
 
 std::size_t value::length() const {
--- a/src/bsoncxx/document/value.cpp
+++ b/src/bsoncxx/document/value.cpp
@@ -48,9 +48,41 @@ value& value::operator=(const value& rhs) {
     return *this;
 }
 
+view::const_iterator value::cbegin() const {
+    return this->view().cbegin();
+}
+
+view::const_iterator value::cend() const {
+    return this->view().cend();
+}
+
+view::const_iterator value::begin() const {
+    return this->view().begin();
+}
+
+view::const_iterator value::end() const {
+    return this->view().end();
+}
+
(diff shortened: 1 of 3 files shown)


src/bsoncxx/document/value.cpp
```diff
@@ -48,11 +48,43 @@ value& value::operator=(const value& rhs) {
...
+    return static_cast<uint8_t*>(_data.get());
+}
+
+std::size_t value::length() const {
+    return _length;
+}
+
+bool value::empty() const {
+    return _length == 5;
+}
```
rayangler, line 86: These functions don't make a call to `view()`. I did this to save the driver from having to create a view to return information about the member fields it already has. Would it be better practice to have these functions call `return this->view().data()` instead to keep things consistent?

bazile-clyde: Nope! This looks right. `value` really shouldn't depend on view for anything, but for it's simpler, for now, to call a `view` method. In fact, you should probably change `begin` and `end` to call `cbegin` and `cend` respectively the way `view` does it. That way, we reduce dependencies between the two.

bazile-clyde (changes requested):

src/bsoncxx/document/value.cpp
```diff
@@ -48,11 +48,43 @@ value& value::operator=(const value& rhs) {
...
+    return this->view().find(key);
+}
+
 element value::operator[](stdx::string_view key) const {
     auto view = this->view();
     return view[key];
 }
 
+const std::uint8_t* value::data() const {
+    return static_cast<uint8_t*>(_data.get());
```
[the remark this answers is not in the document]
rayangler: This should work just fine with `return _data.get()`! I was just following the same form of how `_data` is passed in the `value::view()` [function](https://github.com/mongodb/mongo-cxx-driver/blob/master/src/bsoncxx/document/value.hpp#L126).

src/bsoncxx/document/value.hpp
```diff
@@ -88,6 +125,31 @@ class BSONCXX_API value {
     ///
     element operator[](stdx::string_view key) const;
 
+    ///
+    /// Access the raw bytes of the underlying document.
+    ///
+    /// @return A (non-owning) pointer to the view's buffer.
```
kevinAlbs, line 131:
```suggestion
    /// @return A (non-owning) pointer to the value's buffer.
```

kevinAlbs (approved):
LGTM! I created CXX-2118 to make a similar change to the `array::value` and `bson_value::value`.

src/bsoncxx/document/value.hpp
bazile-clyde: Also, this pointer _is_ owning. So we should remove the _non-owning_ portion of the comment.

rayangler: Done!

bazile-clyde (approved):
LGTM!

PR #733, opened 2020-10-21
CXX-1660 fix doxygen warnings

rcsanchez97:


rcsanchez97:
I just realized that I pushed a broken change here.  I'm investigating now and will follow-up with a revision.

rcsanchez97:
@kevinAlbs This should be ready for review now.

kevinAlbs (approved):
LGTM with a suggested class docs change.

PR #735, opened 2020-10-22
Header file addition for stream

sapi01:
The stream builder and the basic builder are defined in separate headers. We should see the stream builder convenience methods in bsoncxx/builder/stream/helpers.hpp. To use the stream builder in our code, include the following header files.


kevinAlbs:
Thank you for the contribution.

PR #736, opened 2020-10-22
CXX-1697 test with C driver extra alignment on

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/5f91ea0ec9ec440f318bbd19/tasks


.mci.yml
```diff
@@ -260,7 +289,7 @@ functions:
                      export MONGOCRYPTD_PATH=$(cygpath -m $MONGOCRYPTD_PATH)
                   fi
                   cd mongo-cxx-driver
-                  
+
```
kevinAlbs, line 292: Revert whitespace change.

.mci.yml
```diff
@@ -371,6 +400,75 @@ functions:
                   # logic to handle ${test_params}.
                   MONGOC_VERSION=${mongoc_version|master} .evergreen/build_example_projects.sh
 
+                  if [ $(echo ${test_params} | grep -c UBSAN_OPTIONS) -ge 1 ]; then
```
kevinAlbs, line 403: Doubling the C driver compile time and the test time for each task seems like a lot to check for errors due to alignment. Can we create a separate `compile_and_test_without_alignment` task? We can add that to each variant, but it can at least run alongside other tasks instead of adding onto existing tasks.

kevinAlbs (changes requested):

.mci.yml
rcsanchez97: Done

.mci.yml
rcsanchez97: Excellent point.  I reworked the change and it looks much nicer now.  Adding a separate task allows for much better reuse of the Evergreen functions we already have.  The whole diff is much easier to read and much simpler as well.

rcsanchez97:
Full Evergreen patch build: https://spruce.mongodb.com/version/5f9d74ff61837d12e51d657c/tasks

(Note: this PR has a base commit that is prior to Sam's fix that turned all the zSeries tasks green, so all of the zSeries tasks in this PR are red)

kevinAlbs (approved):
LGTM!

PR #738, opened 2020-11-02
move clang-tidy check opts to file

bazile-clyde:
Moving the `check` options out of the script and into a file makes it easier to run clang-tidy locally. The next step would be actually fixing some of these bugs.

To run locally:
```
FIND="find src -type f \( -name \*.hh -o -name \*.hpp -o -name \*.cpp \) -not -path \"*third_party*\""
eval $FIND | xargs clang-tidy -p build
```


kevinAlbs (approved):
LGTM! This does make running clang-tidy locally more convenient.

PR #740, opened 2020-11-03
CXX-2074 implement mongohouse tests

samantharitter:
In order to support these tests, I added a few new methods to the bsoncxx::array::view class, along with tests.  I chose to implement them on the class itself rather than as a test method because I thought they would be generally useful methods for users of the driver.  These changes are in their own commit.  

I also made changes to allow tests to iterate over events in the apm checker, improve some error messaging around unset elements, and fixed a small bug around unset elements.

The mongohouse tests will run on evergreen in their own new linux buildvariant, where the mongohouse tests are the only task that runs.


src/bsoncxx/array/view.cpp
```diff
@@ -139,6 +140,19 @@ element view::operator[](std::uint32_t i) const {
...
+view::const_iterator view::find_value(bsoncxx::types::bson_value::view v) const {
+    for (auto it = this->cbegin(); it != this->cend(); it++) {
+        if (it->get_value() == v) {
+            return it;
+        }
+    }
+    return this->cend();
+}
+
+bool view::has_value(bsoncxx::types::bson_value::view v) const {
```
bazile-clyde, line 152: We could also overload `operator[]` to be consistent with the other find/operator methods for indices. So just:
```
element view::operator[](bsoncxx::types::bson_value::view v) const {
    return *(this->find_value(v));
}
```
would allow for boolean testing.

src/bsoncxx/array/view.hpp
```diff
@@ -25,6 +25,13 @@
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
+
+namespace types {
+namespace bson_value {
+class view;
+}  // namespace bson_value
+}  // namespace types
+
```
bazile-clyde, line 34: Is this used somewhere?

bazile-clyde (changes requested):

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -40,9 +40,14 @@ void apm_checker::compare(bsoncxx::array::view expectations,
     if (_skip_kill_cursors)
         _events.erase(std::remove_if(_events.begin(), _events.end(), is_kill_cursor));
     for (auto expectation : expectations) {
-        REQUIRE(events_iter != _events.end());
-
         auto expected = expectation.get_document().view();
+
+        if (events_iter == _events.end()) {
+            FAIL("Out of events, expected to find event " << to_json(expected) << "\n\n");
```
kevinAlbs, line 46: Nice!

src/mongocxx/test/spec/monitoring.hh
```diff
@@ -41,8 +41,26 @@ class apm_checker {
     void skip_kill_cursors();
     void print_all();
 
+    using event_vector = std::vector<bsoncxx::document::value>;
+    using iterator = event_vector::iterator;
+    using const_iterator = event_vector::const_iterator;
+
+    inline iterator begin() noexcept {
```
kevinAlbs, line 48: Oh good idea. This does make checking capture events more convenient.

src/bsoncxx/array/view.hpp
```diff
@@ -79,6 +86,28 @@ class BSONCXX_API view {
...
+    ///
+    /// Returns the first element that has a value that compares equal to 'v', if one exists in
+    /// the array. Otherwise, returns the past-the-end iterator.
+    ///
+    /// @param v
+    ///   The value to search for in this array.
+    ///
+    /// @return An iterator to the first matching element, or the past-the-end iterator.
+    ///
+    const_iterator find_value(types::bson_value::view v) const;
```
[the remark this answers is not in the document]
samantharitter: Done, I'll change this commits to reference CXX-2125.

src/mongocxx/test/spec/monitoring.hh
samantharitter: Thanks!  It would have been difficult to implement these tests without this.

src/mongocxx/test/spec/monitoring.cpp
samantharitter: Thanks!

src/bsoncxx/document/element.cpp
```diff
@@ -90,6 +92,11 @@ BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_BEGIN
 BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_END
 
 types::b_utf8 element::get_string() const {
+    if (_raw == nullptr) {
```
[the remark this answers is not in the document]
samantharitter: Thanks!  That was a fun bit of debugging...

src/bsoncxx/document/element.cpp
```diff
@@ -59,7 +59,8 @@ std::uint32_t element::keylen() const {
 
 bsoncxx::type element::type() const {
     if (_raw == nullptr) {
-        throw bsoncxx::exception{error_code::k_unset_element};
+        throw bsoncxx::exception{error_code::k_unset_element,
+                                 "cannot return the type of uninitialized element"};
```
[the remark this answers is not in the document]
samantharitter: Glad you think so! :D

src/bsoncxx/array/view.cpp
```diff
@@ -139,6 +140,19 @@ element view::operator[](std::uint32_t i) const {
     return *(this->find(i));
 }
 
+view::const_iterator view::find_value(bsoncxx::types::bson_value::view v) const {
```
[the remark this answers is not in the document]
samantharitter: I thought that might be confusing to users because the two methods work so differently.  The existing find() method looks up an element at the given index--which is not what I expected it to do when I first looked at this class!

Really, I think it would be more clear if the existing `array::find()` method were renamed to `array::at()`, to match the API of `std::array`.  Then we would be free to repurpose `array::find()` to act like`array::find_value()`, which would then also behave more like `std::find` and maybe be more clear to users.

If you and @kevinAlbs agree with all that, then here's what I think we could do:
- rename the existing `array::find()` to `array::at()` and deprecate `array::find()`
- implement `array::find_value()` as shown here and maybe add `array::find_value_if()`
- open a JIRA ticket to deprecate `array::find_value()` and rename it to `array::find()` in the next release (same for `find_value_if`)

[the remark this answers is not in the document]
samantharitter: That's true, I could have implemented this that way!  But unless you have a strong preference for using `std::find_if`, I'll just leave it as-is, as they are essentially the same

src/bsoncxx/array/view.cpp
samantharitter: I also don't think we should overload `operator[]`, I think we should try to keep methods in agreement with the `std::array` API as much as possible.  For `std::array`, `operator[]` functions as our existing one does, by looking up the element at the given index.

src/bsoncxx/array/view.hpp
samantharitter: This forward declaration is necessary in order to make the `find_value` declaration below compile.

src/mongocxx/test/spec/mongohouse.cpp
```diff
@@ -0,0 +1,300 @@
...
+            }
+
+            configure_fail_point(client, test.get_document().value);
+
+            auto perform_op =
+                [&database, &op_runner, &test, &outcome_collection_name](document::view operation) {
+                    optional<document::value> actual_outcome_value;
+                    INFO("Operation: " << bsoncxx::to_json(operation));
+                    try {
+                        actual_outcome_value = op_runner.run(operation);  // here
```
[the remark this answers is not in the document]
samantharitter: Ah whoops, will remove

src/bsoncxx/test/array.cpp
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/array/view.hpp>
+#include <bsoncxx/builder/basic/array.hpp>
+#include <bsoncxx/test_util/catch.hh>
+#include <bsoncxx/types.hpp>
+#include <bsoncxx/types/bson_value/make_value.hpp>
+#include <bsoncxx/types/bson_value/view.hpp>
```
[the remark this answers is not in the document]
samantharitter: Nope, good call!  I will remove it.

src/bsoncxx/array/view.cpp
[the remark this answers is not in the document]
kevinAlbs: > rename the existing array::find() to array::at() and deprecate array::find()

> open a JIRA ticket to deprecate array::find_value() and rename it to array::find() in the next release (same for find_value_if)


I agree with `at` being a better name than `find`. I am in favor of adding `at`.

But I am opposed to deprecating and repurposing `find`.

Given that you can construct a `bson_value` from an int32, that might silently change behavior when a user upgrades.

Suppose a user has this code, intending to look up element at position 3:
```
auto iter = arr.find(3);
```
Then they upgrade to C++ 3.8, and we've change `find` to mean "find by searching bson_value", I think this would still compile, but have a different meaning.

> implement array::find_value() as shown here and maybe add array::find_value_if()

`find_value` seems like a clear enough name to me. As does `search`.

If we want to be consistent with `std::array`, it does not provide `find` or `find_if`. Instead, users use `std::find` and `std::find_if` with iterators. We could take the same approach. If we made a `bsoncxx::element` comparable to `bsoncxx::bson_value`, I think users could use `std::find` and `std::find_if`.

That being said `arr.find_value("test")` is more concise than `std::find (arr.begin(), arr.end(), "test"))`. `find_value` and `has_value` seem like reasonable conveniences. But I think we should draw the line somewhere too.

kevinAlbs (commented):
Changes LGTM pending discussion of `has_value` and `find_value`.

src/bsoncxx/array/view.cpp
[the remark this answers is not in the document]
samantharitter: Perhaps we should rename `find()` to `at()` (in a separate ticket) and leave this as `find_value()` then?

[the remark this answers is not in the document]
kevinAlbs: SGTM if @bazile-clyde agrees.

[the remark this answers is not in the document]
bazile-clyde: > I thought that might be confusing to users because the two methods work so differently. The existing find() method looks up an element at the given index--which is not what I expected it to do when I first looked at this class!

Ah, that makes sense. It’d be really confusing to have a subtle difference between finding by index or value.

> Really, I think it would be more clear if the existing array::find() method were renamed to array::at(), to match the API of std::array.

I think so too. We could (and probably should for consistency with `std::array`) also add bounds checking for `at`. So `throw` instead of returning the end iterator. Especially since we already have a non-throwing "find" with `operator[]`.

> Then we would be free to repurpose array::find() to act like array::find_value(), which would then also behave more like std::find and maybe be more clear to users.

That being said, I’m not sure why we need `find` or `find_value` though. Sequence containers in the STL don’t have `find` methods and instead rely on `std::find` to look for specific elements by value. The `find` methods for associative containers are used to look up elements by keys, not values. 

As is, `find_value` behaves identically to `std::find`. The only difference is how we compare the elements. Why not just create an `operator==` that compares array elements to values? Then users can use `std::find` without a custom compare function, which is essentially what `find_value` is abstracting away.

src/bsoncxx/array/view.cpp
bazile-clyde: That's a good, point. Resolved.

src/mongocxx/test/spec/mongohouse.cpp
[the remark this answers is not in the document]
bazile-clyde: Was the removal pushed?

src/mongocxx/test/spec/mongohouse.cpp
```diff
@@ -0,0 +1,300 @@
...
+using namespace bsoncxx::stdx;
+using namespace bsoncxx::string;
+using namespace mongocxx::spec;
+
+using bsoncxx::builder::basic::kvp;
+using bsoncxx::builder::basic::make_document;
+
+const std::string kMongohouseURI = "mongodb://mhuser:pencil@localhost";
+
+void run_mongohouse_tests_in_file(std::string test_path) {
```
bazile-clyde, line 38: Since this code is identical to `crud.cpp` maybe we should consolidate both chunks into one method and/or use some of the functions in `util.cpp` like the other test runners. It'll be a lot less code to update when we inevitably have to refactor this.

src/bsoncxx/array/view.cpp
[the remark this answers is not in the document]
samantharitter: I think that adding `find_value` is a nice shortcut, and a little bit nicer to use than `std::find` for an operation that I imagine being somewhat common for users.  Especially considering that it might be a little confusing to users that `std::find` would compare `element` and `value` and not two things of the same type.  But, it does differ from the API of an STL container, and if you and @kevinAlbs agree that it would be better to encourage users to use `std::find` for element lookup, I'll change the test to use `std::find` and add public `element` to `value` comparison functions instead.

I've opened a separate ticket to rename `find()` to `at()`.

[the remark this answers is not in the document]
kevinAlbs: I discussed this with @bazile-clyde. Though I do see the convenience of `find_value`, I am in favor of removing `find_value` in and adding an `operator==` overload for `bson_value` and `element` for consistency with STL.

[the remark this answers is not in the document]
samantharitter: Fair enough! Done.

src/mongocxx/test/spec/mongohouse.cpp
```diff
@@ -0,0 +1,300 @@
...
+        if (event.find("command_started_event") != event.end()) {
+            if (event["command_started_event"]["command_name"].get_string().value !=
+                bsoncxx::stdx::string_view{"killCursors"}) {
+                continue;
+            }
+
+            // Validate namespace
+            auto db = event["command_started_event"]["database_name"].get_string().value;
+            auto coll = event["command_started_event"]["command"]["killCursors"].get_string().value;
+            std::string cmd_ns{db};
```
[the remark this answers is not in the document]
samantharitter: Yes, hopefully this is fixed in my most recent commit.  I believe it has to do with the curly bracket initialization, I once found the formal bug report and could find it again if you are interested.

src/mongocxx/test/spec/mongohouse.cpp
bazile-clyde: Following up on this.

samantharitter: Ah whoops, missed this comment.  Will refactor.

bazile-clyde (approved):
LGTM, pending fixes for test failures.

src/mongocxx/test/CMakeLists.txt
```diff
@@ -196,6 +205,10 @@ add_test(transactions_specs test_transactions_specs)
 add_test(retryable_reads_spec test_retryable_reads_specs)
 add_test(read_write_concern_specs test_read_write_concern_specs)
 
+# Adding this as a test will run it as part of the RUN_TESTS command in MSVC.
+# Do not add, since we only test mongohouse on Linux.
```
kevinAlbs, line 209: Why is this only on run on Linux? https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/atlas_data_lake accounts for running in Windows.

I am ok with limiting to Linux now. But I think it is worth testing connectivity on Windows / macOS since TLS goes through a different code path. But we could do this in C driver tests instead, since the C++ driver does not alter TLS / connection behavior.

kevinAlbs (commented):
Code changes look good!

I have a hypothesis on the test failures.

In single threaded mode, a `mongocxx::client` will exit server selection if it is unable to connect after attempting one time.
https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#serverselectiontryonce

mongohouse is started in the background. If the C++ driver tests start running before mongohouse starts listening on sockets, then that could result in the immediate server selection error.

We could solve that by running mongohouse tests with the URI parameter `serverSelectionTryOnce=false` or by running some script before the tests that ensures mongohouse is listening for connections (e.g. loop and sleep for one second until the shell can connect).

src/mongocxx/test/CMakeLists.txt
samantharitter: That... is a great question.  I was using Java's implementation as a reference, and they only run the mongohouse test on an ubuntu-1804.  But now that I've gone looking, I can't find anywhere that states which/how many different buildvariants should be running these tests.  

In the scope document for ADL, there are these two goals:
- For each driver, run the test suite in Evergreen triggered off of commits to ADL's master branch
- For each driver, run the test suite in Evergreen on every commit to the driver.

Nothing specific about what type of environment the test suite should run in on Evergreen.  We are satisfying the second goal, at least.  I'm not sure about how the first one would be triggered, but I didn't see any of the other implementations checking for updates to the ADL master branch.

For now, maybe we can leave this as just linux?  It would be easy to add additional environments later, once there's more clear guidance about which environments should be tested.

samantharitter:
Let's see if serverSelectionTryOnce=false fixes evergreen!  My patch build before this was green, so I think you're right about it being a flaky timing issue.

src/mongocxx/test/CMakeLists.txt
kevinAlbs: > Nothing specific about what type of environment the test suite should run in on Evergreen. We are satisfying the second goal, at least. I'm not sure about how the first one would be triggered, but I didn't see any of the other implementations checking for updates to the ADL master branch.

Oh good catch! I think that is done outside of code changes, by setting an [inter-project dependency](https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#inter-project-dependencies) in [Evergreen project settings for cxx-driver](https://evergreen.mongodb.com/projects##cxx-driver). That can be done after this is merged.

For now, maybe we can leave this as just linux? It would be easy to add additional environments later, once there's more clear guidance about which environments should be tested.

src/mongocxx/test/spec/mongohouse.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+using namespace bsoncxx::stdx;
+using namespace bsoncxx::string;
+using namespace mongocxx::spec;
+
+using bsoncxx::builder::basic::kvp;
+using bsoncxx::builder::basic::make_document;
+
+const std::string kMongohouseURI =
+    "mongodb://mhuser:pencil@localhost/?serverSelectionTryOnce=false";
```
kevinAlbs, line 37: Mind noting with a comment why `serverSelectionTryOnce=false` is set? The racy behavior of spawning mongohouse might not be obvious later.

kevinAlbs (approved):
Tests passed, LGTM with a comment!

src/mongocxx/test/spec/mongohouse.cpp
samantharitter: Yes, definitely!

src/mongocxx/test/CMakeLists.txt
samantharitter: Will do!

samantharitter:
Merged in!

PR #743, opened 2020-11-06
(Debian packaging) CXX-1747 allow building Debian packages with MNMLSTC

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/5fa5a13b9ccd4e7a05535ca8/tasks

Note that after this PR is approved and merged I will cherry-pick the change onto the `releases/v3.6` branch.

### DIFF
--- a/.evergreen/debian_package_build.sh
+++ b/.evergreen/debian_package_build.sh
@@ -17,6 +17,10 @@ on_exit () {
 }
 trap on_exit EXIT
 
+if [ ! -z "${DEB_BUILD_PROFILES}" ]; then
+   echo "DEB_BUILD_PROFILES was set; building with profiles: ${DEB_BUILD_PROFILES}"
+fi
+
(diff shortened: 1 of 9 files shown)


kevinAlbs (approved):
LGTM, just have one question.

rcsanchez97:
> 
>    As a check of my own understanding, we're still fetching MNMLSTC from git
>    as part of our own cmake configuration.
> 
That is correct.

>    Ideally, if MNMLSTC Debian was packaged and available in Debian, we could
>    list it as a dependency and set BSONCXX_POLY_USE_SYSTEM_MNMLSTC for the
>    MNMLSTC build profile.
> 
>    Does that sound right?
> 
Exactly.  We would also be able to dispense with all the header munging
that is a result of shipping the mnmlstc headers in a sub-directory of
/usr/include/bsoncxx.

PR #744, opened 2020-11-09
CXX-2109 Deprecate all references to utf8

bazile-clyde:


src/bsoncxx/types.hpp
```diff
@@ -122,10 +158,14 @@ struct BSONCXX_API b_utf8 {
...
               typename std::enable_if<!std::is_same<b_utf8, typename std::decay<T>::type>::value,
                                       int>::type = 0>
-    BSONCXX_INLINE explicit b_utf8(T&& t) : value(std::forward<T>(t)) {}
+    BSONCXX_INLINE BSONCXX_DEPRECATED explicit b_utf8(T&& t) : value(std::forward<T>(t)) {}
 
     stdx::string_view value;
 
+    operator b_string() {
+        return b_string{value};
+    }
```
bazile-clyde, line 167: You'll notice this can't be round-tripped, i.e., `operator b_utf8` isn't defined in `b_string`. This is because we would have to create a `b_utf8` in the definition and the constructor for the `b_utf8` is deprecated. Again, this is a public header so including the private header containing `BSONCXX_SUPPRESS_DEPRECATION_WARNINGS...` would be a bad idea. I _could_ create a public version of the suppression warnings but that would just allow users to bypass the deprecation warning, assuming that even works.

examples/bsoncxx/builder_basic.cpp
```diff
@@ -35,7 +35,7 @@ int main(int, char**) {
     using bsoncxx::builder::basic::kvp;
 
     doc.append(
-        kvp("foo", "bar"));  // string literal value will be converted to b_utf8 automatically
+        kvp("foo", "bar"));  // string literal value will be converted to b_string automatically
```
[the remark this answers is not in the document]
samantharitter: 👍

examples/bsoncxx/builder_core.cpp
```diff
@@ -61,8 +61,8 @@ int main(int, char**) {
     // Just call append() and pass a value wrapped in the corresponding BSON type.
     auto array = builder::core{true};  // we are building an array
 
-    array.append(types::b_utf8{"hello"});  // append a UTF-8 string
-    array.append(types::b_double{1.234});  // append a double
-    array.append(types::b_int32{1234});    // append an int32
+    array.append(types::b_string{"hello"});  // append a UTF-8 string
+    array.append(types::b_double{1.234});    // append a double
+    array.append(types::b_int32{1234});      // append an int32
```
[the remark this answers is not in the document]
samantharitter: Acknowledged

src/bsoncxx/document/element.cpp
```diff
@@ -78,32 +78,28 @@ stdx::string_view element::key() const {
...
     types::b_##name element::get_##name() const {                   \
         types::bson_value::view v{_raw, _length, _offset, _keylen}; \
         return v.get_##name();                                      \
     }
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
-BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_END
 
-types::b_utf8 element::get_string() const {
+types::b_utf8 element::get_utf8() const {
```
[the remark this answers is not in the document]
samantharitter: Good catch!

src/bsoncxx/document/element.cpp
```diff
@@ -78,32 +78,28 @@ stdx::string_view element::key() const {
...
     types::bson_value::view v{_raw, _length, _offset, _keylen};
-    return v.get_string();
+    BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_BEGIN
+    return v.get_utf8();
```
samantharitter, line 92: Can this call the new `v.get_string()` instead ?

src/bsoncxx/types.hpp
```diff
@@ -122,10 +158,14 @@ struct BSONCXX_API b_utf8 {
     template <typename T,
               typename std::enable_if<!std::is_same<b_utf8, typename std::decay<T>::type>::value,
                                       int>::type = 0>
-    BSONCXX_INLINE explicit b_utf8(T&& t) : value(std::forward<T>(t)) {}
+    BSONCXX_INLINE BSONCXX_DEPRECATED explicit b_utf8(T&& t) : value(std::forward<T>(t)) {}
```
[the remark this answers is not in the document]
samantharitter: We've been able to deprecate a class before gracefully, and I think that the same flow would work for a struct.  Here's what we did for the ssl -> tls rename:

0. Implement new/newly-named NewThing
1. Remove old class Thing
2. Add a deprecated typedef alias for Thing

```
///
/// This class has been renamed to NewThing.
///
/// @deprecated use the NewThing class instead.
///
MONGOCXX_DEPRECATED typedef NewThing Thing;
```

3. Change all instances of Thing -> NewThing in parameters and return values in header files.  Since the alias is there, this shouldn't break user code.

Are there any places that `b_utf8` appears in public headers that you don't think we can get rid of with this strategy?

src/bsoncxx/types.hpp
samantharitter: I agree that we should not create a public version of the suppression warnings.

With the typedef strategy I mentioned in a different comment, I think this issue would go away.

src/bsoncxx/types.hpp
```diff
@@ -668,6 +708,10 @@ BSONCXX_INLINE bool operator==(const b_maxkey&, const b_maxkey&) {
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
 
+BSONCXX_INLINE bool operator!=(const b_utf8& lhs, const b_utf8& rhs) {
```
samantharitter, line 711: also should be covered by the new b_string methods if there is a typedef alias

samantharitter (changes requested):

src/bsoncxx/types.hpp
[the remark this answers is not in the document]
bazile-clyde: That worked! `b_utf8` is now removed.

src/bsoncxx/types.hpp
bazile-clyde: Done

src/bsoncxx/document/element.cpp
bazile-clyde: done

src/bsoncxx/types.hpp
bazile-clyde: removed.

src/bsoncxx/enums/type.hpp
```diff
@@ -17,7 +17,7 @@
 #endif
 
 BSONCXX_ENUM(double, 0x01)
-BSONCXX_ENUM(utf8, 0x02)
+BSONCXX_ENUM(string, 0x02)
```
kevinAlbs, line 20: I believe this renaming causes some API removal. 

grepping through all .hpp files for `#include <bsoncxx/enums/type.hpp>`, I think these would be removed.

1. `type::k_utf8` is defined in `src/bsoncxx/types.hpp`.
2. `error_code::k_need_element_type_k_utf8` and `error_code::k_cannot_append_utf8` are defined. These are integer error codes which I don't believe we can change.

Both cases are enum values.

For those cases, we may need to redefine each at the same integer value.

Ideally we can deprecate the old utf8 value. There may not be a completely portable way of doing so, but `#pragma deprecated` would at least deprecate for compilers that recognize the pragma. Ideally, users are informed by a compiler warning. But at the very least, we can give them notice via release notes.

Here's what I'm thinking.

```
///
/// An enumeration of each BSON type.
/// These x-macros will expand to be of the form:
///    k_double = 0x01,
///    k_string = 0x02,
///    k_document = 0x03,
///    k_array = 0x04 ...
///
enum class type : std::uint8_t {
#define BSONCXX_ENUM(name, val) k_##name = val,
#undef BSONCXX_ENUM
    k_utf8 = 0x02
};

#pragma deprecated(k_utf8)
```

I think this could be a motivation for doing CXX-641 before the next release as well. That way we can double check ABI changes before we release. I moved that to "Needs Triage".

kevinAlbs (commented):
Looking good!

src/bsoncxx/builder/core.cpp
```diff
@@ -288,15 +288,15 @@ core& core::append(const types::b_double& value) {
     return *this;
 }
 
-core& core::append(const types::b_utf8& value) {
+core& core::append(const types::b_string& value) {
```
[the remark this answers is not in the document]
bazile-clyde: Yep

src/bsoncxx/enums/type.hpp
bazile-clyde: Done. I added `k_utf8` and deprecated it using `BSONCXX_DEPRECATED`. For the error codes, I added them back to `error_code.hpp`. They're at the end of the enum to avoid changing other error code values. I set them to the `_string` variants since they share the same values. I did _not_ add them to the switch statements in `error_code.cpp`. You can't add a new case for the same values.

bazile-clyde: So it appears that the ability deprecated enum values didn't appear until GCC-6 ([source](https://gcc.gnu.org/gcc-6/changes.html)):
 
> The C and C++ compilers now support attributes on enumerators. For instance, it is now possible to mark enumerators as deprecated:
> 
> 
>
> ```
> enum {
>   newval,
>   oldval __attribute__ ((deprecated ("too old")))
> };
> ```

The code snippet above (`#pragma deprecated(k_utf8)`) doesn't even compile locally so I'm not sure if it's possible to deprecate this.

bazile-clyde: We might just have to leave this and error code as is until the next major release. @samantharitter any ideas?

bazile-clyde:
@kevinAlbs and @samantharitter unfortunately, to fix merge conflicts, I ended up squash all of the commits into one.

kevinAlbs (approved):
LGTM!

samantharitter (approved):
LGTM as well!

PR #745, opened 2020-11-09
CXX-1985 Rename filter to opts for list_databases

kevinAlbs:
As it turns out, the `list_databases` overload that accepted a `bsoncxx::document::view_or_value` "filter" was really applying it as an options document. So we already support the `authorizedDatabases` option, and all options that can be passed to the `listDatabases` command.

This PR renames the argument and adds a test to ensure we can pass `authorizedDatabases` through.

### DIFF
--- a/src/mongocxx/test/client.cpp
+++ b/src/mongocxx/test/client.cpp
@@ -131,7 +131,7 @@ TEST_CASE("list databases passes authorizedDatabases option", "[client]") {
         bsoncxx::document::value opts = make_document(kvp("authorizedDatabases", true));
         client.list_databases(opts.view());
         REQUIRE(num_called == 1);
-        REQUIRE_BSON_MATCHES(*opts_passed, make_document(kvp("authorizedDatabase", true)));
+        REQUIRE_BSON_MATCHES(*opts_passed, make_document(kvp("authorizedDatabases", true)));
     }
 }
 
--- a/src/mongocxx/test/client.cpp
+++ b/src/mongocxx/test/client.cpp
@@ -73,6 +73,68 @@ TEST_CASE("A client lists its databases with a filter applied", "[client]") {
     REQUIRE(client_list_databases_called);
 }
 
+TEST_CASE("list databases with a filter works", "[client]") {
+    using bsoncxx::builder::basic::kvp;
+    using bsoncxx::builder::basic::make_document;
+
+    instance::current();
+
+    mongocxx::client client{mongocxx::uri{"mongodb://localhost:27017"}};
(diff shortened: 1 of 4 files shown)


src/mongocxx/test/client.cpp
```diff
@@ -73,6 +73,68 @@ TEST_CASE("A client lists its databases with a filter applied", "[client]") {
...
+    }
+}
+
+TEST_CASE("list databases passes authorizedDatabases option", "[client]") {
+    using bsoncxx::builder::basic::kvp;
+    using bsoncxx::builder::basic::make_document;
+    using bsoncxx::to_json;
+
+    auto mock_client_find_databases_with_opts =
+        mongocxx::libmongoc::client_find_databases_with_opts.create_instance();
```
bazile-clyde, line 96: This should probably go in `helpers.hpp` and have the `mock_` prefix removed for consistency.

src/mongocxx/test/client.cpp
```diff
@@ -73,6 +73,68 @@ TEST_CASE("A client lists its databases with a filter applied", "[client]") {
...
+    int num_called = 0;
```
bazile-clyde, line 97: It seems like this is used to assert that `mock_client_find_databases_with_opts` was called. I think there's already `visit` function you could override for functions declared in `helpers.hpp`. An example is on line 177-178: 
```
    bool destroy_called = false;
    client_destroy->visit([&](mongoc_client_t*) { destroy_called = true; });
```

src/mongocxx/test/client.cpp
```diff
@@ -73,6 +73,68 @@ TEST_CASE("A client lists its databases with a filter applied", "[client]") {
...
+    SECTION("list_databases with no arguments") {
+        client.list_databases();
+        REQUIRE(num_called == 1);
+    }
+
+    SECTION("list_databases with filter") {
+        bsoncxx::document::value opts = make_document(kvp("filter", 1));
+        client.list_databases(opts.view());
+        REQUIRE(num_called == 1);
+        REQUIRE_BSON_MATCHES(*opts_passed, make_document(kvp("filter", 1)));
```
bazile-clyde, line 127: Why recreate the doc `{ "filter" : 1 }` ? Why not just ensure `opts_passed` equals `opts`. Here and on line 134.

bazile-clyde (changes requested):

bazile-clyde (approved):
LGTM

kevinAlbs (commented):
Apologies for the sloppy initial commit.

bazile-clyde:
> Apologies for the sloppy initial commit.

No worries!

PR #747, opened 2020-11-11
CXX-2088 add ENABLE_TESTS option

kevinAlbs:


src/mongocxx/CMakeLists.txt
```diff
@@ -187,7 +187,7 @@ include(MongocxxUtil)
 
 if(MONGOCXX_BUILD_SHARED)
     mongocxx_add_library(mongocxx_shared "${MONGOCXX_OUTPUT_BASENAME}" SHARED)
-    target_link_libraries(mongocxx_shared PUBLIC bsoncxx_shared)
+      target_link_libraries(mongocxx_shared PUBLIC bsoncxx_shared)
```
rcsanchez97, line 190: This looks like an accidental whitespace change.

rcsanchez97 (approved):
LGTM (modulo the minor whitespace change)

src/mongocxx/CMakeLists.txt
kevinAlbs: Fixed.

PR #748, opened 2020-11-11
CXX-2127 upgrade Catch2 to v2.13.4

kevinAlbs:
CXX-2127 reported a build issue on arm macOS that was resolved in Catch v2.12.4.

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -1868,7 +1868,6 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
 
             if (test_util::get_max_wire_version(mongodb_client) >= 1) {
                 // The server supports out().
-                REQUIRE(cursor);
                 auto results = get_results(std::move(*cursor));
                 REQUIRE(results.empty());
 
--- a/src/third_party/catch/include/catch.hpp
+++ b/src/third_party/catch/include/catch.hpp
@@ -1,9 +1,9 @@
 /*
- *  Catch v2.10.2
- *  Generated: 2019-10-24 17:49:11.459934
+ *  Catch v2.13.3
+ *  Generated: 2020-10-31 18:20:31.045274
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2019 Two Blue Cubes Ltd. All rights reserved.
+ *  Copyright (c) 2020 Two Blue Cubes Ltd. All rights reserved.
  *
  *  Distributed under the Boost Software License, Version 1.0. (See accompanying
  *  file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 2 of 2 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -1868,7 +1868,6 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
 
             if (test_util::get_max_wire_version(mongodb_client) >= 1) {
                 // The server supports out().
-                REQUIRE(cursor);
```
bazile-clyde, line 1871: Was this intentional?

kevinAlbs: Yes, this was causing a compilation failure on Windows (https://evergreen.mongodb.com/task/cxx_driver_auth__os~windows_2k8_mongodb_version~latest_compile_and_test_auth_with_shared_libs_patch_716240487475b9a104de4eea5f352ba4b255b0da_5fab5de53e8e86119a6aa381_20_11_11_03_43_35)

`cursor` has no `bool` operator overload. The above aggregate will always return a `cursor`, so it seems unnecessary to check here.

bazile-clyde: I don't think that's the issue. `cursor` (the variable, not the class) is declared as an `stdx::optional` in line 1866. On windows, this decomposes to a `boost::optional<mongocxx::v_noabi::cursor>`. `boost::optional` does have a boolean overload that we use pretty often. 

There seem to be unresolved issues with catch2 and boost::optional ([catch2](https://github.com/catchorg/Catch2/issues/1135) and [CXX-1154](https://jira.mongodb.org/browse/CXX-1154)). Weirdly, flipping the boolean check from `REQUIRE(condition)` to `REQUIRE_FALSE(!condition)` works ([evg patch](https://spruce.mongodb.com/version/5fac15989ccd4e4f06bc8551/tasks)).

That being said because `aggregate` returns a plain cursor I'm not sure why we're assigning it to an optional. So I agree, it does seem unnecessary to check here.

bazile-clyde (approved):
LGTM

src/mongocxx/test/collection.cpp
kevinAlbs: Oh good point. I missed that it was being assigned to an optional. `mongocxx::cursor` does not have a default constructor, so the following does not compile:

```
cursor result_cursor;
REQUIRE_NOTHROW(cursor = coll_in.aggregate(pipeline, options));
```
Due to:
```
../src/mongocxx/test/collection.cpp:1866:20: error: no matching constructor for initialization of 'mongocxx::v_noabi::cursor'
            cursor result_cursor;
```

I think that is the reason for the optional, and seems to be a pattern for other assignments in this file.

> Weirdly, flipping the boolean check from REQUIRE(condition) to REQUIRE_FALSE(!condition) works (evg patch).

Oh that is interesting...

> That being said because aggregate returns a plain cursor I'm not sure why we're assigning it to an optional. So I agree, it does seem unnecessary to check here.

Yeah I agree, I think I'd rather leave out the optional check, since it does not add to the test.

PR #750, opened 2020-11-13
CXX-1630 add new example for causal consistency

samantharitter:


examples/mongocxx/causal_consistency.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // Use a causally-consistent session to run some operations.
+    options::client_session session_opts;
+    session_opts.causal_consistency(true);
+
+    auto session_1 = client.start_session(std::move(session_opts));
+
+    bsoncxx::types::b_timestamp current_date;
+    auto time_point = std::chrono::system_clock::now();
+    std::time_t now = std::chrono::system_clock::to_time_t(time_point);
+    current_date.timestamp = static_cast<uint32_t>(now);
```
kevinAlbs, line 61: I think we want `b_date` instead of `b_timestamp`. I think BSON timestamps are generally not for users:
http://bsonspec.org/spec.html
> Timestamp - Special internal type used by MongoDB replication and sharding. First 4 bytes are an increment, second 4 are a timestamp.

examples/mongocxx/causal_consistency.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // Set the cluster time for session_2 to session_1's cluster time,
+    // and set the operation time for session2 to session2's operation time.
+    auto cluster_time_1 = session_1.cluster_time();
+    auto operation_time_1 = session_1.operation_time();
+    session_2.advance_cluster_time(std::move(cluster_time_1));
+    session_2.advance_operation_time(std::move(operation_time_1));
+
+    read_preference rp_secondary;
+    rp_secondary.mode(read_preference::read_mode::k_secondary);
+    wc_majority.majority(std::chrono::milliseconds(1000));
```
kevinAlbs, line 104: This can reuse the `wc_majority` as-is without modifying it.

examples/mongocxx/causal_consistency.cpp
```diff
@@ -0,0 +1,121 @@
...
+int main() {
+    using namespace mongocxx;
+
+    instance inst{};
+    client client{mongocxx::uri{"mongodb://localhost/?replicaSet=replset"}};
+
+    // Start Causal Consistency Example 1
+
+    write_concern wc_majority{};
+    wc_majority.acknowledge_level(write_concern::level::k_majority);
```
kevinAlbs, line 46: You can remove this line, setting `majority` below also sets the acknowledge level.

kevinAlbs (commented):
Looking good! Just a few small suggested tweaks.

examples/mongocxx/causal_consistency.cpp
samantharitter: Done.

examples/mongocxx/causal_consistency.cpp
samantharitter: Done.

examples/mongocxx/causal_consistency.cpp
samantharitter: Ah of course! Done.

kevinAlbs (approved):
LGTM!

samantharitter:
Merged in manually.

PR #751, opened 2020-11-20
Add coding style to contribution guidelines

bazile-clyde:
@kevinAlbs Per our conversation yesterday, this is an attempt to provide clear guidelines for contributing to our codebase. The server's guidelines are well defined, pretty close to how we already code, and default to the Google C++ style guide. I also linked the C++ Core Guidelines. Those guidelines should be a given but I linked them to avoid ambiguity.

### DIFF
--- a/docs/content/contributing/_index.md
+++ b/docs/content/contributing/_index.md
@@ -8,7 +8,7 @@ title = "Contribution Guidelines"
 
 ### Code Style
 
-When contributing code, in addition to following the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines), please follow the same [code style](https://github.com/mongodb/mongo/wiki/Server-Code-Style), [design guidelines](https://github.com/mongodb/mongo/wiki/Server-Design-Guidelines), and [style guidelines](https://github.com/mongodb/mongo/wiki/Style-Guidelines) as [mongodb/mongo](https://github.com/mongodb/mongo). Additions and exceptions are listed below.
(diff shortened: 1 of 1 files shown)


docs/content/contributing/_index.md
```diff
@@ -6,6 +6,10 @@ title = "Contribution Guidelines"
   parent="contributing"
 +++
 
+### Code Style
+
+When contributing code, in addition to following the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines), please follow the same [code style](https://github.com/mongodb/mongo/wiki/Server-Code-Style), [design guidelines](https://github.com/mongodb/mongo/wiki/Server-Design-Guidelines), and [style guidelines](https://github.com/mongodb/mongo/wiki/Style-Guidelines) as [mongodb/mongo](https://github.com/mongodb/mongo). Additions and exceptions are listed below.
```
kevinAlbs, line 11: We may need to remove https://github.com/mongodb/mongo/wiki/Server-Code-Style. Our .clang-format differs from the one in mongodb/mongo. Since we have a lint task that enforces style, we might not need to mention it.

kevinAlbs (commented):
Thanks for adding this! I think this gives us a framework to point to for making some design decisions. I have one note, but otherwise this looks good.

docs/content/contributing/_index.md
samantharitter: Yes, what if we just mention that users should clang-format their code?  Maybe we talk about that somewhere else? 
 Also, in the past we've taken a lot of cues from the Google style guide: https://google.github.io/styleguide/cppguide.html

bazile-clyde: Done

kevinAlbs (approved):
LGTM!

PR #754, opened 2020-11-30
update clang-format

bazile-clyde:
This updates the version of clang-format used by clang_format.py from 3.8 to 7.0.1 (the same version as the server) and corrects include directives according to our contribution guidelines ([here](http://mongocxx.org/contributing/)). 

7.0 has the `IncludeBlocks` key which separates sections of includes into blocks as required.

### DIFF
--- a/src/bsoncxx/array/element.cpp
+++ b/src/bsoncxx/array/element.cpp
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <bsoncxx/array/element.hpp>
-#include <bsoncxx/types/bson_value/view.hpp>
+#include <bsoncxx/config/private/prelude.hh>
 
 #include <stdexcept>
 
-#include <bsoncxx/config/private/prelude.hh>
+#include <bsoncxx/array/element.hpp>
+#include <bsoncxx/types/bson_value/view.hpp>
 
 namespace bsoncxx {
 BSONCXX_INLINE_NAMESPACE_BEGIN
@@ -47,6 +47,6 @@ bool BSONCXX_CALL operator!=(const types::bson_value::view& v, const element& el
     return !(elem == v);
 }
 
-}  // namespace document
+}  // namespace array
 BSONCXX_INLINE_NAMESPACE_END
 }  // namespace bsoncxx
--- a/src/bsoncxx/types/bson_value/private/value.hh
+++ b/src/bsoncxx/types/bson_value/private/value.hh
@@ -17,6 +17,7 @@
 #include <bsoncxx/config/private/prelude.hh>
 
 #include <bsoncxx/private/libbson.hh>
+#include <bsoncxx/types/bson_value/value.hpp>
 #include <bsoncxx/types/bson_value/view.hpp>
 
 namespace bsoncxx {
--- a/src/bsoncxx/types/bson_value/private/value.hh
+++ b/src/bsoncxx/types/bson_value/private/value.hh
@@ -14,10 +14,10 @@
 
 #pragma once
 
+#include <bsoncxx/config/private/prelude.hh>
+
 #include <bsoncxx/private/libbson.hh>
 #include <bsoncxx/types/bson_value/view.hpp>
-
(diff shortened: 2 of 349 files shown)


bazile-clyde:
All of the tasks failing on this patch are also failing on master.

src/mongocxx/options/client_encryption.hpp
```diff
@@ -56,7 +56,7 @@ class MONGOCXX_API client_encryption {
     /// @return
     ///   An optional pointer to the key vault client.
     ///
-    const stdx::optional<client*>& key_vault_client() const;
+    const stdx::optional<mongocxx::client*>& key_vault_client() const;
```
bazile-clyde, line 59: After rearaging the includes, the compiler interpreted `client` in this declaration as `mongocxx::v_noabi::options::client`. However, in the definition (`client_encription.cpp:32`) `client` explicitly refers to a `mongocxx::client`. 

I'm not sure why this didn't generate a compiler warning. Still, there's a few code smells here that could've contributed to this: using a raw pointer (as opposed to the std::unique_ptr, etc.), not including what you use (neither `options/client.hpp` nor `mongocxx/client.hpp` is included above) and a forward declaration of `client`.

[evg task](https://evergreen.mongodb.com/task_log_raw/cxx_driver_ubuntu1604_release_compile_and_test_with_shared_libs_patch_dfe361bf672809beba0f6164fafad9b088d55fef_5fc54d51d6d80a31f05432fb_20_11_30_19_51_56/0?type=T#L2119)

src/bsoncxx/types/bson_value/private/value.hh
```diff
@@ -14,11 +14,12 @@
 
 #pragma once
 
+#include <bsoncxx/config/private/prelude.hh>
+
 #include <bsoncxx/private/libbson.hh>
+#include <bsoncxx/types/bson_value/value.hpp>
```
bazile-clyde, line 20: Not sure `view.hpp` was included but not `value.hpp`. This caused issues [here](src/bsoncxx/array/element.cpp)

src/mongocxx/options/client_encryption.hpp
kevinAlbs: I'm surprised the ambiguity did not generate a compiler warning either.

Those are good points. I don't know of any rationale behind using a raw pointer vs. an optional of unique pointer. Since the compiler does not need to know the definition of a pointer type, I'd wager that was the original intent behind doing a forward declare instead of including `mongocxx/client.hpp` directly.

kevinAlbs (approved):
In general I am a little hesitant to make large formatting changes. I recall when Mark added clang format to the server codebase. Afterwards, most of the git blame was that commit, which made going through history a bit more difficult. But the diffs here are almost exclusively headers so I don't think that will interfere. Considering this caught issues with includes, this seems like a worthwhile change. LGTM!

src/bsoncxx/array/element.cpp
```diff
@@ -47,6 +47,6 @@ bool BSONCXX_CALL operator!=(const types::bson_value::view& v, const element& el
     return !(elem == v);
 }
 
-}  // namespace document
+}  // namespace array
```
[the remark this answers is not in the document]
bazile-clyde: Yep! The option is `FixNamespaceComments` (clang-format 7.0 docs [here](https://releases.llvm.org/7.0.0/tools/clang/docs/ClangFormatStyleOptions.html)). I didn't know it was on by default but it seemed like a good thing to leave on.

bazile-clyde:
> In general I am a little hesitant to make large formatting changes. I recall when Mark added clang format to the server codebase. Afterwards, most of the git blame was that commit, which made going through history a bit more difficult. But the diffs here are almost exclusively headers so I don't think that will interfere. Considering this caught issues with includes, this seems like a worthwhile change. LGTM!

Yeah, I agree. I'm trying to avoid any large changes that aren't include directives. The only other large changes come from clang-tidy, which should just be bugs and code smells, and iwyu, which, again is just headers.

PR #757, opened 2020-12-01
Fix path to tarball

bazile-clyde:
I ran into an issue where `make_release.py` could not find the tarball it just created. Specifying that the tarball is in the build folder fixed the issue. This is already committed on the `releases/v3.6` branch.


rcsanchez97 (approved):
LGTM

kevinAlbs (approved):
LGTM

PR #759, opened 2020-12-02
Fix python on Ubuntu 18.04 hosts

kevinAlbs:
`python -m virtualenv venv` stopped working because `virtualenv` is no longer installed on Ubuntu 18.04 hosts. I think this is related to MONGOCRYPT-297.

This changes to use the version of python available in the MongoDB toolchain on hosts that have it (which should be hosts aside from Windows).

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -284,7 +284,12 @@ functions:
                   fi
                   export CMAKE=${cmake}
 
-                  python -m virtualenv venv
+                  if [ -f /opt/mongodbtoolchain/v3/bin/python3 ]; then
+                      /opt/mongodbtoolchain/v3/bin/python3 -m venv venv
+                  else
+                      python -m virtualenv venv
+                  fi
+
                   cd venv
                   if [ -f bin/activate ]; then
                       . bin/activate
--- a/.mci.yml
+++ b/.mci.yml
@@ -284,10 +284,8 @@ functions:
                   fi
                   export CMAKE=${cmake}
 
-                  if [ -f /opt/mongodbtoolchain/v3/bin/python3 ]; then
+                  if ! python -m virtualenv venv 2>/dev/null; then
                       /opt/mongodbtoolchain/v3/bin/python3 -m venv venv
-                  else
-                      python -m virtualenv venv
                   fi
 
                   cd venv


.mci.yml
```diff
@@ -284,7 +284,12 @@ functions:
                   fi
                   export CMAKE=${cmake}
 
-                  python -m virtualenv venv
+                  if [ -f /opt/mongodbtoolchain/v3/bin/python3 ]; then
```
rcsanchez97, line 287: I would suggest this instead:
```
if ! python -m virtualenv 2>/dev/null; then
    /opt/mongodbtoolchain/v3/bin/python3 -m venv venv
else
    python -m virtualenv venv
fi
```

On a system without the virtualenv module, the command will exit with non-zero status:

```
roberto@debian:~$ python -m virtualenv
/usr/bin/python: No module named virtualenv
roberto@miami:~$ echo $?
1
```

rcsanchez97 (changes requested):

.mci.yml
bazile-clyde: `python -m virtualenv` fails with `too few arguments`. So:
```
if ! python -m virtualenv 2>/dev/null; then   
then> echo "FAILED"                                                                                                                                               
then> else   
else> echo "SUCCESS"                                                                                                                                              
else> fi                                     
SystemExit: 2
FAILED
``` 
fails. But: 
```
python -m virtualenv venv
... output ...
echo $?
0
```
succeeds locally. Maybe I'm doing this wrongly?

rcsanchez97: Interesting.  It might be better to go with `if ! python -m virtualenv venv 2>/dev/null; then`.  Then if the system does have the virtualenv module it doesn't fail with too few arguments.  Then the `else` branch can also be eliminated.

bazile-clyde: Yeah, that looks cleaner.

bazile-clyde (changes requested):
Pending the discussion around the if condition.

bazile-clyde (approved):
LGTM

rcsanchez97 (approved):
LGTM

PR #763, opened 2021-01-22
Fix CMake build for MSVC compiler when tests are disabled

jzakrzewski:


rcsanchez97:
@jzakrzewski thanks very much for your contribution!

PR #764, opened 2021-01-29
CXX-2164 Replace Travis with Evergreen and GitHub actions

kevinAlbs:
CXX-2164 links to a DRIVERS ticket explaining why we must migrate away from Travis.

Travis was testing older versions of clang / gcc than anything in Evergreen. This PR adds additional variants to evergreen so we continue to check compilation on older clang / gcc.


.github/workflows/compile-and-test-latest.yml
```diff
@@ -0,0 +1,26 @@
+name: compile-and-test-latest
```
kevinAlbs, line 1: The action did not run as part of the PR, but it ran on my fork here:
https://github.com/kevinAlbs/mongo-cxx-driver/actions/runs/521114465

.mci.yml
```diff
@@ -971,6 +971,31 @@ buildvariants:
...
+      expansions:
+        build_type: "Debug"
+        tar_options: *linux_tar_options
+        cmake_flags: *linux_cmake_flags
+        mongodb_version: *version_latest
+        compiler: clang++
+      run_on:
+        - ubuntu1404-small
+      tasks:
+        - name: compile_without_tests
```
kevinAlbs, line 998: I have added clang 3.4 and gcc 4.8 to the matrix.

Travis was testing clang 3.6 and gcc 4.9, older than anything we were testing. So I added this to ensure we don't lose coverage against older compilers.

This is using strictly less versions, but I am making an assumption that a successful build on clang 3.4 / gcc 4.8 will imply a successful build on clang 3.6 / gcc 4.9.

etc/calc_release_version.py
```diff
@@ -144,27 +144,24 @@ def main():
...
-                new_version_loose = LooseVersion(str(version_new['major']) + '.' +
-                                                 str(version_new['minor']) + '.' +
-                                                 str(version_new['patch']) + '-' +
-                                                 version_new['prerelease'])
-                if new_version_loose > version_loose:
-                    version_loose = new_version_loose
-                    if DEBUG:
-                        print('Found new best version "' + str(version_loose) \
-                                + '" from tag "' + release_tag_match.group('ver') + '"')
+        tags = check_output (['git', 'describe', '--tags', '--match', 'r*', '--abbrev=0'])
```
kevinAlbs, line 147: The version of git (1.9.1) on Ubuntu 14.04 does not support `--merged`. @rcsanchez97 I think this should be equivalent behavior.

.github/workflows/compile-and-test-latest.yml
bazile-clyde: Nice

.github/workflows/compile-and-test-latest.yml
```diff
@@ -0,0 +1,26 @@
...
+            - run: |
+                INSTALL_PREFIX=$(pwd)/install
+                export LD_LIBRARY_PATH=$INSTALL_PREFIX/lib
+                PREFIX=$INSTALL_PREFIX ./.evergreen/install_c_driver.sh master
+                mkdir cmake-build
+                cd cmake-build
+                cmake -DCMAKE_PREFIX_PATH=$INSTALL_PREFIX -DCMAKE_CXX_FLAGS="--coverage" ..
+                cmake --build . --target all
+                cmake --build . --target test
+                bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"
```
bazile-clyde, line 26: newline?

.github/workflows/compile-and-test-latest.yml
bazile-clyde, line 1: Why do we need this action if we also run code coverage in evergreen?

kevinAlbs: Hah good point. I copied it from Travis, but forgot we had it on evergreen as well. I removed the coverage bit.

In general, the action isn't necessary. It is helpful for external contributors to have some assurance their changes work, since they cannot trigger evergreen patch builds themselves.

.github/workflows/compile-and-test-latest.yml
kevinAlbs: Done.

bazile-clyde (approved):
LGTM. Not sure about the changes in `etc/calc_release_version.py`, though. I'll rely on you and @rcsanchez97 for that.

etc/calc_release_version.py
rcsanchez97: Yeah, it looks to me like it should be equivalent.  I tried to recall why we specifically chose this approach.  My recollection is vague, but it likely had to do with some possible weirdness associated with branching.  That said, I don't know of a single instance where we have encountered a configuration that would not work with the `describe` command.  I tried a few scenarios by hand locally and there was no unusual or unexpected output.

rcsanchez97 (approved):

etc/calc_release_version.py
kevinAlbs: I did the same, it seems equivalent. `git describe` walks backwards to the nearest matching tag. So it gives the same result from `master` and a release branch.

Like before, the result for non-master non-release branches is a little off. I branched directly from master and running gives `3.6.1-20210129+git97cf47e743` when it should probably be `3.7.0-20210129+git97cf47e743`. But... a problem for when we tackle CDRIVER-3315.

PR #765, opened 2021-02-01
CXX-2171 use wc/rc majority in CSFLE prose tests

kevinAlbs:
This is not urgent. It is a small fix to CSFLE tests which prevent some flaky failures when run against replica sets.

### DIFF
--- a/src/mongocxx/test/client_side_encryption.cpp
+++ b/src/mongocxx/test/client_side_encryption.cpp
@@ -203,6 +203,9 @@ void run_datakey_and_double_encryption(Callable create_data_key,
     // Use client to run a find on keyvault.datakeys by querying with the _id
     // set to the datakey_id.
     auto datakeys = setup_client->database("keyvault").collection("datakeys");
+    mongocxx::read_concern rc_majority;
+    rc_majority.acknowledge_level (mongocxx::read_concern::level::k_majority);
+    datakeys.read_concern (rc_majority);
     auto query = make_document(kvp("_id", datakey_id));
     auto cursor = datakeys.find(query.view());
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -203,6 +203,9 @@ void run_datakey_and_double_encryption(Callable create_data_key,
     // Use client to run a find on keyvault.datakeys by querying with the _id
     // set to the datakey_id.
     auto datakeys = setup_client->database("keyvault").collection("datakeys");
+    mongocxx::read_concern rc_majority;
+    rc_majority.acknowledge_level(mongocxx::read_concern::level::k_majority);
+    datakeys.read_concern(rc_majority);
```
kevinAlbs, line 208: The test [readme](https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#prose-tests) notes:

> Perform all applicable operations on key vault collections (e.g. inserting an example data key, or running a find command) with readConcern/writeConcern "majority".

- automatic encryption on a client internally uses read concern majority when reading keys
- creating a data key with a `client_encryption` object internally uses write concern majority

The manual test operations must use majority read/write concern too. Otherwise, a situation could occur like:
- a key is manually inserted with the default `w:1` write concern
- automatic encryption occurs, reading with read concern majority
- the inserted key has not been replicated yet, so automatic encryption fails to find the key

bazile-clyde: Ack.

bazile-clyde (approved):

PR #766, opened 2021-02-12
remove raw pointers from operation class

bazile-clyde:
**Note to reviewers: it'd be easiest to start with `operation.hh`.**

This PR makes two changes to `operation.hh`:
1) replaces raw pointers to a client, database, etc. with a shared pointer
2) adds setters for a client, databases, etc.

I ran into both of these issues frequently while building the new test runner.

**Why change raw pointers to a shared pointer?**
`operation.hh` took a raw pointer to an object. We frequently passed the address of local variables to the operation class. This works until the local variable is destroyed at the end of its scope and leaves the operation class with a pointer to a destroyed value. For example:

```
operation_runner create_op_runner() {
    client client{uri{}};
    auto db = client["my_db"]; 
    return operation_runner{&db, nullptr}; // bad: 'db' will be destroyed at the end  of this function
}

void run_operation() {
   auto doc = make_document(); 
   auto op_runner = create_op_runner();
   op_runner.run(doc); // bad: refers to a destroyed database. Will segfault
}
```
The solution _could_ be to create a real pointer rather than use a local address, but this will require us to manually management that pointer's resources. A shared pointer does this for us.

**Why add setters to `operation.hh`?**
`operation.hh` had two constructors: one that took a collection and one that took five other objects. If you run an operation that only needs a database object, you will still have to pass `nullptr` in place of every other object. The solution to the common problem of having a constructor with a ton of optional arguments is the builder pattern. This is a simplified version of it.

### DIFF
--- a/src/mongocxx/test/spec/operation.cpp
+++ b/src/mongocxx/test/spec/operation.cpp
@@ -122,7 +122,7 @@ bsoncxx::stdx::optional<read_preference> lookup_read_preference(document::view d
     return {};
 }
 
-client_session* operation_runner::_lookup_session(stdx::string_view key) {
+std::shared_ptr<client_session> operation_runner::_lookup_session(stdx::string_view key) {
     if (key.compare("session0") == 0) {
         return _session0;
     } else {
@@ -132,7 +132,7 @@ client_session* operation_runner::_lookup_session(stdx::string_view key) {
     return nullptr;
 }
 
-client_session* operation_runner::_lookup_session(document::view doc) {
+std::shared_ptr<client_session> operation_runner::_lookup_session(document::view doc) {
     if (doc["session"]) {
         stdx::string_view session_name = doc["session"].get_string().value;
         return _lookup_session(session_name);
@@ -156,7 +156,7 @@ document::value operation_runner::_run_aggregate(document::view operation) {
     _set_collection_options(operation);
 
     stdx::optional<cursor> result_cursor;
-    client_session* session = _lookup_session(operation["arguments"].get_document().value);
+    auto session = _lookup_session(operation["arguments"].get_document().value);
 
     if (operation["object"] &&
         operation["object"].get_string().value == stdx::string_view{"database"}) {
@@ -204,7 +204,7 @@ document::value operation_runner::_run_distinct(document::view operation) {
     }
 
     stdx::optional<cursor> result_cursor;
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         result_cursor.emplace(_coll->distinct(*session, field_name, filter, options));
     } else {
         result_cursor.emplace(_coll->distinct(field_name, filter, options));
@@ -291,7 +291,7 @@ document::value operation_runner::_run_find(document::view operation) {
     }
 
     stdx::optional<cursor> result_cursor;
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         result_cursor.emplace(_coll->find(*session, filter, options));
     } else {
         result_cursor.emplace(_coll->find(filter, options));
@@ -326,7 +326,7 @@ document::value operation_runner::_run_delete_many(document::view operation) {
     auto result = builder::basic::document{};
     std::int32_t deleted_count = 0;
 
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         if (auto delete_result = _coll->delete_many(*session, filter, options)) {
             deleted_count = delete_result->deleted_count();
         }
@@ -363,7 +363,7 @@ document::value operation_runner::_run_delete_one(document::view operation) {
     auto result = builder::basic::document{};
     std::int32_t deleted_count = 0;
 
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         if (auto delete_result = _coll->delete_one(*session, filter, options)) {
             deleted_count = delete_result->deleted_count();
         }
@@ -414,7 +414,7 @@ document::value operation_runner::_run_find_one_and_delete(document::view operat
 
     auto result = builder::basic::document{};
     stdx::optional<document::value> document;
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         document = _coll->find_one_and_delete(*session, filter, options);
     } else {
         document = _coll->find_one_and_delete(filter, options);
@@ -493,7 +493,7 @@ document::value operation_runner::_run_find_one_and_replace(document::view opera
 
     auto result = builder::basic::document{};
     stdx::optional<document::value> document;
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
         document = _coll->find_one_and_replace(*session, filter, replacement, options);
     } else {
         document = _coll->find_one_and_replace(filter, replacement, options);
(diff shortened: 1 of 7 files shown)


kevinAlbs:
> I ran into both of these issues frequently while building the new test runner.

Nice, that sounds like good rationale to make these changes.

> The solution could be to create a real pointer rather than use a local address, but this will require us to manually management that pointer's resources. A shared pointer does this for us.

Thanks for including that example!

src/mongocxx/test/spec/operation.hh
```diff
@@ -39,23 +39,29 @@ bsoncxx::stdx::optional<read_preference> lookup_read_preference(document::view d
...
-    operation_runner(database* db,
-                     collection* coll,
-                     client_session* session0 = nullptr,
-                     client_session* session1 = nullptr,
-                     client* client = nullptr);
+    explicit operation_runner() noexcept = default;
+
     document::value run(document::view operation);
 
+    operation_runner& set_collection(std::shared_ptr<collection> coll);
```
kevinAlbs, line 46: Nice, making these chainable makes it simpler for callers to use.

src/mongocxx/test/spec/util.cpp
```diff
@@ -566,7 +566,14 @@ void run_transaction_operations(document::view test,
                 parse_database_options(operation, &db);
                 collection coll = db[coll_name];
                 parse_collection_options(operation, &coll);
-                operation_runner op_runner{&db, &coll, session0, session1, client};
+
+                auto op_runner = operation_runner{}
+                                     .set_database(std::make_shared<database>(db))
+                                     .set_collection(std::make_shared<collection>(coll))
```
kevinAlbs, line 572: I was thinking of suggesting the objects be copied instead of having callers dynamically create a shared pointer. But that cannot be done for a `client` / `session` since those are not copyable. It seems reasonable to have them all be `shared_ptr` for consistency.

kevinAlbs (approved):
LGTM!

src/mongocxx/test/spec/util.cpp
bazile-clyde: Yeah, that was my first thought but our move-only types make it impossible.

src/mongocxx/test/spec/operation.cpp
```diff
@@ -204,7 +204,7 @@ document::value operation_runner::_run_distinct(document::view operation) {
     }
 
     stdx::optional<cursor> result_cursor;
-    if (client_session* session = _lookup_session(operation["arguments"].get_document().value)) {
+    if (auto session = _lookup_session(operation["arguments"].get_document().value)) {
```
samantharitter, line 207: Nice use of auto!  Even if this method wasn't changing auto is definitely better.

src/mongocxx/test/spec/operation.hh
samantharitter: Agreed, and having these as setters rather than as construction parameters seems a lot nicer and more flexible.

samantharitter (commented):
Hey Clyde, thanks for your work on this!

I’m sure those ownership issues in the test framework were really tricky to track down!  And they definitely need to be fixed.  But, I suspect there’s a clearer way to fix the ownership model than applying `shared_ptr`.  I am super wary of introducing broad use of `shared_ptr` as a convention that we start to apply everywhere in the codebase.

In most cases where it is tempting to use a `shared_ptr`, it usually implies that something about the ownership model needs work.  `shared_ptr` is a convenient way to cover up an issue with ownership, but usually there is a more thorough and clear way to address the underlying issue with stronger ownership semantics, like using the `unique_ptr` + bare ptr model.

I disagree with Kevin that we should apply `shared_ptr` to all these types for consistency’s sake.  `shared_ptr`s aren’t free.  Also, having a type that holds 5 different `shared_ptr`s implies another design issue, and perhaps all those shared elements should be encapsulated within their own new class so they can be passed around and used together.  I would encourage you to explore whether there is a different solution.

Each instance of `shared_ptr` adds a hidden performance and memory storage cost to the code.  Granted, this is test code, so it doesn’t matter as much in this particular PR, but it’s really worth considering as a drawback of `shared_ptr`.

Using `shared_ptr` means giving up our control about when resources get freed, and giving up clarity about ownership models.  Managing memory can be a pain, but it’s also one of the things that makes C++ C++, and that makes C++ a performant language.

Another concern about `shared_ptr`’s, when they are used extensively, is that they can create hidden chains of ownership that lead to resources that hold shared_ptrs to each other, and thus cannot be destroyed.  When this happens, it’s pretty nasty, and can be really tricky to debug and untangle!  Again, in this PR, that’s not coming into play right now, but it’s really worth considering as a drawback of introducing more `shared_ptr`s into the codebase.  

This blog post explains more eloquently and completely some of the reasons against over-using `shared_ptr`:
https://www.meetingcpp.com/blog/items/shared_ptr-addiction.html

The answer on this stack overflow post also sums this up well:
https://stackoverflow.com/questions/7657718/when-to-use-shared-ptr-and-when-to-use-raw-pointers

bazile-clyde:
> I am super wary of introducing broad use of shared_ptr as a convention that we start to apply everywhere in the codebase.
I don't have any intention of using this broadly as a convention or applying it everywhere in the codebase. I understand the performance impact and complexity this introduces. I'm using it here explicitly because we were already using a shared pointer implicitly. In other words, this does not change much logic for the caller or callee. It was intended to avoid having the caller delete a pointer while the callee is using it. This is a bug fix.

> I would encourage you to explore whether there is a different solution.
Do you have any suggestions? This is less error-prone than using raw pointers. However, using unique pointers will cause significantly larger changes. Since unique pointers are move-only the operations runner will also become a move-only class. I'll have to introduce getters in order to return these move-only types back to the caller. Furthermore, every function that relies on a local pointer to one of these objects (`parse_database_options` for example) will have to be updated to move that object into and out of it.

samantharitter:
> Do you have any suggestions?

Well, up until this point, the `operation_runner` has only been used as a temporary object in tests.  The current model is that the tests own their client/collection/database objects, so the lifetime model is simple and clear.  Some tests make many `operation_runner`s that all operate over the same set of client/collection/database, and so having the `operation_runner` hold non-owning pointers to those types made logical sense and was the most performant choice.  To my knowledge we have never before used `operation_runner` in a way that required it to own its client/collection/database objects.

The new `create_op_runner()` method you used in your example above is probably to support the Unified Test Runner, right?  `create_op_runner()` introduces a new lifetime model for `operation_runner`, one it wasn't really designed for, with different requirements for how it interacts with the client/collection/database objects.  Maybe `operation_runner` as it currently exists isn't quite the right fit for this new use case.

I don't have the full context for the changes you are trying to make to support Unified, since I haven't seen those changes, but here are some thoughts for alternate designs:

- there could be a new Unified runner or other new test type that uses `operation_runner` in the same way it currently gets used by other test runners, so the client/database/collection objects are owned by an entity other than `operation_runner` that keeps them alive for the duration of the test

- the logic for running test operations (meaning, all those private methods on `operation_runner` like `_run_bulk_write()` etc) could be better moved out of the `operation_runner` class and broken out into free functions so they can be called from different contexts, taking the client/database/collection objects it needs in some new blob type that can be used regardless of who owns those objects.

- `operation_runner` could be an owning-optional type.  You could add unique_ptr members for client/database/collection when an `operation_runner` is created from your new `create_op_runner()` method and needs to own its types.  In the case where `operation_runner` is view-y, like in our existing test runners, the unique_ptrs would be unused, and the bare ptrs used as they are currently used.  When `operation_runner` is value-y, like in your new `create_op_runner()` use case, the unique_ptrs would be populated by the client/database/collection objects it owns, and the bare ptrs would point to those unique_ptrs.  Cleanup is simple, the dtor cleans up the unique_ptr and ignores the raw pointers.  The current view-y lifetime model for `operation_runner` is preserved, but the class becomes flexible enough to own types when it needs to.  This is probably not the ideal solution, either, because it means we're trying to use one type in two pretty different ways, but IMHO it is preferable to using shared_ptrs.

bazile-clyde:
@samantharitter I'll table this PR for now and try out some of those solutions.

kevinAlbs:
> Maybe operation_runner as it currently exists isn't quite the right fit for this new use case.

In the C implementation, I chose not to use the existing operation running code for the existing spec tests (and instead created the gigantic [operation.c](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/unified/operation.c), isolated from other test runners). That may apply here. Getting `operation_runner` to work with existing callers and the unified test runner may be more difficult than having an independent class/functions for running operations that is exclusively used by the unified test runner.

In the grand scheme of things, the other spec test runners ideally will be obsolete. Even if there is some duplication, it may be easier to have the unified test runner a bit isolated.

bazile-clyde:
@kevinAlbs yeah that's probably the best solution for now. Closing this.

PR #768, opened 2021-02-18
cache test network calls

bazile-clyde:
I realized we were making _a lot_ (hundreds) of **isMaster** and **serverStatus** calls to the server per spec test. Profiler results for transaction tests (see images below) show that currently, 16.70% of the time is spent in the function `should_skip_spec_tests`. That function makes multiple roundtrips to the server. After this change, only 2.26% of the time is spent in `should_skip_spec_tests`—a pretty big improvement. 

I also timed `test_transactions_spec` locally before and after caching server results. Here's the results:

```
Before optimization
n=10 | min    | Q1     | med    | mean   | Q3     | max    | std
wall | 28.30  | 28.44  | 28.63  | 28.70  | 28.88  | 29.60  | 0.3865
user | 0.6500 | 0.6550 | 0.6750 | 0.6770 | 0.6975 | 0.7100 | 0.0226
sys  | 0.110  | 0.120  | 0.140  | 0.138  | 0.150  | 0.170  | 0.0230
rss  | 13368  | 13408  | 13556  | 13564  | 13722  | 13740  | 159.2700

After optimization
n=10 | min   | Q1    | med   | mean  | Q3    | max   | std
wall | 27.66 | 27.70 | 27.89 | 28.03 | 28.13 | 28.84 | 0.4206
user | 0.500 | 0.520 | 0.520 | 0.524 | 0.530 | 0.550 | 0.0143
sys  | 0.080 | 0.090 | 0.100 | 0.098 | 0.100 | 0.120 | 0.0140
rss  | 13544 | 13635 | 13740 | 13735 | 13860 | 13908 | 136.2227
```
The transaction tests were run 10 times (n=10) using a script which called `time test_transactions_spec` on Linux and used a Rscript to calculate the mean, median, etc.

Profile results from CLion:
before-optimization
after-optimization

### DIFF
--- a/src/mongocxx/test_util/client_helpers.cpp
+++ b/src/mongocxx/test_util/client_helpers.cpp
@@ -58,7 +58,7 @@ document::value get_is_master(const client& client) {
 }
 
 document::value get_server_status(const client& client) {
-    static auto status = client["test"].run_command(make_document(kvp("serverStatus", 1)));
+    static auto status = client["admin"].run_command(make_document(kvp("serverStatus", 1)));
     return status;
 }
 }  // namespace
--- a/src/mongocxx/test_util/client_helpers.cpp
+++ b/src/mongocxx/test_util/client_helpers.cpp
@@ -50,6 +50,19 @@ using namespace bsoncxx;
 using bsoncxx::builder::basic::kvp;
 using bsoncxx::builder::basic::make_document;
 
+namespace {
+// These frequently used network calls are cached to avoid bottlenecks during tests.
+document::value get_is_master(const client& client) {
+    static auto reply = client["admin"].run_command(make_document(kvp("isMaster", 1)));
+    return reply;
+}
+
+document::value get_server_status(const client& client) {
+    static auto status = client["test"].run_command(make_document(kvp("serverStatus", 1)));
+    return status;
+}
+}  // namespace
+
 std::vector<std::int32_t> parse_version(std::string version) {
     std::vector<std::int32_t> elements;
     std::stringstream ss{version};
@@ -181,7 +194,7 @@ std::basic_string<std::uint8_t> convert_hex_string_to_bytes(stdx::string_view he
 }
 
 std::int32_t get_max_wire_version(const client& client) {
-    auto reply = client["admin"].run_command(make_document(kvp("isMaster", 1)));
+    auto reply = get_is_master(client);
     auto max_wire_version = reply.view()["maxWireVersion"];
     if (!max_wire_version) {
         // If wire version is not available (i.e. server version too old), it is assumed to be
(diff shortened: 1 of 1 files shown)


kevinAlbs:
> After this change, only 2.26% of the time is spent in should_skip_spec_tests—a pretty big improvement.

That is awesome. CLion's flame graph profiler does look really useful for drilling down into perf.

I'm not sure how much this translates into total runtime saved. I looked at tasks in the Evergreen patch, which shows `Duration` and `Base commit duration`. Some of the Windows tasks looked like a possible significant improvement (e.g. [1 minute difference](https://spruce.mongodb.com/task/cxx_driver_windows_2k8_release_compile_and_test_with_shared_libs_extra_alignment_patch_3c7609f04488c547ad7e1607f2eb8b675a107e83_602eec3b3e8e861950994c00_21_02_18_22_37_48/logs?execution=0)). Some other tasks looked like a difference of +/- 30 seconds.

Regardless of the impact on test runtime, this seems like a positive change IMO (to centralize `isMaster` and `serverStatus` calls).

> The transaction tests were run 10 times (n=10) using a script which called time test_transactions_spec on Linux and used a Rscript to calculate the mean, median, etc.

Nice analysis!

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -50,6 +50,19 @@ using namespace bsoncxx;
...
 
+namespace {
+// These frequently used network calls are cached to avoid bottlenecks during tests.
+document::value get_is_master(const client& client) {
+    static auto reply = client["admin"].run_command(make_document(kvp("isMaster", 1)));
+    return reply;
+}
+
+document::value get_server_status(const client& client) {
+    static auto status = client["test"].run_command(make_document(kvp("serverStatus", 1)));
```
kevinAlbs, line 61: Aside: Using the `admin` database is necessary for some administrative commands, like `currentOp`. The database used here for `serverStatus` or `isMaster` could be anything.

kevinAlbs (approved):

src/mongocxx/test_util/client_helpers.cpp
bazile-clyde: Gotcha. Should I change this to "admin" for consistency? I'm not sure if that'd break anything (this was only ever called with the "test" database) but I can change it and let the test re-run to find out.

kevinAlbs: It should be equivalent and safe to change to `admin`. The C test runner `call_ismaster` uses admin. It makes no behavioral difference, so I am fine either way.

bazile-clyde: Ok. Done. Waiting for the tests to pass then merging.

PR #770, opened 2021-03-01
CXX-1871 create entities (client)

bazile-clyde:


src/mongocxx/test/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ set(spec_test_common
     spec/operation.cpp
     spec/monitoring.cpp
     spec/util.cpp
-        )
+)
```
bazile-clyde, line 83: Formatting

src/mongocxx/test/CMakeLists.txt
```diff
@@ -319,15 +320,17 @@ set_dist_list (src_mongocxx_test_DIST
...
    spec/gridfs.cpp
    spec/mongohouse.cpp
    spec/monitoring.cpp
    spec/monitoring.hh
    spec/operation.cpp
    spec/operation.hh
-   spec/unified_test_format.cpp
    spec/read_write_concern.cpp
    spec/retryable-reads.cpp
+   spec/unified_test_format.cpp
```
bazile-clyde, line 333: Sorted.

src/mongocxx/test/spec/change_stream.cpp
```diff
@@ -91,7 +91,7 @@ void run_change_stream_tests_in_file(const std::string& test_path) {
             spec::apm_checker apm_checker;
             options::client client_opts;
             // "Begin monitoring all APM events for `client`"
-            apm_checker.skip_kill_cursors();
+            apm_checker.set_ignore_command_monitoring_event(apm_checker::event::kill_cursors);
```
bazile-clyde, line 94: Since we'll skip more than "kill cursor" commands, I created an enum for different values. My original idea was to use strings, but a typo wouldn't be caught until the specified events aren't skipped, and the test fails.

src/mongocxx/test/spec/entity.hh
```diff
@@ -0,0 +1,57 @@
...
+
+    map() = default;
+
+    map(const map&) = delete;
+    map operator=(const map&) = delete;
+
+    map(map&&) = default;
+    map& operator=(map&&) = default;
+
+    void insert(const key_type& key, mongocxx::client&& client);
```
bazile-clyde, line 46: Will add insert methods for the other entities as well. The idea here was to encapsulate the variant and have the entity class worry about type safety. There are no getters yet. We won't need them until we run operations. The getters will return mongocxx types (client, db, coll, etc) given a key rather than mapped_type.

src/mongocxx/test/spec/unified_test_format.cpp
```diff
@@ -113,6 +111,180 @@ bool has_run_on_requirements(const bsoncxx::document::view test) {
...
+entity::map& get_entity_map() {
+    static auto m = entity::map{};
+    return m;
+}
+
+// TODO: create_change_stream
+// TODO: create_bucket
+// TODO: create_session
+// TODO: create_database
+// TODO: create_collection
```
bazile-clyde, line 240: These are next on my TODO.

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -122,18 +128,18 @@ using xformer_t = std::function<stdx::optional<item_t>(item_t, bsoncxx::builder:
 //
 //   For document elements, the element's key and value will be passed into the function. If
 //   nothing is returned, the element will be removed from the document. Otherwise, the key and
-//   value pair returned will be added to the document. To leave the element as-is, simply return
-//   the same key and value passed into the function. If no key is returned, an exception will be
-//   thrown.
+//   value pair returned will be added to the document. To leave the element as-is, simply
+//   return the same key and value passed into the function. If no key is returned, an exception
+//   will be thrown.
```
bazile-clyde, line 133: clang-format. Here and below.

src/mongocxx/test/spec/change_stream.cpp
kevinAlbs: Good thinking, an enum does seem better.

src/mongocxx/test/spec/entity.hh
kevinAlbs: Nice, I assume `insert` and `get` could be templated?

src/mongocxx/test/spec/unified_test_format.cpp
```diff
@@ -113,6 +111,180 @@ bool has_run_on_requirements(const bsoncxx::document::view test) {
...
+}
+
+std::string uri_options_to_string(document::view object) {
+    // Spec: Optional object. Additional URI options to apply to the test suite's connection string
+    // that is used to create this client. Any keys in this object MUST override conflicting keys in
+    // the connection string.
+    if (!object["uriOptions"])
+        return {};
+
+    // TODO: Spec: if 'readPreferenceTags' is specified in this object, the key will map to an array
```
kevinAlbs, line 154: Nice, it's good that this is called out with an explicit comment. If this does appear in a future test, I think `readPreferenceTags` and `readPreference` could be handled as a special case and removed from input `object` before converting the rest to string with `json_to_uri_opts`.

src/mongocxx/test/spec/unified_test_format.cpp
```diff
@@ -113,6 +111,180 @@ bool has_run_on_requirements(const bsoncxx::document::view test) {
...
+    // Spec: If true and the topology is a sharded cluster, the test runner MUST assert that this
+    // MongoClient connects to multiple mongos hosts (e.g. by inspecting the connection string).
+    if (!object["useMultipleMongoses"] || !object["useMultipleMongoses"].get_bool())
+        return default_uri;
+
+    // from: https://docs.mongodb.com/manual/reference/config-database/#config.shards
+    // If the shard is a replica set, the host field displays the name of the replica set, then a
+    // slash, then a comma-separated list of the hostnames of each member of the replica set, as in
+    // the following example:
+    //      { ... , "host" : "shard0001/localhost:27018,localhost:27019,localhost:27020", ... }
```
kevinAlbs, line 181: Oh good idea. C currently hard codes `localhost:27018` for `useMultipleMongoses`. I noted this in CDRIVER-3285.

kevinAlbs (approved):
Superb work, and thank you for including context in comments. LGTM!

src/mongocxx/test/spec/entity.hh
bazile-clyde: Yep. This class shouldn't get much larger than it is now after adding support for the other entities.

src/mongocxx/test/spec/unified_test_format.cpp
bazile-clyde: Noted.

PR #773, opened 2021-03-11
CXX-1871 load initial data

bazile-clyde:
This PR adds support for the `intialData` top-level field of the unified test format ([here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#top-level-fields)).  

The only part left is actually running and verifying the operations.

### DIFF
--- a/src/mongocxx/test/spec/monitoring.cpp
+++ b/src/mongocxx/test/spec/monitoring.cpp
@@ -156,6 +156,7 @@ void apm_checker::set_ignore_command_monitoring_event(event e) {
 
 void apm_checker::clear() {
     this->_events.clear();
+    this->_ignore.clear();
 }
 
 }  // namespace spec
--- a/src/mongocxx/test/spec/unified_test_format.cpp
+++ b/src/mongocxx/test/spec/unified_test_format.cpp
@@ -429,7 +429,7 @@ std::vector<document::view> array_elements_to_documents(array::view array) {
     return docs;
 }
 
-bool add_data_to_collection(const array::element& data) {
+void add_data_to_collection(const array::element& data) {
     auto db_name = data["databaseName"].get_string().value;
     auto& map = get_entity_map();
     auto& db = map.get_database_by_name(db_name);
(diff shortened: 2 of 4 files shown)


src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -156,6 +156,7 @@ void apm_checker::set_ignore_command_monitoring_event(event e) {
 
 void apm_checker::clear() {
     this->_events.clear();
+    this->_ignore.clear();
```
bazile-clyde, line 159: This should've been clear also when it was added. Doing it now.

src/mongocxx/test/spec/unified_test_format.cpp
```diff
@@ -33,6 +33,11 @@ using schema_versions_t =
     std::array<std::array<int, 3 /* major.minor.patch */>, 1 /* supported version */>;
 constexpr schema_versions_t schema_versions{{{1, 0, 0}}};
 
+entity::map& get_entity_map() {
+    static auto m = entity::map{};
+    return m;
+}
+
```
bazile-clyde, line 40: Move this up. No changes.

benjirewis (approved):
LGTM! Looks like the runner is nearing completion for cpp ❗ 😄

src/mongocxx/test/spec/unified_test_format.cpp
```diff
@@ -423,6 +420,42 @@ std::vector<std::string> versions_to_string(schema_versions_t versions) {
...
+    auto to_insert = array_elements_to_documents(data["documents"].get_array().value);
+    return to_insert.empty() || coll.insert_many(to_insert)->result().inserted_count() != 0;
+}
+
+void load_initial_data(document::view test) {
+    if (!test["initialData"])
+        return;
+
+    auto data = test["initialData"].get_array().value;
+    std::all_of(std::begin(data), std::end(data), add_data_to_collection);
```
sgolemon, line 456: This is an odd way to use std::all_of(), particularly since you're not handling the failure case.  As soon as one element isn't inserted to the collection, the loop is going to stop.  I would think failure should be dealt with somehow, but if failure is permissible then surely skipping the failures and continuing on with the rest would be the better regression, no?  In that case, maybe use std::for_each() without that ignored bool return?

sgolemon (commented):
Doing my best to not pepper with "this should be const" and "why no braces?" since the local custom seems to be obeyed and it'd just be opinionated of me to complain.

Most of this looks 100% reasonable, but I do have a concern with the std::all_of() call in the initial data load at the bottom.  It smells of a bad failure mode.

src/mongocxx/test/spec/unified_test_format.cpp
sgolemon: Actually, just noticed std::for_each() is C++17 which you probably can't force at this point in time. Still the failure possibility should be addressed.

bazile-clyde: Ah, this should've been wrapped in a `REQUIRE` macro since no failures are allowed. I ended up just making `add_data_to_collection` return void and asserting the inserted count in that function to capture more info for debugging.

sgolemon (approved):
LGTM

PR #775, opened 2021-03-31
CXX-1871 run operations and assert results

bazile-clyde:
This runs all of the valid operations in the unified test format and asserts the outcomes, results, etc. The only failing test here is "Client side error in command starting transaction." That test erroneously progresses the transaction state after a client-side error. I remember a C driver ticket to investigate it, and _thought_ I pulled in those changes, but I'll double-check to see what the issue is. Other than that test, everything passes. 

All that's left is to pull in the valid-fail and invalid tests.

Edit: All tests now pass and `$$matchesEntity` and `$$sessionLsid` are implemented. Everything is good to go. To allow the matcher function to use the entity map during its assertions, I moved the matcher function and its dependencies to its own file. It was in `client_helpers.hh` which is included in almost every test. It's difficult for any function in `client_helpers.hh` to include any other test dependencies without creating a circular dependency.


src/mongocxx/client_session.hpp
```diff
@@ -90,8 +98,23 @@ class MONGOCXX_API client_session {
...
+    ///
+    /// @see
+    /// https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst
+    ///
+    bool get_dirty() const noexcept;
+
+    ///
+    /// Returns the current transaction state for this session.
+    ///
+    transaction_state get_transaction_state() const noexcept;
```
bazile-clyde, line 116: These were taken from libmongoc's public API. `get_transaction_state` seems like it should be public but `get_dirty` was only exposed in libmongoc for wrapping drivers I believe. I could try to see if we could keep `get_dirty` private.

src/mongocxx/test/CMakeLists.txt
```diff
@@ -164,8 +164,9 @@ add_executable(test_mongohouse_specs
 add_executable(test_unified_format_spec
     ${PROJECT_SOURCE_DIR}/test_util/client_helpers.cpp
     ${THIRD_PARTY_SOURCE_DIR}/catch/main.cpp
-    spec/unified_test_format.cpp
-    spec/entity.cpp
+    spec/unified_tests/runner.cpp
+    spec/unified_tests/entity.cpp
+    spec/unified_tests/operations.cpp
```
bazile-clyde, line 169: I put all of the unified test runner files in one folder.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -31,6 +33,33 @@ namespace spec {
 using namespace mongocxx;
 using bsoncxx::to_json;
 
+// commands postfixed with "_v2" are used to support the unified test format.
```
bazile-clyde, line 36: Since the unified test format lists event monitoring commands in camelcase, I could either: try to mangle the new logic in with the old or create new functions altogether. It was simpler to create new functions, and it should be easier to get rid of old logic.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -46,38 +75,71 @@ void apm_checker::compare(bsoncxx::array::view expectations,
...
     for (auto expectation : expectations) {
         auto expected = expectation.get_document().view();
-
-        if (events_iter == _events.end()) {
-            FAIL("Out of events, expected to find event " << to_json(expected) << "\n\n");
-            print_all();
-            REQUIRE(false);
-        }
-
-        CAPTURE(to_json(*events_iter), expectation);
```
bazile-clyde, line 56: This info is captured by `REQUIRE_BSON_MATCHES_V`.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -46,38 +75,71 @@ void apm_checker::compare(bsoncxx::array::view expectations,
...
         REQUIRE(std::find_if(_events.begin(), _events.end(), [&](bsoncxx::document::view doc) {
                     return test_util::matches(doc, expected);
                 }) != _events.end());
     }
 }
 
-void apm_checker::print_all() {
-    printf("\n\n");
-    printf("APM Checker contents: \n");
+std::string apm_checker::print_all() {
```
bazile-clyde, line 98: This function used `printf`, which causes issues with `CATCH2`. Since that framework maintains its own stack for output, this is always printed out of order, making it hard to find and piece together.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -136,9 +198,10 @@ apm_checker::event apm_checker::to_event(stdx::string_view s) {
         return apm_checker::event::kill_cursors;
     if (s.to_string() == "getMore")
         return apm_checker::event::get_more;
-    else
-        throw mongocxx::logic_error{error_code::k_invalid_parameter,
-                                    "unrecognized event {" + s.to_string() + "}"};
+    if (s.to_string() == "configureFailPoint")
+        return apm_checker::event::configure_fail_point;
+    throw mongocxx::logic_error{error_code::k_invalid_parameter,
```
bazile-clyde, line 203: No need for an else if the previous if statements all return when entered. My IDE was warning me about this.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -147,15 +210,21 @@ std::string apm_checker::to_string(event e) {
...
+            return "configureFailPoint";
     }
 }
 
 void apm_checker::set_ignore_command_monitoring_event(event e) {
     this->_ignore.push_back(to_string(e));
 }
 
-void apm_checker::clear() {
+void apm_checker::clear_events() {
```
bazile-clyde, line 222: This allows events to be cleared just before running operations without destroying the list of commands we plan to ignore.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,989 @@
+// Copyright 2020 MongoDB Inc.
```
bazile-clyde, line 1: The actual logic to run these operations was copied from `operation.cpp` with a few tweaks. The newer logic is for special commands (e.g., `assertDifferentLsidOnLastTwoCommands`) and commands we didn't implement in the past (e.g., `upload`). The `run` function at the bottom of this class does all of the switchings.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,18 +454,219 @@ void add_data_to_collection(const array::element& data) {
...
+    auto result = ops["expectResult"];
+    test_util::assert_matches(actual_result["result"].get_value(), result.get_value());
+
+    if (ops["saveResultAsEntity"]) {
+        auto key = ops["saveResultAsEntity"].get_string().value.to_string();
+        get_entity_map().insert(key, actual_result);
+    }
+}
+
+void assert_error(const mongocxx::operation_exception& e,
```
bazile-clyde, line 482: Operation exceptions (server errors) contain the raw server error as well (usually). This allows different logic from all other exceptions.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -300,6 +301,198 @@ bool is_numeric(types::bson_value::view value) {
            value.type() == type::k_double;
 }
 
+std::string to_string(types::bson_value::view_or_value val) {
```
bazile-clyde, line 304: This lets us `CAPTURE` BSON values which helped a ton when debugging.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -300,6 +301,198 @@ bool is_numeric(types::bson_value::view value) {
...
+        }
+
+        FAIL("type '" + to_string(actual_t) + "' expected in array '" +
+             to_json(op.get_array().value) + "'");
+    } else if (op.key().to_string() == "$$unsetOrMatches") {
+        auto val = op.get_value();
+        if (is_set(actual))
+            assert_matches(actual, val);
+    } else if (op.key().to_string() == "$$sessionLsid") {
+        // TODO: get sessions IDs from entity map
```
bazile-clyde, line 416: This is the only outstanding logic. I have to figure out a way to call back to the entity map to verify this special operation's values and the one below. Trying to pass the map through `assert_matches` doesn't work. The entity map depends on this, so if this also depended on the entity map, that would create a circular dependency.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -300,6 +301,198 @@ bool is_numeric(types::bson_value::view value) {
...
+    auto actual_arr = actual.get_array().value;
+    auto expected_arr = expected.get_array().value;
+
+    REQUIRE(size(actual_arr) == size(expected_arr));
+    for (auto a = actual_arr.begin(), e = expected_arr.begin(); e != expected_arr.end(); e++, a++) {
+        assert_matches(a->get_value(), e->get_value());
+    }
+}
+
+void assert_matches(types::bson_value::view actual, types::bson_value::view expected) {
```
bazile-clyde, line 473: This models the logic [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#evaluating-matches)

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -382,7 +575,7 @@ bool matches(types::bson_value::view main,
         array::view main_array = main.get_array().value;
         array::view pattern_array = pattern.get_array().value;
 
-        if (main_array.length() < pattern_array.length()) {
+        if (size(main_array) < size(pattern_array)) {
```
bazile-clyde, line 578: We should test the arrays have the same number of elements, not that they are the same length. This caused issues when I originally tried to use this function for the unified test format. I reverted most of the changes to this function after creating my own but left this since it's correct.

src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -317,7 +510,7 @@ bool matches(types::bson_value::view main,
              types::bson_value::view pattern,
              match_visitor visitor_fn) {
     if (auto t = is_type_operator(pattern)) {
-        return t == main.type() ? true : false;
+        return t == main.type();
```
bazile-clyde, line 513: Not sure why this was written the other way.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+        options.upsert(arguments["upsert"].get_bool().value);
+    }
+
+    auto result = builder::basic::document{};
+    stdx::optional<document::value> document;
+
+    switch (arguments["update"].type()) {
+        case bsoncxx::type::k_document: {
+            document::view update = arguments["update"].get_document().value;
+            document = coll.find_one_and_update(filter, update, options);
```
kevinAlbs, line 570: This should check if a session was passed (and the array case).

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+        if (arguments["hint"].type() == bsoncxx::v_noabi::type::k_string)
+            options.hint(hint{arguments["hint"].get_string().value});
+        else
+            options.hint(hint{arguments["hint"].get_document().value});
+    }
+
+    auto result = builder::basic::document{};
+    std::int32_t deleted_count = 0;
+
+    if (auto delete_result = coll.delete_one(filter, options)) {
```
kevinAlbs, line 809: This should check if a session was passed.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    }
+    if (name == "assertSessionTransactionState") {
+        auto session_name = op["arguments"]["session"].get_string().value.to_string();
+        auto& session = map.get_client_session(session_name);
+        return assert_session_transaction_state(session, op_view);
+    }
+    if (name == "dropCollection") {
+        auto coll_name = op["arguments"]["collection"].get_string().value.to_string();
+        auto& db = map.get_database(object);
+        db.collection(coll_name).drop();
```
kevinAlbs, line 939: This should check if a session was passed.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+        return download(bucket, op_view);
+    }
+    if (name == "upload") {
+        return upload(map, op_view);
+    }
+    if (name == "deleteOne") {
+        auto& coll = map.get_collection(object);
+        return delete_one(coll, op_view);
+    }
+
```
kevinAlbs, line 1002: When we test against sharded clusters, we'll need to add `targetedFailpoint`. But I think it is ok to leave as future work, as long as we fail with the "unsupported operation" exception, which should be clear.

kevinAlbs (commented):
Woo, nicely done!

Most of the comments are small. I think the biggest question is whether we need a separate apm_checker per client in case there are multiple client entities listening for events.

kevinAlbs:
I forgot to mention this in my review.

> All that's left is to pull in the valid-fail and invalid tests.

I vote for leaving those out of Evergreen. Neither C nor Go runs them as part of continuous testing. They are useful for manual validation. The reason we did not run them in Evergreen was that it would require additional workarounds to prevent the test runner from aborting on tests that were invalid / expected to fail.

bazile-clyde (commented):
> I forgot to mention this in my review.
> 
> > All that's left is to pull in the valid-fail and invalid tests.
> 
> I vote for leaving those out of Evergreen. Neither C nor Go runs them as part of continuous testing. They are useful for manual validation. The reason we did not run them in Evergreen was that it would require additional workarounds to prevent the test runner from aborting on tests that were invalid / expected to fail.

Got it. So this PR will be it then.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    std::int32_t deleted_count = 0;
+    std::int32_t matched_count = 0;
+    std::int32_t modified_count = 0;
+    std::int32_t upserted_count = 0;
+    result::bulk_write::id_map upserted_ids;
+    std::int32_t inserted_count = 0;
+    bsoncxx::stdx::optional<result::bulk_write> bulk_write_result;
+    bulk_write_result = coll.bulk_write(writes, opt);
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    if (arguments["ordered"]) {
+        insert_options.ordered(arguments["ordered"].get_bool().value);
+    }
+
+    for (auto&& element : documents) {
+        documents_to_insert.push_back(element.get_document().value);
+    }
+
+    stdx::optional<result::insert_many> insert_many_result;
+    insert_many_result = coll.insert_many(documents_to_insert, insert_options);
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    if (arguments["upsert"]) {
+        options.upsert(arguments["upsert"].get_bool().value);
+    }
+
+    std::int32_t matched_count = 0;
+    bsoncxx::stdx::optional<std::int32_t> modified_count;
+    std::int32_t upserted_count = 0;
+    stdx::optional<result::replace_one> replace_result;
+
+    replace_result = coll.replace_one(filter, replacement, options);
```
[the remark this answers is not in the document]
bazile-clyde: Done

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    if (arguments["batchSize"]) {
+        options.batch_size(arguments["batchSize"].get_int32().value);
+    }
+
+    if (arguments["collation"]) {
+        options.collation(arguments["collation"].get_document().value);
+    }
+
+    stdx::optional<cursor> result_cursor;
+    result_cursor.emplace(entity.aggregate(pipeline, options));
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -0,0 +1,1007 @@
...
+    if (args["batchSize"]) {
+        options.batch_size(args["batchSize"].get_int32().value);
+    }
+
+    auto key = operation["saveResultAsEntity"].get_string().value.to_string();
+    CAPTURE(object, to_json(operation), key);
+    auto stream = [&] {
+        const auto& type = map.type(object);
+        if (type == typeid(mongocxx::database))
+            return map.get_database(object).watch(pipeline, options);
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/operations.cpp
bazile-clyde: Done

src/mongocxx/test/spec/unified_tests/operations.cpp
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/operations.cpp
bazile-clyde: Done

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,18 +473,220 @@ void add_data_to_collection(const array::element& data) {
...
+                  const array::element& ops,
+                  document::view res) {
```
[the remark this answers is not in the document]
bazile-clyde: These functions were supposed to correspond to the spec sections "expectError" and "expectResult" but I can change it if you think `assert_result` for all of the function names would be clearer.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,18 +473,220 @@ void add_data_to_collection(const array::element& data) {
...
+            if (!has_run_on_requirements(ele.get_document())) {
+                std::stringstream warning;
+                warning << "test skipped: "
+                        << "none of the runOnRequirements were met" << std::endl
+                        << to_json(ele["runOnRequirements"].get_array().value);
+                WARN(warning.str());
+                continue;
+            }
+
+            REQUIRE_FALSE(/* TODO */ ele["skipReason"]);
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,18 +473,220 @@ void add_data_to_collection(const array::element& data) {
...
+
+            disable_fail_point disable_fail_point_fn{};
+            get_apm_checker().clear_events();
+            for (auto ops : ele["operations"].get_array().value) {
+                try {
+                    auto result = bsoncxx::builder::basic::make_document();
+                    result = operations::run(get_entity_map(), get_apm_checker(), ops);
+
+                    if (ops["object"].get_string().value.to_string() == "testRunner") {
```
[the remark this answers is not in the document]
bazile-clyde: Done.

src/mongocxx/test_util/client_helpers.cpp
bazile-clyde: Yeah. If `length`s are equal that doesn't imply that the `size`s is equal as well. I'm not really sure why we need `length`. Maybe to check to make sure you're under the 16MB limit?

bazile-clyde:
There are only a couple more outstanding comments, but I thought I'd get what I have so far back in review. Most of the requested changes are separated by commits as well.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -29,10 +38,18 @@ using namespace mongocxx;
...
 using namespace spec;
 
+using bsoncxx::builder::basic::kvp;
+using bsoncxx::builder::basic::make_document;
+
 using schema_versions_t =
     std::array<std::array<int, 3 /* major.minor.patch */>, 1 /* supported version */>;
 constexpr schema_versions_t schema_versions{{{1, 0, 0}}};
 
+spec::apm_checker& get_apm_checker() {
```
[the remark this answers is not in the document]
bazile-clyde: Created an APM map that maps client IDs to APM checkers. Also created a function `init_maps` to control the maps' construction and destruction since there is a dependency between the two. The comment in that function gives more detail.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,18 +473,220 @@ void add_data_to_collection(const array::element& data) {
...
+
+    if (expect_error["isError"])
+        return;
+
+    if (auto is_client_error = expect_error["isClientError"]) {
+        REQUIRE(is_client_error.get_bool());
+    }
+
+    // below is only used for server-side errors.
+    REQUIRE_FALSE(expect_error["errorLabelsContain"]);
```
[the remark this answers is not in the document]
bazile-clyde: Done.

bazile-clyde (commented):
All comments should be addressed. I also cherry-picked the workflow commit so this PR's Github action is run in debug mode.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -15,12 +15,21 @@
 #include <fstream>
 #include <regex>
```
kevinAlbs, line 16: Compiling locally, I got this error:
```
../src/mongocxx/test/spec/unified_tests/runner.cpp:443:27: error: implicit instantiation of undefined template 'std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
        std::stringstream v_str;
```

Adding `#include <sstream>` fixed it.

src/mongocxx/client_session.hpp
kevinAlbs: Ah, I missed this comment on first review. While I agree that `get_dirty` does not seem useful to users, I don't see a great alternative for accessing the underlying `mongoc_client_session_t` from tests. These were a couple alternatives I thought of:

1. Keep this public but mark it with the `MONGOCXX_PRIVATE` macro. There is some precedent for that in the APM events:
https://github.com/mongodb/mongo-cxx-driver/blob/a46914ade52a1fb63e79207250458c81da513bf0/src/mongocxx/events/heartbeat_started_event.hpp#L35:L35 That marks it as "linker private" so consumers cannot actually link to the method from their code. However, that also applies to the test executable, so that doesn't work.
2. Mark this as a friend of a class that the test runner can utilize. But including test classes in public headers seems like a dealbreaker.

`get_dirty` isn't useful to users, but if we have no reasonable alternative, then so be it. Let's keep it as-is.

src/mongocxx/test/spec/monitoring.cpp
kevinAlbs: Great, that level of detail will probably save countless hours.

src/mongocxx/test_util/client_helpers.cpp
kevinAlbs: I'm not sure. It could have also been a mistake in the original API.

src/mongocxx/test/spec/unified_tests/runner.cpp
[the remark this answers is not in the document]
kevinAlbs: I don't feel strongly about it. For test comparisons like this I like the `expected` / `actual` naming. I am OK with leaving it if you find this clearer.

kevinAlbs (approved):
LGTM! I only have a couple remaining nits.

Feel free to open another PR to merge `CXX-1871` after merging this. If there are any test issues, I can help with debugging.

src/mongocxx/test/spec/unified_tests/runner.cpp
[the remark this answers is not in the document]
bazile-clyde: Done.

PR #776, opened 2021-04-01
CXX-2111 Support Azure/GCP KMS providers

kevinAlbs:
Summary of changes:
- Supports new Azure / GCP spec and prose tests.
- Updates the minimum C driver required to 1.18.0. The C driver 1.18.0, once released, will require libmongocrypt 1.1.0. libmongocrypt 1.1.0 contains Azure / GCP support.
- Supports the new tests for supporting AWS temporary credentials (CXX-2172).
- Skips captured listCollections events on CSFLE spec tests. The issue described in CXX-2155 is not applicable to the C / C++ drivers.

Spec test updates are on the first commit. You can filter them out when reviewing.


docs/content/mongocxx-v3/client-side-encryption.md
```diff
@@ -71,11 +71,7 @@ JSON Schemas supplied in the `schema_map` only apply to configuring automatic cl
 //         "bsonType" : "object",
 //         "properties" : {
 //            "encryptedFieldName" : {
-<<<<<<< HEAD
```
kevinAlbs, line 74: This is a remnant of a prior merge mishap.

src/mongocxx/test/spec/client_side_encryption.cpp
```diff
@@ -267,12 +327,14 @@ TEST_CASE("Client side encryption spec automated tests", "[client_side_encryptio
 
     std::string test_file;
     while (std::getline(test_files, test_file)) {
-        if (std::find(unsupported_tests.begin(), unsupported_tests.end(), test_file) !=
-            unsupported_tests.end()) {
-            WARN("skipping " << test_file << " due to unsupported operation");
-            continue;
+        SECTION(test_file) {
```
kevinAlbs, line 330: As @bazile-clyde taught me, wrapping this in a `SECTION` makes it much easier to run individual tests. For example, to run just `bypassAutoEncryption.json`:

```
./cmake-build/src/mongocxx/test/test_client_side_encryption_specs --section "bypassAutoEncryption.json"
```

PR #778, opened 2021-04-07
allow underscore and capitals in Github token

bazile-clyde:
This script throws the error "No Github token found in file 'github_token.txt'" if the file contains a token with underscores and/or capital letters. This commit fixes that issue.

### DIFF
--- a/etc/make_release.py
+++ b/etc/make_release.py
@@ -524,7 +524,7 @@ def read_github_creds(github_token_file):
     Read the GitHub token from the specified file and return it as a string.
     """
 
-    token_re = re.compile('^(?:Token - )?(?P<tok>[0-9a-f]{40}).*$')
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Ping @rcsanchez97, is this something we should merge or close?

rcsanchez97:
@kevinAlbs Strangely, I have no recollection of this PR.  Even more
strangely, I have no recollection of #779, which is the same change for
the `releases/v3.6` branch, which I apparently closed without merging.

I think that Clyde requested my review since I was the original author
of the `make_release.py` script.

The change seems OK to me, though I don't execute the actual release
process, so is it OK if I toss this back to you?  I have no objection to
the change, but you might make a note in the release checklist that this
change was made so that whomever makes future releases has a clue if
something related to this goes wrong.

kevinAlbs (approved):
I ran a quick test against my GitHub release token and LGTM.

PR #782, opened 2021-04-14
CXX-1646 Include server replies in all exceptions

benjirewis:
[CXX-1646](https://jira.mongodb.org/browse/CXX-1646)

We want to include a server reply in all thrown exceptions where possible. After a small audit, I've found only one potential instance of an unreturned raw server response after an error, and even this one is arguably unimportant.

I've looked through all calls of `throw_exception` and all creations of `operation_exception` and its implementers: `authentication_exception`, `bulk_write_exception`, `query_exception`, and `write_exception`. I'm fairly certain all exceptions created have no relevant server response to include (many wrap libmongoc functions that do not return the raw server reply) or already include a raw server response (`steal()`ing the BSON error document).

### DIFF
--- a/src/mongocxx/collection.cpp
+++ b/src/mongocxx/collection.cpp
@@ -1149,13 +1149,18 @@ cursor collection::_distinct(const client_session* session,
         throw bsoncxx::exception{bsoncxx::error_code::k_internal_error};
     }
 
-    scoped_bson_t error_document;
-    const bson_t *error_doc = error_document.bson_for_init();
+    const bson_t* error_document;
 
     cursor fake_cursor{libmongoc::cursor_new_from_command_reply_with_opts(
         _get_impl().client_impl->client_t, reply_bson, nullptr)};
-    if (libmongoc::cursor_error_document(fake_cursor._impl->cursor_t, &error, &error_doc)) {
-        throw_exception<operation_exception>(error_document.steal(), error);
+    if (libmongoc::cursor_error_document(fake_cursor._impl->cursor_t, &error, &error_document)) {
+        if (error_document) {
+            bsoncxx::document::value error_doc{
+                bsoncxx::document::view{bson_get_data(error_document), error_document->len}};
+            throw_exception<query_exception>(error_doc, error);
+        } else {
+            throw_exception<query_exception>(error);
+        }
     }
 
     return fake_cursor;
--- a/src/mongocxx/collection.cpp
+++ b/src/mongocxx/collection.cpp
@@ -1149,18 +1149,13 @@ cursor collection::_distinct(const client_session* session,
         throw bsoncxx::exception{bsoncxx::error_code::k_internal_error};
     }
 
-    const bson_t* error_document;
+    scoped_bson_t error_document;
+    const bson_t *error_doc = error_document.bson_for_init();
 
     cursor fake_cursor{libmongoc::cursor_new_from_command_reply_with_opts(
         _get_impl().client_impl->client_t, reply_bson, nullptr)};
-    if (libmongoc::cursor_error_document(fake_cursor._impl->cursor_t, &error, &error_document)) {
-        if (error_document) {
-            bsoncxx::document::value error_doc{
-                bsoncxx::document::view{bson_get_data(error_document), error_document->len}};
-            throw_exception<operation_exception>(error_doc, error);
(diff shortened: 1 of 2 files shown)


src/mongocxx/collection.cpp
```diff
@@ -1149,10 +1149,18 @@ cursor collection::_distinct(const client_session* session,
         throw bsoncxx::exception{bsoncxx::error_code::k_internal_error};
     }
 
+    const bson_t* error_document;
+
     cursor fake_cursor{libmongoc::cursor_new_from_command_reply_with_opts(
         _get_impl().client_impl->client_t, reply_bson, nullptr)};
-    if (libmongoc::cursor_error(fake_cursor._impl->cursor_t, &error)) {
-        throw_exception<operation_exception>(error);
+    if (libmongoc::cursor_error_document(fake_cursor._impl->cursor_t, &error, &error_document)) {
```
benjirewis, line 1156: For context, `cursor_error_document` is being called on a _fake_ cursor created for `distinct`. I don't _think_ it's even possible to have an error in the cursor here, as a legitimate error from sending `distinct` to the server would have surfaced a few blocks earlier...

src/mongocxx/collection.cpp
```diff
@@ -1149,10 +1149,18 @@ cursor collection::_distinct(const client_session* session,
         throw bsoncxx::exception{bsoncxx::error_code::k_internal_error};
     }
 
+    const bson_t* error_document;
```
bazile-clyde, line 1152: TLDR; Here, you could use a `scoped_bson_t` instead of a pointer to a `bson_t`.

Because the lifetime of objects is less obvious in C++ than C we opt to use classes that manage their own life rather than pointers, which we later have to remember to clean up.  

For example, if an exception is thrown between a heap-allocated object's creation and a call to some destructor function then the destructor function will never get called and we'll leak memory. C functions don't throw exceptions but most C++ functions do.

benjirewis: Sounds good! Thanks for the explanation here and offline. `cursor_error_document` does require a `**bson_t`, so I used `bson_for_init` to make another `error_doc` variable first; let me know if there's a better way.

src/mongocxx/collection.cpp
kevinAlbs: I agree that it isn't possible looking at the implementation of `mongoc_cursor_new_from_command_reply_with_opts`. Arguably it should set an error document reply if the reply was a server error, so I think this is still an improvment.

src/mongocxx/collection.cpp
```diff
@@ -1149,10 +1149,13 @@ cursor collection::_distinct(const client_session* session,
         throw bsoncxx::exception{bsoncxx::error_code::k_internal_error};
     }
 
+    scoped_bson_t error_document;
+    const bson_t *error_doc = error_document.bson_for_init();
```
kevinAlbs, line 1153: `scoped_bson_t` is useful for cases where we are initializing a `bson_t` and need to destroy it on scope exit (similar to why defer statement in Go), since it destroys in the destructor.

But I believe in this case, where `mongoc_cursor_error_document` is giving us a pointer to a `const bson_t*`, we don't want to call `bson_destroy` on the `bson_t` it gives us. The cursor manages the lifetime of that `bson_t`. The `const bson_t *` is a non-owning view of it.

I suggest removing `scoped_bson_t` and simplifying this to:

```
const bson_t *error_doc;
```

kevinAlbs (commented):
There are three cases in index_view.hh as well I think can accept a reply:
https://github.com/mongodb/mongo-cxx-driver/blob/8ffd856d0d3d4b112f2e85c9b33f62198fefa7af/src/mongocxx/private/index_view.hh#L166:L166

src/mongocxx/collection.cpp
benjirewis: Gotcha; switched back to just `const bson_t*` since the cursor is managing the lifetime of the error document.

benjirewis:
Good find in `index_view.hh`... I was only searching in cpp files, forgot about the private headers. Those three cases should `throw_exception` with a stolen reply now.

kevinAlbs (approved):
LGTM! The macOS failures are from the host not having boost available (BUILD-13061).

PR #784, opened 2021-04-21
CXX-2230 Upgrade unified test runner to version 1.1

benjirewis:
CXX-2230

Bumps supported UTF schema version to 1.1.0. Allows `serverParameters` in `runOnRequirement`. Drafts test runner changes for defining `serverApi` on a `client` entity (can't actually implement this until CXX-2138). Adds a dozen missing functions to supported UTF operations.

Tested new features above against new versioned API spec tests. `serverParameters` seem to skip correctly and new operations appear to run without error.

### DIFF
--- a/src/mongocxx/test/spec/unified_tests/operations.cpp
+++ b/src/mongocxx/test/spec/unified_tests/operations.cpp
@@ -906,6 +906,121 @@ document::value run_command(database& db, document::view operation) {
     return db.run_command(command);
 }
 
+document::value update_one(collection& coll, document::view operation) {
+    document::view arguments = operation["arguments"].get_document().value;
+    document::view filter = arguments["filter"].get_document().value;
+    document::view update = arguments["update"].get_document().value;
+    options::update options{};
+
+    if (arguments["collation"]) {
+        options.collation(arguments["collation"].get_document().value);
+    }
+    if (arguments["hint"]) {
+        if (arguments["hint"].type() == bsoncxx::v_noabi::type::k_string)
+            options.hint(hint{arguments["hint"].get_string().value});
+        else
+            options.hint(hint{arguments["hint"].get_document().value});
+    }
+    if (arguments["upsert"]) {
+        options.upsert(arguments["upsert"].get_bool().value);
+    }
+
+    std::int32_t matched_count = 0;
+    bsoncxx::stdx::optional<std::int32_t> modified_count;
+    bsoncxx::stdx::optional<types::bson_value::view> upserted_id{};
+
+    auto update_one_result = coll.update_one(filter, update, options);
+    if (update_one_result) {
+        matched_count = update_one_result->matched_count();
+
+        // Server versions below 2.4 do not return an `nModified` count.
+        try {
+            modified_count = update_one_result->modified_count();
+        } catch (const std::exception& e) {
+        }
+
+        if (auto upserted_element = update_one_result->upserted_id()) {
+            upserted_id = upserted_element->get_value();
+        }
+    }
+
+    auto result = builder::basic::document{};
+    result.append(builder::basic::kvp(
+        "result",
+        [matched_count, modified_count, upserted_id](builder::basic::sub_document subdoc) {
+            subdoc.append(builder::basic::kvp("matchedCount", matched_count));
+
+            if (modified_count) {
+                subdoc.append(builder::basic::kvp("modifiedCount", *modified_count));
+            }
+
+            if (upserted_id) {
+                subdoc.append(builder::basic::kvp("upsertedId", *upserted_id));
+            }
+        }));
+
+    return result.extract();
+}
+
+document::value update_many(collection& coll, document::view operation) {
+    document::view arguments = operation["arguments"].get_document().value;
+    document::view filter = arguments["filter"].get_document().value;
+    document::view update = arguments["update"].get_document().value;
+    options::update options{};
+
+    if (arguments["collation"]) {
+        options.collation(arguments["collation"].get_document().value);
+    }
+    if (arguments["hint"]) {
+        if (arguments["hint"].type() == bsoncxx::v_noabi::type::k_string)
+            options.hint(hint{arguments["hint"].get_string().value});
+        else
+            options.hint(hint{arguments["hint"].get_document().value});
+    }
+    if (arguments["upsert"]) {
+        options.upsert(arguments["upsert"].get_bool().value);
+    }
+
+    std::int32_t matched_count = 0;
+    bsoncxx::stdx::optional<std::int32_t> modified_count;
+    std::int32_t upserted_count = 0;
+    bsoncxx::stdx::optional<types::bson_value::view> upserted_id{};
+
+    auto update_many_result = coll.update_many(filter, update, options);
+    if (update_many_result) {
+        matched_count = update_many_result->matched_count();
+
+        // Server versions below 2.4 do not return an `nModified` count.
+        try {
+            modified_count = update_many_result->modified_count();
+        } catch (const std::exception& e) {
+        }
+
+        upserted_count = update_many_result->result().upserted_count();
+
+        if (auto upserted_element = update_many_result->upserted_id()) {
+            upserted_id = upserted_element->get_value();
+        }
+    }
+
+    auto result = builder::basic::document{};
+    result.append(builder::basic::kvp(
+        "result",
+        [matched_count, modified_count, upserted_id](builder::basic::sub_document subdoc) {
+            subdoc.append(builder::basic::kvp("matchedCount", matched_count));
+
+            if (modified_count) {
+                subdoc.append(builder::basic::kvp("modifiedCount", *modified_count));
+            }
+
+            if (upserted_id) {
+                subdoc.append(builder::basic::kvp("upsertedId", *upserted_id));
+            }
+        }));
+
+    return result.extract();
+}
+
 document::value operations::run(entity::map& entity_map,
                                 std::unordered_map<std::string, spec::apm_checker>& apm_map,
                                 const array::element& op) {
@@ -1005,6 +1120,11 @@ document::value operations::run(entity::map& entity_map,
         session.commit_transaction();
         return empty_doc;
     }
+    if (name == "abortTransaction") {
+        auto& session = entity_map.get_client_session(object);
+        session.abort_transaction();
+        return empty_doc;
+    }
     if (name == "assertSessionTransactionState") {
         auto session_name = string::to_string(op["arguments"]["session"].get_string().value);
         auto& session = entity_map.get_client_session(session_name);
@@ -1084,6 +1204,12 @@ document::value operations::run(entity::map& entity_map,
     if (name == "runCommand") {
         auto& db = entity_map.get_database(object);
         return run_command(db, op_view);
+    }
+    if (name == "updateOne") {
+        return update_one(entity_map.get_collection(object), op_view);
+    }
+    if (name == "updateMany") {
+        return update_many(entity_map.get_collection(object), op_view);
     }
 
     throw std::logic_error{"unsupported operation: " + name};
--- a/src/mongocxx/test/spec/unified_tests/operations.cpp
+++ b/src/mongocxx/test/spec/unified_tests/operations.cpp
@@ -976,6 +976,43 @@ document::value delete_one(collection& coll, client_session* session, document::
     return result.extract();
 }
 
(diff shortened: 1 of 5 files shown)


src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -553,6 +553,112 @@ document::value fail_point(entity::map& map, spec::apm_checker& apm, document::v
...
+    }
+    if (arguments["sort"]) {
+        options.sort(arguments["sort"].get_document().value);
+    }
+    if (arguments["upsert"]) {
+        options.upsert(arguments["upsert"].get_bool().value);
+    }
+
+    stdx::optional<document::value> document;
+    switch (arguments["replacement"].type()) {
```
benjirewis, line 636: Could use an `if` here; unsure what's idiomatic in cpp.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -899,6 +1042,209 @@ client_session* get_session(document::view op, entity::map& map) {
...
+            options.hint(hint{arguments["hint"].get_string().value});
+        else
+            options.hint(hint{arguments["hint"].get_document().value});
+    }
+
+    int64_t count = coll.count_documents(filter, options);
+    auto result = builder::basic::document{};
+    result.append(builder::basic::kvp("result", [count](builder::basic::sub_document subdoc) {
+        subdoc.append(builder::basic::kvp("count", count));
+    }));
```
benjirewis, line 1198: I build a document `{"result": {"count": count}}` here and in `estimated_document_count` instead of returning an `int64_t` since `operations::run` always returns a `document::value`.

benjirewis (commented):
Let me know if anything doesn't look like idiomatic CXX or diverges from CXX driver style! Tried to copy existing patterns where I could.

src/mongocxx/test/spec/unified_tests/operations.cpp
kevinAlbs: The [ISO C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-switch-if), which Clyde has linked to in the past, recommends switch when there is an option. Compilers can also warn against unhandled cases.

For one case, I'd naturally do an `if/else`. But I don't have a strong preference.

In this particular case, since we only care that `replacement` is a document, I don't think an `if/else` or `switch` is required. `get_document` will throw an exception if `replacement` is not a document type.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -553,6 +553,112 @@ document::value fail_point(entity::map& map, spec::apm_checker& apm, document::v
...
+            } else {
+                document = coll.find_one_and_replace(filter, replacement, options);
+            }
+            break;
+        }
+        default:
+            throw std::logic_error{"replacement must be a document"};
+    }
+
+    // Server versions below 3.0 sometimes return an empty document rather than null when no
```
kevinAlbs, line 650: Same comment about this not being specific to the server version.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -870,6 +976,43 @@ document::value delete_one(collection& coll, client_session* session, document::
...
+        if (arguments["hint"].type() == bsoncxx::v_noabi::type::k_string)
+            options.hint(hint{arguments["hint"].get_string().value});
+        else
+            options.hint(hint{arguments["hint"].get_document().value});
+    }
+
+    auto result = builder::basic::document{};
+    std::int32_t deleted_count = 0;
+
+    auto delete_result = [&] {
```
kevinAlbs, line 998: Minor: is there a reason to use a lambda? Since `delete_result` is a `stdx::optional` I think it can be default initialized, so this this could follow the pattern of the `find_one_and_*` helpers:

```
stdx::optional<result::delete_result> delete_result;
if (session) {
	delete_result = ...
} else {
	delete_result = ...
}
```

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -899,6 +1042,212 @@ client_session* get_session(document::view op, entity::map& map) {
     return &map.get_client_session(session_name);
 }
 
+document::value run_command(database& db, document::view operation) {
+    document::view arguments = operation["arguments"].get_document().value;
+    document::view command = arguments["command"].get_document().value;
+
+    return db.run_command(command);
```
kevinAlbs, line 1049: `assert_result` assumes that operations return results in a document with the key `result`. Arguably, results should be in its own class. But for the scope of this PR, I think it should follow that pattern to avoid confusing future test failures, and return a document `{"result": <what run_command returns> }`.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -899,6 +1042,212 @@ client_session* get_session(document::view op, entity::map& map) {
...
+
+    std::int32_t matched_count = 0;
+    bsoncxx::stdx::optional<std::int32_t> modified_count;
+    bsoncxx::stdx::optional<types::bson_value::view> upserted_id{};
+
+    auto update_one_result = coll.update_one(filter, update, options);
+    if (update_one_result) {
+        matched_count = update_one_result->matched_count();
+
+        // Server versions below 2.4 do not return an `nModified` count.
```
kevinAlbs, line 1079: I'm not sure about this either. But as of C++ driver 3.3, only 3.0+ is supported: https://docs.mongodb.com/drivers/cxx/#mongodb-compatibility). I suggest removing the `try/catch`. If tests do fail, I bet it is the same culprit (unacknowledged writes).

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -899,6 +1042,212 @@ client_session* get_session(document::view op, entity::map& map) {
...
+    std::int32_t matched_count = 0;
+    bsoncxx::stdx::optional<std::int32_t> modified_count;
+    std::int32_t upserted_count = 0;
+    bsoncxx::stdx::optional<types::bson_value::view> upserted_id{};
+
+    auto update_many_result = coll.update_many(filter, update, options);
+    if (update_many_result) {
+        matched_count = update_many_result->matched_count();
+
+        // Server versions below 2.4 do not return an `nModified` count.
```
kevinAlbs, line 1136: Same comment, I suggest removing this try/catch.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -947,9 +1296,19 @@ document::value operations::run(entity::map& entity_map,
...
+        return find_one_and_delete(
+            entity_map.get_collection(object), get_session(op_view, entity_map), op_view);
+    if (name == "findOneAndReplace")
+        return find_one_and_replace(
+            entity_map.get_collection(object), get_session(op_view, entity_map), op_view);
     if (name == "findOneAndUpdate")
         return find_one_and_update(
             entity_map.get_collection(object), get_session(op_view, entity_map), op_view);
+    if (name == "listCollections") {
+        entity_map.get_database(object).list_collections().begin();
```
kevinAlbs, line 1309: The implementation of C's `mongoc_database_find_collections_with_opts` is one of the few non-lazy cursors, so this will guarantee that `listCollections` gets run.

Assuming there are no tests that inspect the `listCollections` results, this should be OK. We can add to this later when needed.

kevinAlbs (commented):
Looking great!

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -553,6 +553,112 @@ document::value fail_point(entity::map& map, spec::apm_checker& apm, document::v
...
+    }
+
+    stdx::optional<document::value> document;
+    if (session) {
+        document = coll.find_one_and_delete(*session, filter, options);
+    } else {
+        document = coll.find_one_and_delete(filter, options);
+    }
+
+    // Server versions below 3.0 sometimes return an empty document rather than null when no
```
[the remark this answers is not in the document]
benjirewis: Thanks so much for looking into this! Your explanation and repro make sense to me, but is this a bug in the C++ or C driver? Should I link to some ticket for a `TODO` to remove the logic around expecting an empty document instead of `null`? Happy to file a ticket.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -553,6 +553,112 @@ document::value fail_point(entity::map& map, spec::apm_checker& apm, document::v
...
+    if (arguments["returnDocument"]) {
+        auto return_document = string::to_string(arguments["returnDocument"].get_string().value);
+
+        if (return_document == "After") {
+            options.return_document(options::return_document::k_after);
+        }
+
+        if (return_document == "Before") {
+            options.return_document(options::return_document::k_before);
+        }
```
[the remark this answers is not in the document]
benjirewis: Yep! Moved `Before` to an `else if` and added an `else` with that error.

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Gotcha, similar to Go then; the explanation they give about exhausting an enum whenever possible makes sense. Removed the `switch` here though as `get_document` should catch the relevant error.

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Sounds good to me; I just copied from `delete_one`. Technically one line shorter haha, but I do find the normal `if else` easier to read. Switched.

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Ah gotcha, returned a `{"result": ...}` document instead!

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Removed the `try/catch` and tests seem to pass!

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Removed.

src/mongocxx/test/spec/unified_tests/operations.cpp
benjirewis: Great, if we need to examine the `listCollections` results for versioned API, I can just add this in later.

src/mongocxx/test/spec/unified_tests/operations.cpp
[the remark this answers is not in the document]
kevinAlbs: I think the C driver should document that the reply is always initialized to the empty document for unacknowledged writes, (as it does for bulk writes). Created https://github.com/mongodb/mongo-c-driver/pull/789.

I think it is arguable whether the C++ driver should be changed to return a `std::nullopt` instead of an empty document for an unacknowledged write. And I think I'm in favor of not changing it.
- Right now, there is no ambiguity. `std::nullopt` unambiguously means that nothing was matched. For an unacknowledged write, there may or may not have been a match, but the empty document unambiguously signals that there is no server response.
- Changing it is a slight backwards break in the return.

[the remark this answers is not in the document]
kevinAlbs: Can you update this comment, and the others, to note that this only applies to unacknowledged writes (and not the server version)?
```suggestion
    // Using an unacknowledged write returns an empty document rather than null when no
```

kevinAlbs (approved):
LGTM with comment updates!

src/mongocxx/test/spec/unified_tests/operations.cpp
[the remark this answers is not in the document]
benjirewis: Gotcha. The documentation for the C driver looks good to me, and I agree that changing the CXX driver to always return `std::nullopt` is backwards-breaking and not really necessary. Updated comments.

benjirewis (commented):
Changed the returned document of `estimated_document_count` and `count_documents` slightly and added functionality for an expected `errorCode`. Small changes necessary for CXX-2193.

kevinAlbs (approved):
LGTM! I created CXX-2237 to track the test failure on the valgrind Ubuntu 16.04 variant (which is unrelated).

PR #785, opened 2021-04-21
CXX-2234 Expect spaces after base64 in extJSON binary

benjirewis:
CXX-2234

Adds spaces to two extended JSON tests that expect an output with "base64".

This should right a failure caused by the bug fix [CDRIVER-3958](https://jira.mongodb.org/browse/CDRIVER-3958). There are no other occurrences of `"base64":` in the CXX driver repo outside of spec test JSON.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -135,7 +135,7 @@ functions:
                 export DRIVERS_TOOLS=$(pwd)
 
                 sh .evergreen/atlas_data_lake/run-mongohouse-local.sh
-              
+
     "test_mongohouse":
         command: shell.exec
         params:
@@ -157,8 +157,8 @@ functions:
 
                 ulimit -c unlimited || true
 
-                ./src/mongocxx/test/test_mongohouse_specs             
-              
+                ./src/mongocxx/test/test_mongohouse_specs
+
     "start_mongod":
         command: shell.exec
         params:
@@ -353,6 +353,9 @@ functions:
                   export MONGOCXX_TEST_AWS_ACCESS_KEY_ID="${cse_aws_access_key_id}"
                   set -o errexit
                   set -o xtrace
+
+                  # export MONGOC_VERSION for `test_bson` test binary.
+                  export MONGOC_VERSION="${mongoc_version|master}"
 
                   if [ "Windows_NT" == "$OS" ]; then
                       CRUD_TESTS_PATH=$(cygpath -m $CRUD_TESTS_PATH)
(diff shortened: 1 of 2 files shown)


benjirewis:
This patch now causes failures on the "Ubuntu 18.04 with minimum libmongoc (MongoDB Latest)" task since the minimum libmongoc version is still 1.17.0-beta2. I think we'll have to bump the minimum to whatever C driver version contains the [CDRIVER-3958](https://jira.mongodb.org/browse/CDRIVER-3958) fix.

kevinAlbs (approved):
Thanks for fixing this!

.mci.yml
```diff
@@ -354,6 +354,10 @@ functions:
                   set -o errexit
                   set -o xtrace
 
+                  # TODO CXX-2227: Remove export after minimum libmongoc version is bumped.
+                  # export MONGOC_VERSION for `test_bson` test binary.
+                  export MONGOC_VERSION="${mongoc_version|master}"
```
benjirewis, line 359: I added this env. variable to condition the two tests on the underlying libmongoc version. We can eventually remove this logic on any branch with libmongoc minimum version >= 1.18.0 (CXX-2227).

src/bsoncxx/test/json.cpp
```diff
@@ -93,24 +94,58 @@ TEST_CASE("CXX-1246: Legacy Extended JSON (Explicit)") {
...
-    REQUIRE(
-        output ==
-        R"({ "number" : 42, "bin" : { "$binary" : { "base64": "ZGVhZGJlZWY=", "subType" : "04" } } })");
+
+    // TODO CXX-2227: Remove conditional result after minimum libmongoc version is bumped.
+    //
+    // As of libmongoc 1.18.0, "base64" has correct spacing (see CDRIVER-3958) after extJSON
+    // marshalling.
+    const char* expected;
+    if (std::string(mongoc_version).compare("1.18.0") >= 0) {
```
kevinAlbs, line 113: Can you change the comparison to use the integer macros exposed in `mongoc-version.h`?

E.g. `if ( (MONGOC_MAJOR_VERSION == 1 && MONGOC_MINOR_VERSION >= 18) || MONGOC_MAJOR_VERSION > 1)`

`mongoc-version.h` should be included through `mongoc.h` already.

A string comparison won't do a lexicographic compare.

benjirewis: Discussed offline. We actually want to `BSON_VERSION` macros from `bson-version.h` since this is in `bsoncxx`. Used those instead of the string comparison and environment variable (changes to `.mci.yml` removed).

kevinAlbs (approved):
LGTM!

PR #787, opened 2021-04-29
CXX-2193 Sync new estimatedDocumentCount tests

benjirewis:
CXX-2193

Adds new tests for estimatedDocumentCount on 4.9.0+ (edc now uses the `$collStats` aggregation state instead of the `count` command on 4.9.0+). No code changes needed, as the underlying C driver has already been updated in [CDRIVER-3905](https://jira.mongodb.org/browse/CDRIVER-3905). First commit updates spec tests, second commit has changes to CRUD and unified spec test runners.

Moves `crud/v1` and `crud/v2` tests to `crud/legacy/v1` and `crud/legacy/v2` respectively and adds a new `crud/unified` test folder.


.mci.yml
```diff
@@ -333,9 +333,11 @@ functions:
                   export DYLD_LIBRARY_PATH=.:$PREFIX/lib/
                   export PATH=$(pwd)/src/bsoncxx/${build_type}:$(pwd)/src/mongocxx/${build_type}:$PREFIX/bin:$(pwd)/install/bin:$PATH
 
-                  # CRUD_TESTS_PATH and GRIDFS_TESTS_PATH are required to be set by the
-                  # `test_crud_specs` and `test_gridfs_specs` test binaries, respectively.
-                  export CRUD_TESTS_PATH="$(pwd)/../data/crud"
+                  # CRUD_LEGACY_TESTS_PATH, CRUD_UNIFIED_TESTS_PATH and GRIDFS_TESTS_PATH are required to
+                  # be set by the `test_crud_specs`, `test_unified_format_spec` and `test_gridfs_specs` test
+                  # binaries respectively.
```
benjirewis, line 338: Is this comment necessary? All of the below `PATH` variables are for their respective spec test targets; not sure why `CRUD_TESTS_PATH` and `GRIDFS_TESTS_PATH` needed to be singled out.

benjirewis:
The failures on `min-libmongoc` are probably inevitable for the time being, as the minimum libmongoc version for the CXX driver does not contain [CDRIVER-3905](https://jira.mongodb.org/browse/CDRIVER-3905). Looking into the Windows + latest failures...

.mci.yml
```diff
@@ -355,8 +357,10 @@ functions:
                   set -o xtrace
 
                   if [ "Windows_NT" == "$OS" ]; then
-                      CRUD_TESTS_PATH=$(cygpath -m $CRUD_TESTS_PATH)
+                      CRUD_LEGACY_TESTS_PATH=$(cygpath -m $CRUD_LEGACY_TESTS_PATH)
+                      CRUD_UNIFIED_TESTS_PATH=$(cygpath -m $CRUD_UNIFIED_TESTS_PATH)
                       ENCRYPTION_TESTS_PATH=$(cygpath -m $ENCRYPTION_TESTS_PATH)
+                      UNIFIED_FORMAT_TESTS_PATH=$(cygpath -m $UNIFIED_FORMAT_TESTS_PATH)
```
benjirewis, line 363: Windows tests were complaining about being unable to find files at `UNIFIED_FORMAT_TESTS_PATH` so I added this in; unsure how they were passing before...

src/mongocxx/test/CMakeLists.txt
```diff
@@ -260,6 +260,8 @@ set_tests_properties(read_write_concern_specs PROPERTIES
 
 set_tests_properties(unified_format_spec PROPERTIES
         ENVIRONMENT "UNIFIED_FORMAT_TESTS_PATH=${PROJECT_SOURCE_DIR}/../../data/unified-format")
+set_tests_properties(unified_format_spec PROPERTIES
+        ENVIRONMENT "CRUD_UNIFIED_TESTS_PATH=${PROJECT_SOURCE_DIR}/../../data/crud/unified")
```
kevinAlbs, line 264: I think this overwrites the `ENVIRONMENT` property. I think changing this to append to the list (like the `transactions_specs` above) should resolve the issue with ctest on Windows:

```suggestion
set_property(TEST unified_format_spec
        APPEND PROPERTY ENVIRONMENT "CRUD_UNIFIED_TESTS_PATH=${PROJECT_SOURCE_DIR}/../../data/crud/unified")
```

With that I was able to run through ctest locally (TIL how to do that):
```
$ cd cmake-build
$ ctest -R unified_format_spec -V
```

.mci.yml
benjirewis: Removed!

.mci.yml
benjirewis: Ahh clever, thanks for looking into this... Yep, looks like I was just overwriting `UNIFIED_FORMAT_TESTS_PATH` by accident. Removed the path variables here, but kept the `CTEST_OUTPUT_ON_FAILURE` env variables below.

src/mongocxx/test/CMakeLists.txt
benjirewis: Great catch, you're totally right. Looks like appending the property instead of calling `set_tests_properties` again fixes the Windows issues without having to re-export `UNIFIED_FORMAT_TESTS_PATH` in `.mci.yml`.

kevinAlbs (approved):
LGTM with tests passing!

.mci.yml
benjirewis: Looks like `ENCRYPTION_TESTS_PATH` is still needed, too. Not quite sure why, as the same overwriting is not happening in `test/CMakeLists.txt`.

.mci.yml
```diff
@@ -354,6 +354,8 @@ functions:
                   set -o xtrace
 
                   if [ "Windows_NT" == "$OS" ]; then
+                      ENCRYPTION_TESTS_PATH=$(cygpath -m $ENCRYPTION_TESTS_PATH)
```
kevinAlbs, line 357: Ah, I think setting the environment variable is needed for a similar reason. In src/mongocxx/test/CMakeLists.txt this overwrites the environment variable test property:

```
if (MONGOCXX_ENABLE_SLOW_TESTS)
  set_tests_properties(driver PROPERTIES
    ENVIRONMENT "MONGOCXX_ENABLE_SLOW_TESTS=1")
endif()
```

While we are cleaning this up, can you update src/mongocxx/test/CMakeLists.txt to append to the environment on the `driver` target?

benjirewis: 🤦  Yep that's probably it! Appended it instead of overwriting.

kevinAlbs (approved):
LGTM!

benjirewis:
While the mongohouse failures are due to [DRIVERS-1733](https://jira.mongodb.org/browse/DRIVERS-1733), it reminded me that running `getParameter` against an ADL instance can result in error. Accounted for this in `get_server_params`.

kevinAlbs (approved):
LGTM!

PR #788, opened 2021-05-03
CXX-2227 Bump minimum required C driver version

benjirewis:
CXX-2227

Bumps minimum required C driver version (both libmongoc and libbson) to "master". Eventually, we'll want to use the 1.18.0 tag after its release in CXX-2220 instead of the tip of master.


src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1296,11 +1296,9 @@ document::value operations::run(entity::map& entity_map,
         return empty_doc;
     }
     if (name == "assertSessionNotDirty") {
-        // TODO CXX-2227: bump minimum required libmongoc version to 1.18
```
benjirewis, line 1299: Should I implement these unified test runner functions now? They are needed to run the `poc-sessions` test in unified format.

benjirewis: I'll just do them now, actually. Doesn't make much sense to wait.

kevinAlbs (commented):
Looks good! Can you also update `LIBMONGOC_REQUIRED_VERSION` in `src/mongocxx/CMakeLists.txt`? That will have cmake reject any incompatible older versions of libmongoc.

kevinAlbs (approved):
LGTM!

PR #789, opened 2021-05-04
CXX-2237 Remove session count assertion listLocalSessions test

benjirewis:
CXX-2237

Removes unneeded assertion for session count in listLocalSessions test causing intermittent failures.

### DIFF
--- a/src/mongocxx/test/database.cpp
+++ b/src/mongocxx/test/database.cpp
@@ -345,25 +345,17 @@ TEST_CASE("Database integration tests", "[database]") {
             }
 
             auto session1 = mongo_client.start_session();
-            auto session2 = mongo_client.start_session();
 
             pipeline.list_local_sessions({});
-            pipeline.limit(2);
+            pipeline.limit(1);
             pipeline.add_fields(make_document(kvp("name", "Jane")));
             pipeline.project(make_document(kvp("name", 1)));
 
             auto cursor = database.aggregate(pipeline);
             auto results = get_results(std::move(cursor));
 
-            // start_session() does not actually send a startSession command, so only one
-            // session is guaranteed to be active due to aggregate's implicit session.
-            if (results.size() == 2) {
-                REQUIRE(results[0].view()["name"].get_string().value == stdx::string_view("Jane"));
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/database.cpp
```diff
@@ -355,7 +355,6 @@ TEST_CASE("Database integration tests", "[database]") {
             auto cursor = database.aggregate(pipeline);
             auto results = get_results(std::move(cursor));
 
-            REQUIRE(results.size() == 2);
```
kevinAlbs, line 358: This assertion was failing because sometimes only one session was returned. In that case, I think the assertions below would still fail.

Driver sessions can either be started locally by generating a UUID, or by sending a startSession command: https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#generating-a-session-id-locally

The C driver generates sessions locally. If this test is run independently, and the server has no active sessions, the assertion on line 359 could still fail.

Note, I think this test usually passes because prior tests have used sessions that have not been expired.

I would suggest changing this test to limit the results to 1. The `aggregate` operation will use an implicit session. So we are always guaranteed 1 active session server side.

benjirewis: Hmm gotcha. I worry that `REQUIRE(results.size() == 1)` will now intermittently fail if two sessions are actually created. I added some extra logic and an explanatory comment in the test for inspecting `results`, let me know what you think.

kevinAlbs: It should be OK if there is more than one active session server-side. The aggregation pipeline has a limit stage of two.

However, this test does not require starting two sessions and ensuring two are created server side. The test is only to ensure that the `pipeline::list_local_sessions` method works.

For maintainability, and a less confusing test, I'd suggest changing this to:

- only create one session
- change the pipeline limit to 1
- assert that results.size() is 1

kevinAlbs (commented):
Looking good! I have a suggestion for simplifying the test.

src/mongocxx/test/database.cpp
benjirewis: The simplification to a single session sounds good to me! Done.

kevinAlbs (approved):
LGTM!

PR #793, opened 2021-05-11
CXX-2211 Remove usages of xtrace in shell scripts

benjirewis:
CXX-2211

Removes all usages of `xtrace` from shell scripts in the repo.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -84,7 +84,6 @@ functions:
               shell: bash
               script: |
                  set -o errexit
-                 set -o xtrace
                  set -o pipefail
                  rm -rf "mongo-cxx-driver"
                  rm -fr "mongo-c-driver"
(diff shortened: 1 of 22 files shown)


.mci.yml
```diff
@@ -346,12 +333,10 @@ functions:
                   export UNIFIED_FORMAT_TESTS_PATH=$(pwd)/../data/unified-format
 
                   # export environment variables for encryption tests (and redact them)
```
rcsanchez97, line 335: You should probably remove "(and redact them)" since the redaction was necessary due to all the xtrace business, but that has been removed.

kevinAlbs (approved):
LGTM! Can you cherry-pick this to the `releases/v3.6` branch (and only that branch, since we won't be actively testing older branches in the public Evergreen project).

PR #794, opened 2021-05-11
CXX-2251 ship missing src/bsoncxx/types/value.hpp

rcsanchez97:


rcsanchez97:
We do.  However, it would be better to wait and cherry-pick once we know
whether the plan to re-release 3.6.3 or to make a new 3.6.4 release.

PR #795, opened 2021-05-11
CXX-2255 Upgrade to Ubuntu 18.04 for Evergreen tasks

benjirewis:
CXX-2255

Upgrades Ubuntu 16.04 tasks to Ubuntu 18.04 tasks in Evergreen.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -877,16 +877,17 @@ buildvariants:
             - ubuntu1804-build
           - name: uninstall_check
 
-    # TODO CXX-2260 Upgrade std::experimental tasks to Ubuntu 18.04 after bug fixes with g++ 5.5
+    # TODO CXX-2260 Upgrade std::experimental tasks to use Ubuntu 18.04 and mongodb_latest
+    # after bug fixes with g++ 5.5
     - name: ubuntu1604-debug-std-experimental
-      display_name: "Ubuntu 16.04 Debug (std::experimental) (MongoDB Latest)"
+      display_name: "Ubuntu 16.04 Debug (std::experimental) (MongoDB 4.4)"
       expansions:
           build_type: "Debug"
           tar_options: *linux_tar_options
           extra_path: *linux_extra_path
           cmake_flags: *linux_cmake_flags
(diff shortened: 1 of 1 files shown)


benjirewis:
Discussed offline: reverted the `debug-std-experimental` task to Ubuntu 16.04 to avoid compilation errors on g++ 5.5 (the compiler on Ubuntu 18.04) with non-standard polyfills (like EXPERIMENTAL). Filed CXX-2260 (and added a `TODO`) to address these errors and upgrade `debug-std-experimental` to Ubuntu 18.04.

.mci.yml
```diff
@@ -877,6 +877,7 @@ buildvariants:
             - ubuntu1804-build
           - name: uninstall_check
 
+    # TODO CXX-2260 Upgrade std::experimental tasks to Ubuntu 18.04 after bug fixes with g++ 5.5
```
kevinAlbs, line 880: Should this variant also be changed to test against 4.4 instead of latest in the meantime?

```
mongodb_version: *version_44
```

IIRC otherwise versioned API tests against the latest server version published on Ubuntu 16.04 may not pass.

benjirewis: Ah great point. Used `version_44` and updated comment and `display_name`.

kevinAlbs (approved):
LGTM!

.mci.yml
```diff
@@ -877,15 +877,17 @@ buildvariants:
             - ubuntu1804-build
           - name: uninstall_check
 
+    # TODO CXX-2260 Upgrade std::experimental tasks to use Ubuntu 18.04 and mongodb_latest
+    # after bug fixes with g++ 5.5
```
rcsanchez97, line 881: I think that g++5.5 actually worked and it was something around g++7.3 and newer that gave us problems.

rcsanchez97 (approved):
LGTM, module the nit on the comment.

.mci.yml
benjirewis: Ah good point. Made the comment refer generically to `newer g++ versions`.

PR #796, opened 2021-05-20
CXX-2258 Change prelude order and rerun clang-format

benjirewis:
CXX-2258

Changes order of headers to make sure `prelude`s are last, and reruns clang-format (v11.1.0) with `.clang-format` style.


kevinAlbs (approved):
LGTM! I sampled the files for review, since I do not think there were any manual changes aside from the clang format file.

PR #797, opened 2021-05-25
CXX-2258 Change prelude order and rerun clang-format

benjirewis:
CXX-2258

Adds prelude ordering to `.clang-format` for this branch (it didn't exist before), and reruns `clang-format`. There are some other formatting changes, but most are just moving headers around.


benjirewis:
[Here](https://spruce.mongodb.com/version/60ad3f193627e06b024f022e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) is a patch testing against the `cxx-driver-v3.6` project.

src/bsoncxx/types/bson_value/value.cpp
```diff
@@ -12,14 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <bsoncxx/json.hpp>
-
+// This include must come before the other value include.
```
benjirewis, line 15: Here and in `client-encryption.cpp`, certain imports must happen "out of order" from clang's perspective. So, I've used `// clang-format off/on` to disable the formatting for this line.

In this case, `value` in this file refers to `bson_value/value`, but if `bson_value/value` is included later, I _think_ the compiler gets confused about another definition of `value` from some other import.

kevinAlbs: Thank you for pointing this out. Checking locally I get this error when reformatting without the `// clang-format off` directive:

```
In file included from ../src/bsoncxx/types/bson_value/value.cpp:23:
../src/bsoncxx/types/bson_value/private/value.hh:28:7: error: incomplete type 'bsoncxx::v_noabi::types::bson_value::value' named in nested name specifier
class value::impl {
      ^~~~~~~
```

This error is because `src/bsoncxx/types/bson_value/private/value.hh` implicitly depends on `src/bsoncxx/types/bson_value/value.hpp` being included. Clyde fixed this when introducing clang-format:

https://github.com/mongodb/mongo-cxx-driver/pull/754/files#r532911731

Minor: if we are backporting clang-format, I think it seems reasonable to backport the additional `#include <bsoncxx/types/bson_value/view.hpp>` into `value.hh` instead of disabling clang-format.

kevinAlbs (approved):
LGTM with a minor comment.

src/bsoncxx/types/bson_value/value.cpp
benjirewis: Ah great catch, and thank you for the context. That appears to solve the problem here. Left the `// clang-format off/on` in `client_encryption.cpp`.

PR #799, opened 2021-06-10
CXX-2254 Add testing for versioned API

benjirewis:
CXX-2254

Adds two new tasks `test_versioned_api` and `test_versioned_api_accept_version_two`. Adds a new test utility `test_util::add_test_server_api` to append server API options when `MONGODB_API_VERSION` environment variable is set. Uses `add_test_server_api` to add API version to connecting clients in tests where necessary.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -346,7 +346,12 @@ functions:
 
                   if [ "Windows_NT" == "$OS" ]; then
                       CTEST_OUTPUT_ON_FAILURE=1 MSBuild.exe /p:Configuration=${build_type} RUN_TESTS.vcxproj
-                      CTEST_OUTPUT_ON_FAILURE=1 MSBuild.exe /p:Configuration=${build_type} examples/run-examples.vcxproj
+                      # Only run examples if MONGODB_API_VERSION is unset. We do not append
+                      # API version to example clients, so examples will fail when requireApiVersion
+                      # is true.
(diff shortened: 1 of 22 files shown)


.mci.yml
```diff
@@ -752,6 +775,32 @@ tasks:
...
 
+    - name: test_versioned_api
+      commands:
+        - func: "setup"
+        - func: "start_mongod"
+          vars:
+              REQUIRE_API_VERSION: true
+              # Authentication with versioned API should already be tested
+              # in the C driver.
+              AUTH: noauth
```
benjirewis, line 786: While we _should_ be testing with authentication (see versioned API test README [here](https://github.com/mongodb/specifications/blob/master/source/versioned-api/tests/README.rst#testing-with-required-api-version)), we would have to add username and password to every URI used in testing. Versioned API on authentication-related commands is handled entirely through the C driver, so as long as we test with authentication in the C driver, I don't think we need to do so here.

kevinAlbs:
I filed tickets for task failures that are unrelated with the label `failing-on-waterfall` to make it easier to distinguish whether there are any new task failures on a patch build.

I did not file a ticket for the `lint` task, but will fix that in a separate PR.

kevinAlbs (approved):
LGTM

rcsanchez97 (approved):
LGTM

PR #800, opened 2021-06-14
Fix lint

kevinAlbs:
Result of running:
```
python2 ./etc/clang_format.py format
```

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2301,10 +2301,10 @@ TEST_CASE("read_concern is inherited from parent", "[collection]") {
     }
 }
 
-void find_index_and_validate(
-    collection& coll,
-    stdx::string_view index_name,
-    const std::function<void(bsoncxx::document::view)>& validate = [](bsoncxx::document::view) {}) {
+void find_index_and_validate(collection& coll,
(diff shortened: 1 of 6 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -2301,10 +2301,10 @@ TEST_CASE("read_concern is inherited from parent", "[collection]") {
...
 }
 
-void find_index_and_validate(
-    collection& coll,
-    stdx::string_view index_name,
-    const std::function<void(bsoncxx::document::view)>& validate = [](bsoncxx::document::view) {}) {
+void find_index_and_validate(collection& coll,
+                             stdx::string_view index_name,
+                             const std::function<void(bsoncxx::document::view)>& validate =
+                                 [](bsoncxx::document::view) {}) {
```
kevinAlbs, line 2307: Note, `./etc/clang_format.py` expects clang format version 7.0.1, and will download the version of clang.

benjirewis (approved):
LGTM thanks for the quick fix. 

I wonder how I missed these formats before... For posterity, I used this [run-clang-format](https://github.com/Sarcasm/run-clang-format) Python script, when I probably should have used our internal `./etc/clang_format.py`.

PR #803, opened 2021-07-12
CXX-2290: Validate command filtering in event monitoring

vector-of-bool:
No functional changes, but checks that we see filtered events that bubble-up from the C driver. Also cleans up the unified-test-spec runner with some more thorough checking and error messages.

### DIFF
--- a/src/mongocxx/test/spec/monitoring.cpp
+++ b/src/mongocxx/test/spec/monitoring.cpp
@@ -49,7 +49,8 @@ void remove_ignored_command_monitoring_events(apm_checker::event_vector& events,
 void apm_checker::compare_unified(bsoncxx::array::view expectations, entity::map& map) {
     remove_ignored_command_monitoring_events(_events, _ignore);
 
-    auto equal = [&](const bsoncxx::array::element& exp, const bsoncxx::document::view actual) {
+    // This will throw an exception on unmatched fields and return true in all other cases.
+    auto compare = [&](const bsoncxx::array::element& exp, const bsoncxx::document::view actual) {
         CAPTURE(print_all(), to_json(actual), assert::to_string(exp.get_value()));
 
         // Extra fields are only allowed in root-level documents. Here, each k in keys is treated
@@ -66,8 +67,23 @@ void apm_checker::compare_unified(bsoncxx::array::view expectations, entity::map
         return true;
     };
 
-    // This will throw an exception on unmatched fields and return true in all other cases.
-    std::equal(expectations.begin(), expectations.end(), _events.begin(), equal);
+    auto exp_it = expectations.cbegin();
+    auto exp_end = expectations.cend();
+    auto ev_it = _events.cbegin();
+    auto ev_end = _events.cend();
+    for (; exp_it != exp_end && ev_it != ev_end; ++exp_it, ++ev_it) {
+        compare(*exp_it, *ev_it);
+    }
+    if (exp_it != exp_end) {
+        auto next_expected = exp_it->get_document();
+        CAPTURE(to_json(next_expected));
+        FAIL("Not enough events occurred");
+    }
+    if (ev_it != ev_end) {
+        auto next_event = *ev_it;
+        CAPTURE(to_json(next_event));
+        FAIL("Too many events occurred");
+    }
 }
 
 void apm_checker::compare(bsoncxx::array::view expectations,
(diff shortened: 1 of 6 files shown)


src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -66,8 +67,23 @@ void apm_checker::compare_unified(bsoncxx::array::view expectations, entity::map
...
+    auto exp_end = expectations.cend();
+    auto ev_it = _events.cbegin();
+    auto ev_end = _events.cend();
+    for (; exp_it != exp_end && ev_it != ev_end; ++exp_it, ++ev_it) {
+        compare(*exp_it, *ev_it);
+    }
+    if (exp_it != exp_end) {
+        auto next_expected = exp_it->get_document();
+        CAPTURE(to_json(next_expected));
+        FAIL("Not enough events occurred");
```
kevinAlbs, line 80: Was the motivation for this change to capture the last `expectation` / `event` when the lists are not equal lengths? If so, that seems generally helpful.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -116,10 +132,36 @@ std::string apm_checker::print_all() {
...
+        "getnonce",
+        "createUser",
+        "updateUser",
+        "gopydbgetnonce",
+        "copyDbSaslStart",
+        "copydb",
+    };
+    return std::find(std::begin(sensitive_commands),
+                     std::end(sensitive_commands),
+                     event.command_name())  //
```
kevinAlbs, line 154: Is the trailing `//` intentional to break the line?

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -31,6 +32,48 @@ using namespace mongocxx;
 using assert::to_string;
 using bsoncxx::types::bson_value::value;
 
+namespace {
+
+/// When we are matching two documents, we walk down them recursively. To aide test debugging, store
```
kevinAlbs, line 37: Minor:
```suggestion
/// When we are matching two documents, we walk down them recursively. To aid test debugging, store
```

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -45,8 +45,8 @@ using bsoncxx::builder::basic::kvp;
 using bsoncxx::builder::basic::make_document;
 
 using schema_versions_t =
-    std::array<std::array<int, 3 /* major.minor.patch */>, 1 /* supported version */>;
-constexpr schema_versions_t schema_versions{{{{1, 1, 0}}}};
+    std::array<std::array<int, 3 /* major.minor.patch */>, 2 /* supported version */>;
+constexpr schema_versions_t schema_versions{{{{1, 1, 0}}, {{1, 5, 0}}}};
```
kevinAlbs, line 49: Can you add a comment noting that this only partially supports 1.5.0?

It is warranted here since these tests are important to implement now, and do not require full 1.5.0 support. But this may be surprising later when the test runner fails to run other tests in unified test format 1.2.0-1.5.0.

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -31,6 +32,48 @@ using namespace mongocxx;
...
+/// a path as we go, in this variable:
+thread_local std::vector<std::string> S_match_doc_path;
```
kevinAlbs, line 39: Nice, this is a cool idea. Is the rationale for the `thread_local` to make a future parallel test runner easier?

kevinAlbs (commented):
This looks great! I only have minor comments and a few questions.

If CXX-2265 only requires adding the test files, feel free to do that in this PR as well. Though we merge PRs as one commit, as long as the ticket number is in the commit message, the git hook will pick up the commit and comment in to the CXX ticket.

src/mongocxx/test/spec/monitoring.cpp
vector-of-bool: Yes. Previously it would either not check all events or it would crash.

src/mongocxx/test/spec/monitoring.cpp
vector-of-bool: Yeah, it's become a bit of a (weird) habit I use to slap `clang-format` on the wrist to make it do some nicer formatting.

src/mongocxx/test/spec/unified_tests/assert.cpp
vector-of-bool: Yes-ish. While parallel runners in the future would probably use separate OS processes, I like to use `thread_local` to store call-stack-local state Just In Case™ there are ever multiple threads. Passing the path down the call stack would be "the proper way", but I find usage of an internal-linkage thread-local to be cleaner for cases like this.

src/mongocxx/test/spec/monitoring.cpp
kevinAlbs: Ah good to know!

src/mongocxx/test/spec/unified_tests/assert.cpp
kevinAlbs: Ack. That seems like a good precaution.

kevinAlbs (approved):
LGTM!

PR #806, opened 2021-07-22
CXX-2282 build RPMs in Fedora 35 environment

rcsanchez97:
Patch build: https://spruce.mongodb.com/version/60f9f04e0305b94cdeeb8e46/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/spec.patch
+++ b/.evergreen/spec.patch
@@ -9,7 +9,7 @@
  Name:           mongo-cxx-driver
 -Version:        3.4.1
 -Release:        5%{?dist}
-+Version:        3.6.1
++Version:        3.7.0
 +Release:        1%{?dist}
  Summary:        A C++ Driver for MongoDB
  License:        ASL 2.0
--- a/.evergreen/spec.patch
+++ b/.evergreen/spec.patch
@@ -1,15 +1,19 @@
---- mongo-cxx-driver.spec.orig	2020-08-17 16:22:55.490400655 -0400
-+++ mongo-cxx-driver.spec	2020-08-17 16:26:24.490702714 -0400
-@@ -1,6 +1,8 @@
+--- mongo-cxx-driver.spec.orig	2021-07-21 13:05:46.139038889 -0400
++++ mongo-cxx-driver.spec	2021-07-22 18:23:45.099314489 -0400
+@@ -1,17 +1,17 @@
  # for better compatibility with SCL spec file
  %global pkg_name mongo-cxx-driver
  
 +%define _unpackaged_files_terminate_build 0
 +
  Name:           mongo-cxx-driver
- Version:        3.4.1
- Release:        4%{?dist}
-@@ -10,8 +12,6 @@
+-Version:        3.4.1
+-Release:        5%{?dist}
++Version:        3.6.1
++Release:        1%{?dist}
+ Summary:        A C++ Driver for MongoDB
+ License:        ASL 2.0
+ URL:            https://github.com/mongodb/mongo-cxx-driver/wiki
  Source0:        https://github.com/mongodb/%{pkg_name}/archive/%{name}-r%{version}.tar.gz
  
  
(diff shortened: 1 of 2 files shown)


.evergreen/spec.patch
```diff
@@ -1,15 +1,19 @@
...
  
 +%define _unpackaged_files_terminate_build 0
 +
  Name:           mongo-cxx-driver
- Version:        3.4.1
- Release:        4%{?dist}
-@@ -10,8 +12,6 @@
+-Version:        3.4.1
+-Release:        5%{?dist}
++Version:        3.6.1
```
kevinAlbs, line 12: Should this be our latest tagged release 3.6.4?

.evergreen/spec.patch
```diff
@@ -1,15 +1,19 @@
---- mongo-cxx-driver.spec.orig	2020-08-17 16:22:55.490400655 -0400
-+++ mongo-cxx-driver.spec	2020-08-17 16:26:24.490702714 -0400
-@@ -1,6 +1,8 @@
+--- mongo-cxx-driver.spec.orig	2021-07-21 13:05:46.139038889 -0400
++++ mongo-cxx-driver.spec	2021-07-22 18:23:45.099314489 -0400
+@@ -1,17 +1,17 @@
  # for better compatibility with SCL spec file
  %global pkg_name mongo-cxx-driver
  
 +%define _unpackaged_files_terminate_build 0
```
kevinAlbs, line 7: Is this now required to account for the fact that tests are no longer built?

.evergreen/spec.patch
```diff
@@ -27,15 +31,16 @@
  
  %build
  
-@@ -77,6 +75,7 @@
+@@ -77,6 +75,8 @@
  %cmake \
      -DCMAKE_BUILD_TYPE=Release \
      -DBSONCXX_POLY_USE_BOOST=1 \
 +    -DENABLE_UNINSTALL=OFF \
```
kevinAlbs, line 38: Is removing the uninstall script also required to build successfully?

.evergreen/spec.patch
rcsanchez97: I think you are misreading the change.  The `spec.patch` file is itself a diff, so we are looking at a diff of a diff.  The `+%define _unpackaged_files_terminate_build 0` line is just context.  It is not a change related to my commit.  That said, the reason for that directive is that our build creates both shared and static libraries, but the RPM only ships the shared libraries.  This leaves the static libraries in the install tree without them ending up in an RPM at the end of the process.  By default, the RPM build process will fail when this happens, so that the maintainer does not inadvertently leave parts of the package behind.  Remi could probably also just set `-DBUILD_SHARED_AND_STATIC_LIBS=OFF` so that the static libraries are not built.  But either approach is fine.

.evergreen/spec.patch
rcsanchez97: Like the above line, this is also an unchanged context line.  That said, we did add this some time ago and it was probably needed for some obscure reason that escapes me at the moment.

.evergreen/spec.patch
kevinAlbs: 🤦 you are totally right – I thought this was introduced here. Thank you for the additional background. I hadn't realized that RPM only ships shared libraries.

.evergreen/spec.patch
kevinAlbs: This does appear to be introduced in this change. Was this pulled from the latest spec file? Comparing with https://src.fedoraproject.org/rpms/mongo-cxx-driver/blob/rawhide/f/mongo-cxx-driver.spec it appears to still be 3.4.1.

rcsanchez97: Based on our pattern with the C driver, this should really be 3.7.0.

kevinAlbs (approved):

PR #807, opened 2021-07-23
CXX-2202 update ppc64le build platform

rcsanchez97:
Patch build: https://spruce.mongodb.com/version/60fb3c39a4cf47264f5b96fd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/bsoncxx/config/CMakeLists.txt
+++ b/src/bsoncxx/config/CMakeLists.txt
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set (libdir "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
+
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/config.hpp.in
     ${CMAKE_CURRENT_BINARY_DIR}/config.hpp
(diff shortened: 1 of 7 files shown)


src/bsoncxx/config/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set (libdir "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
```
kevinAlbs, line 15: Oh good catch! Is this the C++ analog to CDRIVER-3339? If so, is the observable issue the same (i.e. that `/usr/lib` may be included in `pkg-config`)?

This fix seems orthogonal to CXX-2202. It it OK to put this all in one commit, but I suggest filing a separate CXX ticket and including in the commit message so this fix is easier to track in release notes.

kevinAlbs (approved):
LGTM, mod one question about the libdir fix!

src/bsoncxx/config/CMakeLists.txt
rcsanchez97: You're right that this is similar in nature to CDRIVER-3339 and it is a completely distinct issue to CXX-2202.  Though, CXX-2202 needs this issue to be fixed first.  I went ahead and filed CXX-2327 and updated the commit message to reference it.

PR #809, opened 2021-07-27
CXX-2329 Change type of thrown exception to match documentation

chardan:
The documentation says that we should throw mongocxx::logic_error, but std::logic_error is thrown instead.

Note: at this time mongocxx::logic_error does not have support for error messages (the C++ standard does) 
Note: it's unclear to me why we don't throw mongocxx::invalid_argument instead (the C++ standard has this)

See: https://jira.mongodb.org/browse/CXX-2329

### DIFF
--- a/src/mongocxx/exception/error_code.cpp
+++ b/src/mongocxx/exception/error_code.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015 MongoDB Inc.
+// Copyright 2015-present MongoDB Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -84,6 +84,8 @@ class error_category final : public std::error_category {
                 return "an invalid client session was provided";
             case error_code::k_invalid_transaction_options_object:
                 return "an invalid transactions options object was provided";
+	    case error_code::k_create_resource_fail:
+		return "could not create resource";
             default:
                 return "unknown mongocxx error";
         }
--- a/src/mongocxx/exception/error_code.cpp
+++ b/src/mongocxx/exception/error_code.cpp
@@ -84,8 +84,8 @@ class error_category final : public std::error_category {
                 return "an invalid client session was provided";
             case error_code::k_invalid_transaction_options_object:
                 return "an invalid transactions options object was provided";
-	    case error_code::k_create_resource_fail:
-		return "could not create resource";
+            case error_code::k_create_resource_fail:
+                return "could not create resource";
             default:
                 return "unknown mongocxx error";
         }
--- a/src/mongocxx/options/server_api.cpp
+++ b/src/mongocxx/options/server_api.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <system_error>
+
 #include <bsoncxx/stdx/string_view.hpp>
 #include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/options/server_api.hpp>
@@ -29,7 +31,7 @@ std::string server_api::version_to_string(server_api::version version) {
         case server_api::version::k_version_1:
             return "1";
         default:
-            throw std::logic_error{"invalid server API version"};
+            throw mongocxx::logic_error{std::make_error_code(std::errc::invalid_argument), "invalid server API version"};
     }
 }
 
@@ -37,7 +39,7 @@ server_api::version server_api::version_from_string(stdx::string_view version) {
     if (!version.compare("1")) {
         return server_api::version::k_version_1;
     }
-    throw std::logic_error{"invalid server API version"};
+    throw mongocxx::logic_error{std::make_error_code(std::errc::invalid_argument), "invalid server API version"};
 }
 
 server_api::server_api(server_api::version version) : _version(std::move(version)) {}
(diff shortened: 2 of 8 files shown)


chardan:
If it's not feasible to do this now (for whatever reason, including it breaking existing API), then it seems the documentation should be revised to reflect the correct type of thrown exception.

chardan:
It's worth pointing out that std::logic_error has some interesting rules on what it's allowed to do on copy. I'm not sure if that was what kept the previous implementation from supporting this or not. FWIW, if we aren't simply using the standard implementatoin (as we do for mongocxx::exception()), it seems like we are free to change this.

chardan:
Another update: mongocxx::exception derives from std::system_error, which supports std::error_code along with messages. 

However, it's not clear to me exactly why we want to do this for logic_error (a big part of std::error_code's reason to be is making non-portable errors/standard). I'm waiting to hear back from some folks on what the justification might be.

chardan:
Note: I think it's worth further discussion as to whether or not mongocxx::logic_error is the right type to throw, either-- it's meant to represent an invalid parameter, so I would think that having mongocxx::invalid_parameter might make some more sense. First, this more closely follows the behavior of the standard library (which doesn't throw std::logic_error directly), and also seems like it could provide a better opportunity for applications to recover.

benjirewis:
It looks like there are some failures from `test_driver`; not sure if those are relevant.

For context, I wrote this code and did not have a particular reason for using `std::logic_error`. I think `mongocxx::logic_error` would be preferable, as it includes error codes and seems to be more prevalent throughout the codebase.

You'll probably also want to fix the `std::logic_error`s in [`server_api.hh`](https://github.com/mongodb/mongo-cxx-driver/blob/master/src/mongocxx/options/private/server_api.hh#L36-L37) as well.

chardan:
@benjirewis Ok, I think the picture is becoming a bit more clear-- and, after further exploration, I see that there's an exception/error_code.hpp header that provides custom error codes, which is making the picture more clear-- and there's quite a few uses of mongocxx::error_code::k_invalid_parameter, so that looks like "how we do it". :-) Thank you for your comments!

src/mongocxx/options/server_api.cpp
```diff
@@ -29,15 +32,15 @@ std::string server_api::version_to_string(server_api::version version) {
         case server_api::version::k_version_1:
             return "1";
         default:
-            throw std::logic_error{"invalid server API version"};
+            throw mongocxx::logic_error { mongocxx::error_code::k_invalid_parameter, "invalid server API version" };
```
kevinAlbs, line 35: Please run `python ./etc/clang_format.py format` to format. This should resolve the [lint](https://spruce.mongodb.com/task/cxx_driver_ubuntu1804_release_latest_lint_patch_8f25e2348f1db555cdea8a56036a7563b0817e02_61033720e3c33110c85e91e4_21_07_29_23_27_14/logs?execution=0) task on Evergreen (which was failing on a prior commit).

src/mongocxx/exception/error_code.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 MongoDB Inc.
+// Copyright 2015-present MongoDB Inc.
```
kevinAlbs, line 1: Good eye!

src/mongocxx/exception/error_code.cpp
```diff
@@ -84,6 +84,8 @@ class error_category final : public std::error_category {
                 return "an invalid client session was provided";
             case error_code::k_invalid_transaction_options_object:
                 return "an invalid transactions options object was provided";
+	    case error_code::k_create_resource_fail:
```
benjirewis, line 87: Nice 👍

benjirewis (approved):
LGTM

kevinAlbs (commented):
> If it's not feasible to do this now (for whatever reason, including it breaking existing API), then it seems the documentation should be revised to reflect the correct type of thrown exception.

This is feasible to do now. The was introduced in CXX-2138, which is part of the upcoming 3.7.0 release.

If this was part of a stable release, I think this would be backwards breaking behavior, since client code would need to update exception handling to account for the new type.

> It looks like there are some failures from test_driver; not sure if those are relevant.

In keeping with the pattern of tracking these task failures with tickets in the [failing-on-waterfall](https://jira.mongodb.org/issues/?jql=project%3DCXX%20and%20labels%3Dfailing-on-waterfall%20and%20status%20!%3D%20closed) label:

- The `test_mongohouse` task failure is tracked in CXX-2281.
- I filed CXX-2332 to track the Ubuntu 14.04 task failures.
- CXX-2333 tracks the rpm-package-build failure.

kevinAlbs (approved):
LGTM with a fix to lint!

@chardan, after this is approved by all reviewers and this is OK to merge, note the commit message should follow: https://github.com/mongodb/mongo-cxx-driver/blob/master/CONTRIBUTING.md#commit-messages

src/mongocxx/options/server_api.cpp
chardan: +1, done.

src/mongocxx/exception/error_code.cpp
chardan: Ok, I'll try to resolve these as I run into them! Thanks!

PR #810, opened 2021-07-27
Cxx2257 jfw versioned api examples

chardan:
Provide examples in the C++ driver documentation for calling the versioned API and some utilities, in order to satisfy:
https://jira.mongodb.org/browse/CXX-2257?filter=-1

Whether this is corrected by a documentation change or by the underlying implementation, this depends upon:
https://jira.mongodb.org/browse/CXX-2329

Note: the namespace expansions feel verbose to me, I can be easily convinced to use aliases, as I think (a'la Boost) it
will actually be easier for the reader to understand.

### DIFF
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <iostream>
 #include <vector>
+#include <iostream>
+#include <optional>
 
 #include <bsoncxx/builder/basic/array.hpp>
 #include <bsoncxx/builder/basic/document.hpp>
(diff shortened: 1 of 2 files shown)


examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,13 +1182,63 @@ void delete_examples(mongocxx::database db) {
...
+
+    // There are several helper member functions for converting between strings and API versions:
+    std::string version_string = mongocxx::options::server_api::version_to_string(mongocxx::options::server_api::version::k_version_1);
+
+    // Convert a string to an API version:
+    
+    // Using an invalid version string should throw a mongocxx::logic_error:
+    try
+     {
+         mongocxx::options::server_api::version version =
```
chardan, line 1198: FYI, I've written these types out fully mostly for my own edification-- at the next set of commits (depending on discussion WRT exception types) I'll write them in a more idiomatic (e.g. with auto) fashion.

chardan:
closing, changing branch

PR #811, opened 2021-07-28
CXX-2330 make calc_release_version.py output consistent between Pytho…

rcsanchez97:
Patch build; https://evergreen.mongodb.com/version/6100bb8e32f417783415eb23?redirect_spruce_users=true

### DIFF
--- a/etc/calc_release_version.py
+++ b/etc/calc_release_version.py
@@ -43,13 +43,21 @@ def check_output(args):
     provides a reasonable facsimile.
     """
     if 'check_output' in dir(subprocess):
-        return subprocess.check_output(args)
+        out = subprocess.check_output(args)
+    else:
+        proc = subprocess.Popen(args, stdout=subprocess.PIPE)
+        out, err = proc.communicate()
+        ret = proc.poll()
+        if ret:
+            raise subprocess.CalledProcessError(ret, args[0], output=out)
 
-    proc = subprocess.Popen(args, stdout=subprocess.PIPE)
-    out, err = proc.communicate()
-    ret = proc.poll()
-    if ret:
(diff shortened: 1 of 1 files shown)


etc/calc_release_version.py
```diff
@@ -43,13 +43,21 @@ def check_output(args):
...
+        ret = proc.poll()
+        if ret:
+            raise subprocess.CalledProcessError(ret, args[0], output=out)
+
+    if type(out) is bytes:
+        """
+        git isn't guaranteed to always return UTF-8, but for our purposes
+        this should be fine as tags and hashes should be ASCII only.
+        """
+        out = out.decode('utf-8')
```
kevinAlbs, line 59: I do not understand how this resulted in the difference between python2 / python3 output.

But consistently returning a `str` seems like the right choice.

I confirmed the output on the `master` branch looks correct on python2 and python3 after applying these changes.

As a sanity check, I inspected the latest three release tarballs to ensure the `./build/VERSION_CORRECT` was indeed correct.

kevinAlbs (approved):
LGTM!

rcsanchez97:
> 
>    I do not understand how this resulted in the difference between python2 /
>    python3 output.
> 
>    But consistently returning a str seems like the right choice.
> 
This was puzzling to me as well.  However, the same change had to be
made in the C driver a while back to make sure that Python2 and Python3
invocations both returned the same result.

>    I confirmed the output on the master branch looks correct on python2 and
>    python3 after applying these changes.
> 
>    As a sanity check, I inspected the latest three release tarballs to ensure
>    the ./build/VERSION_CORRECT was indeed correct.
> 
Thanks!

PR #813, opened 2021-07-31
CXX-2257 Add C++ versioned-API examples

chardan:
Provide examples in the C++ driver documentation for calling the versioned API and some utilities, in order to satisfy:
https://jira.mongodb.org/browse/CXX-2257?filter=-1

### DIFF
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <iostream>
+#include <optional>
 #include <vector>
 
 #include <bsoncxx/builder/basic/array.hpp>
@@ -22,6 +23,8 @@
 #include <bsoncxx/stdx/string_view.hpp>
 #include <bsoncxx/types.hpp>
 #include <mongocxx/client.hpp>
+#include <mongocxx/exception/error_code.hpp>
+#include <mongocxx/exception/logic_error.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/options/find.hpp>
 #include <mongocxx/uri.hpp>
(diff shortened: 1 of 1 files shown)


examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,13 +1182,62 @@ void delete_examples(mongocxx::database db) {
...
+         auto version = mongocxx::options::server_api::version_from_string(R"({"version" : "1"}")");
+     }
+    catch(const mongocxx::logic_error&)
+     {
+	 std::cout << "calling version_from_string() with an invalid version string threw an error, as expected: good\n";
+     }
+
+    // Using a valid version string should be ok:
+    {
+         auto version = mongocxx::options::server_api::version_from_string(R"({"version" : "1"}")");
```
chardan, line 1208: Ooops, looks like if I want to show how to handle an invalid version, I should also show what happens with a valid one! Will correct.

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,13 +1183,74 @@ void delete_examples(mongocxx::database db) {
     }
 }
 
+void start_versioned_API_examples() {
```
kevinAlbs, line 1186: This looks like a good example of the API for constructing and using a `mongocxx::server_api`.

I think this needs to be updated to match the ticket description.

```
// Start Versioned API Example 1
--
 
client = MongoClient(
<connection string>,
api=\{"version": "1"}
)
 
// End Versioned API Example 1
```

Should correspond to the C++ example with those comment delimiters, something like:

```
// Start Versioned API Example 1
auto server_api = mongocxx::options::server_api(mongocxx::options::server_api::version::k_version_1);
auto client_opts = mongocxx::options::client ();
opts.server_api_opts (api_opts);
auto client = mongocxx::client(uri, client_opts);
// End Versioned API Example 1
```

The examples are used in language examples for the MongoDB documentation pages ([example, the language selector on the right](https://docs.mongodb.com/manual/changeStreams/)). So they need to closely conform to the pseudo code.

kevinAlbs:
As a heads up, this may have been done in the scope of #813. I added you as a reviewer for #813. We may want to close this PR.

chardan:
Superceded-by: https://github.com/mongodb/mongo-cxx-driver/pull/814

PR #814, opened 2021-08-04
CXX-2314 Add tests+examples for docs on the versioned server API

vector-of-bool:
Closes CXX-2314

### DIFF
--- a/src/mongocxx/test/CMakeLists.txt
+++ b/src/mongocxx/test/CMakeLists.txt
@@ -173,6 +173,13 @@ add_executable(test_unified_format_spec
     ${spec_test_common}
 )
 
+add_executable(test_versioned_api
+    ${PROJECT_SOURCE_DIR}/test_util/client_helpers.cpp
+    ${THIRD_PARTY_SOURCE_DIR}/catch/main.cpp
+    versioned_api.cpp
+    ${spec_test_common}
+    )
+
 target_link_libraries(test_driver mongocxx_mocked ${libmongoc_target})
 target_link_libraries(test_logging mongocxx_mocked ${libmongoc_target})
 target_link_libraries(test_instance mongocxx_mocked ${libmongoc_target})
@@ -186,6 +193,7 @@ target_link_libraries(test_retryable_reads_specs mongocxx_mocked ${libmongoc_tar
 target_link_libraries(test_read_write_concern_specs mongocxx_mocked ${libmongoc_target})
 target_link_libraries(test_mongohouse_specs mongocxx_mocked ${libmongoc_target})
 target_link_libraries(test_unified_format_spec mongocxx_mocked ${libmongoc_target})
+target_link_libraries(test_versioned_api mongocxx_mocked ${libmongoc_target})
 
 target_include_directories(test_driver PRIVATE ${libmongoc_include_directories})
 target_include_directories(test_logging PRIVATE ${libmongoc_include_directories})
@@ -224,6 +232,7 @@ add_test(transactions_specs test_transactions_specs)
 add_test(retryable_reads_spec test_retryable_reads_specs)
 add_test(read_write_concern_specs test_read_write_concern_specs)
 add_test(unified_format_spec test_unified_format_spec)
+add_test(versioned_api test_versioned_api)
 
 # Adding this as a test will run it as part of the RUN_TESTS command in MSVC.
 # Do not add, since we only test mongohouse on Linux.
--- a/src/mongocxx/test/CMakeLists.txt
+++ b/src/mongocxx/test/CMakeLists.txt
@@ -358,5 +358,6 @@ set_dist_list (src_mongocxx_test_DIST
    transactions.cpp
    uri.cpp
    validation_criteria.cpp
+   versioned_api.cpp
    write_concern.cpp
 )
--- a/src/mongocxx/test/versioned_api.cpp
+++ b/src/mongocxx/test/versioned_api.cpp
@@ -24,13 +24,14 @@
 
 using namespace mongocxx;
 
-// We'll format many of these examples by hand
-// clang-format off
-
 static bool is_server_v5_or_newer() {
-    auto wire_version = mongocxx::test_util::get_max_wire_version(client{uri{}});
+    auto wire_version =
+        mongocxx::test_util::get_max_wire_version(client{uri{}, test_util::add_test_server_api()});
     return wire_version >= 13;
 }
+
+// We'll format many of these examples by hand
+// clang-format off
 
 TEST_CASE("Versioned API, non-strict") {
     instance::current();
--- a/src/mongocxx/test/versioned_api.cpp
+++ b/src/mongocxx/test/versioned_api.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <bsoncxx/string/to_string.hpp>
 #include <bsoncxx/test_util/catch.hh>
 #include <mongocxx/client.hpp>
 #include <mongocxx/exception/operation_exception.hpp>
@@ -26,8 +27,23 @@ using namespace mongocxx;
 // We'll format many of these examples by hand
 // clang-format off
 
+static bool is_server_v5_or_newer() {
+    using namespace bsoncxx::builder::basic;
+    client cl{uri{}};
+    auto status = cl["admin"].run_command(make_document(kvp("serverStatus", 1)));
+    auto version = bsoncxx::string::to_string(status["version"].get_string().value);
+    auto dot_pos = version.find('.');
+    REQUIRE(dot_pos != version.npos);
+    auto major_digit_str = version.substr(0, dot_pos);
+    auto major_version = std::stoul(major_digit_str);
+    return major_version >= 5;
+}
+
 TEST_CASE("Versioned API, non-strict") {
     instance::current();
+    if (!is_server_v5_or_newer()) {
+        return;
+    }
     // Start Versioned API Example 1
     using namespace mongocxx;
     uri client_uri{"mongodb://localhost"};
@@ -45,6 +61,9 @@ TEST_CASE("Versioned API, non-strict") {
 
 TEST_CASE("Versioned API, strict") {
     instance::current();
+    if (!is_server_v5_or_newer()) {
+        return;
+    }
     // Start Versioned API Example 2
     using namespace mongocxx;
     uri client_uri{"mongodb://localhost"};
@@ -63,6 +82,9 @@ TEST_CASE("Versioned API, strict") {
 
 TEST_CASE("Versioned API, non-strict, for commands/features outside versioned API") {
     instance::current();
+    if (!is_server_v5_or_newer()) {
+        return;
+    }
     // Start Versioned API Example 3
     using namespace mongocxx;
     uri client_uri{"mongodb://localhost"};
@@ -81,13 +103,15 @@ TEST_CASE("Versioned API, non-strict, for commands/features outside versioned AP
 
 TEST_CASE("Versioned API, non-strict with deprecation errors") {
     instance::current();
+    if (!is_server_v5_or_newer()) {
+        return;
+    }
     // Start Versioned API Example 4
     using namespace mongocxx;
     uri client_uri{"mongodb://localhost"};
     // Create an option set for API v1
     const auto server_api_opts =
         options::server_api{options::server_api::version_from_string("1")}
-            .strict(false)              // Explicitly disable strict mode for the server API
             .deprecation_errors(true);  // Enable deprecation errors
     // Store it in the set of client options
     const auto client_opts =
@@ -101,6 +125,9 @@ TEST_CASE("Versioned API, non-strict with deprecation errors") {
 
 TEST_CASE("Versioned API, with insert-many for 'count' migration") {
     instance::current();
+    if (!is_server_v5_or_newer()) {
+        return;
+    }
 
     using namespace mongocxx;
     mongocxx::client client{
--- a/src/mongocxx/test/versioned_api.cpp
+++ b/src/mongocxx/test/versioned_api.cpp
@@ -28,15 +28,8 @@ using namespace mongocxx;
 // clang-format off
 
 static bool is_server_v5_or_newer() {
-    using namespace bsoncxx::builder::basic;
-    client cl{uri{}};
-    auto status = cl["admin"].run_command(make_document(kvp("serverStatus", 1)));
-    auto version = bsoncxx::string::to_string(status["version"].get_string().value);
-    auto dot_pos = version.find('.');
-    REQUIRE(dot_pos != version.npos);
-    auto major_digit_str = version.substr(0, dot_pos);
-    auto major_version = std::stoul(major_digit_str);
-    return major_version >= 5;
+    auto wire_version = mongocxx::test_util::get_max_wire_version(client{uri{}});
+    return wire_version >= 13;
 }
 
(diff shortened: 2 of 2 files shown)


vector-of-bool:
I used the corresponding Python and Go driver changesets to figure out how this one should look.

src/mongocxx/test/CMakeLists.txt
```diff
@@ -173,6 +173,13 @@ add_executable(test_unified_format_spec
     ${spec_test_common}
 )
 
+add_executable(test_versioned_api
+    ${PROJECT_SOURCE_DIR}/test_util/client_helpers.cpp
+    ${THIRD_PARTY_SOURCE_DIR}/catch/main.cpp
+    versioned_api.cpp
```
kevinAlbs, line 179: This may need to be the full path to `versioned_api.cpp`.

The Evergreen patch build attached to this PR is reporting errors in cmake configuration. Example: https://spruce.mongodb.com/task/cxx_driver_arm_ubuntu1804_50_compile_and_test_with_shared_libs_patch_c2932becd7b69dfa15bfc6a8cda6038dfc33a600_610af93d7742ae1ea40b23f7_21_08_04_20_32_00/logs?execution=0

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <mongocxx/uri.hpp>
+
+using namespace mongocxx;
+
+// We'll format many of these examples by hand
+// clang-format off
+
+TEST_CASE("Versioned API, non-strict") {
+    instance::current();
+    // Start Versioned API Example 1
```
kevinAlbs, line 31: These examples are within the scope of CXX-2257. I think we can close #813 in favor of this.

Can you include CXX-2314 in the commit message so the final commit gets linked to the ticket?

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // Store it in the set of client options
+    const auto client_opts =
+        options::client{}
+            .server_api_opts(server_api_opts); // Set the version
+    // Create a new client with the options
+    mongocxx::client client{client_uri, client_opts};
+    // End Versioned API Example 1
+}
+
+TEST_CASE("Verioned API, strict") {
```
kevinAlbs, line 46:
```suggestion
TEST_CASE("Versioned API, strict") {
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // Store it in the set of client options
+    const auto client_opts =
+        options::client{}
+            .server_api_opts(server_api_opts); // Set the version
+    // Create a new client with the options
+    mongocxx::client client{client_uri, client_opts};
+    // End Versioned API Example 1
+}
+
+TEST_CASE("Verioned API, non-strict, for commands/features outside versioned API") {
```
kevinAlbs, line 64:
```suggestion
TEST_CASE("Versioned API, non-strict, for commands/features outside versioned API") {
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+        options::client{}
+            .server_api_opts(server_api_opts); // Set the version
+    // Create a new client with the options
+    mongocxx::client client{client_uri, client_opts};
+    // End Versioned API Example 1
+}
+
+TEST_CASE("Verioned API, strict") {
+    instance::current();
+    // Start Versioned API Example 1
```
kevinAlbs, line 48:
```suggestion
    // Start Versioned API Example 2
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const auto server_api_opts =
+        options::server_api{options::server_api::version_from_string("1")}
+            .strict(true);  // Enable strict mode for the server API
+    // Store it in the set of client options
+    const auto client_opts =
+        options::client{}
+            .server_api_opts(server_api_opts); // Set the version
+    // Create a new client with the options
+    mongocxx::client client{client_uri, client_opts};
+    // End Versioned API Example 1
```
kevinAlbs, line 61:
```suggestion
    // End Versioned API Example 2
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+TEST_CASE("Verioned API, non-strict, for commands/features outside versioned API") {
+    instance::current();
+    // Start Versioned API Example 1
```
kevinAlbs, line 66:
```suggestion
    // Start Versioned API Example 3
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const auto server_api_opts =
+        options::server_api{options::server_api::version_from_string("1")}
+            .strict(false);  // Explicitly disable strict mode for the server API
+    // Store it in the set of client options
+    const auto client_opts =
+        options::client{}
+            .server_api_opts(server_api_opts); // Set the version
+    // Create a new client with the options
+    mongocxx::client client{client_uri, client_opts};
+    // End Versioned API Example 1
```
kevinAlbs, line 79:
```suggestion
    // End Versioned API Example 3
```

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
```
kevinAlbs, line 80: Please add example 4 from CXX-2257.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+using namespace mongocxx;
+
+// We'll format many of these examples by hand
+// clang-format off
+
+TEST_CASE("Versioned API, non-strict") {
+    instance::current();
+    // Start Versioned API Example 1
+    using namespace mongocxx;
```
chardan, line 32: note: This appears at global scope earlier in the file-- doesn't hurt anything, and makes it clear to the user that it's part of the example code.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,173 @@
...
+    uri client_uri{"mongodb://localhost"};
+    // Create an option set for API v1
+    const auto server_api_opts =
+        options::server_api{options::server_api::version_from_string("1")};
```
chardan, line 36: You might consider showing how to handle errors from a bad version string as well (catching the resultant mongocxx::logic_error and then checking its flag, for instance), as a separate example away from the "happy path".

chardan (commented):
Looks good-- a couple of notes for your consideration.

src/mongocxx/test/CMakeLists.txt
vector-of-bool: Looks like the file needed to be added to the dist list later down in the file.

vector-of-bool:
Looks like the tests need to be skipped on server <5.0

src/mongocxx/test/versioned_api.cpp
```diff
@@ -73,10 +73,29 @@ TEST_CASE("Verioned API, non-strict, for commands/features outside versioned API
...
+
+TEST_CASE("Versioned API, non-strict with deprecation errors") {
+    instance::current();
+    // Start Versioned API Example 4
+    using namespace mongocxx;
+    uri client_uri{"mongodb://localhost"};
+    // Create an option set for API v1
+    const auto server_api_opts =
+        options::server_api{options::server_api::version_from_string("1")}
+            .strict(false)              // Explicitly disable strict mode for the server API
```
kevinAlbs, line 90: The pseudo code in CXX-2257 for `Versioned API Example 4` does not explicitly set the `strict` option. I think this line can be removed.

src/mongocxx/test/versioned_api.cpp
kevinAlbs: To weigh in, that might be helpful for users to know how to handle errors. But the pseudo code in CXX-2257 does not consider error handling. Since these docs examples are used on pages like https://docs.mongodb.com/manual/changeStreams/ where language examples are shown side-by-side, I suggest erring towards consistency.

kevinAlbs (commented):
Changes look good; pending passing tests.

> Looks like the tests need to be skipped on server <5.0

I think so. Versioned API is a 5.0 server feature. And I believe older servers will error on the unrecognized option.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,200 @@
...
+        make_sale(6,  "xyz",  5,     5,   "2021-02-15T12:05:10Z"),
+        make_sale(7,  "xyz",  5,     10,  "2021-02-15T14:12:12Z"),
+        make_sale(8,  "abc",  10,    5,   "2021-03-16T20:20:13Z"),
+    };
+    sales.insert_many(items);
+    // End Versioned API Example 5
+    // clang-format on
+
+    try {
+        db.run_command(make_document(kvp("count", "sales")));
```
kevinAlbs, line 170: I think this is causing a test failure on the `test_versioned_api` variant.

That variant is described in the Versioned API test readme: https://github.com/mongodb/specifications/tree/master/source/versioned-api/tests#testing-with-required-api-version

I believe the cause is:
- The run_command helper does not append server API parameters. This is specified in the Versioned API section [Generic command helper](https://github.com/mongodb/specifications/blob/df6be82f865e9b72444488fd62ae1eb5fca18569/source/versioned-api/versioned-api.rst#generic-command-helper).
- The `test_versioned_api` task configures the server and requires a server API version in all commands. That is configured in the [Evergreen config](https://github.com/mongodb/mongo-cxx-driver/blob/2fdfd62c60370f660c935f82c1f67b7d4b26d9ab/.mci.yml#L806:L806).

I think the solution is to use the deprecated `count` helper. That should append a server API. To avoid the deprecation warnings from failing compile, wrap the call in `BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_BEGIN` and `BSONCXX_SUPPRESS_DEPRECATION_WARNINGS_END`.

This also alerted me that my corresponding change in CDRIVER-4093 did not take this into account, and also used the generic C driver command helper. I moved CDRIVER-4093 back to "In Progress" to address that.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // Start Versioned API Example 5
+    auto db = client.database("db");
+    auto sales = db.collection("sales");
+    auto make_sale =
+        [](int id, std::string item_name, int price, int quanitity, std::string time) {
+            return make_document(kvp("_id", id),
+                                 kvp("item", item_name),
+                                 kvp("price", price),
+                                 kvp("quantity", quanitity),
+                                 kvp("date", time));
```
kevinAlbs, line 152: The `date` should be a BSON date type, not a string. It is represented as a unix timestamp. I don't think converting ISO date strings to Unix timestamps is relevant to the example. But for consistency of types between language examples, I suggest adding a helper like `iso_to_unix` that is defined outside of the example, similar to the one defined in C https://github.com/mongodb/mongo-c-driver/commit/0d6a8ba27ed5cd94a26e4db879f61b545799193f#diff-803a49bf3111998fa92319347cedc6c44fe5070ed21e30d4bcc03b87653d880aR3706. It can bypass any actual conversion, but help signify to the user that a BSON date is being stored.

Aside: https://bsonspec.org/spec.html is the canonical reference on BSON types.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -26,8 +27,23 @@ using namespace mongocxx;
 // We'll format many of these examples by hand
 // clang-format off
 
+static bool is_server_v5_or_newer() {
```
kevinAlbs, line 30: Nice, you can alternatively use the `test_util::get_max_wire_version` to determine if the wire version is greater than 13. The wire version is often used in the driver to perform server support checks.

https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst is a list of wire versions mapping to server versions.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -26,8 +27,23 @@ using namespace mongocxx;
...
+    using namespace bsoncxx::builder::basic;
+    client cl{uri{}};
+    auto status = cl["admin"].run_command(make_document(kvp("serverStatus", 1)));
```
kevinAlbs, line 33: This may have the same caveat as running `run_command` for the `count` command below. The client above can be configured with a conditional version API like so:

```
client mongodb_client{uri{}, test_util::add_test_server_api()};
```

But it may be better to use the `test_util::get_max_wire_version` helper altogether.

kevinAlbs (commented):
Looking good! I have a suggestion on how to address one remaining test failure, and an ask to use a BSON date instead of string for the `date` field.

src/mongocxx/test/versioned_api.cpp
vector-of-bool: What is the deprecated `count` helper that you refer to? I can't seem to find anything that is deprecated with a "count" name.

kevinAlbs: Ah, my apologies, I do not see one either. Some drivers have a deprecated `count` helper, which is distinct from `countDocuments` and `estimatedDocumentCount`. I assumed the C++ driver had one as part of the `mongocxx::collection` class.

For reference the [CRUD spec](https://github.com/mongodb/specifications/blob/df6be82f865e9b72444488fd62ae1eb5fca18569/source/crud/crud.rst#read) describes which underlying MongoDB command is sent for each of the three count helpers.

> The run_command helper does not append server API parameters. This is specified in the Versioned API section Generic command helper.

My understanding was incorrect. The generic command helpers do append API version parameters. After inspecting test failures introduced in [CDRIVER-4093](https://jira.mongodb.org/browse/CDRIVER-4093), I see that libmongoc's `mongoc_database_command_with_opts` applies the version API as expected.

I think there is nothing to change here. I suspect the `The apiVersion parameter is required, please configure your MongoClient's API` error from [the `test_version_api` task](https://spruce.mongodb.com/task/cxx_driver_versioned_api__os~windows_2k8_mongodb_version~latest_test_versioned_api_patch_c2932becd7b69dfa15bfc6a8cda6038dfc33a600_6114951457e85a7025bb1f47_21_08_12_03_27_17/logs?execution=0) may be due to the `is_server_v5_or_newer` check above.

src/mongocxx/test/versioned_api.cpp
```diff
@@ -28,15 +28,8 @@ using namespace mongocxx;
...
-    using namespace bsoncxx::builder::basic;
-    client cl{uri{}};
-    auto status = cl["admin"].run_command(make_document(kvp("serverStatus", 1)));
-    auto version = bsoncxx::string::to_string(status["version"].get_string().value);
-    auto dot_pos = version.find('.');
-    REQUIRE(dot_pos != version.npos);
-    auto major_digit_str = version.substr(0, dot_pos);
-    auto major_version = std::stoul(major_digit_str);
-    return major_version >= 5;
+    auto wire_version = mongocxx::test_util::get_max_wire_version(client{uri{}});
```
kevinAlbs, line 31: Use `client{uri{}, test_util::add_test_server_api()}` to apply the necessary default versioned API test on the `test_versioned_api` task.

kevinAlbs (approved):
Changes LGTM. I have a suggested change to fix the `test_versioned_api` task. Feel free to re-request review if that does not fix that test task.

PR #816, opened 2021-08-10
CXX-2342 Add ServiceID to events in loadbalanced mode.

chardan:
Add serviceId to CommandFailedEvent, CommandSucceededEvent, and CommandStartedEvent

https://jira.mongodb.org/browse/CXX-2342

Part of CXX-2173 LoadBalancer epic.


chardan:
In tests/space/monitoring.cpp, there's remove_ignored_command_monitoring_events(), which currently removes
commandStartedEvent, commandFailedEvent, and commandSucceededEvent. That filtering was added 2021-04-16 (Clyde Bazile). Was this in anticipation of the load-balancing code, and if so should we now add LB based tests for these?

src/mongocxx/events/command_failed_event.hpp
```diff
@@ -75,6 +79,15 @@ class MONGOCXX_API command_failed_event {
     ///
     std::int64_t operation_id() const;
 
+    ///
+    /// Returns the service id.
+    ///
+    /// @note service_id() only returns a value if load balancing is enabled.
+    ///
+    /// @return No value, or the service id if load balancing is enabled.
```
kevinAlbs, line 87: Nit: other documentation in the code base refers to returns with optionals as possibly returning a "disengaged" value. Here is one example: https://github.com/kevinAlbs/mongo-cxx-driver/blob/33416f88f1d59e19eb103fedab59d73399332a7c/src/mongocxx/collection.hpp#L262:L262

"disengaged" always sounded odd to me, but I would consider changing for consistency.

src/mongocxx/events/command_failed_event.hpp
```diff
@@ -16,6 +16,10 @@
 
 #include <memory>
 
+#include <bsoncxx/stdx/optional.hpp>
+
+#include <bsoncxx/private/libbson.hh>
```
kevinAlbs, line 21: The C++ driver cannot use private headers in public headers. libbson / libmongoc symbols should be an implementation detail and not exposed.

src/mongocxx/events/command_failed_event.hpp
```diff
@@ -75,6 +79,15 @@ class MONGOCXX_API command_failed_event {
...
     std::int64_t operation_id() const;
 
+    ///
+    /// Returns the service id.
+    ///
+    /// @note service_id() only returns a value if load balancing is enabled.
+    ///
+    /// @return No value, or the service id if load balancing is enabled.
+    ///
+    bsoncxx::stdx::optional<bson_oid_t> service_id() const;
```
kevinAlbs, line 89: This cannot return a libbson type. To return a BSON ObjectID, you can use the `bsoncxx:oid`. For some prior art, see `mongocxx::events::server_changed_event::topology_id()`, which also returns a BSON ObjectID in public API by converting from a libbson `bson_oid_t`.

Aside: https://bsonspec.org/spec.html is the canonical reference on BSON types.

kevinAlbs (commented):
This looks like a great start! Next up: we can discuss ideas on what to test and how.

> In tests/space/monitoring.cpp, there's remove_ignored_command_monitoring_events(), which currently removes
> commandStartedEvent, commandFailedEvent, and commandSucceededEvent. That filtering was added 2021-04-16 (Clyde Bazile). Was this in anticipation of the load-balancing code, and if so should we now add LB based tests for these?

Per offline discussion, `remove_ignored_command_monitoring_events` was to implement `ignoreCommandMonitoringEvents` from the Unified Test Format. For example of one such test, see: https://github.com/mongodb/specifications/blob/df6be82f865e9b72444488fd62ae1eb5fca18569/source/unified-test-format/tests/valid-pass/poc-change-streams.yml#L13:L13

Aside: when reading specification tests, refer to the YML files from the mongodb/specifications repo. The JSON counterparts are generated, and do not contain comments.

src/mongocxx/events/command_failed_event.cpp
```diff
@@ -52,6 +56,15 @@ std::int64_t command_failed_event::operation_id() const {
         static_cast<const mongoc_apm_command_failed_t*>(_failed_event));
 }
 
+bsoncxx::stdx::optional<bson_oid_t> command_failed_event::service_id() const {
+    if(const bson_oid_t *result = libmongoc::apm_command_failed_get_service_id(
+	      static_cast<const mongoc_apm_command_failed_t*>(_failed_event)); result) {
```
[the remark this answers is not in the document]
chardan: Right, haven't run it yet-- I'll do so during the next cycle.

src/mongocxx/events/command_failed_event.cpp
```diff
@@ -52,6 +56,15 @@ std::int64_t command_failed_event::operation_id() const {
         static_cast<const mongoc_apm_command_failed_t*>(_failed_event));
 }
 
+bsoncxx::stdx::optional<bson_oid_t> command_failed_event::service_id() const {
+    if(const bson_oid_t *result = libmongoc::apm_command_failed_get_service_id(
```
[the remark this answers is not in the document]
chardan: It is indeed. Since C++17, if-statements (and most other conditionals) can have an init-specifier, similar to a for-statement:
if(init; cond) {} else { /* init is still in scope */ }

src/mongocxx/events/command_failed_event.hpp
chardan: The C++ standard does not use this terminology (n4892, 20.6.2), instead using "contained object" and "contains a value". Perhaps the phrasing "@return No contained value, or contains the service id..." is the way to go. I feel that perpetuating "disengaged" might continue to add further confusion to the mix.

src/mongocxx/events/command_failed_event.hpp
chardan: Right, I have been wondering about this one (and I think I even commented somewhere!)-- I wasn't able to otherwise expose bson_oid_t. But, I see you've answered how this works in the C++ driver below! Thank you!

src/mongocxx/events/command_failed_event.hpp
chardan: Ok, thank you!! I'll rework this.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -267,6 +275,10 @@ void apm_checker::set_command_succeeded(options::apm& apm) {
                            make_document(kvp("reply", event.reply()),
                                          kvp("command_name", event.command_name()),
                                          kvp("operation_id", event.operation_id()))));
+
+        // We MUST NOT report a service_id() outside of load-balancing mode:
+        CHECK_FALSE(event.service_id());
```
kevinAlbs, line 280: Nice! Can you add positive test cases that with mocks?

I am thinking like the following pseudo-code for each of the event types:
```
bool apm_command_started_get_service_id_called = false;
bson_oid_t mock_oid;

// TODO: initialize mock_oid

auto apm_command_started_get_service_id =
	libmongoc::apm_command_started_get_service_id.create_instance();
apm_command_started_get_service_id->interpose([]() {
    apm_command_started_get_service_id_called = true;
    return &mock_oid;
})

/* Register command monitoring callbacks. */
bool on_command_started_called = false;
auto apm_opts;
apm_opts.on_command_started ([](mongocxx::events::command_started_event* event) {
   on_command_started_called = true;
   // TODO: compare event.service_id () to mock_oid's representation
})


// Add necessary options for versioned API test environment.
auto client_opts;
mongocxx::test_util::add_test_server_api(client_opts);

// Add command monitoring callbacks.
client_opts.apm_opts (apm_opts);

auto client = mongocxx::client (uri() test_util::add_test_server_api()});
// Send a command (e.g. ping)
client["db"].command(/* ... */)

CHECK_TRUE (apm_command_started_get_service_id_called);
CHECK_TRUE (on_command_started_called);

File: src/mongocxx/test/spec/monitoring.cpp [hunk 7]
```diff
@@ -242,6 +242,10 @@ void apm_checker::set_command_started(options::apm& apm) {
                                          kvp("command_name", event.command_name()),
                                          kvp("operation_id", event.operation_id()),
                                          kvp("database_name", event.database_name()))));
+
+        // We MUST NOT report a service_id() outside of load-balancing mode:
+        CHECK_FALSE(event.service_id());
```
kevinAlbs, line 247: Can you investigate the new test failures on the patch build attached to this PR? I see error logs for the boost polyfill pointing to this file:

https://spruce.mongodb.com/task/cxx_driver_macos_1014_50_compile_and_test_with_shared_libs_extra_alignment_patch_2fdfd62c60370f660c935f82c1f67b7d4b26d9ab_6115b7740305b94d0870d23b_21_08_13_04_14_38/logs?execution=0

chardan: Oh, thank you for catching this! It looks like CHECK_FALSE() doesn't know what to do with the underlying boost::optional<>, so I've included the suggested header. However, I don't think this is the right /place/ to do it: I don't see any other references to boost in the tests. It probably needs to go into wherever we're including Catch, or perhaps in stdx? The candidates that stand out to me are third_party/catch/include/catch.hpp (but I imagine we're generating that via catch elsewhere? I don't yet see where...), or src/bsoncxx/stdx/optional.hpp. Let's see if this corrects it, and then figure out if it's best to leave it where it is (we are clearly already using boost) or if we need to squirrel it away maybe in one of those locations (or elsewhere).

src/mongocxx/test/spec/monitoring.cpp
chardan: Working on this! I'm still trying to get my head around how the test harness works; some of it seems straightforward, but there are a few things I'll probably have to ask about when you have time.

src/mongocxx/test/spec/monitoring.cpp
chardan: Update: I stashed this in the optional<> polyfill support, where Boost is specifically wrangled.

src/mongocxx/test/spec/monitoring.cpp
```diff
@@ -267,6 +284,14 @@ void apm_checker::set_command_succeeded(options::apm& apm) {
...
                                          kvp("command_name", event.command_name()),
                                          kvp("operation_id", event.operation_id()))));
+
+        // We MUST NOT report a service_id() outside of load-balancing mode, but in
+        // load-balancing mode, a value is required:
+        if (test_util::is_load_balanced())
+            CHECK(event.service_id());
+        else
+            CHECK_FALSE(event.service_id());
+
```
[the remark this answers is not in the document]
kevinAlbs: There are two methods of mocking (mocking the C calls, and mocking the `serviceId` from within libmongoc). I think we can still mock the C calls without significant changes, and that would give test coverage.

I set up time to discuss offline. Apologies for the confusing back-and-forth! The two ways of mocking is tricky to write out...

kevinAlbs (commented):
I scheduled time on Monday to pair on approaching tests.

src/mongocxx/test/spec/monitoring.cpp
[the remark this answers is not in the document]
chardan: +1 ok, sorry for the confusion on my end as well! :-)

chardan:
Superceded by:
https://github.com/chardan/mongo-cxx-driver/tree/CXX2342-jfw-add_serviceid_to_events

PR #819, opened 2021-08-31
CXX-2276 more precise empty directory pruning on install

rcsanchez97:


generate_uninstall/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ else ()
          \"\${MONGOCXX_INSTALL_MANIFEST_CONTENT}\")
       execute_process (
          COMMAND
-            find ${CMAKE_INSTALL_PREFIX} -type d -empty -delete
+            find \"\$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/${BSONCXX_HEADER_INSTALL_DIR}\" \"\$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/${MONGOCXX_HEADER_INSTALL_DIR}\" -type d -empty -delete
```
kevinAlbs, line 76: Does this need the same corresponding change on L40 for Windows?

rcsanchez97:
So, I had considered that, and I concluded that the change wasn't
necessary there.  In particular, because the VS-based build on Windows
lacks the capability for specifying DESTDIR, it seems unlikely that we
would encounter this particular issue.  That is, the place where
artifacts physically land on the file system is under the actual
installation prefix that was specified when the build was configured.
There is no possibility of the Windows installation getting confused the
same way that the Linux installation did before this fix.

The other consideration is that because there are no "standard"
locations for things like development tools and libraries on Windows, it
seems unlikely that the C/C++ drivers would be installed under some sort
of shared directory tree structure, making this particular issue
inconsequential there.  Our installation guides essentially have users
build and install to a path like `C:\mongo-cxx-driver`, which I would be
surprised if the user encountered this issue while following that
recommendation.

That said, I don't think it wouldn hurt to replace
"`${CMAKE_INSTALL_PREFIX_WIN32}`" with
"`${CMAKE_INSTALL_PREFIX_WIN32}\\\\${BSONCXX_HEADER_INSTALL_DIR}`" and
then create a second command for the mongocxx headers.  I can do that if
you like, and I would test it on a Windows host to make sure that it
doesn't cause some other sort of problem.

kevinAlbs (approved):
> In particular, because the VS-based build on Windows lacks the capability for specifying DESTDIR, it seems unlikely that we would encounter this particular issue.

That is a good point. LGTM.

PR #820, opened 2021-08-31
default added

pratik-a:


kevinAlbs:
Thank you for the contribution @pratik-a!

src/bsoncxx/types/bson_value/view.cpp
```diff
@@ -89,6 +89,7 @@ view& view::operator=(const view& rhs) noexcept {
         break;
 #include <bsoncxx/enums/type.hpp>
 #undef BSONCXX_ENUM
+default : break;
```
chardan, line 92: optional: I think this is fine and will do what you want; consider also moving it ahead of the include-pasted cases, so that the latter will be indicated as an error instead of your more intentional case. (And, I think it's slightly easier to read, YYMV.)

chardan (approved):
+1 looks ok to me!

PR #823, opened 2021-09-01
CXX-2342 add serviceid to events

chardan:
Add serviceId to CommandFailedEvent, CommandSucceededEvent, and CommandStartedEvent

https://jira.mongodb.org/browse/CXX-2342

Part of CXX-2173 LoadBalancer epic.

Signed-off-by: Jesse Williamson jesse.williamson@mongodb.com


src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,60 @@ TEST_CASE("Database integration tests", "[database]") {
...
+struct check_service_id
+{
+    void operator()(const EventT& event) {
+
+    // We MUST NOT report a service_id() outside of load-balancing mode, but in
+    // load-balancing mode, a value is required:
+    INFO("checking for service_id()")
+    CAPTURE(event.command_name(), test_util::is_load_balanced());
+    auto service_id = event.service_id();
+    if (test_util::is_load_balanced())
```
kevinAlbs, line 509: The environment variable `MONGOC_TEST_LOADBALANCED` is currently not set. `service_id` will only be present exactly when the corresponding  C driver function is mocked to include it.

I suggest adding a field to this struct, like `bool expect_service_id` which the test can set based on whether the corresponding C driver function is mocked.

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,60 @@ TEST_CASE("Database integration tests", "[database]") {
...
+        CHECK(service_id);
+    else
+        CHECK_FALSE(service_id);
+    }
+};
+
+
+TEST_CASE("Test serviceId is included in command monitoring events") {
```
kevinAlbs, line 517: I suggest splitting this into two `TEST_CASE` blocks.

1. Test mocking each C driver function to return a service ID. Expect each of the C++ callbacks to include `service_id`.
2. Test mocking each C driver function to not return a service ID. Expect each of the C++ callbacks to not include `service_id`.

(2) does not strictly require mocking. All current integration tests in Evergreen run against non load balanced servers (i.e. they will not report a `serviceId` in the handshake). When we do add tests against load balanced servers, we'd have to update (2) to check based on the environment.

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,60 @@ TEST_CASE("Database integration tests", "[database]") {
...
+
+    client_opts.apm_opts (apm_opts);
+    // Adding ?loadBalanced=true results in an error in the C driver because
+    // the initial hello response from the server does not include serviceId.
+    client mongo_client(uri("mongodb://localhost:27017/"), client_opts);
+    stdx::string_view database_name{"database"};
+    database database = mongo_client[database_name];
+
+    // Mock mongoc_apm_command_started_get_service_id
+    auto apm_command_started_get_service_id = libmongoc::apm_command_started_get_service_id.create_instance();
```
kevinAlbs, line 537: This only mocks one of the three events. This should mock `libmongoc::apm_command_failed_get_service_id` and `libmongoc::apm_command_succeeded_get_service_id` as well.

src/mongocxx/test/database.cpp
chardan: I used a feature of Catch2 to generate positive and negative test cases in combination with a helper. A bit odd, but I think somewhat better than just duplicating all the code to get a "not".

src/mongocxx/test/database.cpp
chardan: This is in effect what is happening in my latest change, because of the behavior of GENERATE, which creates virtual SECTIONS. (see: "https://github.com/catchorg/Catch2/blob/devel/docs/generators.md").

src/mongocxx/test/database.cpp
chardan: +1 thank you for catching that!

src/bsoncxx/stdx/optional.hpp
```diff
@@ -40,6 +40,7 @@ BSONCXX_INLINE_NAMESPACE_END
 
 #include <boost/none.hpp>
 #include <boost/optional/optional.hpp>
+#include <boost/optional/optional_io.hpp>
```
kevinAlbs, line 43: To check, this include was needed to use `bsoncxx::stdx::optional` in test assertions like `CHECK(service_id)`, right?

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,95 @@ TEST_CASE("Database integration tests", "[database]") {
     }
 }
 
+// As C++11 lacks generic lambdas, and "ordinary" templates can't appear at block scope,
+// we'll have to define our helper for the serviceId tests here. This implementation would
+// be more straightforward in newer versions of C++ (C++14 and on allow generic lambdas),
+// see CXX-2350 (migration to more recent C++ standards); our C++17 optional<> implmentation
```
kevinAlbs, line 498:
```suggestion
// see CXX-2350 (migration to more recent C++ standards); our C++17 optional<> implementation
```

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,95 @@ TEST_CASE("Database integration tests", "[database]") {
...
+// also has a few inconsistencies with the standard, which appear to vary across platforms:
```
kevinAlbs, line 499:
```suggestion
// also has a few inconsistencies with the standard, which appear to vary across platforms
```

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,95 @@ TEST_CASE("Database integration tests", "[database]") {
...
+    // Set Application Performance Monitoring options (APM):
+    mongocxx::options::apm apm_opts;
+    apm_opts.on_command_started(
+        check_service_id<mongocxx::events::command_started_event>{expect_service_id});
+    apm_opts.on_command_succeeded(
+        check_service_id<mongocxx::events::command_succeeded_event>{expect_service_id});
+    apm_opts.on_command_failed(
+        check_service_id<mongocxx::events::command_failed_event>{expect_service_id});
+
+    // Set up mocking for mongoc_apm_command_started_get_service_id:
```
kevinAlbs, line 536: This applies to all three events. I suggest removing this comment since the comments on L544 and L560 note that mocks are being set.

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,95 @@ TEST_CASE("Database integration tests", "[database]") {
...
+    if (expect_service_id) {
+        // Return a bson_oid_t with data where the service_id has some value:
+        struct {
+            bson_oid_t* operator()(const void*) {
+                static bson_oid_t tmp = {0x65};
+                return &tmp;
+            }
+        } make_service_id_bson_oid_t;
+
+        // Add forever() so that the mock function also extends to endSession (thanks, kalbertson!):
```
kevinAlbs, line 554:
```suggestion
        // Add forever() so that the mock function also extends to all commands sent by the client, like "endSessions":
```

I suggest removing the thanks, since I can see it as being confusing to future readers. But you are very welcome :smile:. It is not obvious that "endSessions" is sent internally by clients, and that it appears in command monitoring events.

src/mongocxx/test/database.cpp
kevinAlbs: Ah nice! I did not know about `GENERATE`.

kevinAlbs (approved):
LGTM with some comment tweaks! Since this is a non-trivial change, can you request review from another team member?

src/bsoncxx/stdx/optional.hpp
chardan: Correct, otherwise Catch2 doesn't know how to display them.

src/mongocxx/test/database.cpp
chardan: +1 good catch

src/mongocxx/test/database.cpp
chardan: s/:/.

src/mongocxx/test/database.cpp
chardan: Done-- but it will remain "morally there". ;-D

src/mongocxx/test/database.cpp
chardan: Given the high level of confusion that's possible for new-to-the-test-framework folks, I would prefer to leave these here. I agree it's somewhat redundant once it's clear what's going on, but like many other things in this test I think a bit of clarification might very well help the next reader, or one less familiar with the test harnesses. (Although if other reviewers are of the same opinion, I don't have super strong feelings about it.)

kevinAlbs: Leaving the comment SGTM. It currently only notes `mongoc_apm_command_started_get_service_id` being mocked. Since `mongoc_apm_command_succeeded_get_service_id` and `mongoc_apm_command_failed_get_service_id` are also mocked, should this be edited to note all three?

src/mongocxx/test/database.cpp
```diff
@@ -492,4 +492,95 @@ TEST_CASE("Database integration tests", "[database]") {
...
+
+    // Set up mocked functions that DO emit a service_id:
+    if (expect_service_id) {
+        // Return a bson_oid_t with data where the service_id has some value:
+        struct {
+            bson_oid_t* operator()(const void*) {
+                static bson_oid_t tmp = {0x65};
+                return &tmp;
+            }
+        } make_service_id_bson_oid_t;
```
vector-of-bool, line 552: Can this be replaced with a lambda expression?

src/mongocxx/test/database.cpp
chardan: Ah! I see what you're saying! I've gone ahead and revised the comment. Derp!

src/bsoncxx/private/helpers.hh
```diff
@@ -33,6 +34,20 @@ inline document::value value_from_bson_t(const bson_t* bson) {
...
+Construct an oid from a bson_oid_t (which is a C API type that we don't want to
+expose to the world).
+
+Note: passing a nullptr is unguarded
+Note: Deduction guides aren't yet available to us, so a factory it is! This is
+something that can be improved as part of CXX-2350 (migration to more recent C++
+standards).
+*/
+inline bsoncxx::oid make_oid(const bson_oid_t* bson_oid) {
+    return bsoncxx::oid(reinterpret_cast<const char*>(const_cast<bson_oid_t*>(bson_oid)),
```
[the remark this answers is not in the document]
chardan: As we're not casting away the const-ness, nope! Good eyes, thanks for spotting it.

src/mongocxx/test/database.cpp
chardan: Yes, but unfortunately there's a little complexity hiding there; as C++11 lacks variadic lambdas, you wind up having to write one for each function-- so there are three of:
        const auto make_service_id_bson_oid_t = [](const _mongoc_apm_command_started_t *) -> const bson_oid_t* {
                        static bson_oid_t tmp = { 0x65 };
                        return &tmp;
                };

...so in lieu of being able to use auto-parameters, I don't think the function object is all too bad, by comparison.

chardan: Wait, it might help if I got the signature right, huh? Ok. Done!

vector-of-bool (commented):
LGTM

vector-of-bool (approved):

PR #827, opened 2021-09-08
CXX-2345 Add BSON binary subtype 0x07

eramongodb:
The C++ driver does not appear to have BSON corpus tests corresponding to those proposed in the spec update. Therefore the scope of changes required seems to be limited to adding the new BSON binary subtype constant.


kevinAlbs (approved):
LGTM. The C++ driver does not run BSON corpus tests.

PR #833, opened 2021-09-29
CXX-2297 add connect.sh and Atlas Serverless connectivity test

kevinAlbs:


.evergreen/connect.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
kevinAlbs, line 1: The motivation for moving this into a separate script file is to make it easier to run locally. I build in a directory named `cmake-build` and can run this locally as follows:

```
MONGOC_INSTALL_PREFIX=/Users/kevin.albertson/install/mongo-c-driver-1.18.0 \
MONGOCXX_INSTALL_PREFIX=/Users/kevin.albertson/install/mongo-cxx-driver-dev \
LIB_DIR="lib" \
BUILD_TYPE="Debug" \
BUILD_DIR=$(pwd)/cmake-build \
URI="mongodb://localhost:27017/?" \
    ./.evergreen/connect.sh
```

chardan: Is there a place (even as an extended comment at the beginning of the script, but perhaps as a help screen?) where you can put a scrubbed version of this example? Sounds very useful!

chardan (changes requested):
This looks pretty good, and definitely useful! I am not a shell expert but it seems ok to me-- a little elaboration in-situ on the usage seems like a good idea, but I like that you've broken this out and don't want to add a bunch of mechanism to get in the way of it being available.

.evergreen/connect.sh
kevinAlbs: Good idea, done!

.evergreen/connect.sh
```diff
@@ -0,0 +1,54 @@
...
+LIB_DIR=${LIB_DIR:-"lib"}
+BUILD_TYPE=${BUILD_TYPE:-"Debug"}
+
+print_usage_and_exit () {
+    echo "Error: $1"
+    cat << EOF
+    connect.sh runs the connect example with a MongoDB URI.
+
+    Example usage:
+    MONGOC_INSTALL_PREFIX=/Users/kevin.albertson/install/mongo-c-driver-1.18.0 \\
```
chardan, line 15: Up to you if you want to leave your name in here, or if it should be scrubbed. I am fine with it as-is.

chardan (approved):
Even if more modifications are needed for other platforms, etc., thanks for breaking this out and giving everyone a useful head start! LGTM. +1.

rcsanchez97 (approved):
LGTM!

.evergreen/connect.sh
kevinAlbs: Leaving as-is.

PR #841, opened 2022-01-21
CXX-2262, CXX-2263

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/61ea17d50ae6065f865942e0/tasks?page=0&sorts=STATUS%3AASC


benchmark/single_doc/run_command.hpp
```diff
@@ -45,7 +45,7 @@ class run_command : public microbench {
     mongocxx::database _db;
 };
 void run_command::task() {
-    auto command = make_document(kvp("ismaster", true));
+    auto command = make_document(kvp("ping", 1));
```
kevinAlbs, line 48:
```suggestion
    auto command = make_document(kvp("hello", true));
```

This comes from the [Benchmarking spec](https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst#run-command).

Each benchmark defines a `Dataset size` for measurement. Please update the `0.16` above to `0.13`. I believe `0.16` corresponds to the [previous size](https://github.com/mongodb/specifications/commit/99c9ee188a4bc4fc095837ae0be2451c30761c9b#diff-289c90aed81a3162c93a75dc51bf0f7815a4de7e99283a1007e0624d91c06004R492) of 160,000 bytes.

kevinAlbs (changes requested):
The [grep command](https://jira.mongodb.org/browse/CXX-2177?focusedCommentId=3772606&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3772606) included in the ticket comments can be used to find other occurrences. I modified it slightly to remove results with the master branch. Renaming the master branch is out of scope for oppressive language removal:

```
# Exclude .git.
# Exclude build scripts referencing master branch.
# Exclude build directory, cmake-build directory.
ggrep -PriI '(docs\.mongodb\.com\/master|(?<!/)master(?!.?key)|slave|whitelist|blacklist)' \
 --exclude-dir=.git \
 --exclude-dir="build" \
 --exclude-dir="cmake-build" \
 --exclude="*.sh" \
 --exclude="*.conf" \
 --exclude="*.py" \
 --exclude="*.json" \
 --exclude="*.js" .
```

I request two more changes:

1. Replace all documentation links of the form `https://docs.mongodb.com/master/<...>` with `https://docs.mongodb.com/manual/<...>`
2. Update ./examples/mongocxx/connect.cpp

src/mongocxx/test/client_session.cpp
```diff
@@ -317,10 +317,10 @@ class session_test {
         apm_opts.on_command_started([&](const events::command_started_event& event) {
             std::string command_name{event.command_name().data()};
 
-            // Ignore auth commands like "saslStart", and handshakes with "isMaster".
+            // Ignore auth commands like "saslStart", and handshakes with "hello" (and the legacy "hello" command).
             std::string sasl{"sasl"};
             if (event.command_name().substr(0, sasl.size()).compare(sasl) == 0 ||
-                command_name.compare("isMaster") == 0) {
+                command_name.compare("hello") == 0 || command_name.compare("isMaster") == 0) {
```
chardan, line 323: No need to change this, but I wanted to point out that although std::string::compare() is fine, if you only care about exact equality, it's less work to write (and a bit more familiar to read):

if(command_name == "hello") {}

The main difference-- and point of the compare() member function-- is that operator==() gives a Boolean result (that is. equal-to), whereas compare() gives lexicographic order: less-than and greater-than 0 are distinct.

(Further reference: https://www.cplusplus.com/reference/string/string/compare/)

benchmark/single_doc/run_command.hpp
rcsanchez97: Ah, of course.

rcsanchez97:
> 2. Update ./examples/mongocxx/connect.cpp

Hmm, not sure how I overlooked that.  :-/

src/mongocxx/test/client_session.cpp
rcsanchez97: Ah, that's interesting.  I'm more familiar with string manipulation in Java and Python, so I just followed the existing pattern here.  However, I'll keep this in mind if there is a need to modify this in the future.

benchmark/single_doc/run_command.hpp
kevinAlbs: Also update the `0.16` value on L34 to `0.13`.

kevinAlbs (approved):
LGTM with one update to benchmarking tests.

PR #842, opened 2022-01-31
CXX-2458 Avoid warnings when compiling with -Wfloat-equal

oledahle:
Other code that use the mongo-cxx-driver library might very well be compiled with -Wfloat-equal or even stricter compiler settings. Please include this minor fix to avoid such problems for people using the driver.
Used standard solution in types.hpp to compare double values, to avoid the warnings produced when compiling with -Wfloat-equal.

### DIFF
--- a/src/bsoncxx/types.hpp
+++ b/src/bsoncxx/types.hpp
@@ -16,6 +16,8 @@
 
 #include <chrono>
 #include <cstring>
+#include <float.h>
+#include <math.h>
 
 #include <bsoncxx/array/view.hpp>
 #include <bsoncxx/decimal128.hpp>
@@ -106,7 +108,7 @@ struct BSONCXX_API b_double {
 /// @relatesalso b_double
 ///
 BSONCXX_INLINE bool operator==(const b_double& lhs, const b_double& rhs) {
-    return lhs.value == rhs.value;
+    return fabs(lhs.value - rhs.value) < DBL_EPSILON;
 }
 
 ///
--- a/src/bsoncxx/types.hpp
+++ b/src/bsoncxx/types.hpp
@@ -13,11 +13,16 @@
 // limitations under the License.
 
 #pragma once
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wfloat-equal"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wfloat-equal"
+#endif
 
 #include <chrono>
 #include <cstring>
-#include <cmath>
-#include <limits>
 
 #include <bsoncxx/array/view.hpp>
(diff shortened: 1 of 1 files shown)


src/bsoncxx/types.hpp
```diff
@@ -106,7 +108,7 @@ struct BSONCXX_API b_double {
 /// @relatesalso b_double
 ///
 BSONCXX_INLINE bool operator==(const b_double& lhs, const b_double& rhs) {
-    return lhs.value == rhs.value;
+    return fabs(lhs.value - rhs.value) < DBL_EPSILON;
```
chardan, line 111: Recommend change: use the Standard Library's epsilon from std::numeric_limits. (Although I'm not entirely sure what the guarantees made by b_double are-- perhaps the macro is more appropriate for some internal reason, although that would surprise me.)
https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon

oledahle: OK, switched to using  std::numeric_limits<double>::epsilon(). I also realized that the epsilon needs to be scaled to the order of magnitude of the numbers, so I copied Donald Knuth's solution for that. :-)
https://stackoverflow.com/a/253874/10743364

src/bsoncxx/types.hpp
```diff
@@ -16,6 +16,8 @@
 
 #include <chrono>
 #include <cstring>
+#include <math.h>
```
chardan, line 19: Using <cmath>, <climits>, etc. will ensure that C library symbols are in the std:: namespace-- especially since this is a header file that will be included by other programs.

oledahle: Ok, made the solution more proper C++ by including <cmath> and <limits> instead of .h versions, and using std::fabs explicitly.

kevinAlbs (changes requested):
Thank you for the contribution @oledahle.

After internal discussion, we do not want to proceed with this PR as proposed.

We do not want to make an assumption about the range of what is considered equal for two double values.

Other drivers we polled ((Node, Java, C#, Ruby, Python) with a wrapper type around a BSON double value currently do exact comparison.

libbson's [bson_compare](http://mongoc.org/libbson/current/bson_compare.html) uses memcmp to compare BSON.

To resolve the warnings, I suggest adding wrapping pragma statements like the following:


```
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wfloat-equal"
#elif defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfloat-equal"
#endif
```

```
#if defined(__clang__)
#pragma clang diagnostic pop
#elif defined(__GNUC__)
#pragma GCC diagnostic pop
#endif
```

kevinAlbs (approved):
Thank you again for the contribution @oledahle. Changes LGTM!

chardan (approved):

PR #843, opened 2022-02-07
CXX-2445 replace example.com with doesnotexist.local

kevinAlbs:
See DRIVERS-2191 for motivation for this change.


chardan (approved):
Seems ok to me; another idea to consider would be to generate a random UUID suffix, if somehow in the future /this/ also becomes a registered name. ;-D

PR #845, opened 2022-02-22
CXX-2274 support 'let' option for aggregate command

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/62101824a4cf473735b8a229/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/mongocxx/options/aggregate.cpp
+++ b/src/mongocxx/options/aggregate.cpp
@@ -38,6 +38,10 @@ void aggregate::append(bsoncxx::builder::basic::document& builder) const {
         builder.append(kvp("collation", *this->collation()));
     }
 
+    if (this->let()) {
+        builder.append(kvp("let", *this->let()));
+    }
+
     if (this->max_time()) {
         builder.append(kvp("maxTimeMS", bsoncxx::types::b_int64{this->max_time()->count()}));
     }
@@ -61,6 +65,11 @@ void aggregate::append(bsoncxx::builder::basic::document& builder) const {
 
 aggregate& aggregate::collation(bsoncxx::document::view_or_value collation) {
     _collation = std::move(collation);
+    return *this;
+}
+
+aggregate& aggregate::let(bsoncxx::document::view_or_value let) {
+    _let = std::move(let);
     return *this;
 }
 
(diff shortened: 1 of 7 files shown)


src/mongocxx/options/aggregate.cpp
```diff
@@ -64,6 +68,11 @@ aggregate& aggregate::collation(bsoncxx::document::view_or_value collation) {
     return *this;
 }
 
+aggregate& aggregate::let(bsoncxx::document::view_or_value let) {
+    _let = std::move(let);
```
chardan, line 72: Since "let" has been passed by value, you've already got a copy of it. "_let = let" is probably fine. (But, what you've written won't hurt: move() is a fancy cast, it doesn't actually "move" anything on its own.)

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -144,6 +144,8 @@ type to_type(const Element& type) {
         return bsoncxx::type::k_int64;
     if (type_str == "int")
         return bsoncxx::type::k_int32;
+    if (type_str == "double")
```
chardan, line 147: +1 good use of operator==().

src/mongocxx/options/aggregate.hpp
```diff
@@ -108,6 +108,30 @@ class MONGOCXX_API aggregate {
...
+
+    ///
+    /// Retrieves the current variable mapping for this operation.
+    ///
+    /// @return
+    ///   The current variable mapping.
+    ///
+    /// @see https://docs.mongodb.com/manual/reference/command/aggregate/
+    ///
+    const stdx::optional<bsoncxx::document::view_or_value>& let() const;
```
chardan, line 133: Nice work putting together a tricky signature-- especially cool that you marked it const with respect to the object! 

Looking at the signature for these optional<> members, I assume that the intent is to represent a member that may or may not be "present", not to restrict someone from fiddling with the value. (The returned optional<> is const, but the value it contains is not. So modifying the contents will modify the data in the object. It looks like this is intended, so I think we're ok.)

chardan (approved):
Looks good! Thank you for your work on this.

kevinAlbs (approved):

PR #846, opened 2022-02-24
CXX2183 - expose document validation failure details

chardan:
Ensure that the "details" field is included in errInfo on failure.

https://jira.mongodb.org/browse/CXX-2183

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2337,10 +2337,10 @@ TEST_CASE("read_concern is inherited from parent", "[collection]") {
     }
 }
 
-void find_index_and_validate(
-    collection& coll,
-    stdx::string_view index_name,
-    const std::function<void(bsoncxx::document::view)>& validate = [](bsoncxx::document::view) {}) {
+void find_index_and_validate(collection& coll,
+                             stdx::string_view index_name,
+                             const std::function<void(bsoncxx::document::view)>& validate =
+                                 [](bsoncxx::document::view) {}) {
     auto cursor = coll.list_indexes();
 
     for (auto&& index : cursor) {
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2851,7 +2851,7 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
     // Listen to the insertion-failed event: we want to get a copy of the server's
     // response so that we can compare it to the thrown exception later:
     apm_opts.on_command_succeeded(
-        [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
+        [&writeErrors_well_formed, &insert_succeeded](const mongocxx::events::command_succeeded_event& ev) {
             if (0 != ev.command_name().compare("insert")) {
                 return;
             }
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2813,11 +2813,8 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
 }
 
 TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
-    bool insert_succeeded = false;  // this is checked by side-effect
-
     // A helper for checking that an error document is well-formed according to our requirements:
-    auto writeErrors_well_formed =
-        [&insert_succeeded](const bsoncxx::document::view& reply_view) -> bool {
+    auto writeErrors_well_formed = [](const bsoncxx::document::view& reply_view) -> bool {
         if (!reply_view["writeErrors"]) {
             return false;
         }
@@ -2848,19 +2845,23 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
 
     auto client_opts = test_util::add_test_server_api();
 
+    // We set this by side effect in on_command_succeeded to make sure the callback was actually
+    // triggered:
+    bool insert_succeeded = false;
+
     // Listen to the insertion-failed event: we want to get a copy of the server's
     // response so that we can compare it to the thrown exception later:
-    apm_opts.on_command_succeeded(
-        [&writeErrors_well_formed, &insert_succeeded](const mongocxx::events::command_succeeded_event& ev) {
-            if (0 != ev.command_name().compare("insert")) {
-                return;
-            }
+    apm_opts.on_command_succeeded([&writeErrors_well_formed, &insert_succeeded](
+                                      const mongocxx::events::command_succeeded_event& ev) {
+        if (0 != ev.command_name().compare("insert")) {
+            return;
+        }
 
-            REQUIRE(writeErrors_well_formed(ev.reply()));
+        REQUIRE(writeErrors_well_formed(ev.reply()));
 
-            // Make sure that "we" were actually called:
-            insert_succeeded = true;
-        });
+        // Make sure that "we" were actually called:
+        insert_succeeded = true;
+    });
 
     client_opts.apm_opts(apm_opts);
 
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2813,7 +2813,6 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
 }
 
 TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
-
     // A helper for checking that an error document is well-formed according to our requirements:
     auto writeErrors_well_formed = [](const bsoncxx::document::view& reply_view) -> bool {
         if (!reply_view["writeErrors"]) {
@@ -2844,7 +2843,7 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
 
     mongocxx::options::apm apm_opts;
 
-    auto client_opts = mongocxx::options::client();
+    auto client_opts = test_util::add_test_server_api();
 
     // Listen to the insertion-failed event: we want to get a copy of the server's
     // response so that we can compare it to the thrown exception later:
@@ -2854,12 +2853,17 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
                 return;
             }
 
-            writeErrors_well_formed(ev.reply()); 
+            REQUIRE(writeErrors_well_formed(ev.reply()));
         });
 
     client_opts.apm_opts(apm_opts);
 
     auto mongodb_client = mongocxx::client(uri{}, client_opts);
+
+    if (!test_util::newer_than(mongodb_client, "5.0")) {
+        WARN("skip: test requires MongoDB server 5.0 or newer");
+        return;
+    }
 
     database db = mongodb_client["prose_test_expose_details"];
 
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2337,10 +2337,10 @@ TEST_CASE("read_concern is inherited from parent", "[collection]") {
     }
 }
 
-void find_index_and_validate(collection& coll,
-                             stdx::string_view index_name,
-                             const std::function<void(bsoncxx::document::view)>& validate =
-                                 [](bsoncxx::document::view) {}) {
+void find_index_and_validate(
+    collection& coll,
+    stdx::string_view index_name,
+    const std::function<void(bsoncxx::document::view)>& validate = [](bsoncxx::document::view) {}) {
     auto cursor = coll.list_indexes();
 
     for (auto&& index : cursor) {
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2813,7 +2813,7 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
 }
 
 TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
-    bool insert_succeeded = false;  // this is checked by side-effect in writeErrors_well_formed
+    bool insert_succeeded = false;  // this is checked by side-effect
 
     // A helper for checking that an error document is well-formed according to our requirements:
     auto writeErrors_well_formed =
@@ -2838,8 +2838,6 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
             throw std::runtime_error("no \"details\" field in \"writeErrors\"");
         }
 
-        insert_succeeded = true;
-
         return true;
     };
 
@@ -2859,6 +2857,9 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
             }
 
             REQUIRE(writeErrors_well_formed(ev.reply()));
+
+            // Make sure that "we" were actually called:
+            insert_succeeded = true;
         });
 
     client_opts.apm_opts(apm_opts);
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2813,6 +2813,7 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
 }
 
 TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
+
     // A helper for checking that an error document is well-formed according to our requirements:
     auto writeErrors_well_formed = [](const bsoncxx::document::view& reply_view) -> bool {
         if (!reply_view["writeErrors"]) {
@@ -2831,7 +2832,7 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
         }
 
         // We require the "details" field be present:
-        if (!reply_view["writeErrors"][0]["errInfo"]["details"]) {
+        if (!errdoc["errInfo"]["details"]) {
             throw std::runtime_error("no \"details\" field in \"writeErrors\"");
         }
 
@@ -2845,14 +2846,9 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
 
     // Listen to the insertion-failed event: we want to get a copy of the server's
     // response so that we can compare it to the thrown exception later:
-    mongocxx::options::apm apm_opts;
-    apm_opts.on_command_failed([&writeErrors_well_formed](const events::command_failed_event& ev) {
-        writeErrors_well_formed(ev.failure());
-    });
-
     apm_opts.on_command_succeeded(
         [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
-            writeErrors_well_formed(ev.reply());
+            if(!writeErrors_well_formed(ev.reply()))
         });
 
     client_opts.apm_opts(apm_opts);
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2842,13 +2842,19 @@ TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
     // Set up our test environment:
     instance::current();
 
+    mongocxx::options::apm apm_opts;
+
     auto client_opts = mongocxx::options::client();
 
     // Listen to the insertion-failed event: we want to get a copy of the server's
     // response so that we can compare it to the thrown exception later:
     apm_opts.on_command_succeeded(
         [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
-            if(!writeErrors_well_formed(ev.reply()))
+            if (0 != ev.command_name().compare("insert")) {
+                return;
+            }
+
+            writeErrors_well_formed(ev.reply()); 
         });
 
     client_opts.apm_opts(apm_opts);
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,124 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+...we want to be sure that we got "errInfo" and inside of that a "details" document.
+
+          auto error = e.raw_server_error()->view();
+          auto result = error["writeConcernErrors"][0]["errInfo"];
+          contains_err_info = (err_info == result.get_document().view());
+
+*
+*/
+
+        auto reply_view = ev.reply();
```
kevinAlbs, line 2845: To check for the presence of a field, this is sufficient:

```cpp
if (reply_view["writeErrors"]) {
    // writeErrors field exists.
}

if (reply_view["writeErrors"][0]["errInfo"]["details"]) {
    // writeErrors[0].errInfo.details exists.
}

// To get the value of a field as a document.
auto element = reply_view["writeErrors"][0]["errInfo"]["details"];
if (element && element.type() == bsoncxx::type::k_document) {
    auto document = element.get_document ();
    std::cout << "details document is " << bsoncxx::to_json (document.view()) << std::endl;
}
```

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,124 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+    REQUIRE(coll.insert_one(entry.view()));
+    }
+
+    // But now, cause a type violation on create:
+    {
+    bsoncxx::builder::basic::document entry;
+
+    entry.append(kvp("_id", bsoncxx::oid()), kvp("field_x", 42));
+
+    REQUIRE_THROWS(coll.insert_one(entry.view()));
```
kevinAlbs, line 2901: To satisfy this requirement:

> Then, insert an invalid document (e.g.
``{x: 1}``) and assert that a WriteError occurs, that its code is ``121``
(i.e. DocumentValidationFailure), and that its ``details`` property is
accessible.

I think the test with an invalid insert should updated to use an explicit try / catch. The test should assert that the `details` field is accessible on the exception.

src/mongocxx/test/collection.cpp
chardan: +1 super helpful, thank you!

src/mongocxx/test/collection.cpp
chardan: +1

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,145 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+    // Set up our test environment:
+    instance::current();
+
+    auto client_opts = mongocxx::options::client();
+
+    // Listen to the insertion-failed event: we want to get a copy of the server's
+    // response so that we can compare it to the thrown exception later:
+    mongocxx::options::apm apm_opts;
+    apm_opts.on_command_failed([&writeErrors_well_formed](const events::command_failed_event& ev) {
+        writeErrors_well_formed(ev.failure());
```
chardan, line 2850: ...although I actually think we should never see this? That is, in this particular set of tests, getting a command_failed would be an test failure?

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,145 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+        filter.append(kvp("_id", oid_token));
+        entry.append(kvp("$set", make_document(kvp("field_x", "value_updated"))));
+
+        REQUIRE_NOTHROW(coll.update_one(filter.view(), entry.view()));
+    }
+
+    SECTION("should NOT be able to update valid object with an invalid type") {
+        auto oid_token = bsoncxx::oid();
+
+        // Construct a valid entry to update:
```
chardan, line 2925: It might be nice to put this "construct valid entry" stuff into a function; I'm slightly on the fence at the moment. Currently would rather make sure the test details are right, but it wouldn't take much to convince me.

src/mongocxx/test/collection.cpp
kevinAlbs: Write errors are not considered a command failure. A server reply with `ok: 0` is considered a command failure. A reply with writeErrors still has `ok: 1`:

```
{
  n: 0,
  electionId: ObjectId("7fffffff0000000000000041"),
  opTime: { ts: Timestamp({ t: 1645749523, i: 1 }), t: Long("65") },
  writeErrors: [
    {
      index: 0,
      code: 121,
      errInfo: {
        failingDocumentId: ObjectId("621825131c6f87414554373c"),
        details: {
          operatorName: '$type',
          specifiedAs: { x: { '$type': 'string' } },
          reason: 'type did not match',
          consideredValue: 1,
          consideredType: 'int'
        }
      },
      errmsg: 'Document failed validation'
    }
  ],
  ok: 1,
  '$clusterTime': {
    clusterTime: Timestamp({ t: 1645749523, i: 2 }),
    signature: {
      hash: Binary(Buffer.from("0000000000000000000000000000000000000000", "hex"), 0),
      keyId: Long("0")
    }
  },
  operationTime: Timestamp({ t: 1645749523, i: 1 })
}
```

Listening for a `CommandFailed` event is not a requirement of the test. Only `CommandSucceeded`:

> Additionally, assert that a CommandSucceededEvent was observed and
that the ``writeErrors[0].errInfo`` field in the response document matches the
WriteError's ``details`` property.

chardan: +1 agreed!

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,145 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+
+        // The code should always be 121 (DocumentValidationFailure):
+        if (121 != error_code) {
+            std::ostringstream os;
+            os << "writeErrors expected to have code 121, but had " << error_code << " instead";
+            throw std::runtime_error(os.str());
+        }
+
+        // We require the "details" field be present:
+        if (!reply_view["writeErrors"][0]["errInfo"]["details"]) {
```
chardan, line 2834: Should be errdoc[...]. Will fix.

src/mongocxx/test/collection.cpp
chardan: +1 done.

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,141 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+
+        try {
+            coll.insert_one(entry.view());
+
+            // We should not make it here:
+            CHECK(false);
+        } catch (const operation_exception& e) {
+            auto rse = e.raw_server_error();
+
+            CHECK(rse.has_value());
```
kevinAlbs, line 2892:
```suggestion
            CHECK(rse);
```

Some Evergreen tasks ([example](https://spruce.mongodb.com/task/cxx_driver_arm_ubuntu1804_50_compile_and_test_with_shared_libs_patch_a7be2f22f2cd4c8664e48abae4f78b6f1c9d6e99_621816e77742ae51fc2669a0_22_02_24_23_38_16/logs?execution=0)) do not recognize has_value:

```
[2022/02/24 23:51:53.108] /data/mci/a42c20c4517b667a13c1bc365b509680/mongo-cxx-driver/src/mongocxx/test/collection.cpp:2896:23: error: 'struct core::v1::optional<bsoncxx::v_noabi::document::value>' has no member named 'has_value'; did you mean 'value'?
[2022/02/24 23:51:53.108]              CHECK(rse.has_value());
[2022/02/24 23:51:53.108]                        ^
[2022/02/24 23:51:53.108] In file included from /data/mci/a42c20c4517b667a13c1bc365b509680/mongo-cxx-driver/src/mongocxx/../bsoncxx/test_util/catch.hh:17:0,
[2022/02/24 23:51:53.108]                  from /data/mci/a42c20c4517b667a13c1bc365b509680/mongo-cxx-driver/src/mongocxx/test/collection.cpp:27:
[2022/02/24 23:51:53.108] /data/mci/a42c20c4517b667a13c1bc365b509680/mongo-cxx-driver/src/mongocxx/test/collection.cpp:2896:23: error: 'struct core::v1::optional<bsoncxx::v_noabi::document::value>' has no member named 'has_value'; did you mean 'value'?
[2022/02/24 23:51:53.108]              CHECK(rse.has_value());
[2022/02/24 23:51:53.108]                        ^
[2022/02/24 23:51:53.108] /data/mci/a42c20c4517b667a13c1bc365b509680/mongo-cxx-driver/src/mongocxx/test/collection.cpp:2896:23: error: 'struct core::v1::optional<bsoncxx::v_noabi::document::value>' has no member named 'has_value'; did you mean 'value'?
[2022/02/24 23:51:53.108]              CHECK(rse.has_value());
[2022/02/24 23:51:53.108]     
```

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,141 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+
+            CHECK(writeErrors_well_formed(*rse));
+        } catch (...) {
+            // An exception was thrown, but of the wrong type:
+            CHECK(false);
+        }
+    }
+
+    SECTION("can update valid object with a valid value type") {
```
kevinAlbs, line 2901: The only test case that is necessary to satisfy requirements is "cause a type violation on insert"

> Enable `command monitoring <../../command-monitoring/command-monitoring.rst>`_
to observe CommandSucceededEvents. Then, insert an invalid document (e.g.
``{x: 1}``) and assert that a WriteError occurs, that its code is ``121``
(i.e. DocumentValidationFailure), and that its ``details`` property is
accessible. Additionally, assert that a CommandSucceededEvent was observed and
that the ``writeErrors[0].errInfo`` field in the response document matches the
WriteError's ``details`` property.

Each of the write commands go through `bulk_op.execute()`. I suggest removing the extra tests to simplify.

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,141 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+    // Set up our test environment:
+    instance::current();
+
+    auto client_opts = mongocxx::options::client();
+
+    // Listen to the insertion-failed event: we want to get a copy of the server's
+    // response so that we can compare it to the thrown exception later:
+    apm_opts.on_command_succeeded(
+        [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
+            if(!writeErrors_well_formed(ev.reply()))
```
kevinAlbs, line 2851: This callback will be called on every command run on the client. That includes the "drop" command, the insert in the "insert a valid document" test.

To simplify this check, I suggest limiting this to "insert":
```suggestion
            if (0 != ev.command_name().compare("insert")) {
                return;
            }
            if(!writeErrors_well_formed(ev.reply()))
```

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,141 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+            coll.insert_one(entry.view());
+
+            // We should not make it here:
+            CHECK(false);
+        } catch (const operation_exception& e) {
+            auto rse = e.raw_server_error();
+
+            CHECK(rse.has_value());
+
+            CHECK(writeErrors_well_formed(*rse));
```
kevinAlbs, line 2894: It is an error if no exception is thrown. I suggest adding a boolean flag to assert an `operation_exception` is thrown.  Something like:
```cpp
bool operation_exception_thrown = false;

try {
    coll.insert_one(entry.view());

    // We should not make it here:
    CHECK(false);
} catch (const operation_exception& e) {
    exception_thrown = true;
    // ...
}

CHECK (operation_exception_thrown);
```

src/mongocxx/test/collection.cpp
kevinAlbs: I suggest removing this test. The only necessary test is "cause a type violation on insert".

src/mongocxx/test/collection.cpp
chardan: Ah, shoot, that's because we're using C++11! My local build turns on C++17 (it looks like the reason is that the cxx::* support for whatever reason doesn't compile on my system with:

cmake -DCMAKE_BUILD_TYPE=Debug -DLIBBSON-1.0_DIR=~/work/c-driver-install/lib/cmake/libbson-1.0/ -DCMAKE_INSTALL_PREFIX=~/work/cxx-driver-install -DCMAKE_PREFIX_PATH="~/work/c-driver-install;~/work/cxx-driver-install" -DCMAKE_CXX_STANDARD=11 -DCMAKE_C_COMPILER="/usr/bin/gcc-11" -DCMAKE_CXX_COMPILER="/usr/bin/g++-11" ../

src/mongocxx/test/collection.cpp
chardan: My code above should catch this, under "We should not make it here". Catch2 provides some helpful macros, but unfortunately the version that we'd need in this particular case (since we have to do extra checking on the value) requires more manipulation than I see provided by the library. I /think/ the case is covered, though.

chardan (commented):
I have moved this PR out of draft.

src/mongocxx/test/collection.cpp
chardan: +1

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,90 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+
+        return true;
+    };
+
+    // Set up our test environment:
+    instance::current();
+
+    mongocxx::options::apm apm_opts;
+
+    auto client_opts = mongocxx::options::client();
```
kevinAlbs, line 2847:
```suggestion
    auto client_opts = test_util::add_test_server_api();
```

Like the C driver, the C++ driver conditionally tests with the server API version. This may resolve the task failure on [Ubuntu 18.04 Debug Latest Versioned API](https://spruce.mongodb.com/version/621928ed5623437eeb06c925/tasks?page=0&variant=%5Eversioned%20api__os~ubuntu-1804_mongodb_version~latest%24).

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,90 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+                return;
+            }
+
+            writeErrors_well_formed(ev.reply()); 
+        });
+
+    client_opts.apm_opts(apm_opts);
+
+    auto mongodb_client = mongocxx::client(uri{}, client_opts);
+
```
kevinAlbs, line 2863: The spec says:
> Using a 5.0+ server

This test is failing on pre 5.0 servers.
```suggestion
if (!test_util::newer_than(mongodb_client, "5.0")) {
    WARN("skip: test requires MongoDB server 5.0 or newer");
    return;
}
```

src/mongocxx/test/collection.cpp
kevinAlbs: Good point. Resolving.

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,90 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+    // Listen to the insertion-failed event: we want to get a copy of the server's
+    // response so that we can compare it to the thrown exception later:
+    apm_opts.on_command_succeeded(
+        [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
+            if (0 != ev.command_name().compare("insert")) {
+                return;
+            }
+
+            writeErrors_well_formed(ev.reply());
```
kevinAlbs, line 2857: Is there an assertion to ensure that this callback is called?

kevinAlbs (changes requested):
I have suggested fixes for tests.

In the future, please try to investigate new test failures before requesting review.

src/mongocxx/test/collection.cpp
chardan: +1 I didn't see this locally and indeed put the cart before the horse-- apologies.

src/mongocxx/test/collection.cpp
chardan: +1 done

src/mongocxx/test/collection.cpp
chardan: There is one now. :-)

chardan:
From what I can see in the tests, these are looking better. Please let me know if I have missed something! Thank you.

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,94 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
...
+
+    // Listen to the insertion-failed event: we want to get a copy of the server's
+    // response so that we can compare it to the thrown exception later:
+    apm_opts.on_command_succeeded(
+        [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
+            if (0 != ev.command_name().compare("insert")) {
+                return;
+            }
+
+            REQUIRE(writeErrors_well_formed(ev.reply()));
```
kevinAlbs, line 2856: To satisfy this requirement:

> Additionally, assert that a CommandSucceededEvent was observed

Please add a check to ensure this callback was called.

For example, one suggestion is to add a boolean flag, and check it at the end of the test:

```cpp
bool insert_succeeded = false;

apm_opts.on_command_succeeded(
        [&writeErrors_well_formed](const mongocxx::events::command_succeeded_event& ev) {
            if (0 != ev.command_name().compare("insert")) {
                return;
            }
            insert_succeeded = true;

            REQUIRE(writeErrors_well_formed(ev.reply()));
}
/* ... */
SECTION("cause a type violation on insert") {
        /* ... */
        REQUIRE(insert_succeeded);
}
```


If the command resulted in a different error producing a CommandFailedEvent, this callback may not be called.

kevinAlbs (changes requested):
To fix the `lint` task, please run:

```
python2 ./etc/clang_format.py format
```

I request one additional assert the `on_command_succeeded` callback is called.

Otherwise looks good.

src/mongocxx/test/collection.cpp
chardan: +1 I see what you are asking for now-- done!

kevinAlbs: https://github.com/mongodb/mongo-cxx-driver/pull/846/commits/0d0118245cedebb421cf4ce4ee36753fd8eba3ca#diff-481e7ba83bde51981cec950f1b584f37f027c4cb1ee9806d0cfbb73b90483153R2820 adds `insert_succeeded` to the `writeErrors_well_formed` function. The suggestion is to add it to the callback of `on_command_succeeded`.

`writeErrors_well_formed` is called in both the callback to `on_command_succeeded` and in the catch block on L2896.

Please move the `insert_succeeded = true` from `writeErrors_well_formed` to the callback of `on_command_succeeded`.

src/mongocxx/test/collection.cpp
```diff
@@ -2808,4 +2812,103 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
     REQUIRE(contains_err_info);
 }
 
+TEST_CASE("expose writeErrors[].errInfo", "[collection]") {
+    bool insert_succeeded = false;  // this is checked by side-effect
+
+    // A helper for checking that an error document is well-formed according to our requirements:
+    auto writeErrors_well_formed =
+        [&insert_succeeded](const bsoncxx::document::view& reply_view) -> bool {
```
kevinAlbs, line 2820: The `insert_succeeded` capture is no longer used.

This is causing a test failure on [this task](https://spruce.mongodb.com/task/cxx_driver_ubuntu1804_debug_asan_50_compile_and_test_with_static_libs_extra_alignment_patch_a7be2f22f2cd4c8664e48abae4f78b6f1c9d6e99_621cfd0fc9ec4457e1f6d574_22_02_28_16_49_21/logs?execution=0):

```
[2022/02/28 18:24:14.868] found libmongoc version 1.22.0-pre
[2022/02/28 18:24:14.868] /data/mci/641d255fa2451925acd2c0de5d0c083e/mongo-cxx-driver/src/mongocxx/test/collection.cpp:2820:11: error: lambda capture 'insert_succeeded' is not used [-Werror,-Wunused-lambda-capture]
[2022/02/28 18:24:14.868]         [&insert_succeeded](const bsoncxx::document::view& reply_view) -> bool {
```

kevinAlbs (changes requested):
There is still a test failure introduced by this patch.

Please check tasks that with a failing `Patch Status` but passing `Base Status` to check for regressions before requesting review. You can sort by Patch Status on the [Patch page](https://spruce.mongodb.com/version/621cfd0fc9ec4457e1f6d574/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3AASC).

The remaining task failures look unrelated. I filed CXX-2459 to track the compilation error with ccache. The macOS tasks failing to find boost are tracked in CXX-2359.

src/mongocxx/test/collection.cpp
chardan: Sorry to have missed this, I thought I had looked at all of them-- I had not! I'm also going to try to see why I'm not seeing this locally... thank you for your patience.

chardan: Looks like my local CXX build flags are extremely permissive compared to my CFLAGS-- modifying.

kevinAlbs (approved):
Please run clang-format to fix the [lint task](https://spruce.mongodb.com/task/cxx_driver_ubuntu1804_release_latest_lint_patch_a7be2f22f2cd4c8664e48abae4f78b6f1c9d6e99_621d4ca5850e613d841cbfcb_22_02_28_22_28_55/logs?execution=0) before merging.

Otherwise LGTM.

PR #863, opened 2022-06-01
CXX-2464 Use count instead of collStats for estimatedDocumentCount

eramongodb:
### Description

This PR resolves CXX-2464, which is unblocked by CDRIVER-4309.

In addition to updating the spec test files, the C++ unified test runner had to be updated to handle the `viewOn` and `pipeline` argument fields to `createCollection`, which were previously being ignored.

### Version String Comparison

The unified test runner used the same comparison logic for server version strings and JSON schema version strings. However, server version string comparison _must not_ ignore patch numbers, whereas JSON schema version string comparison _must_ ignore patch numbers. This PR includes a drive-by fix to version string comparison in the test runner to correctly distinguish between the two cases.

### Unsupported Test Files

As a drive-by improvement, the warning indicating an unsupported test file is being skipped was moved into the test section of the test file being skipped to avoid redundant duplication of skip warnings in the test output.

For example, before the change, the following block was printed repeatedly over 50 times with varying subsets of skipped test files in each block. More often than not, all files were included, exacerbating the test output noise. Only the first three test files in a given block is included here for brevity:

```
retryable reads spec tests
.src/mongocxx/test/spec/retryable-reads.cpp:122

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: count.json

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: count-serverErrors.json

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: gridfs-downloadByName.json
  
...

retryable reads spec tests
.src/mongocxx/test/spec/retryable-reads.cpp:122

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: count.json

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: count-serverErrors.json

.src/mongocxx/test/spec/util.cpp:454: warning:
  Skipping unsupported test file: gridfs-downloadByName.json
  
...
```

After the change, each unsupported test file emits a single warning, _once only_, for the whole test run, dramatically reducing the length of the test output:

```
retryable reads spec tests
  count.json
./src/mongocxx/test/spec/util.cpp:453

./src/mongocxx/test/spec/util.cpp:455: warning:
  Skipping unsupported test file: count.json

retryable reads spec tests
  count-serverErrors.json
./src/mongocxx/test/spec/util.cpp:453

./src/mongocxx/test/spec/util.cpp:455: warning:
  Skipping unsupported test file: count-serverErrors.json

retryable reads spec tests
  gridfs-downloadByName.json
./src/mongocxx/test/spec/util.cpp:453

./src/mongocxx/test/spec/util.cpp:455: warning:
  Skipping unsupported test file: gridfs-downloadByName.json

...
```

### Add support for pinning mongo-c-driver version to commit hash

Added support for pinning the mongo-c-driver repository by arbitrary Git blobs (branch, tag, or commit hash) instead of being limited to branch/tag names. This allows pinning to arbitrary commits (i.e. latest commit in mongo-c-driver) without being required to create a corresponding tag in the mongo-c-driver repo.

If the requested version string does not comply with SemVer (e.g. a branch name such as `master` or a commit hash), the mongo-c-driver `BUILD_VERSION` is derived by obtaining the tag name of the latest mongo-c-driver release (which is assumed to be a SemVer string) and bumping the minor version number), where "latest" is as defined by [the GitHub API](https://docs.github.com/en/rest/releases/releases#get-the-latest-release). e.g. if the latest release is `1.21.1` (as it is at time of writing), `BUILD_VERSION` is set to `1.22.0-pre-<version>` where `<version>` is `$VERSION` (which may be a branch, tag name, commit hash, etc.).

### DIFF
--- a/data/retryable-reads/test_files.txt
+++ b/data/retryable-reads/test_files.txt
@@ -15,8 +15,8 @@ distinct.json
 distinct-serverErrors.json
 estimatedDocumentCount-4.9.json
 estimatedDocumentCount-serverErrors-4.9.json
-estimatedDocumentCount-pre4.9.json
-estimatedDocumentCount-serverErrors-pre4.9.json
+estimatedDocumentCount.json
+estimatedDocumentCount-serverErrors.json
 find.json
 findOne.json
 findOne-serverErrors.json
--- a/.evergreen/install_c_driver.sh
+++ b/.evergreen/install_c_driver.sh
@@ -32,7 +32,7 @@ echo "About to install C driver ($VERSION) into $PREFIX"
 
 LIB=mongo-c-driver
 rm -rf $(echo $LIB*)
-curl -o $LIB.zip -L https://github.com/mongodb/$LIB/archive/$VERSION.zip
+curl -sS -o $LIB.zip -L https://github.com/mongodb/$LIB/archive/$VERSION.zip
 unzip -q $LIB.zip
 DIR=$(echo $LIB-*)
 
@@ -45,7 +45,7 @@ else
     # Otherwise, use the tag name of the latest release to construct a prerelease version string.
 
     # Extract "tag_name" from latest Github release.
-    BUILD_VERSION=$(curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/mongodb/mongo-c-driver/releases/latest | jq -r ".tag_name")
+    BUILD_VERSION=$(curl -sS -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/mongodb/mongo-c-driver/releases/latest | jq -r ".tag_name")
 
     # Assert the tag name is a SemVer string.
     test $(echo $BUILD_VERSION | grep -P "$SEMVER_REGEX")
--- a/.evergreen/install_c_driver.sh
+++ b/.evergreen/install_c_driver.sh
@@ -31,18 +31,33 @@ CMAKE_ARGS="
 echo "About to install C driver ($VERSION) into $PREFIX"
 
 LIB=mongo-c-driver
-if [ -n "$(echo "${VERSION}" | grep '^[a-z]' )" ]; then
-    rm -rf $LIB
-    # Must be http as rhel55 has https issues
-    curl -o $LIB.tgz -L http://s3.amazonaws.com/mciuploads/$LIB/$VERSION/$LIB-latest.tar.gz
-    tar --extract --file $LIB.tgz
-    rm -rf $LIB
-    DIR=$(echo $LIB-*)
+rm -rf $(echo $LIB*)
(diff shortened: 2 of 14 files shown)


rcsanchez97 (approved):
LGTM

data/retryable-reads/test_files.txt
```diff
@@ -15,8 +15,8 @@ distinct.json
 distinct-serverErrors.json
 estimatedDocumentCount-4.9.json
 estimatedDocumentCount-serverErrors-4.9.json
-estimatedDocumentCount-pre4.9.json
-estimatedDocumentCount-serverErrors-pre4.9.json
+estimatedDocumentCount.json
+estimatedDocumentCount-serverErrors.json
 find.json
```
kevinAlbs, line 20: The Atlas Data Lake test file has also changed:
https://github.com/mongodb/specifications/commit/c43037606247a689cb5ada0099b8628a2c495bdd#diff-8170be3680c3fc86d77f1c4d3397c6c55c4438bcb6ae1b9a36965272c241044f

In the C++ driver, that file is in ./data/mongohouse/estimatedDocumentCount.json. `mongohouse` is the name of the process used for Atlas Data Lake. The change does not look significant, but I suggest updating it too since it is in the scope of the requested changes.

kevinAlbs (approved):
Nice fix of the version string comparison.

LGTM with minor changes:

- The `lint` task is broken on base too, but I suggest running the clang_format script to fix: `python2 ./etc/clang_format.py format`
- There are [documentation requirements included in the spec change](https://github.com/mongodb/specifications/commit/c43037606247a689cb5ada0099b8628a2c495bdd#diff-2fe314a3da48fca10d3a170b09a5a393a93dbb4e9369d6741c99656a95dd7034R762). I do not think the requirement for release notes applies since the C++ driver has not released a 5.0 compatible release. I suggest adding the other documentation requirements to collection.hpp.
- The `ubuntu1804-min-libmongoc` variant is failing because it uses the C driver 1.19.0. I suggest updating [the tested version to pin to a commit of the C driver](https://github.com/mongodb/mongo-cxx-driver/blob/1afd5102fd63322fa448f490d5f0dbaee78374bb/.mci.yml#L1343) and file a ticket with fix version 3.7.0 to raise the minimum required C driver to 1.22.0 after a stable C 1.22.0 is released.

eramongodb:
Added support for pinning the mongo-c-driver version to an arbitrary Git blob (branch name, tag name, commit hash, etc.) to permit pinning to the latest commit without having to create tags in the mongo-c-driver repository.

PR #866, opened 2022-06-10
CXX-2229 Convert CRUD v2 spec tests to unified test format

eramongodb:
## Description

This PR syncs all CRUD spec tests (both legacy and unified) to the state of the Drivers spec as of [37ce3668](https://github.com/mongodb/specifications/tree/37ce3668631c846ce168c9349f6af5ab484e3598/source/crud/tests) (2021-05-25).

Due to sitting in the backlog for an extended period of time, many unresolved CXX tickets pertaining to CRUD tests have either been invalidated or superceded by newer tickets. Rather than stepping through each CXX ticket individually, this PR instead fast-forwards through all the CXX, DRIVERS, and SPEC tickets that were applied to the CRUD spec tests straight to its state as of [37ce3668](https://github.com/mongodb/specifications/tree/37ce3668631c846ce168c9349f6af5ab484e3598/source/crud/tests) (2021-05-25).

This PR resolves the following CXX tickets:

- CXX-1714 Resync CRUD spec tests to use transaction spec test format
- CXX-1797 Database-level aggregation should be tested on server versions 3.6+
- CXX-1824 Add missing tests for aggregation pipeline in bulk write updates
- CXX-1945 separate BulkWrite tests by operation
- CXX-1975 Tests drivers raise error if hint specified for unacknowledged delete using OP_MSG or OP_DELETE
- CXX-2002 Raise error when hint option is provided on unacknowledged writes against any server version
- CXX-2019 Test arrayFilters on non-supporting server for subsequent ops in bulkWrite
- CXX-2057 Add tests for update/replacement validation in single and bulk write operations
- CXX-2229 Convert CRUD v2 spec tests to unified test format

There may be additional CXX tickets that are applicable that I have missed.

## Allowing hint for unacknowledged writes using OP_MSG

CXX-1975 and CXX-2002 introduce the `unacknowledged-*-hint-clientError` unified test files. These test files are subsequently removed by CXX-2347, which requires syncing up to [59a5dadb](https://github.com/mongodb/specifications/tree/59a5dadb2eeab8bba5a7d9a9acd47aa8c1d594cf/source/crud/tests) (2021-08-31). To keep the timeline of CRUD tests syncs somewhat orderly, CXX-2347 will be applied separately after this PR. Until then, the corresponding `unacknowledged-*-hint-clientError` tests are skipped by the unified test runner.

## Unmodified Tests

The following test files are deliberately left unmodified as they are already synced to newer versions of the spec:

- 3a58fc61 (2021-06-10): unified/aggregate-let.json (CXX-2274)
- 343a56fe (2022-02-23): unified/bulkWrite-*-let.json (CXX-2456)
- 021cbc80 (2022-05-02): unified/estimatedDocumentCount.json (CXX-2464)

## Updates to the Unified Test Runner

There are several changes made to the unified test runner to support or improve the set of CXX tickets being resolved:

### Improve test failure message on missing field

The BSON document match asserter reported a missing required field like so:
```
assert.cpp:249: FAILED:
  REQUIRE( actual_doc[kvp.key()] )
with expansion:
  {?}
with messages:
  description := "bulkWrite-delete-hint"
  print_all() := "..."
```

With this change, the error message looks as follows:
```
assert.cpp:251: FAILED:
explicitly with messages:
  description := "bulkWrite-delete-hint"
  print_all() := "..."
  expected field 'hint' to be present, but it is absent
```

### Fix `get_hostnames()` to account for single server topology

The `REQUIRE(after_slash < std::end(host));` assertion in [get_hostnames()](https://github.com/mongodb/mongo-cxx-driver/blob/master/src/mongocxx/test/spec/unified_tests/runner.cpp#L274) was being triggered when the unified test runner was executed against a single server topology due to the insufficient [test_util::is_replica_set()](https://github.com/mongodb/mongo-cxx-driver/blob/master/src/mongocxx/test/spec/unified_tests/runner.cpp#L259) condition. The condition is improved to actually match the condition described by the associated comment by using a new `test_util::is_sharded_cluster()` function.

### Unified test runner support for features supported by legacy test runner

Several fields or behaviors that are already supported by the C++ driver and the legacy test runner(s) were not implemented by the unified test runner. Some legacy tests that have been converted to unified tests now required support for this behavior in the unified test runner. The fields/behaviors include:

- The hint field was not being forwarded to bulkWrite operations in the unified test runner.
- The `upsertedId` field was not being captured in the operation result by the unified test runner.
- The unified test runner did not support the `update` field in update operations being an array corresponding to pipeline stages.


src/mongocxx/test_util/client_helpers.cpp
```diff
@@ -248,6 +248,11 @@ bool is_replica_set(const client& client) {
     return static_cast<bool>(reply.view()["setName"]);
 }
 
+bool is_sharded_cluster(const client& client) {
+    auto reply = get_is_master(client);
+    return static_cast<bool>(reply.view()["config"]["sharded"]);
```
kevinAlbs, line 253: Suggest using the check from [Server Discovery and Monitoring](https://github.com/mongodb/specifications/blob/ba44473e2488290e13acabbea0f32fa8c2bc4b7a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#type).

```suggestion
    if (!reply.view()["msg"]) {
        return false;
    }
    return 0 == reply.view()["msg"].get_string().value.compare ("isdbgrid");
```

This is also used in the C driver [test_framework_is_mongos](https://github.com/mongodb/mongo-c-driver/blob/0636093de1e11b1259f75f21706b906bb11ae382/src/libmongoc/tests/test-libmongoc.c#L2030-L2046) function.

src/mongocxx/result/update.hpp
```diff
@@ -53,6 +53,12 @@ class MONGOCXX_API update {
     /// @return The number of documents that were modified.
     std::int32_t modified_count() const;
 
+    ///
+    /// Gets the number of documents that were upserted during this operation.
+    ///
+    /// @return The number of documents that were upserted.
+    std::int32_t upserted_count() const;
```
kevinAlbs, line 60: No change requested. Was a `std::int64_t` return type considered?

`std::int32_t` is consistent with `result::bulk_write` and the other fields in this class. The CRUD spec uses `Int64`. I think the proposed `std::int32_t` seems like less surprising API.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1068,54 +1096,54 @@ document::value update_one(collection& coll, client_session* session, document::
...
     }
 
     auto result = builder::basic::document{};
-    result.append(builder::basic::kvp(
-        "result",
-        [matched_count, modified_count, upserted_id](builder::basic::sub_document subdoc) {
-            subdoc.append(builder::basic::kvp("matchedCount", matched_count));
+    result.append(builder::basic::kvp("result", [&](builder::basic::sub_document subdoc) {
+        if (!update_one_result) {
+            subdoc.append(builder::basic::kvp("matchedCount", 0));
```
kevinAlbs, line 1128: Suggest removing the `matchedCount`.

```suggestion
```

`update_one_result` is [only expected to be unset if an unacknowledged write concern is used](https://github.com/mongodb/mongo-cxx-driver/blob/09d9d7e736b86724c21e4e41e57c1112ebf5656b/src/mongocxx/collection.hpp#L1611-L1613).

Returning `"matchedCount": 0` suggests that the server replied. An unacknowledged write concern does not have a server reply.

The [updateOne-hint-unacknowledged.yml](https://github.com/mongodb/specifications/blob/ba44473e2488290e13acabbea0f32fa8c2bc4b7a/source/crud/tests/unified/updateOne-hint-unacknowledged.yml#L69) test has an assertion on the result of an unacknowledged update. It only checks for the optional `acknowledged` field.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1131,40 +1159,45 @@ document::value update_many(collection& coll, document::view operation) {
...
-        [matched_count, modified_count, upserted_count, upserted_id](
-            builder::basic::sub_document subdoc) {
-            subdoc.append(builder::basic::kvp("matchedCount", matched_count));
-
-            if (modified_count) {
-                subdoc.append(builder::basic::kvp("modifiedCount", *modified_count));
-            }
+    result.append(builder::basic::kvp("result", [&](builder::basic::sub_document subdoc) {
+        if (!update_many_result) {
+            subdoc.append(builder::basic::kvp("matchedCount", 0));
```
kevinAlbs, line 1189: Suggest removing `matchedCount`.

```suggestion
```

kevinAlbs (approved):
LGTM. Nicely done. The error message improvement for missing fields seems helpful for debugging test failures. Left suggestions that do not block merge.

src/mongocxx/result/update.hpp
eramongodb: > Was a `std::int64_t` return type considered?

It was not considered. `std::int32_t` was chosen simply because it is the type being returned by `bulk_write::upserted_count()`. If `Int64` needs to be supported, more interfaces than just that of `update` will likely need to be modified.

src/mongocxx/test/spec/unified_tests/operations.cpp
eramongodb: Will apply suggestion.

Note: this line was included to preserve this prior behavior where `matchedCount` was unconditionally included in the result:

```cpp
std::int32_t matched_count = 0;
// ...

if (session) {
    auto update_one_result = coll.update_one(*session, filter, update, options);
    if (update_one_result) {
        matched_count = update_one_result->matched_count();
        // ...
    }
} else { /* ... */ }

auto result = builder::basic::document{};
result.append(builder::basic::kvp(
    "result",
    [/* ... */](builder::basic::sub_document subdoc) {
        subdoc.append(builder::basic::kvp("matchedCount", matched_count));
        // ...
    }));
```

rcsanchez97 (approved):

PR #867, opened 2022-06-22
CXX-2186 Mitigate pain of using field names with dots and dollars

eramongodb:
## Description

This PR resolves CXX-2186.

CRUD unified test files are synced with [44f83106](https://github.com/mongodb/specifications/tree/44f83106033a609f2af40e3e4d99852a03096ba5/source/crud/tests/unified). `estimatedDocumentCount.json` is deliberately left unmodified as it is already synced to newer versions of the spec.

As far as I am aware, there are no restrictions for encoding BSON documents with dollar-prefixed or dotted keys to remove in the C++ driver itself. There exists [interfaces](https://github.com/mongodb/mongo-cxx-driver/blob/03a6b759e4d1946c36dd3cbc686d2ef520eb598d/src/bsoncxx/validate.hpp#L122) and [corresponding tests](https://github.com/mongodb/mongo-cxx-driver/blob/03a6b759e4d1946c36dd3cbc686d2ef520eb598d/src/bsoncxx/test/bson_validate.cpp#L80) to _opt into_ validating or restricting their presence that are left as-is.

As instructed, possible silent rejection of dollar-prefixed or dotted keys by pre-5.0 servers was documented as a warning as part of `write_concern::acknowledge_level()`.

## Matching `$$unsetOrMatches` as Root

The `insertOne-dots_and_dollars.json` and `insertOne-dots_and_dollars.json` test files introduced a novel syntax for the `expectResult` document:

```json
"expectResult": {
  "$$unsetOrMatches": {
    "insertedId": {
      "$$unsetOrMatches": 1
    }
  }
}
```

The unified test runner was incorrectly treating the "matches" document within the root `$$unsetOrMatches` document as a non-root document. The unified test runner was updated to forward the (non-)root status of the `$$unsetOrMatches` document when matching against the "matches" document.

## Unified Test Runner Improvements

Several drive-by improvements to the unified test runner were applied.

- The "extra fields in non-root document" assertion was adjusted to indicate the number of extra fields present in the document.
- Ensured the expected document to be matched against is always printed alongside the actual document being matched.

## Pinning libmongoc and libmongocrypt Versions

Ongoing development of both libmongoc and libmongocrypt were causing build failures on Evergreen due to changes unrelated to the C++ driver itself. This PR seeks to improve the situation by explicitly pinning all versions of libmongoc and libmongocrypt being depended upon by the C++ driver in Evergreen.

Further updates to the C++ driver are expected to bump the new `mongoc_version_default` and `mongocrypt_version_default` Evergreen variables as necessary. They are currently set to the latest commits of each repository with substantial feature changes.

The `mongoc_version_minimum` and `mongocrypt_version_minimum` variables are only used by the `ubuntu1804-min-libmongoc` task, but were moved/defined alongside the `*_version_default` variables for locality.


rcsanchez97 (approved):
LGTM

.evergreen/install_c_driver.sh
```diff
@@ -1,19 +1,20 @@
 #!/bin/bash
 # Usage examples:
-# ./install.sh master
-# ./install.sh 1.3.5
-# PREFIX=/tmp/installdir ./install.sh 1.3.5
+# MONGOC_VERSION=1.21.2  MONGOCRYPT_VERSION=1.4.1 ./install.sh
```
kevinAlbs, line 3:
```suggestion
```

Remove duplicate line.

.mci.yml
```diff
@@ -7,6 +7,13 @@
 #######################################
 variables:
 
+    mongoc_version_default: &mongoc_version_default "fe2bae72" # Accept string arguments for queryType and "index_type"
+    mongocrypt_version_default: &mongocrypt_version_default "638b8090" # Update use of removed API
+
+    # Remember to update .github/workflows/compile-and-test-latest.yml.
```
kevinAlbs, line 13:
```suggestion
```

kevinAlbs (approved):
The test runner changes look helpful for debugging test assertion errors. And I am very much in support of pinning to versions of the C driver and libmongocrypt.

Suggest removing the documentation note "Such keys are reserved for MongoDB internal use." from `check_dollar_keys`.

PR #869, opened 2022-06-22
CXX-2202 Remove variant Ubuntu 18.04 s390x

eramongodb:
## Description

This PR resolves CXX-2202.

The only variant among the list to remove is Ubuntu 18.04 on s390x. All other variants were already removed or were not being tested on.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -1274,23 +1274,6 @@ buildvariants:
       tasks:
           - name: test_mongohouse
 
-    - name: ubuntu1804-zseries
-      display_name: "s390x Ubuntu 18.04 (MongoDB 4.4)"
-      batchtime: 1440 # 1 day
-      expansions:
-          build_type: "Release"
-          cmake: "cmake"
-          tar_options: *linux_tar_options
-          cmake_flags: *linux_cmake_flags
-          mongodb_version: *version_44
-      run_on:
-          - ubuntu1804-zseries-small
-      tasks:
-          - name: compile_and_test_with_shared_libs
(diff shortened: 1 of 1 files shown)


.mci.yml
```diff
@@ -1274,23 +1274,6 @@ buildvariants:
...
-      display_name: "s390x Ubuntu 18.04 (MongoDB 4.4)"
-      batchtime: 1440 # 1 day
-      expansions:
-          build_type: "Release"
-          cmake: "cmake"
-          tar_options: *linux_tar_options
-          cmake_flags: *linux_cmake_flags
-          mongodb_version: *version_44
-      run_on:
-          - ubuntu1804-zseries-small
```
kevinAlbs, line 1287: Can `ubuntu1804-zseries-small` be replaced with `rhel80-zseries-small`? That was suggested in ticket comments to continue testing with zSeries.

eramongodb: I would prefer that to be addressed in a separate PR for [CXX-2492](https://jira.mongodb.org/browse/CXX-2492) instead of this one.

rcsanchez97 (approved):
Personally, I'd like to see RHEL8 zSeries build variant added at the same time.  However, since the Ubuntu zSeries tasks have been failing since essentially forever, I don't see any significant harm in proceeding with the removal right away.

eramongodb:
@rcsanchez97 I intend to resolve the build issues alongside the (re)addition of the Ubuntu zSeries variant as part of CXX-2492. 👍

PR #870, opened 2022-06-27
CXX-2347 Allow hint for unacknowledged writes

eramongodb:
## Description

This PR resolves CXX-2347.

CRUD unified spec tests are synced with [59a5dadb](https://github.com/mongodb/specifications/tree/59a5dadb2eeab8bba5a7d9a9acd47aa8c1d594cf/source/crud/tests/unified). `estimatedDocumentCount.json` is left unmodified as it is already synced to newer versions of the spec. `aggregate-let.json` and `aggregate-out-readConcern.json` are left unmodified to be updated as part of CXX-2296.

As CXX-2347 reverts CXX-1975 and CXX-2002, the `unacknowledge-*-hint-clientError` tests that were added but skipped have been removed entirely. 

## Unified Test Runner Improvements

The unified test runner was modified to handle the possibility of unacknowledged write results in `bulk_write()` and `find_and_modify()`.

The "string" type was added to the set of JSON element types that may be matched against with `$$type`.

## Client-Side Error Assertions

Due to CXX-2377 not yet being resolved, the `bulk_write_exception` thrown to report the _client-side_ error returned from `mongoc_bulk_operation_execute()` is caught and incorrectly reported as a _server-side_ error. An exception is made during unified test runner error matching based on the expected error message string to special-case the expected client-side error due to an unsupported server version.

## Drive-By Fixes

Include directories for libmongoc were not being properly added to some test executables and causing build failures. These have been fixed accordingly.


src/mongocxx/collection.cpp
```diff
@@ -178,7 +178,9 @@ mongocxx::stdx::optional<bsoncxx::document::value> find_and_modify(
 
     bsoncxx::document::view reply_view = reply.view();
 
-    if (reply_view["value"].type() == bsoncxx::type::k_null) {
+    const auto value = reply_view["value"];
+
+    if (!value || value.type() == bsoncxx::type::k_null) {
```
kevinAlbs, line 183: Nice catch. I filed CXX-2539 to track this bug. I think it is worth noting in release notes, and may help with discoverability if users encounter this bug.

rcsanchez97 (approved):
LGTM

kevinAlbs (approved):

PR #878, opened 2022-07-12
CXX-2415 Support setting tag set list with array

eramongodb:
## Description

This PR resolves CXX-2415.

Despite [Read Preference Tags](https://www.mongodb.com/docs/manual/core/read-preference-tags/) being specified as an _array_ of tag sets (aka a tag set list), the C++ Driver does not provide an interface to directly set the tag set list as an array, requiring the user to manually convert a `bsoncxx` array into a document to use the [BSON document overload](https://github.com/eramongodb/mongo-cxx-driver/blob/2f4ca28d9962dab9974806633706e47eeac555a7/src/mongocxx/read_preference.hpp#L179) as a workaround. This PR adds an overload that accepts a `bsoncxx::array::view_or_value` to improve the user experience.

This PR also updates the documentation to be more accurate and consistent when referring to "tag set lists". It also updates tests to correctly set the tag set list as a BSON _array_ of documents `[{key: value, ...}, ...]` according to spec.

This PR also adds missing documentation for the conversion operator from `bsoncxx::array::view` -> `bsoncxx::document::view`.

### DIFF
--- a/src/bsoncxx/array/view.hpp
+++ b/src/bsoncxx/array/view.hpp
@@ -128,6 +128,9 @@ class BSONCXX_API view {
     ///
     bool empty() const;
 
+    ///
+    /// Conversion operator unwrapping a document::view
+    ///
     operator document::view() const;
 
     ///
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):
Nice. Thank you for the documentation fixes too.

vector-of-bool (approved):

PR #881, opened 2022-08-03
Ensure unified test runner tests against primary in replica sets

eramongodb:
## Description

This PR ensures the unified test runner runs commands against the _primary_ in a replica set. This should address flaky tests on Evergreen involving replica sets that occasionally fail due to the server running on `27017` not being the primary as expected.

### DIFF
--- a/src/mongocxx/test_util/client_helpers.hh
+++ b/src/mongocxx/test_util/client_helpers.hh
@@ -110,6 +110,11 @@ bool is_replica_set(const client& client = {uri{}, add_test_server_api()});
 bool is_sharded_cluster(const client& client = {uri{}, add_test_server_api()});
 
 ///
+/// Returns the hostname of the primary in a replica set.
+///
+std::string get_primary(const client& client = {uri{}, add_test_server_api()});
+
+///
 /// Returns "standalone", "replicaset", or "sharded".
 ///
 std::string get_topology(const client& client = {uri{}, add_test_server_api()});
--- a/src/mongocxx/test/spec/unified_tests/runner.cpp
+++ b/src/mongocxx/test/spec/unified_tests/runner.cpp
@@ -254,17 +254,13 @@ std::string uri_options_to_string(document::view object) {
 }
 
 std::string get_hostnames(document::view object) {
-    const auto default_uri = std::string{"localhost:27017"};
-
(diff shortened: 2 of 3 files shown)


eramongodb:
Question: is this behavior (connect to replica set -> find primary -> run commands against primary) supposed to be handled by the client object (all cases) rather than by the unified test runner?

kevinAlbs:
> Question: is this behavior (connect to replica set -> find primary -> run commands against primary) supposed to be handled by the client object (all cases) rather than by the unified test runner?

The client is responsible for finding the primary when requested. When a URI only includes one host, like `mongodb://localhost:27017`, the client starts with a topology type `Single`. Topology type `Single` will not attempt to discover other hosts. This is inconsistent among drivers (see [Driver SDAM behavior survey](https://docs.google.com/spreadsheets/d/1sddTIxDtaB_x6z8c2vp6QbTwJSHm5nEPCRHbhUBLbPs/edit#gid=0)). It can be disabled by using the option `mongodb://localhost:27017?directConnection=false`.

An alternative solution may be to use `?directConnection=false` or to specify multiple hosts when connecting to a replica set.

src/mongocxx/test_util/client_helpers.hh
```diff
@@ -109,6 +109,11 @@ bool is_replica_set(const client& client = {uri{}, add_test_server_api()});
 ///
 bool is_sharded_cluster(const client& client = {uri{}, add_test_server_api()});
 
+///
+/// Returns the hostname of the primary in a replica set.
+///
+std::string get_primary(const client& client = {uri{}, add_test_server_api()});
+
```
kevinAlbs, line 116: This may not be a full solution. Other tests ([example](https://github.com/mongodb/mongo-cxx-driver/blob/c7452871978c1410153bc4da2ca7ea33c145cbf3/src/mongocxx/test/collection.cpp#L170)) with a default `uri` may also connect to a non-primary.

Consider adding a more generic `get_test_uri()` method, that could model the behavior in the C driver's test [URI functions](https://github.com/mongodb/mongo-c-driver/blob/f35c380df69f97e6f6a739707cb9917ad033f900/src/libmongoc/tests/test-libmongoc.c#L1305-L1349).

eramongodb:
Closing to conduct further research into a full solution applicable to the whole test suite.

PR #882, opened 2022-08-04
Fix missing compile options and definitions for polyfill selection

eramongodb:
## Description

This PR adds missing compile options and definitions to example projects to correctly detect and (not) use polyfills as described in the [installation docs](https://github.com/mongodb/mongo-cxx-driver/blob/c7452871978c1410153bc4da2ca7ea33c145cbf3/docs/content/mongocxx-v3/installation/windows.md#step-4-configure-the-driver).


kevinAlbs (approved):
LGTM. Please remove the empty `.editorconfig` file before merging.

I am not sure why the Visual Studio 2017 tasks were passing before.

rcsanchez97 (approved):
LGTM

PR #886, opened 2022-08-16
CXX-2337 Improve granularity of assertions for resumeToken prose test

eramongodb:
## Description

This PR does not necessarily resolve CXX-2337, but attempts to make more discernable the state and conditions when it does fail.

### DIFF
--- a/src/mongocxx/test/change_streams.cpp
+++ b/src/mongocxx/test/change_streams.cpp
@@ -123,7 +123,7 @@ TEST_CASE("Spec Prose Tests") {
     coll.drop();
 
     write_concern wc_majority;
-    wc_majority.majority(std::chrono::milliseconds(0));
+    wc_majority.majority(std::chrono::seconds(30));
 
     // As a sanity check, we implement the first prose test. The behavior tested
     // by the prose tests is implemented and tested by the C driver, so we won't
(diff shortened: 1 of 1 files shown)


eramongodb:
The `wtimeout` value of `0 ms` is suspicious. Raising the timeout to a more common/acceptable value of `100 ms` or removing it altogether (as I do not believe it is required by the prose test) may resolve the flaky test failures.

kevinAlbs:
> The `wtimeout` value of `0 ms` is suspicious. Raising the timeout to a more common/acceptable value of `100 ms` or removing it altogether (as I do not believe it is required by the prose test) may resolve the flaky test failures, as the original test failures suggest the final of the three insertions did not succeed (thus no event recorded -> empty/invalid `token3` document).

I agree. The C driver sets a [30 second `wtimeout`](https://github.com/mongodb/mongo-c-driver/blob/aa940294031177b5faa8623f81dd433a6ab72ed3/src/libmongoc/tests/test-mongoc-change-stream.c#L356). I am not sure if that is the cause of the observed test failure, but I suggest increasing the timeout to 30 seconds.

kevinAlbs (approved):

eramongodb:
Will assume the unreasonably low wtimeout to be the culprit and resolve CXX-2337. Can reopen/create-new-ticket if issue reoccurs.

PR #887, opened 2022-08-17
Add StringMaker specialization for bsoncxx::document::value

eramongodb:
## Description

Improve readability of test suite output by supporting pretty-printing of `bsoncxx::document::value` objects.

As a drive-by fix, removed the redundant `StringMaker<stdx::optional<bsoncxx::document::view>>` specialization (handled by `StringMaker<stdx::optional<T>>`) that inadvertently leads to infinite self-recursion when invoked.


kevinAlbs (approved):
Nice, LGTM

PR #890, opened 2022-08-25
CXX-2530 add missing Debian 10, 11 and Ubuntu 20.04 build variants

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/6307b35a5623437b34e1f3a2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

The main thing to focus on in the patch build is the new build variants.

### DIFF
--- a/.evergreen/find_cmake.sh
+++ b/.evergreen/find_cmake.sh
@@ -11,19 +11,33 @@ find_cmake ()
     CMAKE="/Applications/Cmake.app/Contents/bin/cmake"
   elif [ -f "/opt/cmake/bin/cmake" ]; then
     CMAKE="/opt/cmake/bin/cmake"
-  elif command -v cmake 2>/dev/null; then
+  elif [ -z "$IGNORE_SYSTEM_CMAKE" ] && command -v cmake 2>/dev/null; then
      CMAKE=cmake
   elif uname -a | grep -iq 'x86_64 GNU/Linux'; then
+     if [ -f "$(pwd)/cmake-3.11.0/bin/cmake" ]; then
+      CMAKE="$(pwd)/cmake-3.11.0/bin/cmake"
+      return 0
+     fi
      curl --retry 5 https://cmake.org/files/v3.11/cmake-3.11.0-Linux-x86_64.tar.gz -sS --max-time 120 --fail --output cmake.tar.gz
      mkdir cmake-3.11.0
      tar xzf cmake.tar.gz -C cmake-3.11.0 --strip-components=1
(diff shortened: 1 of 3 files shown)


.evergreen/find_cmake.sh
```diff
@@ -11,20 +11,34 @@ find_cmake ()
...
   elif [ -f "/cygdrive/c/cmake/bin/cmake" ]; then
-    CMAKE=/cygdrive/c/cmake/bin/cmake
-  else
-     # If cmake unavailable, create a BUILD ticket.
-     echo "cmake not available"
-     return 1
+     CMAKE="/cygdrive/c/cmake/bin/cmake"
+  fi
+  if [ -z "$CMAKE" -o -z "$( $CMAKE --version 2>/dev/null )" ]; then
+     # Some images have no cmake yet, or a broken cmake (see: BUILD-8570)
```
eramongodb, line 29: Can we drop the complication of system cmake detection in favor of an unconditional download-build-install a fixed CMake version (effectively `find_cmake` -> `install_cmake`)? This would also allow us to consistently guarantee and develop against a known minimum CMake version across all environments.

eramongodb: BUILD-8570 has been resolved. Are the changes in this PR to `find_cmake` still necessary?

eramongodb (approved):
Minor questions; otherwise LGTM.

rcsanchez97:
TBH, I am not sure.  My instinct is that we should use the system CMake
if it meets our minimum.  Let me think through it a bit and I'll
follow-up again.

eramongodb:
I am generally in favor of avoiding depending on the system environment whenever possible for isolation and reproducibility. However, feel free to consider this issue a non-blocker; it can be addressed in a separate PR.

rcsanchez97:
OK.  I'd like to go ahead with this PR, since keeping the existing
`find_cmake.sh` script in sync across the projects that use it is better
than allowing them to remain out of sync.

I'll create a new ticket to catpure the need to re-think the whole CMake
finding approach across the various projects.

Also, I just noticed that I made a typo in the Jira ticket reference in
my commit messages and in the PR summary.  I am going to fix the commit
messages and force-push to make sure that the correct references make it
in the final merge.

rcsanchez97:
I created CDRIVER-4462 to capture the "re-think find-cmake.sh" work.

rcsanchez97:
I created CDRIVER-4462 to capture the "re-think find-cmake.sh" work.

PR #897, opened 2022-09-30
Improve documentation and scripts relating to CXX Driver releases

eramongodb:
## Description

This PR suggests several improvements to both the documentation and the scripts relating to the process of publishing a new C++ Driver release.

## Minimum required version for C Driver and libmongocrypt

The `install_c_driver.sh` and `etc/make_release.py` scripts were updated to use the minimum required versions for the C Driver and libmongocrypt (as specified in `.mci.yml`).

`etc/make_release.py` was not updated to account for the new `MONGOCRYPT_VERSION` script parameter added as part of https://github.com/mongodb/mongo-cxx-driver/pull/867 in https://github.com/mongodb/mongo-cxx-driver/commit/3581760040f8ca95e5932b2d1cb9ca65ce0a891c. This PR adds a new `--mongocrypt-build-ref` option to the script. `--c-driver-build-ref` and `--mongocrypt-build-ref` now default to `1.22.1` and `1.5.2` respectively rather than `master`. These should be updated alongside the minimum required versions in `.mci.yml` and elsewhere. Alternatively, the release instructions can be updated to explicitly provide the current minimum versions in `.mci.yml` to the script when executing the command (required explicit arguments rather than implicit default arguments).

## etc/requirements.txt

Added the `cryptography` package to the list of required packages in `etc/requirements.txt` to address the following error:

```
AttributeError: module 'jwt.algorithms' has no attribute 'hashes'
```

## Release Instructions

`etc/releasing.md` was updated with clarifications and improvements to the release process:

- All minor and patch versions were updated to describe the expected process for the upcoming minor `3.8.0` and patch `3.7.1` versions.
- Added a new step to ensure the generated GitHub release draft is published.
- Clarified that the documentation changes may require the use of a pull request rather than pushing directly to origin.
- Clarified that the DOCSP ticket to comment on is automatically generated.
- Updated links to the MongoDB Compatibility and Language Compatibility tables.

### DIFF
--- a/etc/make_release.py
+++ b/etc/make_release.py
@@ -86,11 +86,15 @@ ISSUE_TYPE_ID = {'Backport': '10300',
 @click.option('--c-driver-install-dir',
               default=os.getcwd() + '/../mongoc',
               show_default=True,
-              help='When building the C driver, install to this directory')
+              help='When building the C driver and libmongocrypt, install to this directory')
 @click.option('--c-driver-build-ref',
-              default='master',
+              default='1.22.1',
               show_default=True,
               help='When building the C driver, build at this Git reference')
+@click.option('--mongocrypt-build-ref',
+              default='1.5.2',
+              show_default=True,
+              help='When building libmongocrypt, build at this Git reference')
 @click.option('--with-c-driver',
               help='Instead of building the C driver, use the one installed at this path')
 @click.option('--dist-file',
@@ -117,6 +121,7 @@ def release(jira_creds_file,
             remote,
             c_driver_install_dir,
             c_driver_build_ref,
+            mongocrypt_build_ref,
             with_c_driver,
             dist_file,
             skip_distcheck,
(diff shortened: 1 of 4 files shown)


etc/requirements.txt
```diff
@@ -2,4 +2,5 @@
 Click
 GitPython
 PyGithub
-jira
\ No newline at end of file
+jira
+cryptography
```
vector-of-bool, line 6: Recommend version-pinning this to a known-good version, especially as this package is "notable" for breaking changes. (all packages here should be pinned, as well as their transitive dependencies, probably). Refer to: "`pip freeze`"

etc/make_release.py
```diff
@@ -363,9 +367,9 @@ def build_c_driver(c_driver_install_dir, c_driver_build_ref, quiet):
 
     env = os.environ
```
vector-of-bool, line 368: `os.environ` should be copied unless it is intended to modify the process's own environment (Writing to `env` will call `putenv()`)

vector-of-bool (approved):
LGTM, with minor tweaks.

rcsanchez97 (changes requested):
LGTM

rcsanchez97 (approved):
LGTM

PR #898, opened 2022-10-02
CXX-2584 update min libmongoc version in cmake

kevinAlbs:
# Summary
- Update the minimum libmongoc version check in cmake scripts.

# Background & Motivation
The required libmongoc dependency was increased to 1.22.1 in https://github.com/mongodb/mongo-cxx-driver/pull/884.


eramongodb (approved):
I think it is sufficient to document the locations to change in `.mci.yml` only.

PR #904, opened 2022-11-16
Cxx 2454

kkloberdanz:
CXX-2454 Create C++ driver snapshot query example for manual.

related: DRIVERS-2181

### DIFF
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -1325,6 +1325,17 @@ static void snapshot_example2(mongocxx::client& client) {
 
     auto db = client["retail"];
 
+    // auto gt_array = bsoncxx::builder::basic::array{};
+    // gt_array.append("$saleDate");
+    // gt_array.append(make_document(
+    //    kvp("$dateSubtract",
+    //        make_document(kvp("startDate", "$$NOW"), kvp("unit", "day"), kvp("amount", 1)))));
+
+    // array gt_array;
+    // gt_array.append("$saleDate");
+    // gt_array.append(make_document(kvp("$dateSubtract", make_document(kvp("startDate", "$$NOW"),
+    // kvp("unit", "day"), kvp("amount", 1)))));
+
     // client = MongoClient()
     // db = client.retail
     // with client.start_session(snapshot=True) as s:
@@ -1349,28 +1360,29 @@ static void snapshot_example2(mongocxx::client& client) {
     //   ], session=s
     //   ).next()["totalDailySales"]
 
+    // pipeline p;
+
+    // auto gt_array = make_array(
+    //   "$saleDate",
+    //   make_document(
+    //       kvp("$dateSubtract",
+    //           make_document(kvp("startDate", "$$NOW"), kvp("unit", "day"), kvp("amount", 1)))));
+
+    // auto gt_expr = kvp("$gt", gt_array);
+    // auto full_expr = kvp("$expr", gt_expr);
+    // auto doc = make_document(full_expr);
+
     pipeline p;
 
-    // array gt_array;
-    // gt_array.append("$saleDate");
-    // gt_array.append(make_document(kvp("$dateSubtract", make_document(kvp("startDate", "$$NOW"),
-    // kvp("unit", "day"), kvp("amount", 1)))));
-    // auto gt_array = make_array(
-    //    "$saleDate",
-    //    make_document(
-    //        kvp("$dateSubtract",
-    //            make_document(kvp("startDate", "$$NOW"), kvp("unit", "day"), kvp("amount", 1)))));
+    auto doc = make_document(kvp(
+        "$expr",
+        make_document(kvp(
+            "$gt",
+            make_array(
+                "$salesDate",
+                make_document(kvp("startDate", "$$NOW"), kvp("unit", "day"), kvp("amount", 1)))))));
 
-    auto gt_array = bsoncxx::builder::basic::array{};
-    gt_array.append("$saleDate");
-    gt_array.append(make_document(
-        kvp("$dateSubtract",
-            make_document(kvp("startDate", "$$NOW"), kvp("unit", "day"), kvp("amount", 1)))));
-
-    auto gt_expr = kvp("$gt", gt_array);
-    auto full_expr = kvp("$expr", gt_expr);
-
-    p.match(make_document(full_expr)).count("totalDailySales");
+    p.match(std::move(doc)).count("totalDailySales");
     // auto cursor = db["sales"].aggregate(session, p);
 
     // int64_t total_daily_sales = 0;
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -1208,7 +1208,7 @@ static void snapshot_examples(mongocxx::client& client) {
 
     options::insert write_options;
     write_concern wc;
-    wc.majority(std::chrono::milliseconds(10000));
+    wc.majority(std::chrono::seconds(10));
     write_options.write_concern(wc);
 
     auto db = client["pets"];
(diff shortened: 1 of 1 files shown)


examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1180,55 @@ void delete_examples(mongocxx::database db) {
...
+    auto session = client.start_session(opts);
+
+    {
+        pipeline p;
+
+        p.match(make_document(kvp("adoptable", true))).count("adoptableCatsCount");
+        auto cursor = db["cats"].aggregate(session, p);
+
+        for (auto doc : cursor) {
+            adoptable_pets_count += doc.find("adoptableCatsCount")->get_int32();
```
kkloberdanz, line 1210: Is there a clean one liner to get either int32 or int64 and return that value as int64 (similar to bson_iter_as_int64 from the bson C API?)

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
     }
 }
 
+static void snapshot_examples(mongocxx::client& client) {
+    // Write concern "majority" needed to avoid the following error below:
+    //
+    // Unable to read from a snapshot due to pending collection catalog changes;
+    // please retry the operation.
```
kevinAlbs, line 1200: Was this error observed when running this example?

This may suggest there are other conditions where this error appears.

SERVER-41532 includes "insert in a transaction" in the Steps to Reproduce. This example does not start a transaction.
SERVER-41532 notes this applies to sharded clusters. This example is only run against a replica set.

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
+    //
+    // See also: https://jira.mongodb.org/browse/SERVER-41532
+
+    // Start Example 59
+    using namespace mongocxx::v_noabi;
```
kevinAlbs, line 1205:
```suggestion
    using namespace mongocxx;
```

`v_noabi` is an [inline namespace](https://github.com/mongodb/mongo-cxx-driver/blob/e6aeda6575ee0f7338019edf812055370f7f9c8c/src/mongocxx/config/config.hpp.in#L15).

examples/mongocxx/mongodb.com/documentation_examples.cpp
kevinAlbs: AFAIK no.

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
+        p.match(make_document(kvp("adoptable", true))).count("adoptableDogsCount");
+        auto cursor = db["dogs"].aggregate(session, p);
+
+        for (auto doc : cursor) {
+            adoptable_pets_count += doc.find("adoptableDogsCount")->get_int32();
+        }
+    }
+
+    if (adoptable_pets_count != 2) {
+        throw std::logic_error("wrong number of adoptable pets in example 57, expecting 2 got: " +
```
kevinAlbs, line 1251:
```suggestion
        throw std::logic_error("wrong number of adoptable pets in example 59, expecting 2 got: " +
```

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
 
+static void snapshot_examples(mongocxx::client& client) {
+    // Write concern "majority" needed to avoid the following error below:
+    //
+    // Unable to read from a snapshot due to pending collection catalog changes;
+    // please retry the operation.
+    //
+    // See also: https://jira.mongodb.org/browse/SERVER-41532
+
+    // Start Example 59
```
kevinAlbs, line 1204:
```suggestion
    // Start Snapshot Query Example 1
```

I missed this in the C review. These comments are scraped by the docs team. I think they need to be an exact match of what was requested in the ticket.

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
+        for (auto doc : cursor) {
+            adoptable_pets_count += doc.find("adoptableDogsCount")->get_int32();
+        }
+    }
+
+    if (adoptable_pets_count != 2) {
+        throw std::logic_error("wrong number of adoptable pets in example 57, expecting 2 got: " +
+                               std::to_string(adoptable_pets_count));
+    }
+    // End Example 59
```
kevinAlbs, line 1254:
```suggestion
    // End Snapshot Query Example 1
```

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
+}
```
kevinAlbs, line 1255: I missed in the C review. `// Start Snapshot Query Example 2` also needs to be added.

examples/mongocxx/mongodb.com/documentation_examples.cpp
kkloberdanz: I observed this error from this example in CI. More context is below:
```
[2022/11/18 20:29:37.878] terminate called after throwing an instance of 'mongocxx::v_noabi::query_exception'
[2022/11/18 20:29:37.878]   what():  Unable to read from a snapshot due to pending collection catalog changes; please retry the operation. Snapshot timestamp is Timestamp(1668803377, 209). Collection minimum is Timestamp(1668803377, 210): generic server error
[2022/11/18 20:29:37.884] bash: line 200: 431523 Aborted                 (core dumped) $test
[2022/11/18 20:29:37.884] Command ''shell.exec' in "test"' failed: shell script encountered problem: exit code 134.
```
I was able to work around it by adding a short sleep between inserting to the new collection and starting the snapshot, but setting write concern to majority also solved this.

examples/mongocxx/mongodb.com/documentation_examples.cpp
kkloberdanz: Woops 😅

examples/mongocxx/mongodb.com/documentation_examples.cpp
kkloberdanz: Got it! How about I also update the C driver example accordingly.

rcsanchez97 (approved):

examples/mongocxx/mongodb.com/documentation_examples.cpp
kkloberdanz: Out of curiosity, does this behavior surprise you? Do you think this could be a bug?

kevinAlbs: The error suggests it is a transient. The Python implementation includes a [helper to wait for the snapshot to become available](https://github.com/mongodb/mongo-python-driver/commit/e325b24b78e431cb889c5902d00b8f4af2c700c3#diff-c5d782e261f04fca18024ab18c3ed38fb45ede24cde4f9092e012f6fcbbe0df5R1368). I suggest adding that helper.

examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -1180,6 +1193,85 @@ void delete_examples(mongocxx::database db) {
...
 
+static void snapshot_examples(mongocxx::client& client) {
+    // Write concern "majority" needed to avoid the following error below:
+    //
+    // Unable to read from a snapshot due to pending collection catalog changes;
+    // please retry the operation.
+    //
+    // See also: https://jira.mongodb.org/browse/SERVER-41532
+
+    // Start Snapshot Query Example 1
```
kevinAlbs, line 1204: Suggest doing the `insert_one` operations outside of the `// Start Snapshot Query Example 1` comment to match the example in the drivers ticket.

kevinAlbs (approved):
Nicely done. LGTM with two comments addressed.

PR #924, opened 2023-01-10
CXX-2629 Remove requirement to set `/Zc:__cplusplus` in MSVC flags

kevinAlbs:
# Summary
- Remove requirement for /Zc:__cplusplus

# Background & Motivation

The current [documented install instructions](https://mongocxx.org/mongocxx-v3/installation/windows/#:~:text=For%20building%20with%20Visual%20Studio%202017) note:

> For building with Visual Studio 2017 (without a C++17 polyfill), it is necessary to configure with an additional option, /Zc:__cplusplus to opt into the correct definition of `__cplusplus` (problem described here):
> 
> ```
> 'C:\Program Files (x86)\CMake\bin\cmake.exe' .. \
>     -G "Visual Studio 15 2017 Win64"            \
>     -DCMAKE_CXX_STANDARD=17                     \
>     -DCMAKE_CXX_FLAGS="/Zc:__cplusplus"         \
>     -DCMAKE_PREFIX_PATH=C:\mongo-c-driver       \
>     -DCMAKE_INSTALL_PREFIX=C:\mongo-cxx-driver  \
> ```

Setting `CMAKE_CXX_FLAGS` appears to have the side-effect of overwriting the default MSVC flags set by cmake.
For example, if `CMAKE_CXX_FLAGS` is not set, this is the output I get with cmake 3.23.1 on my Windows machine:
```
CMAKE_CXX_FLAGS=/DWIN32 /D_WINDOWS /W3 /GR /EHsc
```

Following the C++ driver instructions results in `CMAKE_CXX_FLAGS` only including `/Zc:__cplusplus`.

Excluding `/EHsc` appears particularly problematic. `/EHsc` is the [documented default](https://learn.microsoft.com/en-us/cpp/cpp/exception-handling-in-visual-cpp?view=msvc-170) for the exception handling model:

> Use an /EH compiler option to specify the exception handling model to use in a C++ project. Standard C++ exception handling (/EHsc) is the default in new C++ projects in Visual Studio.

Not including the `/EHsc` flag results in warnings during compilation:
```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.34.31933\include\ostream(305,1): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
```

Setting `/EHsc` appears to fix an `Internal compiler error` reported in CXX-2616.

---

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/ notes:

> The _MSVC_LANG predefined macro continues to report the Standard version switch regardless of the value of /Zc:__cplusplus. _MSVC_LANG is set whether or not the /Zc:__cplusplus option is enabled. When /Zc:__cplusplus is enabled, __cplusplus == _MSVC_LANG.

This PR adds a check for `_MSVC_LANG`, to remove the requirement to pass `/Zc:__cplusplus`. Following the documentation no longer results in omitting the `/EHsc` flag.

### DIFF
--- a/src/bsoncxx/stdx/make_unique.hpp
+++ b/src/bsoncxx/stdx/make_unique.hpp
@@ -48,7 +48,7 @@ using ::boost::make_unique;
 BSONCXX_INLINE_NAMESPACE_END
 }  // namespace bsoncxx
 
-#elif __cplusplus >= 201402L
+#elif __cplusplus >= 201402L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201402L)
 
 #include <memory>
 
--- a/.mci.yml
+++ b/.mci.yml
@@ -43,11 +43,7 @@ variables:
         ubsan_cmake_flags: &ubsan_cmake_flags -DCMAKE_C_FLAGS="-Wall -Wextra -Wno-attributes -Werror -Wno-missing-field-initializers $ignore_deprecated" -DCMAKE_CXX_COMPILER="/usr/bin/clang++" -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0 -fsanitize=undefined -fsanitize-blacklist=$(pwd)/etc/ubsan.ignorelist -fno-sanitize-recover=undefined -O1 -g -fno-omit-frame-pointer -Wall -Wextra -Wconversion -Wnarrowing -pedantic -Werror $ignore_deprecated"
         msvc2015_cmake_flags: &msvc2015_cmake_flags -DBOOST_ROOT=c:/local/boost_1_60_0
         msvc2015_generator: &msvc2015_generator Visual Studio 14 2015 Win64
-        # The VS option /Zc:__cplusplus is necessary to build with VS2017's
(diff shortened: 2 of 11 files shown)


src/bsoncxx/stdx/make_unique.hpp
```diff
@@ -48,7 +48,7 @@ using ::boost::make_unique;
 BSONCXX_INLINE_NAMESPACE_END
 }  // namespace bsoncxx
 
-#elif __cplusplus >= 201402L
+#elif __cplusplus >= 201402L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201402L)
```
eramongodb, line 51: Was concerned about testing `__cplusplus` before `_MSVC_LANG`, but it seems [Catch2](https://github.com/catchorg/Catch2/blob/e932bcf7a349f1c2fd02369df8c923afccb4638c/src/catch2/internal/catch_compiler_capabilities.hpp#L30) does this as well, so I believe this should be fine.

eramongodb (approved):

PR #926, opened 2023-01-17
Options to limit number of mongos servers used

kkloberdanz:
CXX-2318

Provide options to limit number of mongos servers used in connecting to sharded clusters.

Introduces an srvMaxHosts MongoClient option, which is used to limit the number of seeds derived from SRV records (and only supported with the mongodb+srv:// scheme). Also limits the number of mongos seeds that may be picked up during SRV polling.

Introduces the following test changes:
- New spec tests for Initial DNS Seedlist Discovery and URI options specs.
- New numSeeds and numHosts fields to the Initial DNS Seedlist Discovery spec test format.
- New SRV Polling prose tests

Drivers should sync with [mongodb/specifications@b508f6d](https://github.com/mongodb/specifications/commit/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5).


.mci.yml
```diff
@@ -409,6 +409,7 @@ functions:
                   export CRUD_UNIFIED_TESTS_PATH="$(pwd)/../data/crud/unified"
                   export CHANGE_STREAMS_UNIFIED_TESTS_PATH="$(pwd)/../data/change-streams/unified"
                   export ENCRYPTION_TESTS_PATH="$(pwd)/../data/client_side_encryption"
+                  export TEST_DATA_PATH="$(pwd)/../data"
```
eramongodb, line 412:
```suggestion
                  export URI_OPTIONS_TESTS_PATH="$(pwd)/../data/uri-options"
```
For consistency with other env vars and data file organization, suggest renaming the variable and relocating `data/srv-options.json` accordingly.

src/mongocxx/uri.cpp
```diff
@@ -166,6 +167,18 @@ static stdx::optional<bool> _bool_option(mongoc_uri_t* uri, std::string opt_name
...
+stdx::optional<bsoncxx::document::view> uri::credentials() {
+    const bson_t* options_bson = libmongoc::uri_get_credentials(_impl->uri_t);
+    const uint8_t* data = bson_get_data(options_bson);
+    uint32_t len = options_bson->len;
+
+    if (len) {
+        return bsoncxx::document::view(data, len);
+    } else {
+        return {};
+    }
```
eramongodb, line 179:
```suggestion
    
    return bsoncxx::document::view(data, options_bson->len);
```
Length check should not be necessary.

src/mongocxx/test/uri.cpp
```diff
@@ -295,6 +313,228 @@ TEST_CASE("uri::wait_queue_timeout_ms()", "[uri]") {
...
+        auto result = admin.run_command(make_document(kvp("ping", 1)));
+        REQUIRE(!result.empty());
+        REQUIRE(result.begin()->get_double() == 1.0);
+        return true;
+    } catch (mongocxx::exception& e) {
+        return false;
+    }
+}
+
+TEST_CASE("uri::test_srv_max_hosts", "[uri]") {
```
eramongodb, line 494:
```suggestion
TEST_CASE("uri::test_srv_max_hosts", "[uri]") {
   mongocxx::instance::current();

```
Ensure libmongoc is initialized before executing tests that use libmongoc functions.

src/mongocxx/test/uri.cpp
```diff
@@ -295,6 +313,228 @@ TEST_CASE("uri::wait_queue_timeout_ms()", "[uri]") {
...
+        REQUIRE(!expect_exception);
+        if (should_ping) {
+            validate_srv_max_hosts(std::move(client), test_doc, mtx, new_hosts);
+        }
+    } catch (mongocxx::exception& e) {
+        REQUIRE(expect_exception);
+    }
+}
+
+static void iterate_srv_max_hosts_tests(std::string dir, std::vector<std::string> files) {
```
eramongodb, line 459: `src/mongocxx/test/uri.cpp` is reserved for URI tests that do not require a live server and is part of `test_driver`. For consistency with other unified spec tests, suggest moving these tests (that are not testing the URI format itself, but testing behavior against a live server) into a new `src/mongocxx/test/spec/uri.cpp` file (see other test files in that directory for reference) as part of a new `test_uri_options` test executable (see [other executables](https://github.com/mongodb/mongo-cxx-driver/blob/cb3193179199bd764ff9706f3a5fc19f5d8f4349/src/mongocxx/test/CMakeLists.txt#L106) for reference).

eramongodb (changes requested):

src/mongocxx/test/uri.cpp
```diff
@@ -295,6 +314,230 @@ TEST_CASE("uri::wait_queue_timeout_ms()", "[uri]") {
                        [](mongocxx::uri& uri) { return uri.wait_queue_timeout_ms(); });
 }
 
+static void compare_elements(bsoncxx::document::element expected_option,
```
vector-of-bool, line 317: I would name `require_elements_equal` or `assert_elements_equal` to make it clear that this function can terminate the test.

src/mongocxx/test/uri.cpp
```diff
@@ -295,6 +314,230 @@ TEST_CASE("uri::wait_queue_timeout_ms()", "[uri]") {
...
+            key = std::string("tls");
+        }
+
+        bsoncxx::document::element my_value = my_options_view[key];
+        if (my_value) {
+            compare_elements(expected_option, my_value);
+        } else if (creds && creds.value()[key]) {
+            compare_elements(expected_option, creds.value()[key]);
+        } else {
+            REQUIRE(false);
```
vector-of-bool, line 354: Use `FAIL()` or `FAIL_CHECK()`, which lets you insert a message string.

src/mongocxx/test/uri.cpp
```diff
@@ -295,6 +314,230 @@ TEST_CASE("uri::wait_queue_timeout_ms()", "[uri]") {
...
+        };
+
+        auto admin = client["admin"];
+        auto result = admin.run_command(make_document(kvp("ping", 1)));
+        REQUIRE(!result.empty());
+        REQUIRE(result.begin()->get_double() == 1.0);
+        return true;
+    } catch (mongocxx::exception& e) {
+        return false;
+    }
```
vector-of-bool, line 492: Can this exception be made more precise to only handle the exact error we are looking for?

src/mongocxx/test/spec/uri.cpp
```diff
@@ -0,0 +1,270 @@
...
+    std::string path = std::string(test_path) + sub_path.data();
+    CAPTURE(path);
+
+    std::ifstream file{path};
+    REQUIRE(file);
+
+    std::string file_contents((std::istreambuf_iterator<char>(file)),
+                              std::istreambuf_iterator<char>());
+
+    return bsoncxx::from_json(file_contents);
```
kevinAlbs, line 40: The `seeds`, `numSeeds`, `parsed_options`, `numHosts` and `comment` fields in the test format [are not handled](https://github.com/mongodb/specifications/blob/0a44c5bb3133d7b3e156276a9cba45daccb9266d/source/initial-dns-seedlist-discovery/tests/README.rst#test-format-and-use).

That may be OK, since the subset of tests do not use those fields.

Suggest parsing the fields explicitly. Erroring on unhandled fields will inform us when a test is added with new fields:

```c++
struct InitialDNSSeedlistTest {
    bsoncxx::stdx::string_view uri;
    bsoncxx::array::view hosts;
    int32_t num_hosts;
    bsoncxx::document::view options;
    bool error;
    bool ping;

    static InitialDNSSeedlistTest parse(bsoncxx::document::view test_doc) {
        InitialDNSSeedlistTest test;

        for (auto el : test_doc) {
            if (0 == el.key().compare("uri")) {
                test.uri = el.get_string().value;
            } else if (0 == el.key().compare("seeds") || 0 == el.key().compare("numSeeds")) {
                // The 'seeds' and 'numSeeds' assertions are explicitly skipped. The C++ driver does
                // not have access to the initial seedlist populated in the C driver.
            } else if (0 == el.key().compare("hosts")) {
                test.hosts = el.get_array().value;
            } else if (0 == el.key().compare("numHosts")) {
                test.num_hosts = el.get_int32().value;
            } else if (0 == el.key().compare("options")) {
                test.options = el.get_document().value;
            } else if (0 == el.key().compare("error")) {
                test.error = el.get_bool().value;
            } else if (0 == el.key().compare("comment")) {
                // Ignore comment.
            } else if (0 == el.key().compare("ping")) {
                test.ping = el.get_bool().value;
            } else {
                FAIL("InitialDNSSeedlistTest does not understand the field: '"
                     << el.key() << "'. Please add support.");
            }
        }
        return test;
    }
};
```

src/mongocxx/test/spec/uri.cpp
```diff
@@ -0,0 +1,270 @@
...
+    if (expected_hosts) {
+        size_t count = 0;
+        while (!hosts_are_equal(expected_hosts.get_array().value, new_hosts, mtx)) {
+            count++;
+            if (count > 4) {
+                REQUIRE(hosts_are_equal(expected_hosts.get_array().value, new_hosts, mtx));
+            }
+            std::this_thread::sleep_for(std::chrono::seconds(count));
+        }
+    }
```
kevinAlbs, line 132: Also check for "numHosts".

kevinAlbs (changes requested):

.mci.yml
```diff
@@ -429,6 +430,8 @@ functions:
                   fi
                   popd # ../../drivers-evergreen-tools
 
+                  export MONGOCXX_TEST_TLS_CA_FILE=$DRIVERS_TOOLS/.evergreen/x509gen/ca.pem
+
```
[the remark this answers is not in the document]
kkloberdanz: That makes sense. For the purpose of this ticket how shall I proceed? Should I put this PR on ice until CXX-2645 is complete, or should we still get it merged?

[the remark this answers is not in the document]
kevinAlbs: I suggest proceeding with this PR. I suspect much of the URI logic is platform independent, and is getting test coverage in the C driver.

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,320 @@
...
+    } catch (mongocxx::exception& e) {
+        REQUIRE(test.error);
+    }
+}
+
+static void iterate_srv_max_hosts_tests(std::string dir, std::vector<std::string> files) {
+    for (const auto& file : files) {
+        auto test_doc = _doc_from_file("/" + dir + "/" + file);
+        auto test = InitialDNSSeedlistTest::parse(test_doc);
+        run_srv_max_hosts_test_file(test);
```
kevinAlbs, line 237:
```suggestion
        SECTION (file) {
            run_srv_max_hosts_test_file(test);
        }
```

Suggest adding a section for file names to enable running one test file:

```
./cmake-build/src/mongocxx/test/test_driver  "uri::test_srv_max_hosts" --section "replica-set" --section "srvMaxHosts-less_than_srv_records.json"
```

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,320 @@
...
+    auto result = admin.run_command(make_document(kvp("ping", 1)));
+    REQUIRE(result.begin()->get_double() == 1.0);
+
+    auto uri = client.uri();
+    auto my_options = uri.options();
+    auto creds = uri.credentials();
+    compare_options(test.options, my_options, creds);
+
+    bool has_hosts_list = !test.hosts.empty();
+    if (test.num_hosts || has_hosts_list) {
```
kevinAlbs, line 162: Tests specify "numHosts" instead of "hosts". "numHosts" is used when it is not possible to compare the hosts returned.

Here is one test:
https://github.com/mongodb/specifications/blob/31bbf60c4f3008d77fc5ca5287b165e5ea97a4a6/source/initial-dns-seedlist-discovery/tests/sharded/srvMaxHosts-less_than_srv_records.json#L4

Since no tests are asserting "numHosts", I suggest removing the "numHosts" logic. Do not parse "numHosts" in `InitialDNSSeedlistTest::parse` so an error will be thrown when tests are added with the "numHosts" assertion.

src/mongocxx/test/spec/uri.cpp
kevinAlbs: Disregard. No tests are exercising "numHosts".

src/mongocxx/uri.cpp
```diff
@@ -173,6 +173,10 @@ stdx::optional<bsoncxx::document::view> uri::credentials() {
     return bsoncxx::document::view(data, options_bson->len);
 }
 
+stdx::optional<std::int32_t> uri::srv_max_hosts() const {
+    return libmongoc::uri_get_option_as_int32(_impl->uri_t, MONGOC_URI_SRVMAXHOSTS, 0);
```
kevinAlbs, line 177:
```suggestion
    return _int32_option(_impl->uri_t, MONGOC_URI_SRVMAXHOSTS);
```

`mongoc_uri_get_option_as_int32` will return the fallback value (`0` in this case) if the option is not set. Using `_int32_option` will return `stdx::nullopt` if the option is not set.

kevinAlbs (commented):
`URI_OPTIONS_TESTS_PATH` and `INITIAL_DNS_SEEDLIST_DISCOVERY_TESTS_PATH` may also need to [CMakeLists.txt](https://github.com/mongodb/mongo-cxx-driver/blob/1da6947b4818858c4a0dc49bf12ac2391484bfb6/src/mongocxx/test/CMakeLists.txt#L236-L269) to fix Windows tasks.

Windows tasks use the [RUN_TESTS target to run tests](https://github.com/mongodb/mongo-cxx-driver/blob/1da6947b4818858c4a0dc49bf12ac2391484bfb6/.mci.yml#L532). The paths set in [environment variables in .mci.yml](https://github.com/mongodb/mongo-cxx-driver/blob/1da6947b4818858c4a0dc49bf12ac2391484bfb6/.mci.yml#L408-L421) may be cygwin paths. The paths may not be the correct form for Windows and need to be run through `cygpath`.

kevinAlbs (approved):

eramongodb (changes requested):

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,315 @@
...
+// TODO(CXX-2087) extend this test to run all Initial DNS Seedlist Discovery tests.
+TEST_CASE("uri::test_srv_max_hosts", "[uri]") {
+    mongocxx::instance::current();
+
+    if (!std::getenv("MONGOCXX_TEST_TLS_CA_FILE")) {
+        std::cerr << "The environment variable 'MONGOCXX_TEST_TLS_CA_FILE' must be present to run "
+                     "the test 'uri::test_srv_max_hosts'"
+                  << std::endl;
+        return;
+    }
```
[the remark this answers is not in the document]
kkloberdanz: There's just a slight conflict between this suggestion and a previous suggestion from @kevinAlbs. I took out defining `MONGOCXX_TEST_TLS_CA_FILE` from this PR by Kevin's suggestion, which was to check if this variable was set in order to run the tests, and to leave it undefined for tests where TLS is not enabled. I'm fine with doing it either way, and this suggestion makes sense to me, but I would like to hear what Kevin would prefer.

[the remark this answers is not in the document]
eramongodb: > which was to check if this variable was set in order to run the tests, and to leave it undefined for tests where TLS is not enabled

I do not think my suggestion conflicts with @kevinAlbs in intent. However, the use of TLS CA certificates is not exclusive to Initial DNS Seedlist Discovery tests. Therefore I suggest using a dedicated variable `MONGOCXX_TEST_DNS` to indicate whether the DNS tests should be executed rather than `MONGOCXX_TEST_TLS_CA_FILE`, and leaving `MONGOCXX_TEST_DNS` undefined instead. A ticket similar to CDRIVER-4230 can then refer to `MONGOCXX_TEST_DNS` regarding enabling of tests rather than `MONGOCXX_TEST_TLS_CA_FILE`.

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,310 @@
...
+    bsoncxx::stdx::string_view uri;
+    bsoncxx::array::view hosts;
+    bsoncxx::document::view options;
+    bool error = false;
+    bool ping = true;
+
+    static initial_dns_seedlist_test parse(bsoncxx::document::view test_doc) {
+        initial_dns_seedlist_test test;
+
+        for (auto el : test_doc) {
```
eramongodb, line 39:
```suggestion
        for (auto el : test_doc) {
            const auto key = el.key();
```
Avoid repeated `.key()` lookup.

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+static void compare_options(bsoncxx::document::view_or_value expected_options,
+                            bsoncxx::document::view_or_value my_options,
+                            bsoncxx::stdx::optional<bsoncxx::document::view> creds) {
+    auto my_options_view = my_options.view();
+    for (const auto& expected_option : expected_options.view()) {
+        auto key = std::string(expected_option.key());
+        std::transform(key.begin(), key.end(), key.begin(), [](int c) { return std::tolower(c); });
```
eramongodb, line 105:
```suggestion
        auto key = mongocxx::test_util::tolowercase(expected_option.key());
```

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,310 @@
...
+}
+
+static void run_srv_max_hosts_test_file(const initial_dns_seedlist_test& test) {
+    using namespace mongocxx;
+
+    bool should_ping = true;
+    if (!test.ping) {
+        should_ping = false;
+    }
+    should_ping = should_ping && !test.error;
```
eramongodb, line 174:
```suggestion
    const bool should_ping = test.ping && !test.error;
```

src/mongocxx/test/spec/initial_dns_seedlist_discovery.cpp
```diff
@@ -0,0 +1,310 @@
...
+        for (const auto& i : servers) {
+            new_hosts.push_back(std::string(i.host()) + ":" + std::to_string(i.port()));
+        }
+    });
+
+    mongocxx::options::tls tls_options;
+    mongocxx::options::client client_options;
+
+    auto ca_file_path = std::getenv("MONGOCXX_TEST_TLS_CA_FILE");
+    REQUIRE(ca_file_path);
```
eramongodb, line 194:
```suggestion
    auto ca_file_path = test_util::getenv_or_fail("MONGOCXX_TEST_TLS_CA_FILE");
```

eramongodb (approved):
Some minor suggestions remaining; otherwise, LGTM.

vector-of-bool (approved):

PR #930, opened 2023-02-01
CXX-2184

kkloberdanz:
CXX-2184

Support $merge and $out executing on secondaries

Sync spec test files.

Fix bug in test suite.

Previously:
- read_preference was left out
- read_concern and write_concern were overwritten with empty objects, causing failures from the APM checker.

The fix is to only set these options if they are explicitly stated so in the test spec files.

### DIFF
--- a/src/mongocxx/test/spec/unified_tests/runner.cpp
+++ b/src/mongocxx/test/spec/unified_tests/runner.cpp
@@ -342,18 +342,21 @@ options::server_api create_server_api(document::view object) {
 
 read_preference get_read_preference(const document::element& opts) {
     read_preference rp;
-    auto read_pref = opts["readPreference"];
-    auto doc = read_pref.get_document().value;
-    auto mode = std::string(doc["mode"].get_string().value);
-    if (doc["maxStalenessSeconds"]) {
-        auto max_staleness_seconds = doc["maxStalenessSeconds"].get_int32().value;
-        rp.max_staleness(std::chrono::seconds(max_staleness_seconds));
+
+    const auto read_pref = opts["readPreference"];
+
+    if (const auto mss = read_pref["maxStalenessSeconds"]) {
+        rp.max_staleness(std::chrono::seconds(mss.get_int32().value));
     }
-    if (mode == "secondaryPreferred") {
+
+    const auto mode = read_pref["mode"].get_string().value;
+
+    if (mode.compare("secondaryPreferred") == 0) {
         rp.mode(read_preference::read_mode::k_secondary_preferred);
     } else {
-        FAIL("unhandled readPreference mode: " + mode);
+        FAIL("unhandled readPreference mode: " << mode);
     }
+
     return rp;
 }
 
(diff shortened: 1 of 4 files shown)


src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -340,10 +340,22 @@ options::server_api create_server_api(document::view object) {
...
-write_concern get_write_concern(const document::element& opts) {
-    if (!opts["writeConcern"])
-        return {};
+read_preference get_read_preference(const document::element& opts) {
+    read_preference rp;
+    auto read_pref = opts["readPreference"];
+    auto doc = read_pref.get_document().value;
+    auto mode = std::string(doc["mode"].get_string().value);
+    auto max_staleness_seconds = doc["maxStalenessSeconds"].get_int32().value;
+    rp.max_staleness(std::chrono::seconds(max_staleness_seconds));
```
kevinAlbs, line 349: `maxStalenessSeconds` is an optional field. See the [Unified Test Format](https://github.com/mongodb/specifications/blob/31bbf60c4f3008d77fc5ca5287b165e5ea97a4a6/source/unified-test-format/unified-test-format.rst#commonoptions-readpreference) for reference.

```suggestion
    if (doc["maxStalenessSeconds"]) {
        auto max_staleness_seconds = doc["maxStalenessSeconds"].get_int32().value;
        rp.max_staleness(std::chrono::seconds(max_staleness_seconds));
    }
```

kevinAlbs (approved):

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -340,10 +340,24 @@ options::server_api create_server_api(document::view object) {
...
+        auto max_staleness_seconds = doc["maxStalenessSeconds"].get_int32().value;
+        rp.max_staleness(std::chrono::seconds(max_staleness_seconds));
+    }
+    if (mode == "secondaryPreferred") {
+        rp.mode(read_preference::read_mode::k_secondary_preferred);
+    } else {
+        FAIL("unhandled readPreference mode: " + mode);
+    }
+    return rp;
+}
```
eramongodb, line 358:
```suggestion
read_preference get_read_preference(const document::element& opts) {
    read_preference rp;

    const auto read_pref = opts["readPreference"];

    if (const auto mss = read_pref["maxStalenessSeconds"]) {
        rp.max_staleness(std::chrono::seconds(mss.get_int32().value));
    }

    const auto mode = read_pref["mode"].get_string().value;

    if (mode.compare("secondaryPreferred") == 0) {
        rp.mode(read_preference::read_mode::k_secondary_preferred);
    } else {
        FAIL("unhandled readPreference mode: " << mode);
    }

    return rp;
}
```

eramongodb (approved):
Suggesting some improvements to the `get_read_preference()` definition; otherwise, LGTM.

vector-of-bool (approved):
Sorry for the delay. LGTM

PR #935, opened 2023-02-22
Key Management API

kkloberdanz:
CXX-2425

- Add support for new [custom key material](https://github.com/mongodb/specifications/blob/f941df836c97a19820879afd726d550541a78cbe/source/client-side-encryption/client-side-encryption.rst#keymaterial) in DataKeyOpts.
- Add support for new [Key Management API functions](https://github.com/mongodb/specifications/blob/f941df836c97a19820879afd726d550541a78cbe/source/client-side-encryption/client-side-encryption.rst#clientencryption) in ClientEncryption.

### DIFF
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -165,7 +165,8 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKey/
     ///
-    stdx::optional<bsoncxx::document::value> get_key(bsoncxx::types::bson_value::view_or_value id);
+    stdx::optional<bsoncxx::document::view_or_value> get_key(
+        bsoncxx::types::bson_value::view_or_value id);
 
     ///
     /// Finds all documents in the key vault collection.
@@ -192,7 +193,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.addKeyAlternateName/
     ///
-    stdx::optional<bsoncxx::document::value> add_key_alt_name(
+    stdx::optional<bsoncxx::document::view_or_value> add_key_alt_name(
         bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
 
     ///
@@ -209,7 +210,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.removeKeyAlternateName/
     ///
-    stdx::optional<bsoncxx::document::value> remove_key_alt_name(
+    stdx::optional<bsoncxx::document::view_or_value> remove_key_alt_name(
         bsoncxx::types::bson_value::view_or_value id, const std::string& key_alt_name);
 
     ///
@@ -223,7 +224,8 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKeyByAltName/
     ///
-    stdx::optional<bsoncxx::document::value> get_key_by_alt_name(const std::string& key_alt_name);
+    stdx::optional<bsoncxx::document::view_or_value> get_key_by_alt_name(
+        const std::string& key_alt_name);
 
    private:
     class MONGOCXX_PRIVATE impl;
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -138,7 +138,7 @@ class MONGOCXX_API client_encryption {
     /// https://www.mongodb.com/docs/manual/reference/method/KeyVault.rewrapManyDataKey/
     ///
     result::rewrap_many_datakey rewrap_many_datakey(bsoncxx::document::view_or_value filter,
-                                                    options::rewrap_many_datakey& opts);
+                                                    const options::rewrap_many_datakey& opts);
 
     ///
     /// Removes the key document with the given UUID (BSON binary subtype 0x04)
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -194,7 +194,6 @@ class MONGOCXX_API client_encryption {
     ///
     stdx::optional<bsoncxx::document::value> add_key_alt_name(
         bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
-        bsoncxx::types::bson_value::view_or_value id, const std::string& key_alt_name);
 
     ///
     /// Removes a keyAltName from the keyAltNames array of the key document in
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -193,6 +193,7 @@ class MONGOCXX_API client_encryption {
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.addKeyAlternateName/
     ///
     stdx::optional<bsoncxx::document::value> add_key_alt_name(
+        bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
         bsoncxx::types::bson_value::view_or_value id, const std::string& key_alt_name);
 
     ///
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -199,7 +199,7 @@ class MONGOCXX_API client_encryption {
     /// Removes a keyAltName from the keyAltNames array of the key document in
     /// the key vault collection with the given UUID (BSON binary subtype 0x04).
     ///
-    /// @param id Binary id of which key to delete
+    /// @param id Binary id of the key to remove the key alternate name from
     ///
     /// @param key_alt_name String alternative name for the key
     ///
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -132,7 +132,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @return a RewrapManyDataKeyResult.
     ///
-    /// @throws mongocxx::exception if there is an error rewraping the key.
+    /// @throws mongocxx::exception if there is an error rewrapping the key.
     ///
     /// @see
     /// https://www.mongodb.com/docs/manual/reference/method/KeyVault.rewrapManyDataKey/
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -211,7 +211,7 @@ class MONGOCXX_API client_encryption {
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.removeKeyAlternateName/
     ///
     stdx::optional<bsoncxx::document::view_or_value> remove_key_alt_name(
-        bsoncxx::types::bson_value::view_or_value id, const std::string& key_alt_name);
+        bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
 
     ///
     /// Get the key document from the key vault collection with the provided name.
@@ -225,7 +225,7 @@ class MONGOCXX_API client_encryption {
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKeyByAltName/
     ///
     stdx::optional<bsoncxx::document::view_or_value> get_key_by_alt_name(
-        const std::string& key_alt_name);
+        bsoncxx::string::view_or_value key_alt_name);
 
    private:
     class MONGOCXX_PRIVATE impl;
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -182,7 +182,7 @@ class MONGOCXX_API client_encryption {
     /// Adds a keyAltName to the keyAltNames array of the key document in the
     /// key vault collection with the given UUID (BSON binary subtype 0x04).
     ///
-    /// @param id Binary id of which key to delete
+    /// @param id Binary id of the key to add the key alternate name to
     ///
     /// @param key_alt_name String alternative name for the key
     ///
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -165,8 +165,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKey/
     ///
-    stdx::optional<bsoncxx::document::view_or_value> get_key(
-        bsoncxx::types::bson_value::view_or_value id);
+    stdx::optional<bsoncxx::document::value> get_key(bsoncxx::types::bson_value::view_or_value id);
 
     ///
     /// Finds all documents in the key vault collection.
@@ -193,7 +192,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.addKeyAlternateName/
     ///
-    stdx::optional<bsoncxx::document::view_or_value> add_key_alt_name(
+    stdx::optional<bsoncxx::document::value> add_key_alt_name(
         bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
 
     ///
@@ -210,7 +209,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.removeKeyAlternateName/
     ///
-    stdx::optional<bsoncxx::document::view_or_value> remove_key_alt_name(
+    stdx::optional<bsoncxx::document::value> remove_key_alt_name(
         bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
 
     ///
@@ -224,7 +223,7 @@ class MONGOCXX_API client_encryption {
     ///
     /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKeyByAltName/
     ///
-    stdx::optional<bsoncxx::document::view_or_value> get_key_by_alt_name(
+    stdx::optional<bsoncxx::document::value> get_key_by_alt_name(
         bsoncxx::string::view_or_value key_alt_name);
 
    private:
--- a/src/mongocxx/client_encryption.hpp
+++ b/src/mongocxx/client_encryption.hpp
@@ -127,7 +127,7 @@ class MONGOCXX_API client_encryption {
     ///   Document to filter which keys get re-wrapped.
     ///
     /// @param opts
-    ///   Options to specify which provider stores the key and an optional
+    ///   Options to specify which provider to encrypt the data keys and an optional
     ///   master key document.
     ///
     /// @return a RewrapManyDataKeyResult.
--- a/src/mongocxx/options/data_key.hpp
+++ b/src/mongocxx/options/data_key.hpp
@@ -30,8 +30,6 @@ class client_encryption;
 
 namespace options {
 
-using key_material_type = std::vector<uint8_t>;
-
 ///
 /// Class representing options for data key generation for encryption.
 ///
@@ -141,6 +139,7 @@ class MONGOCXX_API data_key {
     ///
     /// @see https://www.mongodb.com/docs/v6.0/reference/method/KeyVault.createKey/
     ///
+    using key_material_type = std::vector<uint8_t>;
     data_key& key_material(key_material_type key_material);
 
     ///
--- a/src/mongocxx/options/rewrap_many_datakey.hpp
+++ b/src/mongocxx/options/rewrap_many_datakey.hpp
@@ -31,7 +31,7 @@ namespace options {
 class MONGOCXX_API rewrap_many_datakey {
    public:
     ///
-    /// Set the KMS provider
+    /// Set the optional KMS provider use to encrypt the data keys. Do not set to use the current KMS provider(s).
     ///
     /// A KMS provider (AWS KMS, Azure Key Vault, GCP KMS, the local provider,
     /// or KMIP) is used to decrypt data keys after fetching from the MongoDB
--- a/src/mongocxx/options/rewrap_many_datakey.hpp
+++ b/src/mongocxx/options/rewrap_many_datakey.hpp
@@ -31,7 +31,8 @@ namespace options {
 class MONGOCXX_API rewrap_many_datakey {
    public:
     ///
-    /// Set the optional KMS provider use to encrypt the data keys. Do not set to use the current KMS provider(s).
+    /// Set the optional KMS provider use to encrypt the data keys. Do not set to use the current
+    /// KMS provider(s).
     ///
     /// A KMS provider (AWS KMS, Azure Key Vault, GCP KMS, the local provider,
     /// or KMIP) is used to decrypt data keys after fetching from the MongoDB
@@ -62,7 +63,7 @@ class MONGOCXX_API rewrap_many_datakey {
     /// @see
     /// https://www.mongodb.com/docs/manual/core/csfle/reference/kms-providers/#std-label-csfle-reference-kms-providers
     ///
-    const stdx::optional<std::string>& provider();
+    const stdx::optional<std::string>& provider() const;
 
     ///
     /// Set the masterKey option.
@@ -94,7 +95,7 @@ class MONGOCXX_API rewrap_many_datakey {
     /// @see
     /// https://www.mongodb.com/docs/manual/core/csfle/reference/kms-providers/#std-label-csfle-reference-kms-providers-create-and-store
     ///
-    const stdx::optional<bsoncxx::document::view_or_value>& master_key();
+    const stdx::optional<bsoncxx::document::view_or_value>& master_key() const;
 
    private:
     friend class mongocxx::client_encryption;
--- a/src/mongocxx/options/rewrap_many_datakey.hpp
+++ b/src/mongocxx/options/rewrap_many_datakey.hpp
@@ -46,7 +46,7 @@ class MONGOCXX_API rewrap_many_datakey {
     /// @see
     /// https://www.mongodb.com/docs/manual/core/csfle/reference/kms-providers/#std-label-csfle-reference-kms-providers
     ///
-    rewrap_many_datakey& provider(std::string provider);
+    rewrap_many_datakey& provider(bsoncxx::string::view_or_value provider);
 
     ///
     /// Get the KMS provider
--- a/src/mongocxx/options/rewrap_many_datakey.hpp
+++ b/src/mongocxx/options/rewrap_many_datakey.hpp
@@ -36,8 +36,7 @@ class MONGOCXX_API rewrap_many_datakey {
     ///
     /// A KMS provider (AWS KMS, Azure Key Vault, GCP KMS, the local provider,
     /// or KMIP) is used to decrypt data keys after fetching from the MongoDB
-    /// Key Vault, and encrypt newly created data keys. Refer to KMSProviders
-    /// for the shape of the KMS provider options.
+    /// Key Vault, and encrypt newly created data keys.
     ///
     /// @param provider String name of the provider.
     ///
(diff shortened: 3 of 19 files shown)


src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,6 +482,25 @@ options::client_session get_session_options(document::view object) {
     return session_opts;
 }
 
+options::client_encryption get_client_encryption_options(document::view object,
+                                                         client* key_vault_client) {
```
eramongodb, line 486:
```suggestion
options::client_encryption get_client_encryption_options(document::view, client* key_vault_client) {
```
Leave parameter unnamed instead of `(void)object`.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -445,6 +482,25 @@ options::client_session get_session_options(document::view object) {
...
+    options::client_encryption ce_opts;
+    ce_opts.key_vault_client(key_vault_client);
+    ce_opts.key_vault_namespace({"keyvault", "datakeys"});
+    ce_opts.kms_providers(make_kms_doc());
```
eramongodb, line 490: `keyVaultClient` and `kmsProviders` are parameters that are specified by the client encryption entity definition. Additionally, TLS opts below should not be set unless the KMIP KMS provider is provided. See [C Driver implementation](https://github.com/mongodb/mongo-c-driver/blob/72cba009ec70fd734dc76426c5385487a1c6930d/src/libmongoc/tests/unified/entity-map.c#L799) for reference.

kevinAlbs:
Commits in DRIVERS-2017 also include tests to validate the unified test runner implementation.
I suggest copying the following test files to `data/unified-valid-pass` from the last modified commit (https://github.com/mongodb/specifications/commit/d8be9cee7a96d921303983bcb40ee80d533e6b34) which includes the test simplification of CXX-2538:

```
source/unified-test-format/tests/valid-pass/kmsProviders-explicit_kms_credentials.json
source/unified-test-format/tests/valid-pass/kmsProviders-placeholder_kms_credentials.json
source/unified-test-format/tests/valid-pass/kmsProviders-unconfigured_kms.json
source/unified-test-format/tests/valid-pass/kmsProviders-mixed_kms_credential_fields.json
```

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -172,6 +172,24 @@ bool is_special(T doc) {
            string::to_string(doc.get_document().value.begin()->key()).rfind("$$", 0) == 0;
 }
 
+// The C driver returns different names than the spec test. This will return the
+// name as the C driver declares it.
+std::string get_alternate_name(const std::string& key) {
```
eramongodb, line 177:
```suggestion
bsoncxx::stdx::string_view get_alternate_name(bsoncxx::stdx::string_view key) {
```
Use `bsoncxx::stdx::string_view` to avoid redundant allocations.

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -190,26 +208,34 @@ void matches_document(types::bson_value::view actual,
...
 
-        assert::matches(actual_doc[kvp.key()].get_value(), kvp.get_value(), map, false);
-        --extra_fields;
-    }
+        if (actual_doc[key]) {
+            assert::matches(actual_doc[key].get_value(), kvp.get_value(), map, false);
+        } else if (actual_doc[original_key]) {
+            assert::matches(actual_doc[original_key].get_value(), kvp.get_value(), map, false);
+        } else {
+            FAIL("expected field '" + key + "' to be present, but it is absent");
```
eramongodb, line 235:
```suggestion
            FAIL("expected field '" << key << "' to be present, but it is absent");
```

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -190,26 +208,34 @@ void matches_document(types::bson_value::view actual,
...
+            }
         }
 
-        assert::matches(actual_doc[kvp.key()].get_value(), kvp.get_value(), map, false);
-        --extra_fields;
-    }
+        if (actual_doc[key]) {
+            assert::matches(actual_doc[key].get_value(), kvp.get_value(), map, false);
+        } else if (actual_doc[original_key]) {
+            assert::matches(actual_doc[original_key].get_value(), kvp.get_value(), map, false);
```
eramongodb, line 233:
```suggestion
        if (const auto doc = actual_doc[key]) {
            assert::matches(doc.get_value(), kvp.get_value(), map, false);
        } else if (const auto doc = actual_doc[original_key]) {
            assert::matches(doc.get_value(), kvp.get_value(), map, false);
```
Avoid repeated field lookups.

eramongodb (changes requested):

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -46,8 +47,8 @@ using bsoncxx::builder::basic::kvp;
 using bsoncxx::builder::basic::make_document;
 
 using schema_versions_t =
-    std::array<std::array<int, 3 /* major.minor.patch */>, 2 /* supported version */>;
-constexpr schema_versions_t schema_versions{{{{1, 1, 0}}, {{1, 7, 0}}}};
+    std::array<std::array<int, 3 /* major.minor.patch */>, 3 /* supported version */>;
+constexpr schema_versions_t schema_versions{{{{1, 1, 0}}, {{1, 7, 0}}, {{1, 8, 0}}}};
```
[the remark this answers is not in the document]
kkloberdanz: Would bumping the version number be the preferred way?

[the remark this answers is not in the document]
eramongodb: It's what we've [done prior](https://github.com/mongodb/mongo-cxx-driver/commit/51e570163744f36f28a106967577369b65be3ea4#diff-143f9d59281ec9632dcf24fa820b9595c463cd4f3cf2c66f185cb0503f0c53a3R50). tbh so long as it is able to signal when new features are required to be implemented, I suppose it should work just fine. Our support for a given unified spec version is not comprehensive anyways. 🤷

src/mongocxx/test/spec/unified_tests/assert.cpp
```diff
@@ -172,6 +172,24 @@ bool is_special(T doc) {
            string::to_string(doc.get_document().value.begin()->key()).rfind("$$", 0) == 0;
 }
 
+// The C driver returns different names than the spec test. This will return the
+// name as the C driver declares it.
+bsoncxx::stdx::string_view get_alternate_name(bsoncxx::stdx::string_view key) {
```
eramongodb, line 177:
```suggestion
std::string get_alternate_name(const std::string& key) {
```
I retract my earlier suggestion to use `bsoncxx::stdx::string_view` after realizing that the mnmlstc implementation of `string_view` for v1.1.0 (what the CXX Driver currently depends on) is [utterly broken](https://github.com/mnmlstc/core/blob/72cccd977409e3581f05228ac027761ca4cea5dd/include/core/string.hpp#L363) (hashes using the value of the _internal pointer_ instead of the string value itself). It appears to have been fixed in [this commit](https://github.com/mnmlstc/core/commit/42a34a9b6168733d60f1334d875247d67cc364fc#diff-14a62deac1406c596d24f2f3ebdb019dab87a1caa9e35cb92e9fea83596662cb), but there is no release version containing this commit. 😐

kevinAlbs (changes requested):
Also add the prose tests added in commits for DRIVERS-2017:
- https://github.com/mongodb/specifications/commit/eee051cd868659556c2ea6b9b558f0d52e63e3db#diff-c57f446f49351e6f7432a2d3028c880a728f1a33031510a7c376aed157d6de00R377
- https://github.com/mongodb/specifications/commit/f941df836c97a19820879afd726d550541a78cbe#diff-c57f446f49351e6f7432a2d3028c880a728f1a33031510a7c376aed157d6de00R1777

The prose tests for CSFLE in the C++ driver are in `src/mongocxx/test/client_side_encryption.cpp`.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1032,4 +1130,18 @@ TEST_CASE("collection management spec automated tests", "[unified_format_spec]")
...
 }
 
+// See:
+// https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst
+TEST_CASE("client side encryption unified format spec automated tests", "[unified_format_spec]") {
+    if (!mongocxx::test_util::should_run_client_side_encryption_test()) {
+        WARN("Skipping - client side encryption unified tests");
+        return;
+    }
+    if (!mongocxx::test_util::newer_than({uri{}}, "4.2")) {
```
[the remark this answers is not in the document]
kkloberdanz: Will do! Is there a good way of checking if CSFLE is enabled on the server?
> If false, tests MUST only run if CSFLE is not enabled.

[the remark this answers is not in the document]
kevinAlbs: > Will do! Is there a good way of checking if CSFLE is enabled on the server?

A server supports CSFLE if it is version 4.2.0 or higher. The `newer_than` function could be used when evaluating the `csfle` requirement. (I realize the original comment sounds like it suggests not to use `newer_than` at all 🤦 ).

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1548,10 +1548,25 @@ document::value rewrap_many_datakey(entity::map& map,
...
+        basic_builder.append(kvp("upsertedCount", bulk_write_result.upserted_count()));
+        basic_builder.append(kvp("upsertedIds", make_document()));
+        if (matched) {
+            basic_builder.append(kvp("matchedCount", matched));
+        }
+        if (modified) {
+            basic_builder.append(kvp("modifiedCount", modified));
+        }
+        auto doc = make_document(
+            kvp("result", make_document(kvp("bulkWriteResult", basic_builder.extract()))));
```
kevinAlbs, line 1569: Converting a bulk write result to a BSON document occurs elsewhere in this file. Consider adding a helper to convert bulk write result to a BSON document, which includes the `upsertedIds`:

```c++
document::value bulk_write_result_to_document(const mongocxx::result::bulk_write& result) {
    builder::basic::document upserted_ids_builder;
    for (auto&& index_and_id : result.upserted_ids()) {
        upserted_ids_builder.append(
            kvp(std::to_string(index_and_id.first), index_and_id.second.get_int32().value));
    }

    return make_document(kvp("matchedCount", result.matched_count()),
                         kvp("modifiedCount", result.matched_count()),
                         kvp("deletedCount", result.deleted_count()),
                         kvp("insertedCount", result.inserted_count()),
                         kvp("upsertedCount", result.upserted_count()),
                         kvp("upsertedIds", upserted_ids_builder.extract()));
}
```

The helper can be used here:

```suggestion
        auto doc = make_document(
            kvp("result",
                make_document(
                    kvp("bulkWriteResult", bulk_write_result_to_document(bulk_write_result)))));
        return doc;
```

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -110,6 +110,41 @@ bsoncxx::document::value make_kms_doc() {
...
+        auto variables_doc = bsoncxx::builder::basic::document{};
+        auto variables = test_kms_doc_view[provider].get_document().view();
+        for (const auto& i : variables) {
+            auto variable = i.key();
+            auto actual_value = kms_values[provider][variable];
+            if (!kms_values[provider][variable]) {
+                FAIL("FAIL: expecting to find variable: '"
+                     << variable << "' in KMS doc for provider: '" << provider << "'");
+            }
+            switch (actual_value.type()) {
```
kevinAlbs, line 131: Suggested change to handle the case when the [input is not a placeholder](https://github.com/mongodb/specifications/blob/baf3724155500acbb557e158b8a5d4334e7f3512/source/unified-test-format/unified-test-format.rst#entity).

> Otherwise, drivers MUST configure the KMS provider with the explicit value of KMS credentials field given in the test file (e.g. kmsProviders: { aws: { accessKeyId: abc, secretAccessKey: def } }). This is to permit testing conditions where invalid KMS credentials are provided.

```suggestion
            bool is_placeholder = false;
            if (i.type() == bsoncxx::type::k_document &&
                i.get_document().value == make_document(kvp("$$placeholder", 1))) {
                is_placeholder = true;
            }
            if (!is_placeholder) {
                // Append value as-is.
                variables_doc.append(kvp(variable, i.get_value()));
                continue;
            }
            // A placeholder was specified. Append the credential from the environment.
            switch (actual_value.type()) {
```

I do not think this behavior is tested. But it can be verified by replacing a `{ "$$placeholder": 1}` with `"foobar"` in `rewrapManyDataKey.json` and expecting an error.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -226,6 +261,21 @@ bool compatible_with_server(const bsoncxx::array::element& requirement) {
...
+    if (auto csfle = requirement["csfle"]) {
+        // csfle: Optional boolean. If true, the tests MUST only run if the
+        // driver and server support Client-Side Field Level Encryption. A
+        // server supports CSFLE if it is version 4.2.0 or higher. If false,
+        // tests MUST only run if CSFLE is not enabled. If this field is
+        // omitted, there is no CSFLE requirement.
+        std::vector<int> requires_at_lest{4, 2, 0};
+        bool is_csfle = csfle.get_bool().value;
+        if (is_csfle) {
+            if (!is_compatible_version(requires_at_lest, expected)) {
```
kevinAlbs, line 274:
```suggestion
        std::vector<int> requires_at_least{4, 2, 0};
        bool is_csfle = csfle.get_bool().value;
        if (is_csfle) {
            if (!is_compatible_version(requires_at_least, expected)) {
```

kevinAlbs (approved):

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -73,6 +74,88 @@ entity::map& get_entity_map() {
...
+                kvp("email", test_util::getenv_or_fail("MONGOCXX_TEST_GCP_EMAIL")),
+                kvp("privateKey", test_util::getenv_or_fail("MONGOCXX_TEST_GCP_PRIVATEKEY")))),
+        kvp("kmip", make_document(kvp("endpoint", "localhost:5698"))),
+        kvp("local", make_document(kvp("key", local_master_key))));
+
+    return kms_doc;
+}
+
+bsoncxx::document::value parse_kms_doc(bsoncxx::document::view_or_value test_kms_doc) {
+    auto kms_values = get_kms_values();
```
eramongodb, line 114:
```suggestion
    const auto kms_values = get_kms_values();
```
Const-qualify local variables that are not modified within this function.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -114,6 +114,21 @@ options::find make_find_options(document::view arguments) {
...
 }
 
+document::value bulk_write_result_to_document(const mongocxx::result::bulk_write& result) {
+    builder::basic::document upserted_ids_builder;
+    for (auto&& index_and_id : result.upserted_ids()) {
+        upserted_ids_builder.append(
+            kvp(std::to_string(index_and_id.first), index_and_id.second.get_int32().value));
+    }
+
+    return make_document(kvp("matchedCount", result.matched_count()),
```
eramongodb, line 124: Per note [here](https://github.com/mongodb/mongo-cxx-driver/blob/57e37617421a8cb7210856ca491ce2131d9b4127/src/mongocxx/test/spec/operation.cpp#L1127), may want to include an `insertedIds: {}` field.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -2390,4 +2393,261 @@ TEST_CASE("Explicit Encryption", "[client_side_encryption]") {
     }
 }
 
+TEST_CASE("Unique Index on keyAltNames", "[client_side_encryption]") {
+    instance::current();
+
+    if (!mongocxx::test_util::should_run_client_side_encryption_test()) {
+        std::cerr << "Skipping Unique Index on keyAltNames prose tests" << std::endl;
```
eramongodb, line 2400:
```suggestion
```
`should_run_client_side_encryption_test` already emits skip reason.

src/mongocxx/test/client_side_encryption.cpp
```diff
@@ -2390,4 +2393,261 @@ TEST_CASE("Explicit Encryption", "[client_side_encryption]") {
...
+        return;
+    }
+
+    if (!test_util::newer_than(uri{}, "4.2")) {
+        std::cerr << "Unique Index on keyAltNames tests require MongoDB server 4.2+." << std::endl;
```
eramongodb, line 2405:
```suggestion
        WARN ("Skipping - requires MongoDB server 4.2+");
```
Use Catch2 `WARN` instead of `std::cerr`. Context (test section) will be automatically included in test output.

src/mongocxx/options/rewrap_many_datakey.hpp
```diff
@@ -0,0 +1,108 @@
...
+    /// or KMIP) is used to decrypt data keys after fetching from the MongoDB
+    /// Key Vault, and encrypt newly created data keys.
+    ///
+    /// @return
+    ///   An optional string name of the provider.
+    ///
+    /// @see
+    /// https://www.mongodb.com/docs/manual/core/csfle/reference/kms-providers/#std-label-csfle-reference-kms-providers
+    ///
+    const stdx::optional<std::string>& provider() const;
```
eramongodb, line 64:
```suggestion
    bsoncxx::string::view_or_value provider() const;
```
`provider` is not optional; suggest removing the `stdx::optional` for `provider` from this class. For consistency with other `options` classes, use `bsoncxx::string::view_or_value` 
instead of `const std::string&` or `stdx::string_view`.

src/mongocxx/options/data_key.hpp
```diff
@@ -30,6 +30,8 @@ class client_encryption;
 
 namespace options {
 
+using key_material_type = std::vector<uint8_t>;
```
eramongodb, line 33:
```suggestion
using key_material_type = std::array<uint8_t, 96>;
```
Suggest non-allocating and fixed-length type for `key_material_type` (key material length is [specified by spec](https://github.com/mongodb/specifications/blob/baf3724155500acbb557e158b8a5d4334e7f3512/source/client-side-encryption/client-side-encryption.rst#keymaterial) to be exactly 96 bytes).

Also suggest moving type alias definition into `data_key` class for consistency with other `options` classes such as: https://github.com/mongodb/mongo-cxx-driver/blob/57e37617421a8cb7210856ca491ce2131d9b4127/src/mongocxx/options/client_encryption.hpp#L77

CC @kevinAlbs I think we discussed this before but I do not remember the details. Is this suggestion appropriate, or did we conclude that we want to leave the length open to change in the future (in which case `std::vector` may be appropriate)?

src/mongocxx/client_encryption.hpp
```diff
@@ -112,6 +116,115 @@ class MONGOCXX_API client_encryption {
...
+    ///
+    /// @param key_alt_name String alternative name for the key
+    ///
+    /// @throws mongocxx::exception if there is an error adding the key alt name.
+    ///
+    /// @return the previous version of the key document.
+    ///
+    /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.addKeyAlternateName/
+    ///
+    stdx::optional<bsoncxx::document::value> add_key_alt_name(
```
eramongodb, line 195:
```suggestion
    stdx::optional<bsoncxx::document::value> add_key_alt_name(
        bsoncxx::types::bson_value::view_or_value id, bsoncxx::string::view_or_value key_alt_name);
```
Suggest using `bsoncxx::string::view_or_value` in API when able for consistency with other public interfaces.

src/mongocxx/options/rewrap_many_datakey.hpp
```diff
@@ -0,0 +1,108 @@
...
+    ///
+    /// @param provider String name of the provider.
+    ///
+    /// @return
+    ///   An optional document containing the TLS options.
+    ///
+    /// @see
+    /// https://www.mongodb.com/docs/manual/core/csfle/reference/kms-providers/#std-label-csfle-reference-kms-providers
+    ///
+    rewrap_many_datakey& provider(std::string provider);
```
eramongodb, line 49:
```suggestion
    rewrap_many_datakey& provider(bsoncxx::string::view_or_value provider);
```
Suggest using `bsoncxx::string::view_or_value` in public API when able for consistency with other public interfaces.

src/mongocxx/options/data_key.hpp
kkloberdanz: My understanding of that conversation was the conclusion was to make the key material type variable and leave all validation of length to libmongocrypt. Thread here: https://mongodb.slack.com/archives/GMY98JS86/p1677166844729879?thread_ts=1677164857.739539&cid=GMY98JS86

src/mongocxx/client_encryption.hpp
```diff
@@ -112,6 +116,115 @@ class MONGOCXX_API client_encryption {
...
+    /// @param key_alt_name String alternative name for the key
+    ///
+    /// @throws mongocxx::exception if there is an error removing the key alt name.
+    ///
+    /// @return The previous version of the key document.
+    ///
+    /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.removeKeyAlternateName/
+    ///
+    stdx::optional<bsoncxx::document::value> remove_key_alt_name(
+        bsoncxx::types::bson_value::view_or_value id, const std::string& key_alt_name);
```
eramongodb, line 213: Use `bsoncxx::string::view_or_value` in _all_ new public API functions for consistency.

src/mongocxx/client_encryption.hpp
```diff
@@ -112,6 +116,115 @@ class MONGOCXX_API client_encryption {
...
+    ///
+    /// @param key_alt_name String alternative name for the key
+    ///
+    /// @throws mongocxx::exception if there is an error getting the key by alt name.
+    ///
+    /// @return A key document in the key vault collection with the given keyAltName.
+    ///
+    /// @see https://www.mongodb.com/docs/manual/reference/method/KeyVault.getKeyByAltName/
+    ///
+    stdx::optional<bsoncxx::document::value> get_key_by_alt_name(const std::string& key_alt_name);
```
eramongodb, line 226: Use `bsoncxx::string::view_or_value` in _all_ new public API functions for consistency.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -73,6 +74,88 @@ entity::map& get_entity_map() {
...
+        kvp("kmip", make_document(kvp("endpoint", "localhost:5698"))),
+        kvp("local", make_document(kvp("key", local_master_key))));
+
+    return kms_doc;
+}
+
+bsoncxx::document::value parse_kms_doc(bsoncxx::document::view_or_value test_kms_doc) {
+    const auto kms_values = get_kms_values();
+    auto doc = bsoncxx::builder::basic::document{};
+    auto test_kms_doc_view = test_kms_doc.view();
```
eramongodb, line 116: Apply `const`-correctness to _all_ local variables that are not modified, not just `kms_values`.

src/mongocxx/private/client_encryption.hh
```diff
@@ -142,6 +144,194 @@ class client_encryption::impl {
...
+    result::rewrap_many_datakey rewrap_many_datakey(bsoncxx::document::view_or_value filter,
+                                                    const options::rewrap_many_datakey& opts) {
+        bson_error_t error;
+
+        std::unique_ptr<mongoc_client_encryption_rewrap_many_datakey_result_t,
+                        decltype(libmongoc::client_encryption_rewrap_many_datakey_result_destroy)>
+            result_ptr(libmongoc::client_encryption_rewrap_many_datakey_result_new(),
+                       libmongoc::client_encryption_rewrap_many_datakey_result_destroy);
+
+        auto provider = opts.provider();
```
eramongodb, line 156: Apply `const`-correctness to local variables that are not modified.

eramongodb (changes requested):

src/mongocxx/private/client_encryption.hh
```diff
@@ -142,6 +144,200 @@ class client_encryption::impl {
...
+        const bson_t* bulk_write_result =
+            libmongoc::client_encryption_rewrap_many_datakey_result_get_bulk_write_result(
+                result_ptr.get());
+
+        if (bulk_write_result) {
+            const auto doc =
+                bsoncxx::document::view(bson_get_data(bulk_write_result), bulk_write_result->len);
+            const auto val = bsoncxx::document::value(doc);
+            result::bulk_write bulk_write_result{val};
+            return result::rewrap_many_datakey(bulk_write_result);
```
eramongodb, line 192:
```suggestion
            return result::rewrap_many_datakey(result::bulk_write(bsoncxx::document::value(doc)));
```
Avoid redundant copies.

src/mongocxx/private/client_encryption.hh
```diff
@@ -142,6 +144,200 @@ class client_encryption::impl {
...
+        const auto r = libmongoc::client_encryption_delete_key(
+            _client_encryption_t, &libbson_key, reply_ptr.bson_for_init(), &error);
+
+        bson_value_destroy(&libbson_key);
+
+        if (!r) {
+            throw_exception<operation_exception>(error);
+        }
+
+        const bsoncxx::document::value val = reply_ptr.steal();
```
eramongodb, line 218:
```suggestion
        const auto val = reply_ptr.view();
```
Steal not necessary; only reads required.

src/mongocxx/private/client_encryption.hh
```diff
@@ -142,6 +144,200 @@ class client_encryption::impl {
...
+        // The C driver calls this field "deletedCount", but the C++ driver
+        // refers to this as "nRemoved". Make a new document with the field name
+        // changed to get around this.
+        //
+        // See: mongo-cxx-driver/src/mongocxx/result/bulk_write.cpp
+        // Function: std::int32_t bulk_write::deleted_count() const {
+        //     return view()["nRemoved"].get_int32();
+        // }
+        const auto new_val = make_document(kvp("nRemoved", val["deletedCount"].get_int32()));
+        return result::delete_result(result::bulk_write(new_val));
```
eramongodb, line 229:
```suggestion
        return result::delete_result(
            result::bulk_write(make_document(kvp("nRemoved", val["deletedCount"].get_int32()))));
```
Avoid redundant copies.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -73,6 +74,88 @@ entity::map& get_entity_map() {
...
+    "\x4f\x69\x37\x36\x36\x4a\x7a\x58\x5a\x42\x64\x42\x64\x62\x64\x4d\x75\x72"
+    "\x64\x6f\x6e\x4a\x31\x64";
+
+bsoncxx::document::value get_kms_values() {
+    char key_storage[96];
+    memcpy(&(key_storage[0]), kLocalMasterKey, 96);
+    const bsoncxx::types::b_binary local_master_key{
+        bsoncxx::binary_sub_type::k_binary, 96, (const uint8_t*)&key_storage};
+
+    const auto kms_doc = make_document(
```
eramongodb, line 91:
```suggestion
    auto kms_doc = make_document(
```
Do not prevent NRVO.

src/mongocxx/private/client_encryption.hh
```diff
@@ -142,6 +144,200 @@ class client_encryption::impl {
...
+        if (!r) {
+            cleanup();
+            throw_exception<operation_exception>(error);
+        }
+
+        const bsoncxx::document::view_or_value val(key_doc.steal());
+
+        cleanup();
+        return val.view().empty() ? stdx::nullopt
+                                  : stdx::optional<bsoncxx::document::view_or_value>{val};
```
eramongodb, line 255:
```cpp
cleanup();

return key_doc.view().empty()
            ? stdx::nullopt
            : stdx::optional<bsoncxx::document::value>(key_doc.steal());
```
Avoid redundant copies in functions that should return `stdx::optional<bsoncxx::document::value>` in this file.

eramongodb (approved):
Please replace the `const std::string&` parameters for `add_key_alt_name`, `get_key_by_alt_name`, and `remove_key_alt_name` with `bsoncxx::string::view_or_value` and revert the return type of these functions + the `get_key` function to `stdx::optional<bsoncxx::document::value>`. Consistency with public API was specific to the string-like parameters, not the return type. `stdx::optional<bsoncxx::document::value>` was the correct return type for these functions.

In addition to some minor suggestions to avoid redundant copies, otherwise LGTM.

kevinAlbs (approved):

PR #939, opened 2023-03-08
CXX-2581 add example of extracting a string

kevinAlbs:
# Summary

- add example of extracting a string

# Background & Motivation

Motivated by request in CXX-2581.


rcsanchez97 (approved):
LGTM

PR #946, opened 2023-03-29
CXX-2393 Sync retryable handshake error unified spec tests

eramongodb:
## Description

This PR resolves CXX-2393 and CXX-2610. Verified by [this patch](https://spruce.mongodb.com/version/642496a232f417a848864293).

## Unified Spec Tests

The old retryable-reads spec test files were moved into a new `legacy` subdirectory to make room for the new `unified` spec test files. The `RETRYABLE_READS_TESTS_PATH` environment variable was renamed to `RETRYABLE_READS_LEGACY_TESTS_PATH` accordingly to distinguish it from the new `RETRYABLE_READS_UNIFIED_TESTS_PATH` environment variable.

## Spec Test Update Scripts

I've elected to remove the `update-.*-spec-tests.sh` scripts as their utility is limited by the typical necessity of being controlled and selective about which test files are being synced, including the particular version of said test files. The update scripts blindly syncs test files to their latest state, which is undesirable given how test file syncs are currently conducted.

## Unified Test Runner

There are only two major changes to the unified test runner that were necessary to support the new handshake error tests.

### Skipping Test Cases

The _optional_ `collection.findOne` and `collection.listIndexNames` helper functions are not implemented by the CXX Driver. Similar to [what is done in the C Driver](https://github.com/mongodb/mongo-c-driver/blob/e24a7a2eb6486d142dd13a0472d76ff25188ed12/src/libmongoc/tests/unified/runner.c#L37), a new `should_skip_test_cases` map was added to `runner.cpp` that specifies `(test escription, test case description)` pairs that should be skipped along with corresponding rationale.

### listCollectionNames and listDatabaseNames

These two operations _are_ implemented by the CXX Driver, and so the unified test runner was updated to invoke the corresponding operations accordingly.

## Miscellaneous Changes

- The `configureFailPoint` command was being ignored _after_ the command was already executed, which was letting some events sneak through and affecting `expectEvents` assertions. The command is now ignored _before_ it is executed.

- Updated `add_observe_events()` to avoid repeated O(n) searches through the `observeEvents` array, and also added logs capturing ignored events due to lack of support (only emitted on failure to reduce verbosity).


src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -185,6 +211,18 @@ document::value list_databases(entity::map& map,
     return result.extract();
 }
 
+document::value list_database_names(entity::map& map, const std::string& object) {
+    auto cursor = map.get_client(object).list_database_names();
```
eramongodb, line 215: It is not entirely clear if/when either the `session` or `filter` arguments should be forwarded to the relevant operations. As currently written, `list_collection_names` above conditionally forwards both `session` and `filter` as done in `list_collections`, but unlike `list_databases`, `list_database_names` does not currently have an overload that accepts a `client_session`. Should such an overload be added to `mongocxx::client` to support passing a `session` argument (in addition to a `filter` argument)?

Also note, `list_databases` currently doesn't have an overload that accepts _both_ a session and an opts argument either.

kevinAlbs: > Should such an overload be added to `mongocxx::client` to support passing a `session` argument (in addition to a `filter` argument)?

I think "yes". I expect not having an overload accepting `session` was an oversight.

`session` is not mentioned in the pspecification describing listDatabases or listDatabaseNames](https://github.com/mongodb/specifications/blob/master/source/enumerate-databases.rst#enumerating-database-names). But I see no reason to exclude `session`. All drivers except Rust [appear to accept session in listDatabaseNames](https://docs.google.com/spreadsheets/d/1JEQfCE1CnAt47tqFkDgrJ7eNenM_0bMbvgNYK-y8iME/edit#gid=0).

> Also note, `list_databases` currently doesn't have an overload that accepts _both_ a `session` and `opts` argument either.

I expect that was also an oversight.

I suggest adding these overloads as part of this PR, or filing a CXX ticket to request the additions.

kevinAlbs (approved):
I appreciate the the removal of `update-.*-spec-tests.sh`.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/64271ad7306615c0eb88e82d/tasks).

Note: the new `list_database_names` overload accepts a _filter_ as its argument for consistency with the existing `list_database_names` overload. This is _not_ the same as the `list_databases` overloads, which accept an _options_ document, within which the `"filter"` field may be specified.

Although the `list_databases` overloads accepting an _options_ document is the odd one out of the `list_*` functions, it is nevertheless also the one I think is most "correct", as it permits easily extending the APIs of the existing overloads which only accept a `filter` argument (one of several fields that may be present in the _options_ document) to support other possible option fields (such as `comment`). But that is a problem for another PR.

PR #950, opened 2023-04-05
CXX-2517 Use C Driver scripts to install libmongoc and libmongocrypt

eramongodb:
## Description

This PR unblocks CXX-2517 and other CXX Driver work that depend on libmongocrypt 1.7.0. Verified by [this patch](https://spruce.mongodb.com/version/642d80740305b96c2d9a1f10).

## Minimum C Driver Version

The minimum C Driver version was bumped to the current latest commit (https://github.com/mongodb/mongo-c-driver/commit/9ef1cc99517a6fb89cc59d34ac6c5999d044b8e1) in anticipation of upcoming CXX-2517 work. The minimum C Driver version should eventually be updated to be 1.24.0 once it is released (documented as a TODO comment).

## EVG Config

The `add_expansions_to_env` parameter is used to simplify the `install_c_driver` EVG function by removing redundant (re)definitions of EVG expansions as environment variables. The EVG function can now invoke the script directly without additional setup required.

## Building libmongocrypt and libmongoc

The `install_c_driver.sh` script was refactored to use the new C Driver [compile-libmongocrypt.sh script](https://github.com/mongodb/mongo-c-driver/blob/9ef1cc99517a6fb89cc59d34ac6c5999d044b8e1/.evergreen/scripts/compile-libmongocrypt.sh) (introduced in https://github.com/mongodb/mongo-c-driver/commit/c7e43c15bb13b69f832ac283320466f092d19137) to satisfy the libmongocrypt dependency. This allowed deferring the libbson and libmongocrypt compile-and-install routine to `compile-libmongocrypt.sh`.

This also allowed removal of specifying minimum libmongocrypt version in the CXX Driver. Instead, the CXX Driver now transitively depends on the C Driver's minimum libmongocrypt version as defined within `compile-libmongocrypt.sh`. This required slightly modifying the `install_c_driver.sh` script to generate a `VERSION_CURRENT` file that is required when building libbson via libmongocrypt to specify/determine the `BUILD_VERSION`.

This refactor also uses the C Driver's `find-cmake-latest.sh` script to obtain a newer CMake version to satisfy libmongocrypt's CMake 3.12 requirement. This also allows the CXX Driver to use newer versions of C++ and corresponding compilers (i.e. VS 2019 and newer).

### DIFF
--- a/.evergreen/install_c_driver.sh
+++ b/.evergreen/install_c_driver.sh
@@ -27,7 +27,7 @@ else
   mongoc_install_idir="${mongoc_dir}"
 fi
 : "${mongoc_idir:?}"
-: "${mongoc_install_idir}"
+: "${mongoc_install_idir:?}"
 
 echo "libmongoc version: ${mongoc_version}"
 echo "libmongocrypt version: ${mongocrypt_version}"
--- a/.evergreen/install_c_driver.sh
+++ b/.evergreen/install_c_driver.sh
@@ -7,20 +7,26 @@ set -o errexit
 set -o pipefail
 
 print_usage() {
-    echo "usage: MONGOC_VERSION=<version> MONGOCRYPT_VERSION=<version> ./install.sh"
+  echo "usage: MONGOC_VERSION=<version> MONGOCRYPT_VERSION=<version> ./install.sh"
 }
 
-if [[ -z $MONGOC_VERSION ]]; then print_usage; exit 2; fi
-if [[ -z $MONGOCRYPT_VERSION ]]; then print_usage; exit 2; fi
+if [[ -z $MONGOC_VERSION ]]; then
+  print_usage
+  exit 2
+fi
+if [[ -z $MONGOCRYPT_VERSION ]]; then
+  print_usage
+  exit 2
+fi
 
 VERSION=$MONGOC_VERSION
 PREFIX=${PREFIX:-$(pwd)"/../mongoc/"}
 OS=$(uname -s | tr '[:upper:]' '[:lower:]')
 
 if [ "$BSON_EXTRA_ALIGNMENT" = "1" ]; then
-    ENABLE_EXTRA_ALIGNMENT="ON"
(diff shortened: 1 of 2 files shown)


kkloberdanz (approved):

.evergreen/install_c_driver.sh
```diff
@@ -1,127 +1,133 @@
...
+  mongoc_idir="${mongoc_dir}"
+  mongoc_install_idir="${mongoc_dir}"
 fi
-CMAKE_ARGS="
-  -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF -DENABLE_SHM_COUNTERS=OFF
-  -DENABLE_TESTS=OFF -DENABLE_EXAMPLES=OFF -DENABLE_STATIC=ON
-  -DENABLE_EXTRA_ALIGNMENT=${ENABLE_EXTRA_ALIGNMENT} -DCMAKE_MACOSX_RPATH=ON
-  -DCMAKE_INSTALL_PREFIX=$PREFIX -DCMAKE_PREFIX_PATH=$PREFIX"
+: "${mongoc_idir:?}"
+: "${mongoc_install_idir}"
```
kevinAlbs, line 29: Assuming the intent is to error if this is unset.
```suggestion
: "${mongoc_install_idir:?}"
```

.evergreen/install_c_driver.sh
```diff
@@ -1,127 +1,133 @@
...
+
+echo "libmongoc version: ${mongoc_version}"
+echo "libmongocrypt version: ${mongocrypt_version}"
```
kevinAlbs, line 32:
```suggestion
```

`mongocrypt_version` is no longer expected.

kevinAlbs (approved):
LGTM with minor comments. The removal of the extra compile of libbson is much appreciated.

rcsanchez97 (approved):
LGTM!

Thanks for figuring out the Windows paths and a newer CMake.

PR #953, opened 2023-04-10
CXX-2656 Update tests for QEv2 compatibility

eramongodb:
## Description

Resolves CXX-2656. Verified by [this patch](https://spruce.mongodb.com/version/6434576b7742ae6f5a6efddb/tasks).

Resolves current Evergreen task failures when testing against latest replica sets due to the following error:

```
Encountered a Queryable Encryption insert/update payload type that is no longer supported
```

## Legacy CSE Tests

This PR moves legacy CSE tests into a proper `legacy` subdirectory for consistency with other unified spec tests with legacy/unified components. Updated CSE test environment variables accordingly.

## CXX-2656 Test Files

This PR syncs the legacy test files with https://github.com/mongodb/specifications/commit/64deb2837a2355f6002775c49b9b6c50c9dc560f with the exception of the following files:

- count.json: see CXX-1940.
- timeoutMS.json: see CXX-2181.
- create-and-createIndexes.json: see CXX-2513.
- create-and-createIndexes.json: see CXX-2513.
- validatorAndPartialFieldExpression.json: see CXX-2513.

## Minimum libmongoc Version

Bumped to https://github.com/mongodb/mongo-c-driver/commit/8aced03a963e70182b79d883e881f239766b09bf to include dependent changes introduced by https://github.com/mongodb/mongo-c-driver/pull/1228, including use of libmongocrypt 1.8.0-alpha0.

## Snapshot Documentation Examples

This PR also introduces a skip condition for snapshot documentation examples. See CXX-2671 for details.

## Miscellaneous

Improved quality of error messages on unexpected test operation success (when an error was expected instead) by emitting the result of the successful command.

PR #955, opened 2023-04-13
CXX-2517 Add support for Range Explicit Encryption

eramongodb:
## Description

Resolves CXX-2517. Verified by [this patch](https://spruce.mongodb.com/version/64381e257742aef6796b772e).

## Test Spec Files

Relevant legacy CSE test files are already synced as part of https://github.com/mongodb/mongo-cxx-driver/pull/953.

Files under https://github.com/mongodb/specifications/tree/c94fefe3e14d163bedcab75a761bc14a77ec689f/source/client-side-encryption/etc/data required by Range Explicit Encryption prose tests were added to the `data/client_side_encryption/explicit_encryption` directory. The `encryptedFields-Range-<Type>.json` are not required (used to generate other test files) and were thus excluded.

### ClientEncryption.encryptExpression()

A straightforward addition to the `client_encryption` class' interface as required by CXX-2517.

### RangeOpts

A new `options::range` class was added corresponding to the new [RangeOpts class](https://github.com/mongodb/specifications/blob/6018dda99f00d39056814dae0572a05bbd9b4868/source/client-side-encryption/client-side-encryption.rst#encryptopts).

### RangePreview Options

The `options::encrypt::encryption_algorithm` and `options::encrypt::encryption_query_type` were updated to support the Range algorithm.

### EncryptOpts.rangeOpts

Added `range_opts()` accessors to `options::encrypt` that depend on the new `options::range` class.

The `options::encrypt::convert()` function was updated to call relevant `client_encryption_encrypt_range_opts_*()` functions for the appropriate options in `options::range`.

Note a new potential `error_code::k_invalid_parameter` exception was introduced to validate that `min` and `max` are either both set or both unset to satisfy `mongoc_client_encryption_encrypt_range_opts_set_min_max`. This may be removed once CDRIVER-4610 is resolved.

## Range Explicit Encryption Prose Tests

A new test case was added corresponding to [Prose Test 22: Range Explicit Encryption](https://github.com/mongodb/specifications/blob/6018dda99f00d39056814dae0572a05bbd9b4868/source/client-side-encryption/tests/README.rst#range-explicit-encryption). The structure of the test is as follows (pseudocode):

```
enum struct RangeFieldType : int {
    DecimalNoPrecision,
    DecimalPrecision,
    DoubleNoPrecision,
    DoublePrecision,
    Date,
    Int,
    Long,
};

TEST_CASE("Range Explicit Encryption", ...) {
  // Run condition checks.

  for (auto field_type : RangeFieldTypes) {
    DYNAMIC_SECTION("Field Type - " << field_type) {
      // Setup

      SECTION("Case 1") { ... }
      SECTION("Case 2") { ... }
      // ...
    }
  }
}
```

The `RangeFieldType` enumeration is used to define and iterate over the supported types to be tested while ensuring fields and values have the appropriate corresponding types (e.g. via `to_type_str`, `to_field_value`, and `to_range_opts`).

The `range_explicit_encryption_setup` function is used to encapsulate common setup routines, returning `range_explicit_encryption_objects` containing the entities required by the following test cases.

The `field_type_values` struct is used to define common field values (0, 6, 30, and 200 with the appropriate type corresponding to the `RangeFieldType` being tested) used by multiple test cases as a convenience.

### Public Technical Preview Warnings

Documentation was updated to mark Public Technical Preview notes for Queryable Encryption as Doxygen warnings. Their scope (entities which the warning is being applied to) was also reduced to only the relevant entities as applicable.

### fprintf in run_encryption_tests_in_file

Some stray `fprintf` that were being unconditionally printed by `run_encryption_tests_in_file` were replaced with `UNSCOPED_INFO` so they are only emitted when necessary.

### file.is_open()

A minor change to improve context should the `REQUIRES` assertion fail.

### Member Order in options::encrypt

The `key_id` function was relocated (both in header and source) to match the pattern of the rest of the member functions.

### RAII Helpers

The options/encrypt.cpp and private/client_encryption.hh files were refactored to remove all instances of raw owning pointers in the implementation.

Notable patterns include:

#### scoped_bson_value

A helper class similar to `scoped_bson_t` that wraps `bson_value_t`. Given its limited use, it is currently defined (note: twice) in `options/encrypt.cpp` and `private/client_encryption.hh`. It may be worth eventually extracting this helper into a proper a proper component (probably under `src/mongocxx/private`?), but that work has been deferred for now.

#### Deleter and std::unique_ptr

Raw owning pointers with corresponding `cleanup` routines replaced with `std::unique_ptr` that define custom deleters. This fixes potential leaks when unexpected exceptions may be thrown by sub-operations (e.g. by calls to `make_document()` or `doc.get_<type>()`) before `cleanup` is invoked.

#### Temporary Objects

Thanks to `scoped_bson_value` (as well as `scoped_bson_t`), the following frequent pattern:

```cpp
template <typename T>
void example(const T& param) {
  bson_value_t value;
  convert_to_libbson(param, &value);
  libmongoc::some_func(&value);
  bson_value_destroy(&value);
}
```

has been refactored to:

```cpp
template <typename T>
void example(const T& param) {
  libmongoc::some_func(scoped_bson_value(param).get());
}
```


src/mongocxx/options/encrypt.hpp
```diff
@@ -100,24 +105,41 @@ class MONGOCXX_API encrypt {
         ///
         /// Use unindexed encryption.
         ///
-        k_unindexed
+        /// @warning Queryable Encryption is in Public Technical Preview. Queryable Encryption
```
kevinAlbs, line 108: Copy the warning to above `k_indexed`.

src/mongocxx/options/encrypt.cpp
```diff
@@ -79,33 +81,83 @@ const stdx::optional<options::range>& encrypt::range_opts() const {
...
+    bson_value_t value = {};
+
+    // Allow obtaining a pointer to this->value even in rvalue expressions.
+    bson_value_t* get() noexcept {
+        return &value;
+    }
+
+    // Ensure this->value can be initialized/overwritten properly.
+    bson_value_t* value_for_init() noexcept {
+        bson_value_destroy(&value);
```
kevinAlbs, line 96: As implemented, `bson_value_destroy` is not safe to call repeatedly without resetting the `value_type`. Calling repeatedly could result in a double free.

If `value_for_init` is intended to be safe to call repeatedly, suggest zeroing `value` with `memset(&value, 0, sizeof(bson_value_t));` or equivalent.

kevinAlbs (approved):
The addition of RAII types seems helpful and less error prone. TIL about `DYNAMIC_SECTION`.

LGTM with minor comments addressed.

src/mongocxx/options/encrypt.cpp
eramongodb: It is probably excessive accomodation to support reinitialization of `scoped_bson_value`. I will opt to remove code that implies reinitialization is supported.

src/mongocxx/options/encrypt.cpp
```diff
@@ -79,33 +81,83 @@ const stdx::optional<options::range>& encrypt::range_opts() const {
...
+        return &this->value;
+    }
+
+    template <typename T>
+    auto convert(const T& value)
+        // Use trailing return type syntax to SFINAE without triggering GCC -Wignored-attributes
+        // warnings due to using decltype within template parameters.
+        -> decltype(bsoncxx::types::convert_to_libbson(std::declval<const T&>(),
+                                                       std::declval<bson_value_t*>())) {
```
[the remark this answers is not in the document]
eramongodb: Yes. For example, as written, given an lvalue argument of type `bsoncxx::document::value`:
```cpp
void example() {
    auto v = bsoncxx::builder::basic::make_document();
    (void)scoped_bson_value(v); // Missing `.view()`.
}
```
The error produced by GCC is:
```
src/mongocxx/options/encrypt.cpp: In instantiation of ‘mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::scoped_bson_value(const T&) [with T = bsoncxx::v_noabi::document::value]’:
src/mongocxx/options/encrypt.cpp:135:30:   required from here
src/mongocxx/options/encrypt.cpp:112:16: error: no matching function for call to ‘mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::convert(const bsoncxx::v_noabi::document::value&)’
  112 |         convert(value);
      |         ~~~~~~~^~~~~~~
src/mongocxx/options/encrypt.cpp:102:10: note: candidate: ‘template<class T> decltype (bsoncxx::v_noabi::types::convert_to_libbson(declval<const T&>(), declval<bson_value_t*>())) mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::convert(const T&)’
  102 |     auto convert(const T& value)
      |          ^~~~~~~
src/mongocxx/options/encrypt.cpp:102:10: note:   template argument deduction/substitution failed:
src/mongocxx/options/encrypt.cpp: In substitution of ‘template<class T> decltype (bsoncxx::v_noabi::types::convert_to_libbson(declval<const T&>(), declval<bson_value_t*>())) mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::convert(const T&) [with T = bsoncxx::v_noabi::document::value]’:
src/mongocxx/options/encrypt.cpp:112:16:   required from ‘mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::scoped_bson_value(const T&) [with T = bsoncxx::v_noabi::document::value]’
src/mongocxx/options/encrypt.cpp:135:30:   required from here
```
and by Clang is:
```
src/mongocxx/options/encrypt.cpp:112:9: error: no matching member function for call to 'convert'
        convert(value);
        ^~~~~~~
src/mongocxx/options/encrypt.cpp:135:11: note: in instantiation of function template specialization 'mongocxx::options::(anonymous namespace)::scoped_bson_value::scoped_bson_value<bsoncxx::document::value>' requested here
    (void)scoped_bson_value(v);
          ^
src/mongocxx/options/encrypt.cpp:102:10: note: candidate template ignored: substitution failure [with T = bsoncxx::document::value]: no matching function for call to 'convert_to_libbson'
    auto convert(const T& value)
         ^
```

If `convert_to_libbson` is directly invoked by the constructor without the `convert()` helper function to SFINAE, the error produced by GCC looks like:
```
src/mongocxx/options/encrypt.cpp: In instantiation of ‘mongocxx::v_noabi::options::{anonymous}::scoped_bson_value::scoped_bson_value(const T&) [with T = bsoncxx::v_noabi::document::value]’:
src/mongocxx/options/encrypt.cpp:126:30:   required from here
src/mongocxx/options/encrypt.cpp:103:43: error: no matching function for call to ‘convert_to_libbson(const bsoncxx::v_noabi::document::value&, bson_value_t*)’
  103 |         bsoncxx::types::convert_to_libbson(value, &this->value);
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
In file included from src/mongocxx/options/encrypt.cpp:17:
src/mongocxx/../bsoncxx/types/private/convert.hh:43:21: note: candidate: ‘void bsoncxx::v_noabi::types::convert_to_libbson(const bsoncxx::v_noabi::types::b_binary&, bson_value_t*)’
   43 | BSONCXX_INLINE void convert_to_libbson(const bsoncxx::types::b_binary& binary, bson_value_t* v) {
      |                     ^~~~~~~~~~~~~~~~~~
<repeats 'note: candidate:' message every overload of `convert_to_libbson`>
```
and by Clang:
```
src/mongocxx/options/encrypt.cpp:103:9: error: no matching function for call to 'convert_to_libbson'
        bsoncxx::types::convert_to_libbson(value, &this->value);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/mongocxx/options/encrypt.cpp:126:11: note: in instantiation of function template specialization 'mongocxx::options::(anonymous namespace)::scoped_bson_value::scoped_bson_value<bsoncxx::document::value>' requested here
    (void)scoped_bson_value(v);
          ^
src/mongocxx/../bsoncxx/types/private/convert.hh:43:21: note: candidate function not viable: no known conversion from 'const bsoncxx::document::value' to 'const bsoncxx::types::b_binary' for 1st argument
BSONCXX_INLINE void convert_to_libbson(const bsoncxx::types::b_binary& binary, bson_value_t* v) {
                    ^
<repeats 'note: candidate:' message every overload of `convert_to_libbson`>
```

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/643d7268850e6141fdb52ee8).

Realized I missed the addition of [a clause specifying](https://github.com/mongodb/specifications/commit/c94fefe3e14d163bedcab75a761bc14a77ec689f#diff-c57f446f49351e6f7432a2d3028c880a728f1a33031510a7c376aed157d6de00R2634) `DecimalNoPrecision` tests should only be run against replica sets. Added the skip condition accordingly.

kkloberdanz (approved):

PR #958, opened 2023-04-24
CXX-2683 Replace codecov-bash with new Codecov uploader

eramongodb:
## Description

Resolves CXX-2683 and includes several miscellaneous improvements to the CXX Driver EVG config. Verified by [this patch](https://spruce.mongodb.com/version/6446bda557e85a3d68e40063).

The primary motivation for this PR is to reduce output verbosity in CXX Driver EVG patches. Recommend comparing outputs in the linked patch against those in the base commit.

The combination of applied output filters, verbosity reduction, and output-on-failure behavior described below reduces the total lines of output (relative to the base commit) for the following tasks by:

- 11721 -> 4970 (-57.6%): compile_and_test_with_shared_libs in Ubuntu 18.04 Debug (MongoDB Latest)
- 7831 -> 4913 (-37.3%): compile_and_test_with_static_libs in Ubuntu 18.04 Release (MongoDB Latest)
- 7721 -> 1769 (-77.1%): compile_and_test_with_static_libs in Windows (VS 2015) Release (MongoDB 4.2)
- 7527 -> 1842 (-75.5%): compile_and_test_with_shared_libs in Windows (VS 2017) Debug (MongoDB Latest)
- 2189 -> 1136 (-48.1%): build_example_with_add_subdirectory in Ubuntu 20.04 Release (MongoDB Latest)

Note: setup failures on Ubuntu 14.04 do not appear to be related to changes in this PR.

## codecov-bash

The [Codecov Bash Uploader](https://github.com/codecov/codecov-bash) is deprecated and no longer maintained as of February 1, 2022. A new [Codecov Uploader](https://github.com/codecov/uploader) is recommended in its place. This PR updates the code coverage EVG function to use the new uploader.

Note, the new uploader requires downloading the appropriate binary from https://uploader.codecov.io/. The available binaries are limited to x86_64 MacOS, Linux, and Windows binaries + AArch64 (Linux?) binaries, but since we are currently only enabling code coverage on ubuntu1804 distros, the script is left minimal to avoid unnecessary complexity (+ I do not expect we will be expanding code coverage to additional distros anytime soon).

A filter (via `perl`) is applied to the output of the Uploader to avoid the spammy list of "(Found|Processing) <gcov file>" messages generated by the Uploader that often occupied the ending of EVG tasks output, obscuring the failures or errors of actual interest.

## Ensure Replica Set Member on Port 27017 is Primary

A long-time cause of flaky test failures in the CXX Driver waterfall has been "not primary" errors due to the test suite assuming the replica set member on port 27017 is primary by directly connecting to it via the default URI (which does not list the other replset members). A routine was added to the `start_mongod` EVG function that triggers a replset reconfiguration + waits for the member on port 27017 to become primary to address this issue. Because mongo-orchestration does not appear to support directly designating a replset member as primary, `mongosh` is used instead to send the appropriate commands.

## Miscellaneous Output Reduction

Various modifications were applied to reduce verbosity of task output, much of which is not of interest or relevance during regular development workflows:

- Removed printing contents of the mongoc directory in the `compile` EVG function.
- Removed the `VERBOSE` option from invocations of GNU Make.
- Added the `/verbosity:minimal` option to invocations of MSBuild.
- Silenced standard output when building and installing libmongocrypt and libmongoc.
- Silenced standard output when registering CA certificates.
- Only print output of building and running example projects and executables on failure.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -615,7 +615,7 @@ functions:
                       # is true.
                       if [[ -z "$MONGODB_API_VERSION" ]]; then
                           for test in $EXAMPLES; do
-                            echo "Running example $test"
+                            echo "Running $test"
                             case "$test" in
                               *encryption*)
                                 echo " - Skipping client side encryption example"


examples/projects/bsoncxx/cmake-deprecated/shared/build.sh
```diff
@@ -10,13 +10,12 @@ rm -rf build/*
 cd build
 if [ -z "$MSVC" ]; then
     "$CMAKE" -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" -DCMAKE_CXX_STANDARD="${CXX_STANDARD}" ..
-    make run VERBOSE=1
```
kevinAlbs, line 13: Replace with `make run` to run example. Same comment applies to other files under `examples/projects`.

.mci.yml
```diff
@@ -191,6 +191,48 @@ functions:
                 echo "{ \"releases\": { \"default\": \"$MONGODB_BINARIES\" }}" > $MONGO_ORCHESTRATION_HOME/orchestration.config
                 ./.evergreen/run-orchestration.sh
 
+                # Ensure server on port 27017 is the primary server.
```
kevinAlbs, line 194: > by directly connecting to it via the default URI (which does not list the other replset members)

Another alternative to consider: the C driver's `test_framework_get_uri_str_no_auth` constructs a URI with multiple hosts after checking the hello response. Using multiple hosts in the URI may be more robust if tests are expected to step down the primary. AFAICT the C++ driver has no such tests, but there are [specification tests that use replSetStepDown](https://github.com/mongodb/specifications/blob/c4f3fabe9047e5e7a87877d642fa0d885a47351c/source/server-discovery-and-monitoring/tests/unified/rediscover-quickly-after-step-down.yml#L88).

Though setting the priority still seems like an improvement, and I expect should resolve the "not primary" errors.

kevinAlbs (approved):
The reduced verbosity and fix to "not primary" errors are very much appreciated. LGTM with the re-addition of `make run`.

.mci.yml
eramongodb: I considered adding the replset hosts to the URIs used by the test framework, but decided against it as many tests use the default-constructed `mongocxx::uri` object to obtain the URI, and updating the test suite to use a test-framework URI constructor like the C Driver seemed too extensive a change for this PR.

PR #960, opened 2023-04-28
CXX-2637 Remove use of `collStats` and `currentOp`

kevinAlbs:
# Summary

- Remove use of `collStats` and `currentOp`

# Background & Motivation

See DRIVERS-2232:

> collStats and currentOp will produce a deprecation message in 6.0 and are candidates for removal in another major release.

The `runCommand Example 2` example, which used `collStats`, was removed. This example was added for the documentation team in DRIVERS-448. DOCSP-21320 requests removing `runCommand Example 2` from documentation.

### DIFF
--- a/data/client_side_encryption/legacy/bypassedCommand.json
+++ b/data/client_side_encryption/legacy/bypassedCommand.json
@@ -78,7 +78,7 @@
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


kkloberdanz (approved):
LGTM

PR #970, opened 2023-05-25
CXX-2694 use setuptools rather than distutils (deprecated) for version calc

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/646e3e3e9ccd4e1799389cce/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/etc/calc_release_version.py
+++ b/etc/calc_release_version.py
@@ -185,7 +185,7 @@ def get_branch_tags(active_branch_name):
 def process_and_sort_tags(tags):
     """
     Given a string (as returned from get_branch_tags), return a sorted list of
-    zero or more tags (sorted based on the LooseVersion comparison) which meet
+    zero or more tags (sorted based on the Version comparison) which meet
     the following criteria:
         - a final release tag (i.e., r3.x.y without any pre-release suffix)
         - a pre-release tag which is not superseded by a release tag (i.e.,
--- a/etc/calc_release_version.py
+++ b/etc/calc_release_version.py
@@ -196,16 +196,19 @@ def process_and_sort_tags(tags):
     for tag in raw_tags:
         release_tag_match = RELEASE_TAG_RE.match(tag)
         if release_tag_match and not release_tag_match.group('verpre'):
-            processed_and_sorted_tags.append(tag)
+            # strip leading "r" for version comparison
+            processed_and_sorted_tags.append(tag[1:])
     # collect together final release tags and pre-release tags for
     # versions that have not yet had a final release
     for tag in raw_tags:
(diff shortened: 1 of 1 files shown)


etc/calc_release_version.py
```diff
@@ -190,7 +203,7 @@ def process_and_sort_tags(tags):
         tag_parts = tag.split('-')
         if len(tag_parts) >= 2 and tag_parts[0] not in processed_and_sorted_tags:
             processed_and_sorted_tags.append(tag)
-    processed_and_sorted_tags.sort(key=LooseVersion)
+    processed_and_sorted_tags.sort(key=Version)
```
kevinAlbs, line 206: Tasks appear to fail on this line:

https://spruce.mongodb.com/task/cxx_driver_debian10_release_latest_compile_and_test_with_static_libs_extra_alignment_patch_ed5a02fb0bf6135e25c45723ba5d768d59cb785e_646f9256d6d80a32b136916f_23_05_25_16_52_39/logs?execution=0

rcsanchez97: The failure seems to be a result of Evergreen executing the build while checked out to branch other than `master` (namely, the branch from the PR or `cxx-2694`). I'm working out a fix.

rcsanchez97:
The latest commit seems to have resolved the issues. The tasks in the patch build appear to be succeeding now.

etc/calc_release_version.py
```diff
@@ -196,16 +196,19 @@ def process_and_sort_tags(tags):
     for tag in raw_tags:
```
kevinAlbs, line 196: Update comment above to replace `LooseVersion comparison` with `Version comparison`

kevinAlbs (commented):
Latest changes look good. I suggest waiting to merge until C driver tasks failures ([example](https://evergreen.mongodb.com/task/mongo_c_driver_windows_2017_debug_compile_nosasl_nossl_patch_6cbb08bd7cdd1dd7e7d30d374212a002955d0789_64776889d1fe07f4502985de_23_05_31_15_32_26)) related to https://github.com/mongodb/mongo-c-driver/commit/d9569499eaccd304f0f96766717729123c6b3428 are resolved (noted in slack here). That may suggest an additional change is needed.

etc/calc_release_version.py
rcsanchez97: I pushed a change based on mongodb/mongo-c-driver@c35519cc5fb0

rcsanchez97: And another commit to fix the comment.

kevinAlbs (approved):

PR #972, opened 2023-06-02
Add GCC diagnostic guards

darktohka:
Building mongo-cxx-driver with the GitHub Actions ubuntu-22.04 image yields the following error during compilation:

```
error: unknown warning group '-Wmaybe-uninitialized', ignored
  [-Werror,-Wunknown-pragmas]
                    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```

To alleviate this issue, I've added __has_warning guards to all GCC diagnostic ignore pragmas. This resolves the compilation issue.

### DIFF
--- a/src/bsoncxx/document/value.hpp
+++ b/src/bsoncxx/document/value.hpp
@@ -234,7 +234,9 @@ class BSONCXX_API value {
 #if defined(__GNUC__) && (__cplusplus >= 201709L)
 // Silence false positive with g++ 10.2.1 on Debian 11.
 #pragma GCC diagnostic push
+#if !defined(__has_warning) || __has_warning("-Wmaybe-uninitialized")
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
 #endif
 BSONCXX_INLINE document::view value::view() const noexcept {
     return document::view{static_cast<uint8_t*>(_data.get()), _length};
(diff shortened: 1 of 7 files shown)


rcsanchez97:
@darktohka your proposed changes break compilation on basically every configuration in our CI. After doing a bit of research and looking at the code you proposed changing, it seems like the issue you reported is actually that in most instances where pragmas are being manipulated there is one code path for clang and a separate code path for gcc. However, in some recently added pragmas, there is only a single code path, for gcc, and since clang emulates gcc at some level, it would appear that those gcc-specific pragma manipulations are triggering an error in clang builds. This is all based on my research into the error message that you cited and some guesswork on my part, because you gave no information about what compiler toolchain you are using in the instance where the build fails or how your executing the build.

All of that said, I think that a less drastic change is to protect the recently added (and apparently problematic) pragma manipulations with an added `!defined(__clang__)` condition. In place of the changes in this PR, I have proposed the changes in the attached patch. I was not able to confirm that they resolve your specific issue as the lack of information prevented me from being able to reproduce the failure in order to confirm that my changes fix it. However, I have tested the changes I am proposing and they build cleanly in our CI. If you could test the changes I am proposing and tell me if they resolve your issue, I can move forward with them.

[pragma.patch.txt](https://github.com/mongodb/mongo-cxx-driver/files/11668104/pragma.patch.txt)

rcsanchez97:
@bazineta Thanks for the confirmation. @darktohka Do you have any thoughts on the alternative I proposed?

kevinAlbs:
@darktohka  is this PR still applicable? Does the alternative proposed by @rcsanchez97 seem OK?

rcsanchez97:
Closing, since we decided to go with #1014 instead.

PR #975, opened 2023-06-12
CXX-2690 Deprecate sharded-replicaset topology type

joshbsiegel:
- Searched through the changes made in https://jira.mongodb.org/browse/DRIVERS-2609 and updated the CXX tests where necessary without adding any tests that didn't exist before
- Modified two instances where we previously had to decipher between `sharded` or `sharded-replicaset` type topology

### DIFF
--- a/src/mongocxx/test/spec/unified_tests/runner.cpp
+++ b/src/mongocxx/test/spec/unified_tests/runner.cpp
@@ -220,8 +220,7 @@ bool equals_server_topology(const document::element& topologies) {
     // test file.
     const static std::string actual = test_util::get_topology();
     const auto equals = [&](const bsoncxx::array::element& expected) {
-        return expected == value(actual) ||
-               (expected == value("sharded") && actual == "sharded-replicaset");
+        return expected == value(actual);
     };
 
     const auto t = topologies.get_array().value;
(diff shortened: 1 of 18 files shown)


src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -220,8 +220,7 @@ bool equals_server_topology(const document::element& topologies) {
     // test file.
     const static std::string actual = test_util::get_topology();
     const auto equals = [&](const bsoncxx::array::element& expected) {
-        return expected == value(actual) ||
-               (expected == value("sharded") && actual == "sharded-replicaset");
+        return expected == value(actual);
```
kevinAlbs, line 223: I suggest reverting the changes in the .cpp files for handling `sharded-replicaset`. The `sharded-replicaset` value is [deprecated but not removed](https://github.com/mongodb/specifications/blob/840e6d49c354656bff11b2622f0d3001b39d9403/source/unified-test-format/unified-test-format.rst) from the Unified Test Format:

> The "sharded-replicaset" topology type is deprecated. MongoDB 3.6+ requires that all shard servers be replica sets (see: [release notes](https://www.mongodb.com/docs/manual/release-notes/3.6-upgrade-sharded-cluster/#shard-replica-sets)). Therefore, tests SHOULD use "sharded" instead of "sharded-replicaset" when targeting 3.6+ server versions in order to avoid unnecessary overhead.

The [runCommand.yml test](https://github.com/mongodb/specifications/blob/840e6d49c354656bff11b2622f0d3001b39d9403/source/run-command/tests/unified/runCommand.yml#L122) still references sharded-replicaset.

PR #976, opened 2023-06-12
CXX-1712 Overload to_json to correctly handle bson arrays

joshbsiegel:
- Overloaded to_json so that it handles `bsoncxx::array::view` and outputs the correct string
- Added test cases to adequately test the main four scenarios of an array being passed through to the function (legacy implicit, legacy explicit, relaxed, canonical)
- Tested locally in conjunction with https://github.com/mongodb/mongo-c-driver/pull/1305


kevinAlbs (commented):
I suggest waiting until the C driver changes are merged before proceeding with this PR.

Once the C driver changes are merged, the C++ driver test and release scripts can be updated to use the unreleased commit of the C driver. I expect that requires changes in [.mci.yml here](https://github.com/mongodb/mongo-cxx-driver/blob/4d6685bf9c4e4acb0b853d0f0edcfcc77b09b0f3/.mci.yml#L10-L15) and [make_release.py here](https://github.com/mongodb/mongo-cxx-driver/blob/4d6685bf9c4e4acb0b853d0f0edcfcc77b09b0f3/etc/make_release.py#L86-L89)

src/mongocxx/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ message ("mongocxx version: ${MONGOCXX_VERSION}")
 set(MONGOCXX_INLINE_NAMESPACE "v${MONGOCXX_ABI_VERSION}")
 set(MONGOCXX_HEADER_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/mongocxx/${MONGOCXX_INLINE_NAMESPACE}" CACHE INTERNAL "")
 
-set(LIBMONGOC_REQUIRED_VERSION 1.22.1) # TODO: update to 1.24.0 once released.
+set(LIBMONGOC_REQUIRED_VERSION 1.24.0) # TODO: update to 1.24.0 once released.
```
kevinAlbs, line 32: Aside: I expect this is not necessary for this PR. `LIBMONGOC_REQUIRED_VERSION` controls the required version of libmongoc in cmake. It does not control the version of libmongoc obtained in testing. But this seems OK to include in this PR.

kevinAlbs (approved):

eramongodb (approved):
Minor suggestions/comments; otherwise, LGTM.

PR #979, opened 2023-06-15
add self-test for calc_release_version.py

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/648b7bf29ccd4ed2ea437daf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Note that this implementation is such that it won't run on a patch build. As a result, the patch build above includes these additional changes to simulate the effect of what this task will accomplish when executed on the waterfall:

```
diff --git a/.evergreen/calc_release_version_selftest.sh b/.evergreen/calc_release_version_selftest.sh
index 3ebdc851b..d4fc1c908 100644
--- a/.evergreen/calc_release_version_selftest.sh
+++ b/.evergreen/calc_release_version_selftest.sh
@@ -6,6 +6,7 @@
 
 set -o errexit
 set -o pipefail
+git stash
 
 function assert_eq () {
     a="$1"
diff --git a/.mci.yml b/.mci.yml
index 01efe6bf1..5e950c0ca 100644
--- a/.mci.yml
+++ b/.mci.yml
@@ -973,6 +973,8 @@ tasks:
               cd etc
               bash -x ../.evergreen/calc_release_version_selftest.sh
             fi
+            cd etc
+            bash -x ../.evergreen/calc_release_version_selftest.sh
 
     - name: uninstall_check
       commands:
```


.evergreen/calc_release_version_selftest.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# calc_release_version_selftest.sh is used to test output of calc_release_version.py.
+# run with:
+# cd etc
```
kevinAlbs, line 4: Along with the comment suggesting not to run the self-test in CI, suggest moving `calc_release_version_selftest.sh` to the `etc` directory.

.mci.yml
```diff
@@ -960,6 +960,20 @@ tasks:
         - func: "run_kms_servers"
         - func: "test"
 
+    - name: calc_release_version_selftest
```
kevinAlbs, line 963: The additional step in the release process and additional task in CI may add some maintainence burden.

IMO the self test is helpful to verify changes in `calc_release_version.py`. I expect the self tests may not be helpful for the vast majority of all other commits.

I suggest:
- Remove the task from Evergreen and removing the updates to the release process.
- Add a comment in `calc_release_version.py` to suggest running the self test when making an update. Suggestion:
```python
# Updates to calc_release_version.py can be manually tested by running the calc_release_version_selftest.sh script
```

kevinAlbs (commented):
Thank you for the PR. Suggested removing the script from CI and the release process.

.mci.yml
rcsanchez97: Good point about the maintenance burden. That thought did cross my mind, but then I wasn't entirely convinced one way or the other. I can definitely agree with removing it from CI and documenting a manual test procedure inside of `calc_release_version.py`. I'll work on getting that updated soon.

rcsanchez97: Done

rcsanchez97:
@kevinAlbs Assuming that you like the modifications I made, after merging this PR, I will update the parallel PR in libmongocrypt and then I will go ahead with implementing in the C driver as well.

.evergreen/calc_release_version_selftest.sh
kevinAlbs, line 1: Suggest running `chmod u+x` on this file to add execute permissions to this file.

On checkout, running gives "permission denied" on my machine:
```sh
% ../.evergreen/calc_release_version_selftest.sh
zsh: permission denied: ../.evergreen/calc_release_version_selftest.sh
```

etc/calc_release_version.py
```diff
@@ -22,6 +22,18 @@
 current Git commit.
 """
 
+# XXX NOTE XXX NOTE XXX NOTE XXX
+# After modifying this script it is advisable to execute the self-test.
+#
+# This is done by starting in the directory containing this script and then
+# executing a separate self-test script, like this:
+#
+# $ bash ../.evergreen/calc_release_version_selftest.sh
```
kevinAlbs, line 31: Consider moving `calc_release_version_selftest.sh` to the `etc` directory so it is located next to `calc_release_version.py`. Since `calc_release_version_selftest.sh` is not run in Evergreen, I see no reason to keep it in the `.evergreen` directory.

kevinAlbs (approved):
LGTM with two comments to consider.

> @kevinAlbs Assuming that you like the modifications I made, after merging this PR, I will update the parallel PR in libmongocrypt and then I will go ahead with implementing in the C driver as well.

SGTM

PR #984, opened 2023-06-27
CXX-2120 Handle nonexistent document fields

joshbsiegel:
# Summary
This pull request will include the name of the nonexistent key in the exception message thrown when the value is attempted to get fetched.

# Background
Prior to this PR, when the user would attempt to fetch the value of a nonexistent key, such as
```
auto some_message = doc["sceret"].get_string().value;
```
the error message would be as follows:
```
cannot get string from an uninitialized element: unset document::element"
```
This is unhelpful and quite frustrating for the user as it doesn't provide any information as to what key was uninitialized. If the user has a bunch of `get_string()` or `get_value()` calls across their process, it would be tedious to have to go through each one to figure out where exactly the program failed.

# Getting the Key
It's important to remember that the exception above is not thrown when a nonexistent field is accessed with `doc["sceret"]`, but rather when the value of that nonexistent field is attempted to be accessed (`get_#name()` or `get_value()`). This is because we still want the user to be able to conveniently check for existence, such as 
```
if (doc["some_field"]) {
    // some_field exists
}
```
So, getting the nonexistent key from the `operator[]` function is not an option.

The other half of this problem is that when a nonexistent key is passed through to the `operator[]` function, a "null" value is *not* attached to that key or added to the BSON data structure. It would just go away. Because of this, there was no existing way to get the faulty key though the `get_#name()` or `get_value()` functions as it wasn't stored anywhere, it's a completely separate function call than the `operator[]` call.

The solution to this problem was adding an `stdx::optional<stdx::string_view> _key` member variable to the `element` class. Where the `operator[]` call would attempt to find the nonexistent key before and return an iterator to a completely empty element, we now return an iterator to an element with only the optional `_key` field set. Now when the class attempts to get the value of a field and sees that it is nullptr (`_raw == nullptr`) we can access `_key` to include the key in the exception thrown.

# What's New
- Added `stdx::optional<stdx::string_view> _key` to the `element` class
- Overloaded `element` constructor to take in a `key` and set all other `element` fields to `nullptr` or 0
- Changed exception messages to include `_key`
- Test case that calls both `get_string()` and `get_value()`

### DIFF
--- a/src/bsoncxx/array/view.cpp
+++ b/src/bsoncxx/array/view.cpp
@@ -126,7 +126,7 @@ view::const_iterator view::find(std::uint32_t i) const {
     }
 
     if (!bson_iter_init_find(&iter, &b, key.c_str())) {
-        return cend();
+        return const_iterator(element(key.c_str()));
     }
 
     return const_iterator(element(data(),
--- a/src/bsoncxx/array/view.cpp
+++ b/src/bsoncxx/array/view.cpp
@@ -118,11 +118,11 @@ view::const_iterator view::find(std::uint32_t i) const {
     bson_iter_t iter;
 
     if (!bson_init_static(&b, data(), length())) {
-        return cend();
+        return const_iterator(element(key.c_str()));
     }
 
     if (!bson_iter_init(&iter, &b)) {
-        return cend();
+        return const_iterator(element(key.c_str()));
     }
 
     if (!bson_iter_init_find(&iter, &b, key.c_str())) {
--- a/src/bsoncxx/document/element.cpp
+++ b/src/bsoncxx/document/element.cpp
@@ -42,7 +42,7 @@ element::element(const std::uint8_t* raw,
     : _raw(raw), _length(length), _offset(offset), _keylen(keylen) {}
 
 element::element(const stdx::string_view key)
-    : _key(key), _raw(nullptr), _length(0), _offset(0), _keylen(0) {}
+    : _raw(nullptr), _length(0), _offset(0), _keylen(0), _key(key) {}
 
 const std::uint8_t* element::raw() const {
     return _raw;
--- a/src/bsoncxx/document/element.cpp
+++ b/src/bsoncxx/document/element.cpp
@@ -63,7 +63,7 @@ bsoncxx::type element::type() const {
     if (_raw == nullptr) {
         throw bsoncxx::exception{error_code::k_unset_element,
                                  "cannot return the type of uninitialized element with key \"" +
-                                     std::string(_key.value().data()) + "\""};
+                                     std::string(_key.value_or("(none)").data()) + "\""};
     }
 
     BSONCXX_CITER;
@@ -74,7 +74,7 @@ stdx::string_view element::key() const {
     if (_raw == nullptr) {
         throw bsoncxx::exception{error_code::k_unset_element,
                                  "cannot return the key from an uninitialized element with key \"" +
-                                     std::string(_key.value().data()) + "\""};
+                                     std::string(_key.value_or("(none)").data()) + "\""};
     }
 
     BSONCXX_CITER;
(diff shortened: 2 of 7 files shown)


src/bsoncxx/document/element.cpp
```diff
@@ -69,7 +73,8 @@ bsoncxx::type element::type() const {
 stdx::string_view element::key() const {
     if (_raw == nullptr) {
         throw bsoncxx::exception{error_code::k_unset_element,
-                                 "cannot return the key from an uninitialized element"};
+                                 "cannot return the key from an uninitialized element with key \"" +
```
joshbsiegel, line 76: The phrasing here is kind of ridiculous but I think that changing the behavior of `key()` to no longer thrown an exception when a nonexistent value is fetched would be problematic. Maybe we can just revert this particular message to what it was before the PR.

src/bsoncxx/document/element.cpp
```diff
@@ -59,7 +62,8 @@ std::uint32_t element::keylen() const {
 bsoncxx::type element::type() const {
     if (_raw == nullptr) {
         throw bsoncxx::exception{error_code::k_unset_element,
-                                 "cannot return the type of uninitialized element"};
+                                 "cannot return the type of uninitialized element with key \"" +
```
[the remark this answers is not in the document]
kevinAlbs: Suggest conditioning the message on whether `_key` is set:

```c++
        if (!_key) {
            throw bsoncxx::exception{error_code::k_unset_element,
                                     "cannot return the type of uninitialized element"};
        }
        throw bsoncxx::exception{error_code::k_unset_element,
                                 "cannot return the type of uninitialized element with key \"" +
                                     std::string(*_key) + "\""};
```

Suggest applying a similar change to other exceptions.

src/bsoncxx/document/element.cpp
kevinAlbs: > to no longer throw an exception when a nonexistent value is fetched would be problematic

I am not sure if there is a practical use-case to rely on `key()` throwing an exception. I think there is a risk changing `key()` to no longer throw an exception could break user assumptions. Example:

```c++
// insertAuthenticated inserts `doc` if `doc` contains the "authenticated" field.
bool insertAuthenticated(mongocxx::collection coll, bsoncxx::document::view doc) {
    try {
        // Check for the presence of the "authenticated" key before inserting.
        doc["authenticated"].key();
        coll.insert_one(doc);
    } catch (const bsoncxx::exception& e) {
        return false;
    }
}
```

src/bsoncxx/document/element.cpp
```diff
@@ -58,8 +61,10 @@ std::uint32_t element::keylen() const {
 
 bsoncxx::type element::type() const {
     if (_raw == nullptr) {
-        throw bsoncxx::exception{error_code::k_unset_element,
-                                 "cannot return the type of uninitialized element"};
+        throw bsoncxx::exception{
+            error_code::k_unset_element,
+            "cannot return the type of uninitialized element" +
+                std::string(_key ? " with key \"" + std::string(_key.value().data()) + "\"" : "")};
```
joshbsiegel, line 67: I had to wrap the second part of the string in `std::string` in order to get the ternary operator to work.

src/bsoncxx/array/view.cpp
```diff
@@ -126,7 +126,7 @@ view::const_iterator view::find(std::uint32_t i) const {
     }
 
     if (!bson_iter_init_find(&iter, &b, key.c_str())) {
```
kevinAlbs, line 128: Suggest also updating the `return cend();` calls above in this function with `const_iterator(element(key.c_str()))` to include the key in the exception.

kevinAlbs (approved):
Nicely done. LGTM with one comment.

kkloberdanz (approved):
Looks good to me! This issue caused me some headaches in the past, so it will be awesome to have this in place!

PR #986, opened 2023-07-11
CXX-2697 Add search index spec tests and management helpers

joshbsiegel:
# Summary
This PR will add the search index spec tests and respective functionality needed to pass the spec tests.

# Background
Due to the spec test additions made in https://github.com/mongodb/specifications/pull/1423, it was necessary to sync the spec test changes made and then ensure that the driver can pass those tests. To accomplish this, we chose to add an index view API and referenced the pseudocode function definitions found in the specifications PR.

# Design
The `search_index_view` structure is very similar to the existing `index_view` structure. All of the public API functions call their private counterpart which constructs a BSON command and sends it to the server. Both `search_index_view` and `search_index_model` are designed with the PIMPL pattern.

# Testing
At the moment, there are only spec tests for search indexes from this [commit](https://github.com/mongodb/specifications/commit/ae35908e5eadc6cda2d8f5c4e33dada25ba8fa98). There is no capability for e2e testing yet but when there is a separate PR will be made with additions to `test/search_index_view.cpp`.

# What's New
- 5 new spec tests
- Add functionality to return a `search_index_view` from a collection
- `search_index_model.hpp` and `search_index_model.cpp`
- `search_index_view.hpp` and `search_index_view.cpp`
- `private/search_index_view.hh`
- `private/search_index_model.hh`
- 5 new operation handlers in `operations.cpp`
- Updating CMake configuration to recognize the new files
- `private/append_aggregate_options`: a helper function for `options::aggregate::append` and `search_index_view::impl::list`


src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,67 @@
...
+    ///
+    bsoncxx::stdx::optional<std::string> get_name() const;
+
+    ///
+    /// Retrieves definition of a search_index_model.
+    ///
+    bsoncxx::document::view get_definition() const;
+
+   private:
+    bsoncxx::stdx::optional<std::string> _name;
```
joshbsiegel, line 60: Should I make this `stdx::string_view` instead? From what I understand, `string_view` is better for performance but I'm not familiar enough with it to determine whether it's necessary here.

src/mongocxx/search_index_model.hpp
```diff
@@ -32,7 +32,7 @@ class MONGOCXX_API search_index_model {
     ///
     /// Move assigns a search_index_model.
     ///
-    search_index_model& operator=(search_index_model&&) noexcept;
```
joshbsiegel: Marking a function as `noexcept` requires that all members of the class must also have `noexcept` on the corresponding function. While the `_definition` field is a `bsoncxx::document::value` which has its move assign operator as `noexcept`, the ` bsoncxx::stdx::optional<std::string> _name` field does not have the same tag on its move assign operator. Instead of making changes to `stdx::optional` which could affect the entire codebase, it's easier to just remove the `noexcept` from the move assign operator.

src/mongocxx/options/search_index_view.hpp
joshbsiegel, line 1: This can be populated in the future if there are additional features that need to be added to `search_index_view`. Having the additional features in the `options::search_index_view` will ensure that they do not cause a break in the API.

src/mongocxx/test/search_index_view.cpp
joshbsiegel, line 1: This file is for future testing that cannot be done at the moment.

src/mongocxx/private/search_index_view.hh
```diff
@@ -0,0 +1,166 @@
...
+                                         const options::search_index_view& options) {
+        BSON_UNUSED(options);
+
+        using namespace bsoncxx;
+
+        builder::basic::array search_index_arr;
+
+        for (auto&& model : search_indexes) {
+            builder::basic::document search_index_doc;
+            // model may or may not have a name attached to it (even though the server will)
```
kevinAlbs, line 84:
```suggestion
            // model may or may not have a name attached to it. The server will create the name if it is not set.
```

src/mongocxx/search_index_view.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    ///
+    /// @{
+    ///
+    /// Returns a cursor over all the search indexes.
+    ///
+    cursor list(const bsoncxx::document::view& aggregation_opts = bsoncxx::document::view{});
+
+    cursor list(const std::string name,
+                const bsoncxx::document::view& aggregation_opts = bsoncxx::document::view{});
```
kevinAlbs, line 38:
```suggestion
                const bsoncxx::document::view& aggregation_opts = bsoncxx::document::view{});
                
    ///
    /// @}
    ///
```

Comment to close the doxygen group.

Doxygen docs can be built and viewed locally:
```
mkdir -p ./build/docs/api/current  
cmake --build cmake-build --target doxygen-current
open ./build/docs/api/current/index.html  
```

src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    ///
+    /// Destroys a search_index_model.
+    ///
+    ~search_index_model();
+
+    ///
+    /// Retrieves name of a search_index_model.
+    ///
+    bsoncxx::stdx::optional<std::string> get_name() const;
```
kevinAlbs, line 52:
```suggestion
    bsoncxx::stdx::optional<std::string> name() const;
```

Suggest removing `get_` for consistency with other class getter naming (e.g. `index_model::keys`).

src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    ///
+    /// Retrieves definition of a search_index_model.
+    ///
+    bsoncxx::document::view get_definition() const;
```
kevinAlbs, line 57:
```suggestion
    bsoncxx::document::view definition() const;
```
Suggest removing `get_` for consistency with other class getter naming (e.g. `index_model::keys`).

src/mongocxx/search_index_view.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
```
kevinAlbs, line 40: Suggest use `///` comments for consistency with most other class comments.

src/mongocxx/private/append_aggregate_options.hh
joshbsiegel, line 1: This file contains the code that was previously in `options::aggregate::append`. In order to reduce code duplication, the best practice would be to use `append` to extract the options from the `options::aggregate` variables passed through to `list`. However, because `append` is private, `search_index_view` would need to become a `friend` of `options::aggregate`. Because the code in `append` only accesses public variables and functions in `options::aggregate` and it seemed unneccessary to give `search_index_view` access to all of `options::aggregate`, we opted to move the function to a private file.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1794,6 +1795,96 @@ document::value create_find_cursor(entity::map& map,
...
+
+document::value list_search_indexes(collection& coll, const document::view& operation) {
+    auto arguments = operation["arguments"];
+    bsoncxx::document::view aggregation_options;
+    options::aggregate options;
+    if (arguments["aggregationOptions"]) {
+        aggregation_options = arguments["aggregationOptions"].get_document().view();
+        // search index spec tests only set batchSize, so if aggregation options exist then it must
+        // be batch size.
+        options.batch_size(aggregation_options["batchSize"].get_int32().value);
```
kevinAlbs, line 1856:
```suggestion
        for (auto&& element : aggregation_options) {
            if (element.key() == "batchSize") {
                options.batch_size(element.get_int32().value);
            } else {
                throw std::logic_error{"unsupported aggregateOptions field: " +
                                       std::string(element.key())};
            }
        }
```

Suggest iterating over all elements and throwing an exception on unrecognized options. This may help inform when tests are introduced with new unrecognized options.

kevinAlbs (approved):
Nice work. LGTM with one comment.

src/mongocxx/search_index_view.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+void search_index_view::update_one(const client_session& session,
+                                   bsoncxx::string::view_or_value name,
+                                   const bsoncxx::document::view_or_value& definition) {
+    _get_impl().update_one(&session, name, definition);
+}
+
+search_index_view::impl& search_index_view::_get_impl() {
+    return *_impl;
```
eramongodb, line 102: Recommend following the [`_get_impl()` pattern](https://github.com/mongodb/mongo-cxx-driver/blob/1bf9c8b53f1b81affb845f1d184c5da40678abf4/src/mongocxx/client.cpp#L326) for const + non-const overloads containing a null check -> exception on violation.

src/mongocxx/search_index_view.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+MONGOCXX_INLINE_NAMESPACE_END
+}  // namespace mongocxx
```
eramongodb, line 105:
```suggestion
}  // namespace mongocxx

```
Missing EOF newline.

src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+///
+/// Class representing a search index on a MongoDB server.
+///
+class MONGOCXX_API search_index_model {
+   public:
+    ///
+    /// Initializes a new search_index_model over a mongocxx::collection.
+    ///
+    search_index_model(const bsoncxx::document::view_or_value& definition);
```
eramongodb, line 23:
```suggestion
    search_index_model(bsoncxx::document::view_or_value definition);
```
view_or_value parameters should not be `const&`.

src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <mongocxx/config/prelude.hpp>
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+///
+/// Class representing a search index on a MongoDB server.
+///
+class MONGOCXX_API search_index_model {
```
eramongodb, line 18: As currently written, there does not appear to be compelling reason to restrict access to members, as there are no invariants being maintained. This class could in principle just be:
```cpp
struct MONGOCXX_API search_index_model {
    bsoncxx::stdx::optional<bsoncxx::string::view_or_value> _name;
    bsoncxx::document::view_or_value _definition;
};
```
Instead, consider applying the PIMPL pattern to this class as well for consistency with (most) other public interface class.

src/mongocxx/private/search_index_view.hh
```diff
@@ -0,0 +1,192 @@
...
+    impl(mongoc_collection_t* collection, mongoc_client_t* client)
+        : _coll{collection}, _client{client} {}
+
+    impl(const impl& i) = default;
+
+    impl(impl&& i) = default;
+
+    ~impl() = default;
+
+    impl& operator=(const impl& i) = default;
```
eramongodb, line 36:
```suggestion
    ~impl() = default;
    impl(impl&& i) = delete;
    impl& operator=(impl&& i) = delete;
    impl(const impl& i) = delete;
    impl& operator=(const impl& i) = delete;
```
All `= default` -> none necessary. That being said, would recommend deleting copy/move special members as the `impl` class is not expected to be directly copied or moved.

src/mongocxx/private/search_index_view.hh
```diff
@@ -0,0 +1,192 @@
...
+
+        append_aggregate_options(options, opts_doc);
+
+        if (session) {
+            opts_doc.append(bsoncxx::builder::concatenate_doc{session->_get_impl().to_document()});
+        }
+
+        if (options.read_preference()) {
+            rp_ptr = options.read_preference()->_impl->read_preference_t;
+        }
```
eramongodb, line 68:
```suggestion
        const mongoc_read_prefs_t* const rp_ptr =
            options.read_preference() ? options.read_preference()->_impl->read_preference_t
                                      : nullptr;
```
Prefer `const` when able over set-and-maybe-modify patterns.

src/mongocxx/private/search_index_view.hh
```diff
@@ -0,0 +1,192 @@
...
+
+        libbson::scoped_bson_t opts_bson(opts_doc.view());
+
+        return libmongoc::collection_aggregate(
+            _coll, static_cast<::mongoc_query_flags_t>(0), stages.bson(), opts_bson.bson(), rp_ptr);
```
eramongodb, line 73:
```suggestion
            _coll, mongoc_query_flags_t(), stages.bson(), opts_bson.bson(), rp_ptr);
```
`static_cast` not necessary.

src/mongocxx/private/search_index_view.hh
```diff
@@ -0,0 +1,192 @@
...
+        bsoncxx::document::view result_view = result.view();
+
+        return bsoncxx::stdx::make_optional(
+            bsoncxx::string::view_or_value(result_view["indexesCreated"]
+                                               .get_array()
+                                               .value.begin()
+                                               ->get_document()
+                                               .value["name"]
+                                               .get_string()
+                                               .value));
```
eramongodb, line 89:
```suggestion
        const auto result = create_many(session, std::vector<search_index_model>{model});
        return bsoncxx::stdx::make_optional(bsoncxx::string::view_or_value(result["indexesCreated"]
                                                                               .get_array()
                                                                               .value.begin()
                                                                               ->get_document()
                                                                               .value["name"]
                                                                               .get_string()
                                                                               .value));
```
Intermediate `result_view` does not seem necessary.

This function appears to always return a string value, otherwise `.get_string()` or any other document access prior will throw an exception. The [spec](https://github.com/mongodb/specifications/blob/9d7ceed8c37bafb5769787c26c8f8c79670f94f8/source/index-management/index-management.rst#index-view-api) also suggests the result is not expected to be optional. Consider changing return type from `bsoncxx::stdx::optional<bsoncxx::string::view_or_value>` to `bsoncxx::string::value`.

eramongodb (changes requested):

src/mongocxx/private/search_index_view.hh
eramongodb: Whoops. Intended to mean plain `std::string` (the `value` case of `view_or_value`).

joshbsiegel: Right, sometimes I take things too literally :sweat_smile:

src/mongocxx/private/search_index_model.hh
```diff
@@ -0,0 +1,47 @@
...
+    impl(bsoncxx::string::view_or_value name, bsoncxx::document::view_or_value definition)
+        : _name(name), _definition(definition.view()) {}
+
+    ~impl() = default;
+
+    impl(impl&& i) = delete;
+
+    impl& operator=(impl&& i) = delete;
+
+    impl(const impl& i) : _name(i._name), _definition(i._definition) {}
```
joshbsiegel, line 25: I'm not sure if it's better practice for this to be ` _definition(i._definition)` or `_definition(i._definition.view())`. I believe both would work, as the constructor for `_impl` does the latter.

eramongodb: `search_index_model::impl` doesn't require user-provided default, delete, or definition of special copy/move member functions. Suggest removing them all and keeping only the ctors (to support `make_*` and `emplace*` functions). The data members could even be made public, making the accessors unnecessary, e.g. as done for `read_preference::impl`: https://github.com/mongodb/mongo-cxx-driver/blob/1bf9c8b53f1b81affb845f1d184c5da40678abf4/src/mongocxx/private/read_preference.hh#L25

src/mongocxx/search_index_model.cpp
```diff
@@ -0,0 +1,56 @@
...
+search_index_model::search_index_model(const search_index_model& other)
+    : _impl(bsoncxx::stdx::make_unique<impl>(*other._impl)) {}
+
+search_index_model& search_index_model::operator=(const search_index_model& other) {
+    assert(other._impl);
+    if (_impl) {
+        *_impl = *other._impl;
+    } else {
+        _impl = bsoncxx::stdx::make_unique<impl>(*other._impl);
+    }
```
eramongodb, line 29:
```suggestion
    _get_impl() = other._get_impl();
```
Not-null invariant means this can be simplified to a simple copy of the `impl` object.

src/mongocxx/search_index_model.cpp
```diff
@@ -0,0 +1,56 @@
...
+search_index_model::search_index_model(bsoncxx::string::view_or_value name,
+                                       bsoncxx::document::view_or_value definition)
+    : _impl{bsoncxx::stdx::make_unique<impl>(name, definition)} {}
+
+search_index_model::search_index_model(search_index_model&&) noexcept = default;
+
+search_index_model& search_index_model::operator=(search_index_model&&) = default;
+
+search_index_model::search_index_model(const search_index_model& other)
+    : _impl(bsoncxx::stdx::make_unique<impl>(*other._impl)) {}
```
eramongodb, line 21:
```suggestion
    : _impl(bsoncxx::stdx::make_unique<impl>(other._get_impl())) {}
```
Ensure the not-null invariant is satisfied.

src/mongocxx/search_index_model.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return _get_impl().name();
+}
+
+bsoncxx::document::view search_index_model::definition() const {
+    return _get_impl().definition();
+}
+
+const search_index_model::impl& search_index_model::_get_impl() const {
+    if (!_impl) {
+        throw logic_error{error_code::k_invalid_client_object};
```
eramongodb, line 45: `error_code::k_invalid_client_object` is for `mongocxx::client`. Add a new `k_invalid_search_index_model` enumerator to the `error_code` enumeration instead.

src/mongocxx/search_index_model.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    ///
+    /// Move constructs a search_index_model.
+    ///
+    search_index_model(search_index_model&&) noexcept;
+
+    ///
+    /// Move assigns a search_index_model.
+    ///
+    search_index_model& operator=(search_index_model&&);
```
eramongodb, line 37:
```suggestion
    search_index_model& operator=(search_index_model&&) noexcept;
```

src/mongocxx/search_index_view.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+namespace mongocxx {
+MONGOCXX_INLINE_NAMESPACE_BEGIN
+
+///
+/// Class representing a MongoDB search index view.
+///
+class MONGOCXX_API search_index_view {
+   public:
+    search_index_view(search_index_view&&) noexcept;
```
eramongodb, line 21:
```suggestion
    search_index_view(const search_index_view&) = delete;
    search_index_view& operator=(const search_index_view&) = delete;

    search_index_view(search_index_view&&) noexcept;
```
Rule of 5: explicitly delete the remaining special (copy) member funtions.

eramongodb: Alternatively, support copyability as done for `search_index_model` by applying the same `_get_impl() = other._get_impl()` pattern and removing the `= delete` of special copy/move members in `search_index_view::impl`.

src/mongocxx/search_index_view.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+void search_index_view::update_one(const client_session& session,
+                                   bsoncxx::string::view_or_value name,
+                                   bsoncxx::document::view_or_value definition) {
+    _get_impl().update_one(&session, name, definition);
+}
+
+const search_index_view::impl& search_index_view::_get_impl() const {
+    if (!_impl) {
+        throw logic_error{error_code::k_invalid_client_object};
```
eramongodb, line 109: Needs similar fix as `search_index_view` w.r.t. `error_code` enumerator.

eramongodb (approved):
Small feedback remaining; otherwise, LGTM. 👍

src/mongocxx/private/search_index_model.hh
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <bsoncxx/document/view_or_value.hpp>
+#include <mongocxx/private/client_session.hh>
+#include <mongocxx/private/libbson.hh>
+#include <mongocxx/private/libmongoc.hh>
```
kevinAlbs, line 6:
```suggestion
#include <mongocxx/search_index_model.hpp>
```

Include `search_index_model.hpp` directly. Remove unnecessary headers.

src/mongocxx/search_index_model.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return _get_impl()._name;
+}
+
+bsoncxx::document::view search_index_model::definition() const {
+    return _get_impl()._definition.view();
+}
+
+const search_index_model::impl& search_index_model::_get_impl() const {
+    if (!_impl) {
+        throw logic_error{error_code::k_invalid_search_index_model};
```
kevinAlbs, line 40:
```suggestion
        throw mongocxx::logic_error{error_code::k_invalid_search_index_model};
```

`search_index_model.cpp` appears to rely on headers included through`client_session.hh`. Update `search_index_model.cpp` to include headers needed for throwing error:

```c++
#include <mongocxx/exception/error_code.hpp>
```

kevinAlbs (commented):
LGTM with minor comments addressed.

PR #993, opened 2023-07-26
Skip building tests by default

kkloberdanz:
CXX-1792
CXX-2702

Tests will not be compiled by default.
To enable building tests, build with `-DBUILD_TESTING=ON`


src/bsoncxx/test/CMakeLists.txt
```diff
@@ -12,24 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
vector-of-bool, line 14: I'm curious if this will work, but I haven't tested it:

```cmake
if(NOT BUILD_TESTING)
  set_property(DIRECTORY PROPERTY EXCLUDE_FROM_ALL TRUE)
endif()
```

(and then the remainder of the file is reverted to as-it-was). IIUC, this should exclude every added target from the implicit `all`, but still defined them and are available for building by being named explicitly?

src/bsoncxx/test/CMakeLists.txt
```diff
@@ -12,24 +12,28 @@
...
 
-include_directories(
-    ${MONGO_CXX_DRIVER_SOURCE_DIR}/src/third_party/catch/include
-)
+if (${ENABLE_TESTS})
```
vector-of-bool, line 15:
```suggestion
if (ENABLE_TESTS)
```

Fun `if` parsing oddities again.

src/bsoncxx/test/CMakeLists.txt
kevinAlbs: I agree with the suggestion to exclude test targets from `all`, but still have them defined.

Example: Targets in the examples directory are currently [excluded from the all target](https://github.com/mongodb/mongo-cxx-driver/blob/effc286c962c93b14406ef41d314558e5cad7b5f/CMakeLists.txt#L346) with:
```cmake
add_custom_target(examples DEPENDS ${MONGOCXX_EXAMPLE_EXECUTABLES} ${BSONCXX_EXAMPLE_EXECUTABLES})
```

But the example targets are still defined. CMake targets can be listed with `cmake --build cmake-build --target help`. An example target can be built by including the `--target` argument: `cmake --build cmake-build --target aggregation_examples`

kkloberdanz: This doesn't seem to work when I try it. The test programs are built by default when I implement this like so:

`CMakeLists.txt`:
```
if(NOT DEFINED BUILD_TESTING)
  set(BUILD_TESTING OFF)  # Set this to OFF by default
endif()
```

`src/bsoncxx/test/CMakeLists.txt` and `src/mongocxx/test/CMakeLists.txt`:
```
if(NOT BUILD_TESTING)
  set_property(DIRECTORY PROPERTY EXCLUDE_FROM_ALL TRUE)
endif()

include_directories(
    ${MONGO_CXX_DRIVER_SOURCE_DIR}/src/third_party/catch/include
)
```

vector-of-bool: Not sure if I can figure why the directory property doesn't work, so just discard that suggestion for now. On the other hand, the BUILD_TESTING variable should work? Are you sure to clear it from your cache before configuring?

src/bsoncxx/test/CMakeLists.txt
kkloberdanz: I'm unresolving this because after applying this suggestion, the bsoncxx tests are being built by default. Should I revert this suggestion?

```
ls -l src/bsoncxx/test/test_bson
```

src/bsoncxx/test/CMakeLists.txt
kkloberdanz: I think I found the issue, and after applying the fix below, your suggestion then works. I pushed these changes and the tests are green.
```diff
-        set(ENABLE_TESTS OLD_ENABLE_TESTS)
-        set(BUILD_TESTING OLD_BUILD_TESTING)
+        set(ENABLE_TESTS ${OLD_ENABLE_TESTS})
+        set(BUILD_TESTING ${OLD_BUILD_TESTING})
```

CMakeLists.txt
```diff
@@ -14,6 +14,10 @@
 
 cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 
+if(NOT DEFINED BUILD_TESTING)
+  set(BUILD_TESTING OFF)  # Set this to OFF by default
+endif()
```
vector-of-bool, line 19: Suggest: Move this below `project()`. Will also want to `include(CTest)`.

vector-of-bool (approved):
LGTM. One more comment.

CMakeLists.txt
kkloberdanz: Based off our conversation in slack, we'll leave off the `include(CTest)` for now since it is currently causing the tests to be built unconditionally.

kevinAlbs (approved):

PR #994, opened 2023-07-26
Opt in to the default CMake install path

kkloberdanz:
CXX-2703

Allow user to opt in to the default CMake install path by setting `-DUSE_DEFAULT_INSTALL_PATH=ON`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,9 +130,12 @@ To build static libraries only, set both BUILD_SHARED_LIBS and BUILD_SHARED_AND_
 endif()
 
 
-# If the user did not customize the install prefix,
-# set it to live under build so we don't inadvertently pollute /usr/local
-if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+option(USE_DEFAULT_INSTALL_PATH "Override the CMake default installation path" OFF)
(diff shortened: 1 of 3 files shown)


kevinAlbs (approved):

rcsanchez97 (approved):

CMakeLists.txt
```diff
@@ -130,9 +130,12 @@ To build static libraries only, set both BUILD_SHARED_LIBS and BUILD_SHARED_AND_
 endif()
 
 
-# If the user did not customize the install prefix,
-# set it to live under build so we don't inadvertently pollute /usr/local
-if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+option(USE_DEFAULT_INSTALL_PATH "Override the CMake default installation path" OFF)
+if(NOT CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND USE_DEFAULT_INSTALL_PATH)
```
vector-of-bool, line 134: `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` is only TRUE on the very first execution of configure, so setting `USE_DEFAULT_INSTALL_PATH=TRUE` will trigger this branch on every subsequent configure, preventing it from succeeding until `CMAKE_INSTALL_PREFIX` is un-set on the CLI with `-U`.

PR #995, opened 2023-07-27
CXX-2722 Add binary data type Sensitive to the BSON bindata subtypes enum

wqian94:
Based on #827:
> The C++ driver does not appear to have BSON corpus tests corresponding to those proposed in the spec update. Therefore the scope of changes required seems to be limited to adding the new BSON binary subtype constant.


wqian94:
The genny workload fails because of IDL validation. Since the genny workload here uses the `bsoncxx` library to construct its BSON binary objects, it's probably a good idea to add support for the new binary data type.

kevinAlbs (approved):
LGTM. Thank you for the PR.

Is BSON binary subtype 08 expected to be created or parsed by typical user applications? If yes, I may create a DRIVERS ticket to inform other driver teams to add support.

PR #1002, opened 2023-07-31
CXX-2700 Add search index management e2e testing against Atlas

joshbsiegel:
# Summary
This PR will add 5 prose tests that will test search index management against a live Atlas cluster, modify the existing Evergreen config to automatically run these tests in the CI pipeline, and make small changes to the existing spec tests to sync with the [drivers ticket](https://github.com/mongodb/specifications/pull/1442).

# Background
[This](https://github.com/mongodb/specifications/pull/1442) ticket (mainly the changes made in the README) gives the vast majority of context needed to understand what prose tests should be added, why they should be added, and how to run them in Evergreen.

# How to Test Locally
Being that running the prose tests could take around 20 minutes in total, there was not a lot of emphasis placed on making them easy to run locally. However, if you do wish to run them locally, here are the steps that you'll have to take:
1. Spawn an Atlas cluster. The easiest way to do this is by running the `setup-atlas-cluster.sh` script in `drivers-evergreen-tools/.evergreen/atlas`. You'll need to `export` some of the environment variables mentioned in the comments of the script, they can be found on the Evergreen variables page. Spawning the cluster can take 7-10 minutes.
2. Export the `MONGODB_URI` environment variable as the cluster URI (printed to `atlas-expansion.yml` by the script)
3. Compile and run the tests. The test name is `atlas search indexes prose tests`. This will take 5-10 minutes.
4. Teardown the cluster using `teardown-atlas-cluster.sh`. You'll need some environment variables for this one too, it's all mentioned in the comments of the script.

# Evergreen
We chose to use a task group to run the search index prose tests. What's unique about task groups is that they have a `setup_group` and a `teardown_group`, which are commands that will run before the tasks attached to the group are run (setup) and after they are run (teardown). Both the setup commands and the teardown commands will run regardless of the outcome of the actual tests. This is important in this use case because we need to ensure that the Atlas cluster gets torn down even if the tests don't pass. Failure to teardown the Atlas cluster is inefficient and will cause problems the next time that we try to spawn an Atlas cluster. 

# What's New
- Minor changes to the spec tests
- Added another overload to `create_one` that only takes in a definition (see comment for explanation)
- Added 5 prose tests in `test/search_index_view.cpp`
- Updated Evergreen config to automatically spawn a cluster, run the tests, and teardown the cluster

### DIFF
--- a/src/mongocxx/search_index_view.cpp
+++ b/src/mongocxx/search_index_view.cpp
@@ -65,23 +65,24 @@ std::string search_index_view::create_one(const client_session& session,
     return _get_impl().create_one(&session, model);
 }
 
-std::vector<bsoncxx::string::view_or_value> search_index_view::create_many(
+std::vector<std::string> search_index_view::create_many(
     const std::vector<search_index_model>& models) {
     auto response = _get_impl().create_many(nullptr, models);
     return _create_many_helper(response["indexesCreated"].get_array().value);
 }
 
-std::vector<bsoncxx::string::view_or_value> search_index_view::create_many(
+std::vector<std::string> search_index_view::create_many(
     const client_session& session, const std::vector<search_index_model>& models) {
     auto response = _get_impl().create_many(&session, models);
     return _create_many_helper(response["indexesCreated"].get_array().value);
 }
 
-std::vector<bsoncxx::string::view_or_value> search_index_view::_create_many_helper(
+std::vector<std::string> search_index_view::_create_many_helper(
     bsoncxx::array::view created_indexes) {
-    std::vector<bsoncxx::string::view_or_value> search_index_names;
+    std::vector<std::string> search_index_names;
     for (auto&& index : created_indexes) {
-        search_index_names.push_back(index.get_document().value["name"].get_string().value);
+        search_index_names.push_back(
+            bsoncxx::string::to_string(index.get_document().value["name"].get_string().value));
     }
     return search_index_names;
 }
--- a/src/mongocxx/search_index_view.cpp
+++ b/src/mongocxx/search_index_view.cpp
@@ -45,6 +45,15 @@ cursor search_index_view::list(const client_session& session,
     return _get_impl().list(&session, name, options);
 }
 
+std::string search_index_view::create_one(bsoncxx::document::view_or_value definition) {
+    return create_one(search_index_model(definition));
+}
+
+std::string search_index_view::create_one(const client_session& session,
+                                          bsoncxx::document::view_or_value definition) {
+    return create_one(session, search_index_model(definition));
+}
+
 std::string search_index_view::create_one(bsoncxx::string::view_or_value name,
                                           bsoncxx::document::view_or_value definition) {
     return create_one(search_index_model(name, definition));
--- a/src/mongocxx/search_index_view.hpp
+++ b/src/mongocxx/search_index_view.hpp
@@ -151,8 +151,7 @@ class MONGOCXX_API search_index_view {
     ///
     /// @return The names of the created indexes.
     ///
-    std::vector<bsoncxx::string::view_or_value> create_many(
-        const std::vector<search_index_model>& models);
+    std::vector<std::string> create_many(const std::vector<search_index_model>& models);
 
     ///
     /// Creates multiple search indexes in the collection.
@@ -164,8 +163,8 @@ class MONGOCXX_API search_index_view {
     ///
     /// @return The names of the created indexes.
     ///
-    std::vector<bsoncxx::string::view_or_value> create_many(
-        const client_session& session, const std::vector<search_index_model>& models);
+    std::vector<std::string> create_many(const client_session& session,
+                                         const std::vector<search_index_model>& models);
 
     ///
     /// @}
@@ -232,7 +231,7 @@ class MONGOCXX_API search_index_view {
 
     MONGOCXX_PRIVATE search_index_view(void* coll, void* client);
 
-    MONGOCXX_PRIVATE std::vector<bsoncxx::string::view_or_value> _create_many_helper(
+    MONGOCXX_PRIVATE std::vector<std::string> _create_many_helper(
         bsoncxx::array::view created_indexes);
 
     MONGOCXX_PRIVATE const impl& _get_impl() const;
--- a/src/mongocxx/search_index_view.hpp
+++ b/src/mongocxx/search_index_view.hpp
@@ -89,6 +89,27 @@ class MONGOCXX_API search_index_view {
     ///
     /// This is a convenience method for creating a single search index.
     ///
+    /// @param definition
+    ///    The document describing the search index to be created.
+    ///
+    /// @return The name of the created search index.
+    ///
+    std::string create_one(bsoncxx::document::view_or_value definition);
+
+    ///
+    /// This is a convenience method for creating a single search index.
+    ///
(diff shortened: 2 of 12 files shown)


src/mongocxx/test/search_index_view.cpp
```diff
@@ -1 +1,211 @@
...
+        auto new_definition = make_document(kvp("mappings", make_document(kvp("dynamic", true))));
+        auto new_model = search_index_model(name, new_definition.view());
+        siv.update_one(name, new_definition.view());
+
+        wait_for_search_index(siv, new_model, true, true);
+
+        std::cout << "update one SUCCESS" << std::endl;
+    }
+
+    // SECTION("drop one suppress namespace not found") {
```
joshbsiegel, line 204: This test case comes from https://github.com/mongodb/specifications/commit/267a54d1bb9af894d05d3dd1a6bda0fa53e20a3f#diff-7f618607e8afab57c7ebfd80f0e91802e2f705ec79bfb946947d026a4fa97ba0R189. I'm not sure how to create a driver-side collection object in the CXX. I see this private function in `collection.hpp` that seems to be the best option at this point.
```
    MONGOCXX_PRIVATE collection(const database& database,
                                bsoncxx::string::view_or_value collection_name);
```
Currently, the commented out test case returns this
```
invalid use of default constructed or moved-from mongocxx::collection object
```
due to there not being a valid `_impl` initialized with the collection object.

src/mongocxx/test/search_index_view.cpp
```diff
@@ -1 +1,211 @@
...
+    return false;
+}
+
+void wait_for_search_index(search_index_view& siv,
+                           search_index_model& model,
+                           bool to_exist,
+                           bool with_status) {
+    auto c = siv.list();
+    while (does_search_index_exist_on_cursor(c, model, with_status) == !to_exist) {
+        std::this_thread::sleep_for(std::chrono::seconds(5));
```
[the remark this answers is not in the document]
joshbsiegel: [This](https://github.com/mongodb/specifications/pull/1442/files#diff-7f618607e8afab57c7ebfd80f0e91802e2f705ec79bfb946947d026a4fa97ba0R61) refers to the Evergreen timeout, but I ended up adding a 5 minute timeout on this function as well.

src/mongocxx/search_index_view.hpp
```diff
@@ -86,6 +86,27 @@ class MONGOCXX_API search_index_view {
...
+    /// This is a convenience method for creating a single search index.
+    ///
+    /// @param definition
+    ///    The document describing the search index to be created.
+    ///
+    /// @return The name of the created search index.
+    ///
+    std::string create_one(const client_session& session,
+                           bsoncxx::document::view_or_value definition);
+
```
joshbsiegel, line 109: These two overloads were added due to changes made in [DRIVERS-2630](https://github.com/mongodb/specifications/commit/267a54d1bb9af894d05d3dd1a6bda0fa53e20a3f#diff-8012d6483484505d56bd53efe58a4f4affbc1512d15d09fcd003241c6ae2fcbeR1049). This change allows for `create_one` to be called while only passing through a definition, no name.

src/mongocxx/search_index_view.cpp
```diff
@@ -65,23 +74,24 @@ std::string search_index_view::create_one(const client_session& session,
     return _get_impl().create_one(&session, model);
 }
 
-std::vector<bsoncxx::string::view_or_value> search_index_view::create_many(
+std::vector<std::string> search_index_view::create_many(
```
joshbsiegel, line 77: I ran into some issues with reading the values of the `std::vector<bsoncxx::string::view_or_value>` returned from this function due to ownership semantics. After talking to @eramongodb, we came to the conclusion that it would be easier and more clear to just make this a `std::vector<std::string>`. This minor change seen across this PR.

src/mongocxx/test/search_index_view.cpp
```diff
@@ -1 +1,211 @@
...
+        auto definition = make_document(kvp("mappings", make_document(kvp("dynamic", false))));
+        auto model = search_index_model(name, definition.view());
+
+        REQUIRE(siv.create_one(model) == "test-search-index");
+
+        REQUIRE(wait_for_search_index(siv, model, true, false));
+
+        siv.drop_one(name);
+
+        REQUIRE(wait_for_search_index(siv, model, false, false));
```
kevinAlbs, line 168: The spec prose tests say to check for empty results:

> #. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until ``listSearchIndexes`` returns an empty array.

As implemented, `does_search_index_exist_on_cursor` only returns `true` if some indexes are returned. I expect the check may time out if the results are empty before they change state.

Consider replacing `wait_for_search_index` with a more generic `assert_soon` helper to separate the logic to wait for a condition:

```c++
// `assert_soon` repeatedly calls `fn` and asserts `fn` eventually returns true.
// `fn` is called every five seconds. Fails if `fn` does not return true within five minutes.
void assert_soon(std::function<bool()> fn) {
    auto start = std::chrono::high_resolution_clock::now();
    while (std::chrono::high_resolution_clock::now() - start < std::chrono::minutes(5)) {
        if (fn()) {
            return;
        }
        std::this_thread::sleep_for(std::chrono::seconds(5));
    }

    FAIL("Expected function to return true within five minutes, but did not");
}
```

Then, the "wait for empty" condition could be implemented as:

```c++
assert_soon([&siv](void) -> bool {
    auto cursor = siv.list();
    // Return true if empty results are returned.
    return cursor.begin() == cursor.end();
});
```

src/mongocxx/test/search_index_view.cpp
```diff
@@ -1 +1,211 @@
...
+
+    database db = mongodb_client[std::getenv("FUNCTION_NAME")];
+
+    SECTION("create one with name and definition") {
+        // use a randomly generated collection name as there's a server side limitation that
+        // prevents multiple search indexes with the same name, definition and collection name
+        // from being created.
+        bsoncxx::oid id;
+        db.create_collection(id.to_string());
+        collection coll = db[id.to_string()];
```
kevinAlbs, line 69:
```suggestion
        auto coll = db.create_collection(id.to_string());
```

Suggested simplification. Also applies to other tests.

src/mongocxx/test/search_index_view.cpp
joshbsiegel: I don't believe that this works either. I get this error returned: 
```
  Collection 'test.64cbcd93b74f63a863033401' does not exist.: generic server
  error
```
It's failing on the `search_indexes` call, not the `drop_one` call.

kevinAlbs: > It's failing on the `search_indexes` call, not the `drop_one` call.

I think it is failing on the `drop_one` call.

The specification notes a requirement to ignore the "NamespaceNotFound" error.

> Drivers MUST suppress NamespaceNotFound errors for the `dropSearchIndex` helper.

I expect `drop_one` needs to be updated. Try applying this commit: https://github.com/mongodb/mongo-cxx-driver/commit/23c8bd33014acc9a1107b0d05e351cbf819b95a2

joshbsiegel: Got it, thanks. Fixed now.

kevinAlbs (approved):

PR #991, opened 2023-07-20
CXX-2661 Remove tasks on macos-1014

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/974 and resolves CXX-2661. Verified by [this patch](https://spruce.mongodb.com/version/64b98f6361837d8919781cb9).

Addresses some miscellaneous EVG validation warnings along the way:

- Sets top-level task timeout to 1 hour. All the current tasks seem to finish under 30 minutes.
- The ubuntu1804-debug variant's display name conflicted with one of the integrations matrix variants. Added a simple " (Extra)" to the name to avoid the conflict.
- The `mongocryptd` matrix was attempting to define the `use_mongocryptd` EVG expansion where it is not supported by the EVG configuration syntax. Moved the expansion into a new matrix axis where it's definition _is_ supported. Note, this breaks the EVG task history for the given variant (same display name as before, but actually a different matrix spec).

PR #967, opened 2023-05-18
Auto download C Driver

kkloberdanz:

PR #1007, opened 2023-08-03
CXX-1848 Add sharded cluster test coverage to EVG config

eramongodb:
## Description

This PR primarily resolves CXX-1848. Verified by [this patch](https://spruce.mongodb.com/version/64cbbfd2850e61d843791438).

This PR syncs both the legacy and unified transactions spec test files, and adds mongos pinning prose tests.

The syncing of spec test files also resolves, or contributes to the partial resolution of, CXX-1547, CXX-2406, CXX-1976, CXX-2221, CXX-2233, CXX-2235, CXX-2393, CXX-2552, and CXX-2690.

### Raw Response Document

There is only one notable change to the behavior of the CXX Driver, which is a bugfix that addresses a discrepancy in behavior of a sharded cluster compared to a single server or replica set when creating index views, as documented in SERVER-78611.

The `index_view::create_*()` functions return an optional that is empty if the index(es) to be created already exist (note: this is _not_ consistent with the [current Drivers spec](https://github.com/mongodb/specifications/blob/267a54d1bb9af894d05d3dd1a6bda0fa53e20a3f/source/index-management/index-management.rst#index-view-api), which does not return an optional). This does not behave as intended with sharded clusters due to the response fields being wrapped in a "raw" response document instead of being top-level fields.

This PR fixes the check such that the optional is empty as intended if the index views to be created already exist (consistency with single servers and replica sets).

### $listLocalSessions and Non-Existent Databases

Sharded clusters appear to behave differently to single servers and replica sets when running `$listLocalSessions` on a database that does not yet exist, as documented in SERVER-79306. The only change required to account for this discrepancy was updating the aggregation example code to ensure teh database exists prior to running aggregation commands. This _may_ affect more aggregation operations, but they were not exposed in testing.

## Legacy Test Runner Changes

The addition of sharded cluster test coverage exposed a multitude of issues with the legacy test runner . These issues include:

- Use of the obsolete `"sharded-replicaset"` topology (followup to https://github.com/mongodb/mongo-cxx-driver/pull/975).
- Prose tests missing minimum server requirements (failCommand on mongos pre-4.1, multi-document transactions on pre-4.2).
- Flaky tailable await cursor tests due to extremely small `maxAwaitTimeMS`.
- Failure to ignore `configureFailPoint` APM events.
- Failure to check outcome of collection with `readPreference=primary` + `readConcern=local`
- Failure to account for bulkWrite client error exception inconsistencies (see CXX-1679).
- Flaky test failures due to failure to properly disable regular failPoints on test failure (exception thrown -> disable code not executed).

The only test cases that are skipped is due to CXX-2678 not yet being implemented, which prevents the legacy test runner from being able to specify the `readPreference` for a `runCommand` operation.

Note: testing ran into the same issue as CDRIVER-4351. Identified the cause to be the preceeding `commitTransaction retry fails on new mongos` test in `mongos-recovery-token.json` failing to properly disable the targetedFailPoint due to failing to trigger the `isMaster`/`hello` failPoint as intended (retries were not actually sending a command to the server). Creating a new client object on each attempt resolved the flaky failures. This solution may also be required in the C Driver to address CDRIVER-4351.

## Unified Test Runner Changes

Issues that were fixed include:

- Failure to check outcome of collection with `readPreference=primary` + `readConcern=local`.
- Incorrect specification of hosts in test client connection strings (simplified to hardcoded topology assumptions corresponding to current mongo-orchestration configuration files).


src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -1963,6 +1963,36 @@ document::value operations::run(entity::map& entity_map,
...
+        auto& client = session.client();
+
+        // If the driver exposes an API to target a specific server for a command, the test runner
+        // SHOULD use the client entity associated with the session to execute the
+        // configureFailPoint command.
+        client["admin"].run_command(fail_point, server_id);
+
+        return make_document(kvp("uri", client.uri().to_string()),
+                             kvp("failPoint", fail_point["configureFailPoint"].get_string()),
+                             kvp("serverId", static_cast<std::int64_t>(server_id)));
```
kevinAlbs, line 1994:
```suggestion
        return empty_doc;
```

I expect no result is needed for `targetedFailPoint`.

eramongodb: I realize I dropped some changes to the unified test runner to use scope guards to disable failpoints during one of my rebases. These values are intended to be used by the targetedFailPoint scope guard. Will push those changes shortly.

src/mongocxx/test/collection.cpp
```diff
@@ -2769,6 +2769,12 @@ TEST_CASE("Ensure that the WriteConcernError 'errInfo' object is propagated", "[
 
     client mongodb_client{uri{}, test_util::add_test_server_api()};
 
+    if (test_util::get_topology(mongodb_client) == "sharded" &&
+        test_util::get_server_version(mongodb_client) < "4.1.0") {
```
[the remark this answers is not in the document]
eramongodb: Good catch. I was misled by other instances of relational comparison with the result of `get_server_version()`. Will fix accordingly.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/64d4050661837df5baad0805). Merged of upstream branch included redefining sharded cluster tasks to match the new `_with_libmongocrypt` tasks. Also fixed what appears to have been a mis-merge of https://github.com/mongodb/mongo-cxx-driver/pull/991 and https://github.com/mongodb/mongo-cxx-driver/pull/967 concerning EVG config warnings for the `mongocryptd` matrix variant (invalid expansion placement).

src/mongocxx/test/spec/util.cpp
```diff
@@ -665,134 +835,161 @@ void run_transaction_operations(document::view test,
...
+                            REQUIRE(main);
+                            REQUIRE(main->type() == type::k_document);
+                            auto session_name = pattern.get_string().value;
+                            if (session_name.compare("session0") == 0) {
+                                REQUIRE(
+                                    test_util::matches(session_lsid0, main->get_document().value));
+                            } else {
+                                REQUIRE(
+                                    test_util::matches(session_lsid1, main->get_document().value));
+                            }
```
kevinAlbs, line 959:
```suggestion
                            }
                            return test_util::match_action::k_skip;
```

Skip when matching `lsid`. The pattern value is a string "session0" / "session1". Actual value is the "lsid" document.

src/mongocxx/test/spec/util.cpp
```diff
@@ -665,134 +835,161 @@ void run_transaction_operations(document::view test,
...
+                        } else if (pattern.type() == type::k_null) {
+                            if (main) {
+                                return test_util::match_action::k_not_equal;
+                            }
+                            return test_util::match_action::k_skip;
+                        } else if (key.compare("upsert") == 0 || key.compare("multi")) {
```
kevinAlbs, line 965:
```suggestion
                        } else if (key.compare("upsert") == 0 || key.compare("multi") == 0) {
```

I goofed in the [previous suggestion](https://github.com/mongodb/mongo-cxx-driver/pull/1007#discussion_r1288986161). Fixing this results in other test failures in matching.

The test failures may suggest missing options or operations. [This change](https://github.com/mongodb/mongo-cxx-driver/commit/b1c1ba33b2f95ff9ebf4c3de9b4e300139879062) fixes the "causal consistency disabled" test.

src/mongocxx/test/spec/util.cpp
```diff
@@ -665,134 +835,161 @@ void run_transaction_operations(document::view test,
...
-
-        /* individual test may contain a skipReason */
-        if (should_skip_spec_test(client, test.get_document())) {
-            continue;
-        }
+        const auto description = string::to_string(test["description"].get_string().value);
 
-        options::client_session session0_opts;
-        options::client_session session1_opts;
+        SECTION(description) {
```
kevinAlbs, line 856:
```suggestion
        INFO("Test description: " << description);
        SECTION(description) {
```

Suggestion is intended to include test description on failure output. Example:

```bash
/Users/kevin.albertson/review/mongo-cxx-driver-1007/src/mongocxx/test/spec/monitoring.cpp:128: FAILED:
explicitly with messages:
  checking for path from environment variable: TRANSACTIONS_LEGACY_TESTS_PATH
  test path is: /Users/kevin.albertson/review/mongo-cxx-driver-1007/data/
  transactions/legacy
  Test path: /Users/kevin.albertson/review/mongo-cxx-driver-1007/data/
  transactions/legacy/error-labels.json
  Test description: do not add UnknownTransactionCommitResult label to
  MaxTimeMSExpired inside transactions
  print_all() := "
```

Which can then be used to select the section to run:
```bash
./cmake-build/src/mongocxx/test/test_transactions_specs \
    "Transactions spec automated tests" \
    --section "Legacy" \
    --section "error-labels.json" \
    --section "do not add UnknownTransactionCommitResult label to MaxTimeMSExpired inside transactions"
```

eramongodb: The use of the test description as the section name already ensures the description will be included in failure output as the section name, so this should not be necessary (re-emits the same test description twice). For example:

```cpp
TEST_CASE("Example") {
  SECTION("Success Section") {
    SUCCEED("Deliberate success");
  }

  SECTION("Failure Section") {
    FAIL("Deliberate failure");
  }
}
```
yields the following output (with `-s` to include the successful test in output):
```cpp
Example
  Success Section
main.cpp:9

main.cpp:10: PASSED:
with message:
  Deliberate success

Example
  Failure Section
main.cpp:13

main.cpp:14: FAILED:
explicitly with message:
  Deliberate failure

===============================================================================
test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed
```

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/64d53042d6d80a0b4adbea61).

src/mongocxx/test/spec/util.cpp
kevinAlbs: Oh nice, thank you. I missed the section name in the output.

kevinAlbs (approved):
The test runner improvements (e.g. capture of APM events on result failure, use of failpoint scope guards) are much appreciated. LGTM!

kkloberdanz (approved):
Looks good!

PR #1009, opened 2023-08-10
Create docker images for C++ driver

kkloberdanz:
CXX-2663

This creates a CI build and test for docker images based on the following distros:
- Alpine
- Debian
- Red Hat UBI
- Ubuntu

Each distro version is in its own directory. This directory contains a Makefile, a test C++ program to ensure that linking works properly, and a TestDockerfile to ensure that the image was built properly.

To build an image locally, navigate to the directory of the distro/version that you wish to build, then type:
```
make
```


extras/docker/alpine3.18/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF -DMONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION=ON -DCMAKE_INSTALL_PREFIX=/install -DCMAKE_PREFIX_PATH=/install .. && \
+    make -j $(nproc) && \
+    make install -j $(nproc)
+
+RUN wget https://github.com/mongodb/mongo-cxx-driver/releases/download/r${MONGOCXX_VERSION}/mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
+    tar xf mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
+    cd mongo-cxx-driver-r${MONGOCXX_VERSION}/build && \
+    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_INSTALL_PREFIX=/install -DCMAKE_PREFIX_PATH=/install && \
+    make -j $(nproc) && \
+    make install -j $(nproc)
```
vector-of-bool:
```suggestion
    tar xf mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
    cmake -S mongo-cxx-driver-r${MONGOCXX_VERSION} -B mongocxx-build -G "Ninja Multi-Config" \
        -D CMAKE_CXX_STANDARD=17 \
        -D mongocrypt_ROOT=/opt/libmongocrypt \
        -D bson-1.0_ROOT=/opt/mongo-c-driver \
        -D mongoc-1.0_ROOT=/opt/mongo-c-driver && \
    cmake --build mongocxx-build --config Release && \
    cmake --install mongocxx-build --config Release --prefix /opt/mongo-cxx-driver
```

extras/docker/alpine3.18/Dockerfile
vector-of-bool, line 1: I recommend consolidating the Makefiles, Dockerfiles, and TestDockerfiles together into a single [Earthfile](https://docs.earthly.dev/docs/earthfile), which can be used to create container images as output artifacts, and will facilitate parallel execution of independent container builds. Code that performs the actual build steps can then also be de-duplicated using Earthfile `COMMAND`.

extras/docker/alpine3.18/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF -DMONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION=ON -DCMAKE_INSTALL_PREFIX=/install -DCMAKE_PREFIX_PATH=/install .. && \
+    make -j $(nproc) && \
+    make install -j $(nproc)
+
+RUN wget https://github.com/mongodb/mongo-cxx-driver/releases/download/r${MONGOCXX_VERSION}/mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
+    tar xf mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
+    cd mongo-cxx-driver-r${MONGOCXX_VERSION}/build && \
+    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=/install -DCMAKE_PREFIX_PATH=/install && \
+    make -j $(nproc) && \
+    make install -j $(nproc)
```
vector-of-bool, line 39:
```suggestion
    tar xf mongo-cxx-driver-r${MONGOCXX_VERSION}.tar.gz && \
    cmake -S mongo-cxx-driver-r${MONGOCXX_VERSION} -B mongocxx-build -G "Ninja Multi-Config" \
        -D CMAKE_CXX_STANDARD=17 \
        -D CMAKE_PREFIX_PATH=/opt/mongocxx && \
    cmake --build mongocxx-build --config Release && \
    cmake --install mongocxx-build --config Release --prefix /opt/mongocxx
```

vector-of-bool (commented):
Recommendations for code deduplication. Also recommended to rename `extras/docker` to `extras/container` or `extras/oci`, since the results are not Docker-specific.

extras/docker/alpine3.18/Dockerfile
kkloberdanz: I am hesitant to use Earthly for this because it seems to add unnecessary complexity. I also think that in the long term it would make sense for this project to make its way to [Docker's official repository](https://github.com/docker-library), much like how Python, Postgres, Redis, etc are setup. I based my approach partially on Python -- but without using code gen for Dockerfiles. I think that sticking with basic Docker tooling will make this migration much easier in the long run and also keep the code and docker build simple in the short run.

vector-of-bool: My primary concern is about code locality and duplication.

To address this with only Docker itself, I recommend moving all Dockerfiles into the same directory, just use different names for the files. Then the build commands can be de-duplicated into scripts that can be ADDed into the containers as part of the builds.

extras/docker/alpine3.18/Makefile
```diff
@@ -0,0 +1,23 @@
...
+	docker build . \
+		-t mongo-cxx-driver:$(MONGOCXX_VERSION)-alpine3.18 \
+		--build-arg="MONGOCXX_VERSION=$(MONGOCXX_VERSION)"
+
+.PHONY: cleanbuild
+cleanbuild:
+	docker build . \
+		--no-cache \
+		-t mongo-cxx-driver:$(MONGOCXX_VERSION)-alpine3.18 \
+		--build-arg="MONGOCXX_VERSION=$(MONGOCXX_VERSION)"
```
vector-of-bool, line 14: Since the image names are fixed, I would recommend splitting to a separate `clean` target, and use `docker rmi $(imagename) || :` to delete the image, and replace the `cleanbuild` with `clean build` plus a "weak dependency" from `build -> clean` (see other comment).

extras/docker/alpine3.18/Dockerfile
kkloberdanz: One thing that I'm going for here is to follow the standard pattern that is used by all other images in the official docker repository, which is to have a directory per distro, for example, looking at the official [Python repo](https://github.com/docker-library/python/tree/master/3.11), we see that each distro they build for is in its own directory with its own Dockerfile. If we're concerned about code duplication, I can put together a codegen script to populate any code that had been copied, which is the same way that all of the other official docker projects that I have looked at handle things.

extras/docker/alpine3.18/Makefile
kkloberdanz: In my opinion, doing a `--no-cache` build is a more idiomatic docker way to achieve this affect. It has the added benefit that if you are running this for local debugging, which is something that I intended with this approach, then it does not unexpectedly globally delete existing images on your machine.

Also, this suggestion has a bug. Suppose I'm currently running this image in a terminal, then at the same time I attempt to run a build for this image in the background, which is a workflow that I commonly find myself in. The call to delete the image will fail because the container using the image is currently running, for example see the error output below:
```
$ docker rmi mongo-cxx-driver:3.8.0-bookworm
Error response from daemon: conflict: unable to remove repository reference "mongo-cxx-driver:3.8.0-bookworm" (must force) - container 51d14d38441d is using its referenced image ce2044a5abb0
```
If I do a `--force rmi`, then it does not delete the image, but rather untags it, which means that if I run a build, then it will use the cached image again, which is not what is desired for a clean build.

If the name `cleanbuild` is confusing, then I can rename it to `nocachebuild`.

kkloberdanz: This also would cause a race condition in CI if there are two builds happening on the same evergreen host. The two builds would share a docker daemon and encounter the same bug as described above if one of them is running the test at the same time that the other is building the image.

extras/docker/alpine3.18/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+RUN git clone https://github.com/mongodb/libmongocrypt.git --branch ${MONGOCRYPT_BRANCH} && \
+    cd libmongocrypt && \
+    mkdir cmake-build && \
+    cd cmake-build && \
+    cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/install && \
+    make -j $(nproc) && \
+    make install -j $(nproc)
+
+RUN wget https://github.com/mongodb/mongo-c-driver/releases/download/${MONGOC_VERSION}/mongo-c-driver-${MONGOC_VERSION}.tar.gz && \
```
kevinAlbs, line 25: As of CDRIVER-4640, there is no longer a source archive produced in the C driver.

Suggest using the GitHub produced source archive, so this command will work with the eventual C driver 1.25.0 release and newer.
```suggestion
RUN wget https://github.com/mongodb/mongo-c-driver/archive/refs/tags/${MONGOC_VERSION}.tar.gz && \
```

If this change is applied, also add `BUILD_VERSION` in the cmake configuration command `-D BUILD_VERSION=${MONGOC_VERSION}`. There is no VERSION_CURRENT file present in the GitHub produced source archive. Without specifying `BUILD_VERSION` I expect the calculated version will be 0.0.0.

Suggestion also applies to other Dockerfiles.

kkloberdanz: I've updated all of the dependencies (i.e., mongoc, mongocxx, and libmongocrypt) to use this archive path.

kevinAlbs (approved):
Nicely done. LGTM

.mci.yml
```diff
@@ -1176,6 +1192,13 @@ tasks:
         - func: "run_kms_servers"
         - func: "test"
 
+    - name: docker_build
+      run_on: &docker-distros
+        - ubuntu2004
```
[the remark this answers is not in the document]
kkloberdanz: My mistake, good eye catching that 😉

etc/make_release.py
```diff
@@ -224,6 +227,65 @@ def release(jira_creds_file,
...
+        "./extras/docker/bookworm/Dockerfile",
+        "./extras/docker/jammy/Dockerfile",
+        "./extras/docker/redhat-ubi-9.2/Dockerfile",
+    ]
+    makefiles = [
+        "./extras/docker/alpine3.18/Makefile",
+        "./extras/docker/bookworm/Makefile",
+        "./extras/docker/jammy/Makefile",
+        "./extras/docker/redhat-ubi-9.2/Makefile",
+    ]
```
rcsanchez97, line 247: I wonder if something like this might be better:

```
extras_docker = pathlib.Path("./extras/docker")
dockerfiles = extras_docker.rglob("Dockerfile")
makefiles = extras_docker.rglob("Makefile")
```

That should make additions of other distros work automagically, and it should allow the code that follows below (i.e., the `for` loops) to work without further modification. It's also more concise.

extras/docker/alpine3.18/Dockerfile
rcsanchez97: I can see both sides of the point here. If these were meant purely for internal consumption, I would lean towards Colby's suggestion. However, with these being intended for public consumption via a public repo, I think that following the established pattern is more useful to potential uses.

rcsanchez97 (approved):
LGTM.

I don't feel especially strongly about the pathlib suggestion, so I am approving. If you decide to incorporate that change, I am fine without needing an additional review.

etc/make_release.py
kkloberdanz: I like it, I pushed this change.

vector-of-bool (approved):

PR #1013, opened 2023-08-24
Updates to calc_release_version.py

vector-of-bool:
This PR brings over the changes from the C driver for calc_release_version.py. No CMake code is updated in this PR.

### DIFF
--- a/etc/calc_release_version.py
+++ b/etc/calc_release_version.py
@@ -219,7 +219,7 @@ def get_branch_tags(active_branch_name):  # type: (str) -> list[str]
             debug('Got tags: {0}'.format(got))
             return got
     # No tags
-    print('(No tags found for this branch)')
+    debug('(No tags found for this branch)')
     return []
 
 
@@ -232,9 +232,9 @@ def iter_tag_lines():
     output = check_output(['git', 'tag', '--list', '--format=%(*objectname)|%(objectname)|%(refname:strip=2)'])
     lines = output.splitlines()
     for l in lines:
-        obj, tag = l.split('|', 1)
+        obj, tagobj, tag = l.split('|', 2)
         if re.match(r'r\d+\.\d+', tag):
-            yield obj, tag
+            yield obj, tagobj, tag
 
 
 def get_object_tags():  # type: () -> dict[str, list[str]]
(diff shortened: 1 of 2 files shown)


etc/calc_release_version.py
```diff
@@ -152,77 +180,104 @@ def get_next_minor(prerelease_marker):
...
+    debug('Getting tags for branch {0}'.format(active_branch_name))
     if active_branch_name == 'master':
         raise Exception('this method is not meant to be called while on "master"')
-    tags = ''
+
     release_branch_match = RELEASE_BRANCH_RE.match(active_branch_name)
     if release_branch_match:
         # This is a release branch, so look for tags only on this branch
-        tag_glob = 'r' + release_branch_match.group('vermaj') + '.' \
+        tag_glob = 'r'+  release_branch_match.group('vermaj') + '.' \
```
rcsanchez97, line 207:
```suggestion
        tag_glob = 'r' + release_branch_match.group('vermaj') + '.' \
```
I'm pretty sure that this was an inadvertent change.

etc/calc_release_version.py
```diff
@@ -152,77 +180,104 @@ def get_next_minor(prerelease_marker):
...
+    return ret
+
+
+def process_and_sort_tags(tags):  # type: (list[str]) -> list[str]
     """
     Given a string (as returned from get_branch_tags), return a sorted list of
     zero or more tags (sorted based on the Version comparison) which meet
     the following criteria:
-        - a final release tag (i.e., r3.x.y without any pre-release suffix)
+        - a final release tag (i.e., 1.x.y without any pre-release suffix)
```
rcsanchez97, line 258:
```suggestion
        - a final release tag (i.e., r3.x.y without any pre-release suffix)
```

etc/calc_release_version.py
```diff
@@ -152,77 +180,104 @@ def get_next_minor(prerelease_marker):
...
         - a pre-release tag which is not superseded by a release tag (i.e.,
-          r3.x.y-preX iff r3.x.y does not already exist)
+          1.x.y-preX iff 1.x.y does not already exist)
```
rcsanchez97, line 260:
```suggestion
          r3.x.y-preX iff r3.x.y does not already exist)
```
It would be good to preserve the C++ driver-style tag names in the comments

rcsanchez97 (approved):
Looks good, with a few typo fix suggestions.

etc/calc_release_version.py
vector-of-bool: I copy-pasted the script over from C driver and had to fix it up to work with the C++ style branches and tags, so some things were missed.

---

Aside: In `iter_tag_lines`, I discovered that some C++ repo tags are pointing to the commits directly, while others are pointing to separate "tag objects." This is a Git "feature" (quirk?) I was unaware of. The C driver uses exclusively discrete tag objects rather than tags glued directly to commits, so it took me took a while to discover why it wasn't working in this repository. I have no idea how/when this distinction occurs. It appears that only the `debian/*` tags have separate tag objects in the C++ repo.

PR #1015, opened 2023-08-30
CXX-2723 Explicitly document throwing exception from APM callback is undefined behavior

eramongodb:
Resolves CXX-2723. Verified by [this patch](https://spruce.mongodb.com/version/64ef67353627e0da95a52a06/).

Adds explicit documentation in the form of a warning that throwing an exception from an APM callback is undefined behavior. This requirement has always existed but was not properly documented.

As a QoI improvement, despite (now) being explicitly documented as UB, this PR proposes adding `noexcept` specifiers to the internal APM callback functions (the actual callback that invokes the registered user-provided callback) to reduce the scope of runtime errors should this requirement be violated, as control flow should not have returned from libmongoc via an exception to begin with (the program is already in an invalid state at this point). Any user code that currently violates the nothrow requirement will now terminate via `noexcept` rather than be allowed to continue with potentially unbounded libmongoc runtime errors. If this is considered too excessive (consider: [our own test cases](https://github.com/mongodb/mongo-cxx-driver/blob/eae761bb770b7a25101a5fd8a44471471cc8e222/src/mongocxx/test/sdam-monitoring.cpp#L71) violate this requirement in failure conditions, as Catch2 uses exceptions to report test assertion failures), we can omit the addition of `noexcept` specifiers.

As a drive-by improvement, ran `doxygen -u` to update the Doxyfile to address obsolete tag warnings (`TCL_SUBST`, `COLS_IN_ALPHA_INDEX`, `PERL_PATH`, `MSCGEN_PATH`).


vector-of-bool (approved):
LGTM

---

re: Catch2 issues: Only `REQUIRE` throws exceptions, and looking at the file I would guess that it should be using `CHECK()` or `CHECKED_IF()` instead of `REQUIRE` for *every condition*.

src/mongocxx/options/apm.hpp
```diff
@@ -44,6 +44,9 @@ class MONGOCXX_API apm {
     /// Set the command started monitoring callback. The callback takes a reference to a
     /// command_started_event which will only contain valid data for the duration of the callback.
     ///
+    /// @warning
+    ///   If the callback throws an exception, the behavior is undefined.
```
kevinAlbs, line 48:
```suggestion
    ///   Callback is expected not to throw. If the callback throws an exception, `std::terminate` is invoked.
```

Suggest updating comments in this file to note throwing an exception is defined to terminate due to `noexcept` on the internal APM helper.

src/mongocxx/options/private/apm.hh
```diff
@@ -26,73 +26,73 @@ namespace options {
...
     mongocxx::events::server_changed_event e(static_cast<const void*>(event));
     auto context = static_cast<apm*>(libmongoc::apm_server_changed_get_context(event));
     context->server_changed()(e);
 }
 
-static void server_opening(const mongoc_apm_server_opening_t* event) {
+static void server_opening(const mongoc_apm_server_opening_t* event) noexcept {
     mongocxx::events::server_opening_event e(static_cast<const void*>(event));
     auto context = static_cast<apm*>(libmongoc::apm_server_opening_get_context(event));
     context->server_opening()(e);
```
kevinAlbs, line 62: If applications are currently incorrectly throwing exceptions, I expect it may be difficult to diagnose a newly encountered terminate.

Consider catching the exception before terminating to print a message. That may help users diagnose the cause of the terminate:

```suggestion
    try {
        context->server_opening()(e);
    } catch (const std::exception& e) {
        std::cerr << "Unexpected exception thrown from `server_opening` callback: " << e.what();
        std::terminate();
    }
```

kevinAlbs (commented):
> If this is considered too excessive (consider: [our own test cases](https://github.com/mongodb/mongo-cxx-driver/blob/eae761bb770b7a25101a5fd8a44471471cc8e222/src/mongocxx/test/sdam-monitoring.cpp#L71) violate this requirement in failure conditions, as Catch2 uses exceptions to report test assertion failures), we can omit the addition of `noexcept` specifiers.

I prefer the well defined terminate to undefined behavior.

src/mongocxx/options/apm.hpp
eramongodb: For consistency with past/current behavior and to preserve freedom of implementation, I would prefer to leave it publically documented as "undefined behavior" so we do not explicitly guarantee the exact manner in which we will handle (or don't) this scenario for QoI purposes.

src/mongocxx/options/private/apm.hh
```diff
@@ -37,7 +37,7 @@ static void exception_guard(const char* source, Fn fn) noexcept {
     } catch (...) {
         std::cerr << "fatal error: APM callback " << source << " exited via an exception"
                   << std::endl;
-        std::terminate();
+        throw;
```
kevinAlbs, line 40: Revert to using `std::terminate()`? The `throw` results in a warning causing Evergreen task failures:

```
[2023/08/31 20:59:06.981] /data/mci/a62ad8c4c1a22cdebf42866d38f8d790/mongo-cxx-driver/src/mongocxx/../mongocxx/options/private/apm.hh:40:9: error: 'throw' will always call 'terminate' [-Werror=terminate]
[2023/08/31 20:59:06.981]    40 |         throw;
```

eramongodb: The intent was to use `noexcept` to invoke `std::terminate()` to take advantage of implicit code paths that might not apply to an explicit call to `std::terminate()`, but I guess that wasn't sufficiently motivated.

kevinAlbs (approved):

PR #1016, opened 2023-08-30
CXX-2736 document data compression

rcsanchez97:
Before this PR is merged, mongodb/mongo-c-driver#1392 should be merged and the location of [the page](https://mongoc.org/libmongoc/current/data-compression.html) linked in this PR should be confirmed.


kevinAlbs (approved):
> and the location of [the page](https://mongoc.org/libmongoc/current/data-compression.html) linked in this PR should be confirmed.

Note: mongoc.org is updated on releases. The data-compression page may not appear until the next release containing the page (1.25.0, or 1.24.x if the commit is backported). I think it is OK to merge now. I expect the next C driver release will occur before the next C++ driver release.

PR #1019, opened 2023-09-11
CXX-2749 Use FetchContent and find_package() to obtain mnmlstc/core

eramongodb:
Resolves CXX-2749. Verified by [this patch](https://spruce.mongodb.com/version/64ff7612306615063d4d8686). (Note: Debian packaging tasks are failing due to patch-specific code paths; expected to pass on waterfall once merged).

This PR proposes using FetchContent instead of ExternalProject to acquire the downloaded mnmlstc/core library when mnmlstc/core is selected as the polyfill library. This moves the download+configure+build of mnmlstc/core from the build step to the configure step for the CXX Driver. Additionally, this PR removes the patching of include directives in mnmlstc/core header files, opting to add the `third_party` include path to targets/flags instead. The use of the CMake FetchContent module requires bumping the minimum CMake version to 3.11.

These changes were motivated by the issue of mnmlstc/core files being prematurely installed during the CXX Driver CMake build step due to the ExternalProject [custom fix-includes step](https://github.com/mongodb/mongo-cxx-driver/blob/38e6c3ca3609618186110d82d971cf025251865a/src/bsoncxx/third_party/CMakeLists.txt#L34) depending on the install step. This PR proposes instead installing mnmlstc/core to a local install directory (in the CXX Driver binary directory) and importing CMake targets via `find_package()` to maintain isolation of CMake project configurations. The local install directory is added to the CXX Driver's installation rules via `install(DIRECTORY)`. (Note: mnmlstc/core's `share` files are currently excluded, but can be added back in if necessary.)

The result is that the installed CXX Driver files look identical to the target branch, but with the following exceptions:

- Include directives in mnmlstc/core headers no longer have the `bsoncxx/third_party/mnmlstc/` prefix (unmodified).
- The `${CMAKE_INSTALL_INCLUDEDIR}/bsoncxx/v_noabi/bsoncxx/third_party/mnmlstc` directory is now added to the set of include directories for the bsoncxx library via CMake target properties, CMake variables, and pkg-config cflags as appropriate.
- Debian packaging scripts no longer require patches to support/remove the mnmlstc/core include directive edits.

As a drive-by improvement, `build_example_projects.sh` now print the name of the example project on failure so it is clear which CMake project configuration+build+test attempt caused the failure.


src/bsoncxx/third_party/CMakeLists.txt
```diff
@@ -12,30 +12,97 @@
...
+           COMMAND
+              "${CMAKE_COMMAND}"
+                 "-S" "${core-src}"
+                 "-B" "${core-build}"
+                 "-DCMAKE_INSTALL_PREFIX=${core-install}"
+                 "-DBUILD_TESTING=OFF"
+                 "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+                 "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
+           RESULT_VARIABLE retval
+        )
```
kevinAlbs, line 63: The `-S` option was added in [CMake 3.13](https://cmake.org/cmake/help/latest/release/3.13.html):

> The [cmake(1)](https://cmake.org/cmake/help/latest/manual/cmake.1.html#manual:cmake(1)) command gained the -S <source_dir> command line option

Suggest using command compatible with minimum supported CMake version 3.11:

```cmake
        file(MAKE_DIRECTORY "${core-build}")
        execute_process (
           WORKING_DIRECTORY "${core-build}"
           COMMAND
              "${CMAKE_COMMAND}"
                 "-DCMAKE_INSTALL_PREFIX=${core-install}"
                 "-DBUILD_TESTING=OFF"
                 "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
                 "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
                 "${core-src}"
           RESULT_VARIABLE retval
        )
```

src/bsoncxx/third_party/CMakeLists.txt
```diff
@@ -12,30 +12,97 @@
...
+        message (STATUS "Building mnmlstc/core... done.")
+
+        message (STATUS "Installing mnmlstc/core...")
+        execute_process (
+           COMMAND
+              "${CMAKE_COMMAND}"
+                 --install "${core-build}"
+                 --config "${CMAKE_BUILD_TYPE}"
+           RESULT_VARIABLE retval
+        )
```
kevinAlbs, line 89: The `--install` option was added in [CMake 3.15](https://cmake.org/cmake/help/latest/release/3.15.html):

> The [cmake(1)](https://cmake.org/cmake/help/latest/manual/cmake.1.html#manual:cmake(1)) command gained a new --install option.

Suggest using command compatible with minimum supported CMake version 3.11:

```cmake
        execute_process (
           COMMAND
              "${CMAKE_COMMAND}"
                 --build "${core-build}"
                 --target install
                 --config "${CMAKE_BUILD_TYPE}"
           RESULT_VARIABLE retval
        )
```

kevinAlbs (commented):
Fix is much appreciated. Suggested changes for supporting CMake 3.11.

rcsanchez97 (approved):
I concur with Kevin's comments, but otherwise LGTM.

eramongodb:
Raised the bumped-to minimum CMake version to 3.15 for consistency with the C Driver (already required for the auto-download configuration). This addresses both issues raised concerning use of `-S` and `--install`. Easy-to-spot pre-3.15 CMake code paths have been removed for modernization.

src/bsoncxx/third_party/CMakeLists.txt
kevinAlbs: Resolving since minimum CMake version is bumped to to 3.15.

src/bsoncxx/CMakeLists.txt
```diff
@@ -73,6 +73,24 @@ message ("bsoncxx version: ${BSONCXX_VERSION}")
...
 set(BSONCXX_HEADER_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/bsoncxx/${BSONCXX_INLINE_NAMESPACE}" CACHE INTERNAL "")
 
+if (BSONCXX_POLY_USE_MNMLSTC AND NOT BSONCXX_POLY_USE_SYSTEM_MNMLSTC)
+    # For deprecated CMake package config files.
+    set(BSONCXX_POLY_MNMLSTC_DEPRECATED_INCLUDE_DIRS
+        "list(PREPEND LIBBSONCXX_INCLUDE_DIRS \"\${PACKAGE_PREFIX_DIR}/include/bsoncxx/v_noabi/bsoncxx/third_party/mnmlstc\")"
+    )
+    set(BSONCXX_POLY_MNMLSTC_DEPRECATED_STATIC_INCLUDE_DIRS
+        "list(PREPEND LIBBSONCXX_STATIC_INCLUDE_DIRS \"\${PACKAGE_PREFIX_DIR}/include/bsoncxx/v_noabi/bsoncxx/third_party/mnmlstc\")"
+    )
```
vector-of-bool, line 83: Use bracket-strings to avoid the need to escape specials:
```suggestion
    set(BSONCXX_POLY_MNMLSTC_DEPRECATED_INCLUDE_DIRS
        [[list(PREPEND LIBBSONCXX_INCLUDE_DIRS "${PACKAGE_PREFIX_DIR}/include/bsoncxx/v_noabi/bsoncxx/third_party/mnmlstc")]]
    )
    set(BSONCXX_POLY_MNMLSTC_DEPRECATED_STATIC_INCLUDE_DIRS
        [[list(PREPEND LIBBSONCXX_STATIC_INCLUDE_DIRS "${PACKAGE_PREFIX_DIR}/include/bsoncxx/v_noabi/bsoncxx/third_party/mnmlstc")]]
    )
```

vector-of-bool (approved):
Sorry to delay. A very welcome upgrade! One small comment that isn't important.

PR #1020, opened 2023-09-12
CXX-2751 Fix reference to VERSION_CURRENT file in distribution list

eramongodb:
Addresses CXX-2751. Verified by [this patch](https://spruce.mongodb.com/version/6500c1582a60ede1f0422073/).

Ensures `build/VERSION_CURRENT` is not unconditionally included in the distribution list when it may not actually exist (`BUILD_VERSION` set manually or `VERSION_CURRENT` found in the binary directory). This addresses missing rule issues with `dist` target dependency requirements for some CMake generators (i.e. Ninja).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,11 +222,13 @@ set(CMAKE_MODULE_PATH
 include(GNUInstallDirs)
 include (ParseVersion)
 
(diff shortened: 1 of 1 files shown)


rcsanchez97 (approved):
LGTM

kevinAlbs (approved):

PR #1021, opened 2023-09-13
CXX-2754 Add MONGOCXX_COMPILER_* macros to push-pop headers

eramongodb:
Resolves CXX-2754. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/915. CXX-2284 added `MONGOCXX_COMPILER_ID` and `MONGOCXX_COMPILER_VERSION` to [config.hh.in](https://github.com/mongodb/mongo-cxx-driver/commit/2c1c5489fa34233c84ebca479db6b127c0fc8424#diff-42db76007ba7a5f030d8fff16fc8dafd1869a4b2dd03703bcd027d9a8e5af666) but did not add corresponding push-pop guards to prelude.hh and postlude.hh.


kevinAlbs (approved):
Thank you for the fix!

PR #1041, opened 2023-10-19
CXX-2753 Move test files into test subdirectory

eramongodb:
This PR is a followup to https://github.com/mongodb/mongo-cxx-driver/pull/1037 and the last expected PR related to CXX-2753. Verified by [this patch](https://spruce.mongodb.com/version/6531483861837dc29434ba48).

As with the prior PRs, this PR does **NOT** affect binary or source compatibility. It only moves around test files and modifies CMake config and test code accordingly.

Specifically, the bsoncxx `catch.hh` and `to_string.hh` headers formerly located under `lib/bsoncxx/v_noabi/bsoncxx/test_util` is moved into the `test` subdirectory. These headers are only used by test code and are not required by the bsoncxx library itself. The mongocxx `client_helpers` component formerly under `lib/mongocxx/v_noabi/mongocxx/test_util` is moved into the `test` subdirectory for the same reason. This is also supported by the fact that all code that needed to be updated with `test_util -> test` are already under the `test` subdirectories for each project.

To support this relocation, the `${CMAKE_CURRENT_SOURCE_DIR}/../..` directory (which corresponds to `<repo-root>/src`) is added to the include directories of test targets. This is to allow inclusion of test headers via `<bsoncxx/test/...>` and `<third_party/...>`. This was seen as preferable to simply `<header.(hpp|hh)>` without any prefix. This technically also permits `<bsoncxx/include/...>`, `<bsoncxx/lib/...>`, and etc. in test code, but this is not an intended or supported pattern and should be rejected in future code reviews.


src/mongocxx/test/CMakeLists.txt
```diff
@@ -40,7 +37,7 @@ set(spec_test_common
     spec/util.cpp
 )
 
-set(client_helpers_cpp ${PROJECT_SOURCE_DIR}/lib/mongocxx/v_noabi/mongocxx/test_util/client_helpers.cpp)
+set(client_helpers_cpp client_helpers.cpp)
```
kevinAlbs, line 40: Suggest removing the `client_helpers_cpp` variable and replace references with `client_helpers.cpp`.

kevinAlbs (approved):
LGTM. Suggested removal of `client_helpers_cpp` variable.

PR #1042, opened 2023-10-19
[CXX-2625] More type traits + invoke()

vector-of-bool:
(CXX-2625) This changeset adds many additional type traits from C++14 and beyond, especially for easier "SFINAE trickery". Many are C++14/17 traits, some are from library fundamentals, and some are neither.

This changeset does not fully replace any existing dependencies, but will make it easier to do so going forward, as well as cleaning up some of our own metaprogramming trickery.


eramongodb:
Some type traits in this PR uses alias templates, e.g.:
```cpp
template <typename Type, typename... Traits>
using requires_t = enable_if_t<conjunction<Traits...>::value, Type>;
```
Additionally, some constraints are expressed as return types (the traditional `std::enable_if` pattern):
```cpp
template <typename T>
bsoncxx::stdx::requires_t<void, bsoncxx::stdx::is_alike<T, int>> //
only_int(T) {} // Example A
```

Instead, I suggest (1) going all-in on defining type traits using named structs whenever possible (including the implementation of `_t` variants) instead of alias templates, as is already being done for some type traits (e.g. `bsoncxx::stdx::conjunction`):
```cpp
template <typename Type, typename... Traits>
struct requires : std::enable_if<conjunction<Traits...>::value, Type> {};

template <typename Type, typename... Traits>
using requires_t = typename requires<Type, Traits...>::type;
```
And (2) moving the constraints into the template parameter list instead of the return type (as is already being done for some ctors, e.g. `bsoncxx::document::value::value()`):
```cpp
template <typename T, bsoncxx::stdx::requires_t<void, bsoncxx::stdx::is_alike<T, int>> = 0>
void only_int(T) {} // Example B
```

Both suggestions are to improve the quality of error messages. The currently proposed changes emit the following error message on constraint failure for the `only_int` example (note: no meaningful information about the failed constraint is visible in the message):
```
error: no matching function for call to 'only_int'
    only_int(0.0);
    ^~~~~~~~
note: candidate template ignored: substitution failure [with T = double]: no type named 'type' in 'std::enable_if<false>'
only_int(T) {} // Example A
^
```

Suggestion 1 ensures the named type trait is embedded in the error message, rather than its obscure substituted implementation details (suggestion applied to both `requires` and `is_alike` for this example):
```
error: no matching function for call to 'only_int'
    only_int(0.0);
    ^~~~~~~~
note: candidate template ignored: substitution failure [with T = double]: no type named 'type' in 'bsoncxx::stdx::requires<void, bsoncxx::stdx::is_alike<double, int>>'
only_int(T) {} // Example A
^
```

Suggestion 2 avoids the noisy mess that obscures the return type in the function declaration (+ the noisy repetition of the constraint when the return type happens to be on the same line as the function name), while still preserving clean SFINAE semantics that is consistent across all functions (including ctors), and also being friendlier to ClangFormat even for complex constraints (no `//`-injection workarounds required):
```
error: no matching function for call to 'only_int'
    only_int(0.0);
    ^~~~~~~~
note: candidate template ignored: substitution failure [with T = double]: no type named 'type' in 'bsoncxx::stdx::requires<void, bsoncxx::stdx::is_alike<double, int>>'
void only_int(T) {} // Example B
     ^
```

Additionally, a suggestion 3: despite the convenience of `requires_t`, I am partial to using `typename std::enable_if<cond..., type>::type` despite its verbosity due to the quality of error messages by compilers that special-case substition failure messages for `std::enable_if`. e.g. given (suggestions 1 and 2 are applied here; note the clean ClangFormat results):
```cpp
template <typename T,
          bsoncxx::stdx::requires_t<void,
                                    bsoncxx::stdx::is_alike<T, double>,
                                    bsoncxx::stdx::is_alike<T, int>> = 0>
void impossible(T) {}
```
The error message looks as follows (the entire constraint is emitted as-is):
```
error: no matching function for call to 'impossible'
    impossible(0.0);
    ^~~~~~~~~~
note: candidate template ignored: substitution failure [with T = double]: no type named 'type' in 'bsoncxx::stdx::requires<void, bsoncxx::stdx::is_alike<double, double>, bsoncxx::stdx::is_alike<double, int>>'
void impossible(T) {}
     ^
```
Whereas if using explicit `std::enable_if`:
```cpp
template <typename T,
          typename std::enable_if<bsoncxx::stdx::is_alike<T, double>::value &&
                                  bsoncxx::stdx::is_alike<T, int>::value>::type = 0>
void impossible(T) {}
```
The error message looks as follows (more meaningful and specific reason for constraint failure):
```
error: no matching function for call to 'impossible'
    impossible(0.0);
    ^~~~~~~~~~
note: candidate template ignored: requirement 'bsoncxx::stdx::is_alike<double, int>::value' was not satisfied [with T = double]
void impossible(T) {}
     ^
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -0,0 +1,368 @@
...
+
+#include <type_traits>
+
+#include <bsoncxx/config/prelude.hpp>
+
+namespace bsoncxx {
+
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace stdx {
+inline namespace type_traits {
```
kevinAlbs, line 11: Providing public API for type traits may add maintenance cost for API that is less related to the goals of bsoncxx library. I suggest discouraging consumers from using directly:

```suggestion
// API in the `type_traits_private` namespace is unstable. Do not use directly.
namespace type_traits_private {
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/view.hpp
```diff
@@ -306,32 +307,33 @@ class BSONCXX_API view {
...
 
-// sfinae in the bool return to avoid competing with the value == value
-// operators.
 template <typename T>
-using not_view = typename std::enable_if<
-    std::is_constructible<bson_value::view, T>::value &&
-        !std::is_same<typename std::decay<T>::type, bson_value::view>::value &&
-        !std::is_same<typename std::decay<T>::type, bson_value::value>::value,
-    bool>::type;
+using is_bson_view_compatible =
```
kevinAlbs, line 311:
```suggestion
using not_view =
```

Suggest renaming back to `not_view` to preserve the existing public API.

vector-of-bool: Restored a `not_view` alias.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
vector-of-bool: I tweaked the namespace layout to be `bsoncxx::_traits`, to keep it easy to type but also "obviously not public". Bike-shedding over the namespace name is TBD.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * @brief Holds a list of types.
+ *
+ * This template is never defined, so cannot be used in contexts that require a complete type.
+ */
+template <typename...>
+struct mp_list;
+
+/// ## Implementation of the C++11 detection idiom
```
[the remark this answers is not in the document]
vector-of-bool: (See other namespace comment)

vector-of-bool:
@eramongodb The avoidance of instantiating class templates is intentional since it explodes can explode compilation time. I was also unsatisfied with the opaque diagnostic messages on constraint failure, so I've added some changes that will produce better error messages on almost all compilers, and often better than the compilers' special-cases for enable_if.

Most oddities in the placement of the SFINAE constraints are specifically to appease VS2015, which has horribly broken SFINAE behavior. When we drop VS2015 support, we will be safe to relocate the constraints to the template parameter lists in (almost?) all contexts (without needing to add the `typename=void` disambiguators).

src/bsoncxx/test/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ else()
     bsoncxx_add_library(bsoncxx_testing "bsoncxx-testing" STATIC)
 endif()
 
+file(GLOB src_bsoncxx_test_DIST_cpps CONFIGURE_DEPENDS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *.cpp)
```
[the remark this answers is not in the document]
kevinAlbs: Not yet addressed.

kevinAlbs (approved):

eramongodb (changes requested):
I find it difficult to review what traits are implemented in such a manner to avoid VS 2015 compatibility issues vs. what are not, but the intent/context is understood.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -0,0 +1,428 @@
+#pragma once
+
+#include <type_traits>
+
+#include <bsoncxx/config/prelude.hpp>
+
+namespace bsoncxx {
+
+BSONCXX_INLINE_NAMESPACE_BEGIN
+namespace _traits {
```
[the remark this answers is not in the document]
eramongodb: To clarify/elaborate, this recommendation is to ensure _all_ the polyfill type trait declarations we use are declared in the `details` namespace to document+ensure we do not become beholden to maintaining the stability of our type trait implementations as part of the public API/ABI (they are just a means to an end; we do not want to be a type traits polyfill library provider).

Ideally this should extend to other polyfill entities as well, but those (optional and string_view in particular) are too baked into the API/ABI atm to allow for this treatment.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -0,0 +1,428 @@
...
+std::true_type is_detected_f(mp_list<Args...>*);
+
+// Failure case for is_detected. Because this function takes an elipsis, this is
+// less preferred than the above overload that accepts a pointer type directly.
+template <bsoncxx_ttparam Oper, typename... Args>
+std::false_type is_detected_f(...);
+
+// Provides the detected_or impl
+template <bool IsDetected>
+struct detection;
```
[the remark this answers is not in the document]
vector-of-bool: The implementation in N4502 is only expository. The design of the impl here is based on more modern techniques to aid in diagnostics and especially compile times (i.e. 1 class template instantiation per `is_detected`, and _zero_ template instantiations for `detected_or`/`detected_t`)

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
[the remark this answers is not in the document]
vector-of-bool: I simplified/flattened the namespace structure further. I wanted to avoid heavy nesting due to diagnostics sizes and name resolution issues. Having a flat `detail` is sufficient, since we control the namespace and won't collide with ourself.

eramongodb (approved):

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -0,0 +1,430 @@
+#pragma once
+
+#include <type_traits>
+
+#include <bsoncxx/config/prelude.hpp>
+
+namespace bsoncxx {
+
+BSONCXX_INLINE_NAMESPACE_BEGIN
```
eramongodb, line 9:
```suggestion
inline namespace v_noabi {
```
Consistency with changes in https://github.com/mongodb/mongo-cxx-driver/pull/1039.

vector-of-bool: Could that macro be removed completely, or is it too much part of the public API?

PR #1048, opened 2023-10-25
Arm64 docker build

kkloberdanz:
CXX-2766

In order to better support ARM hosts, such as recent Apple computers that use ARM chips, we will provide both AMD64 and ARM64 images on Docker Hub.

libmongocrypt does not currently build on Alpine on ARM64, so the Alpine build is being temporarily removed. The build is fixed with the following [PR](https://github.com/mongodb/libmongocrypt/pull/706) and the Alpine build should be re-enabled once a new libmongocrypt release is done that has that PR.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -840,6 +840,9 @@ functions:
             set -o errexit
             set -o pipefail
             set -x
+            # TODO: Re-enable Alpine test once MONGOCRYPT-601 is released.
+            # echo "Building Alpine Docker image"
+            # make -C extras/docker/alpine3.18 nocachebuild test
             echo "Building Debian Docker image"
             make -C extras/docker/bookworm nocachebuild test
             echo "Building Red Hat UBI Docker image"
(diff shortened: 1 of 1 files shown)


.mci.yml
```diff
@@ -840,8 +840,6 @@ functions:
             set -o errexit
             set -o pipefail
             set -x
-            echo "Building Alpine Docker image"
-            make -C extras/docker/alpine3.18 nocachebuild test
             echo "Building Debian Docker image"
```
kevinAlbs, line 843: Suggest commenting out the alpine build with a comment referencing MONGOCRYPT-601:

```yml
            # TODO: Re-enable Alpine test once MONGOCRYPT-601 is released.
            # echo "Building Alpine Docker image"
            # make -C extras/docker/alpine3.18 nocachebuild test
```

.mci.yml
```diff
@@ -1408,6 +1413,11 @@ buildvariants:
       tasks:
           - name: docker_build
```
kevinAlbs, line 1414: Suggest combining the "Docker builder" variants:

```yml
    - name: docker-builder
      display_name: "Docker builder"
      tasks:
          - name: docker_build
          - name: docker_build_arm
```

.mci.yml
```diff
@@ -1236,6 +1234,13 @@ tasks:
         - func: "run_kms_servers"
         - func: "test"
 
+    - name: docker_build_arm
+      run_on: &docker-distros
```
kevinAlbs, line 1238:
```suggestion
      run_on:
```

Suggest removing unused YAML anchor `&docker-distros` here and in the `docker_build` task.

kevinAlbs (approved):

eramongodb (approved):

PR #1049, opened 2023-10-31
Miscellaneous Improvements to EVG Config

eramongodb:
## Description

This PR aims to make numerous overdue improvements to the EVG config for the CXX Driver. Verified by [this patch](https://spruce.mongodb.com/version/654140060ae6062e7c4c9754).

Due to volume of diffs, navigating the changes by-commit is recommended.

## EVG Matrix Variants

The "integration task" matrix is currently defined using "matrix variants" (`axis:` + `matrix_*`), which have long since been not actively supported. Per [documentation](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files/#matrix-variant-definition):

> The matrix syntax is deprecated [...]. **Evergreen is unlikely to do further development on matrix variant definitions.** [...] please do not add new matrix variant definitions.

Instead of using EVG matrix variants, the "integration matrix" variants have been expanded and defined explicitly, identified by the `integration-` prefix in their names. Due to their names being different from the prior config (even though their display names remain unchanged), EVG considers these variants distinct from the prior config and thus does not preserve variant/task history.

YAML "merge keys" (`<<:`) are used to reduce duplication across the "integration tasks" as much as possible. These keys are deliberately limited to the specification of tasks and preserving existing platform-specific behavior. This is part of a greater effort to reduce the complexity of the EVG variant/task definitions by reducing the use of EVG YAML anchors. The only YAML anchors now in use is `mongoc_minimum_required` and `integration_matrix_*`.

These changes also addresses ongoing task failures by resolving CXX-2772 (Ubuntu 18.04 -> Ubuntu 20.04 for 7.0 and Latest ). The "Ubuntu 18.04 Release (MongoDB Latest)" variant is thus obsoleted and removed, reducing the total task count by 8 (235 -> 227). The `lint` task was moved into its own variant.

## YAML Anchors

YAML anchors defined under `variables:` have been almost entirely inlined and moved into related scripts. This is to reduce the complexity of variant/task definitions. These changes include:

- `tar_options`: removed: obsolete (unused?)
- `poly_mnmlstc`: removed: obsolete (unused?)
- `mongoc_version_default`: replaced with `mongoc_version_minimum` (little/no reason for them not to be synced).
- `version_*`: inlined into `mongodb_version:` expansion definitions (no significant benefit to their use).
- `example_projects_*`: inlined into the just 5 variants it is used by (asan/ubsan + experimental std).
- `poly_flags`: inlined into the just 4 variants it is used by (MacOS + experimental std).
- `extra_path`: inlined into the "compile" and "test" task definition (specified by platform detection).
- `*_test_params`: inlined into the "test" task definition (specified by `TEST_WITH_(ASAN|UBSAN|VALGRIND)`).
- `*_cmake_flags`: inlined into the "compile" task definition (specified by platform detection, `TEST_WITH_`, and CMake `generator`).

Therefore, the complexity of specifying CMake configuration and compile flags has been moved into the `compile.sh` and `test.sh` scripts, where Bash arrays can be used to much more easily and readably specify the flags required per task and platform (_especially_ the CC/CXX compile flags).

## Compile and Test

These two EVG functions were occupying a significant portion of the EVG config file. Due to their importance and length, their entire definitions have been moved into the `compile.sh` and `test.sh` scripts. This significantly reduces the size of the EVG config file, offsetting the increased line count due to the integration matrix expansion. This also allows applying tools such as formatting and ShellCheck to the compile and task scripts.

To further facilitate maintainability, rather than dumping all expansions via `add_expansions_to_env`, the specific expansions required by individual functions/scripts are explicitly documented via `include_expansions_in_env:` and `env:`. Although a bit verbose, this should help keep track of what expansions are required by what functions/scripts in the future.

## Miscellaneous

- The `-Wno-aligned-new` flag was dropped due to conflicts with Clang tasks + appearing to cause no trouble for existing tasks (not entirely clear what motivated its addition or if it still applies; can revert if necessary).
- Remove redundant `run_on:` in variant definitions. Moved into task definition if able.
- Although auto-downloaded C Driver's CMake config might find ccache, the CXX Driver's CMake config does not due to variable scoping. The CXX Driver CMake config now opts into use of ccache via the `CMAKE_(C|CXX)_COMPILER_LAUNCHER` environment variables.
- The `compiler` expansion intended to select GCC vs. Clang was completely unused (bug since its introduction?), meaning the Clang-specific tasks were not using Clang at all. The new `cc_compiler` and `cxx_compiler` expansions are used to define `CC` and `CXX` environment variables to direct CMake to use the specified compiler(s), fixing the GCC and Clang tasks. Note: these tasks are moved from Ubuntu 22.04 -> Ubuntu 20.04 due to Ubuntu 22.04 apparently not providing a Clang compiler.
- Move towards using `nproc` and `CMAKE_BUILD_PARALLEL_LEVEL` to enable build parallelism at the CMake level (more consistently and better forwarded to underlying generator without depending on compile flags).
- The distcheck command was not properly selecting the polyfill library consistently with the parent build-and-install routines due to CMake config variables not being forwarded properly in `cmake -P` commands. Additionally, the distcheck commands were unnecessarily specific to the GNU Make generator. The required "polyfill flags" are now properly forwarded, and `cmake --build` is used instead of `make`. These changes now allow the distcheck test to be run on Windows, but this addition was deliberately omitted (see: `_RUN_DISTCHECK` in `compile.sh`) to avoid increasing the already-lengthy run time of Windows tasks.


kevinAlbs (approved):
The removal of most YAML anchors, and movement of scripts from YAML to separate files is a much appreciated simplification.

LGTM with minor suggestions.

.evergreen/compile.sh
```diff
@@ -6,64 +6,243 @@
...
 
 set -o errexit
 set -o pipefail
 
-if [ "$BUILD_TYPE" != "Debug" -a "$BUILD_TYPE" != "Release" ]; then
-    echo "$0: expected BUILD_TYPE environment variable to be set to 'Debug' or 'Release'" >&2
-    exit 1
+: "${branch_name:?}"
+: "${build_type:?}"
+: "${distro_id:?}"
```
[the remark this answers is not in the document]
eramongodb: Required by `find-cmake-latest.sh`. Added a comment to make this indirect requirement clearer.

.evergreen/test.sh
```diff
@@ -0,0 +1,364 @@
...
+
+: "${build_type:?}"
+: "${cse_aws_access_key_id:?}"
+: "${cse_aws_secret_access_key:?}"
+: "${cse_azure_client_id:?}"
+: "${cse_azure_client_secret:?}"
+: "${cse_azure_tenant_id:?}"
+: "${cse_gcp_email:?}"
+: "${cse_gcp_privatekey:?}"
+: "${distro_id:?}"
```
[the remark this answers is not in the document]
eramongodb: Required by `find-cmake-latest.sh`. Added a comment to make this indirect requirement clearer.

eramongodb:
Removed the minimum libmongoc variant. Latest changes verified by [this patch](https://spruce.mongodb.com/version/6543fc799ccd4e4f52fe0689).

cmake/make_dist/MakeDistCheck.cmake
```diff
@@ -36,43 +28,68 @@ function (RUN_DIST_CHECK PACKAGE_PREFIX EXT)
...
    # Run make in the build directory
-   if (DEFINED ENV{DISTCHECK_BUILD_OPTS})
-      set (build_opts $ENV{DISTCHECK_BUILD_OPTS})
-   else ()
-      set (build_opts "-j 8")
-   endif ()
    separate_arguments (build_opts)
    execute_process_and_check_result (
-      COMMAND ${MY_CMAKE_COMMAND} ${MAKE_COMMAND} ${build_opts}
+      COMMAND ${CMAKE_COMMAND} --build .
```
vector-of-bool, line 78:
```suggestion
      COMMAND ${CMAKE_COMMAND} --build . --parallel
```

If `--parallel` is given with no argument value, CMake will infer the parallelism level automatically (may be useful in other contexts, but is a useful shorthand in general).

vector-of-bool (approved):

cmake/make_dist/MakeDistCheck.cmake
eramongodb: Confirming, is inferring parallelism via `--parallel` preferable/different to using the value of the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable (which AFAIK should be visible/inherited by the `execute_process()` commands given the changes in this PR)?

eramongodb: Confirmed: `CMAKE_BUILD_PARALLEL_LEVEL` suffices.

PR #1050, opened 2023-11-03
add Debian post build instructions

kkloberdanz:


kevinAlbs (approved):
Much appreciated. LGTM

etc/releasing.md
```diff
@@ -450,6 +452,28 @@ $ DEBEMAIL='my-email@mongodb.com' DEBFULLNAME='FIRSTNAME LASTNAME' dch -r -D exp
...
+After the package has been accepted into Debian, make sure to check if there are
+any bugs or build failures. Debian tests builds on more architectures than we
+officially support and test on in Evergreen, so there are occasionally
+surprising build failures on exotic architectures.
+
+Bug tracking can be found here for the following projects:
+- [libmongocrypt](https://bugs.debian.org/cgi-bin/pkgreport.cgi?src=libmongocrypt)
+- [mongo-c-driver](https://bugs.debian.org/cgi-bin/pkgreport.cgi?src=mongo-c-driver)
+- [mongo-cxx-driver](https://bugs.debian.org/cgi-bin/pkgreport.cgi?src=mongo-cxx-driver)
+
```
rcsanchez97, line 466:
```suggestion
Any Debian bugs that were specifically fixed in the release being made should be noted in the `debian/changelog` entry. It should look something like this:

    `* New upstream release (Closes: #1234567)`
```

rcsanchez97 (approved):
It would be good if we had a specific reminder about annotating bug closures in the changelog. You may need to tweak the formatting, as I didn't look to see how my suggestion would render.

But after that's added, LGTM

etc/releasing.md
kkloberdanz: I like it! I'll find a way to add it in. Thanks!

PR #1053, opened 2023-11-07
Update changelog for 3.9.0

kevinAlbs:
Incorporates changes from https://docs.google.com/document/d/1qh2-7Q3Zrmdp-h88H2uOKAmkS0bxYLV7Fz_FmmVeswU/edit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   for backwards-compatibility).
 - Add API to manage [Atlas Search Indexes](https://www.mongodb.com/docs/atlas/atlas-search/).
 - Automatically download C driver dependency if not provided.
+- Add VERSIONINFO resource to bsoncxx.dll and mongocxx.dll.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,17 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- The C++ driver container image is now available on [Docker hub](https://hub.docker.com/r/mongodb/mongo-cxx-driver).
 - Add CMake option `MONGOCXX_OVERRIDE_DEFAULT_INSTALL_PREFIX` (default is `TRUE`
```
bisht2050, line 13: Suggest to add its purpose. Something like:

Add CMake option `MONGOCXX_OVERRIDE_DEFAULT_INSTALL_PREFIX` (default is `TRUE` for backwards-compatibility). If enabled, `mongocxx` will set a default `CMAKE_INSTALL_PREFIX` if one is not already defined.

bisht2050 (approved):
LGTM with minor comments

CHANGELOG.md
kevinAlbs: I like that idea. Updated.

PR #1054, opened 2023-11-07
Use consistent pattern for type_traits header include directives

eramongodb:
Overlooked during review of https://github.com/mongodb/mongo-cxx-driver/pull/1042. Applies consistent use of `#include <bsoncxx/stdx/type_traits.hpp>` pattern to directives for consistency with other bsoncxx public header include directives.


vector-of-bool (approved):
These look to be directives automatically added by `clangd` and I missed fixing them. I wonder if there's an option to change that...

PR #1055, opened 2023-11-08
CXX-2785 do not apply `readConcern` or `writeConcern` to search index commands

kevinAlbs:
# Summary

Do not apply `readConcern` or `writeConcern` to search index commands

Tests verified with this patch: https://spruce.mongodb.com/version/654b89333e8e86f7a940c6e3

# Background & Motivation
This is a follow-up to https://github.com/mongodb/mongo-cxx-driver/issues/986

Search index commands return an error if `readConcern` or `writeConcern` is included.

Prior to the fix, applying a non-default `readConcern` or `writeConcern` to the `mongocxx::collection`, `mongocxx::database`, or `mongocxx::client` resulted in the `readConcern` or `writeConcern` being included in the search index command.

Prior to the fix, the added test "create and list search indexes with non-default readConcern and writeConcern" fails when run against an Atlas cluster:

```
/Users/kevin.albertson/code/tasks/mongo-cxx-driver-CXX-2697/src/mongocxx/test/search_index_view.cpp:246: FAILED:
  REQUIRE( siv.create_one(name, definition.view()) == "test-search-index-case6" )
due to unexpected exception with message:
  Command does not support writeConcern: generic server error
```

These changes are intended to be included in 3.9.0. The API added in https://github.com/mongodb/mongo-cxx-driver/pull/986 has not yet been released. The error may give a poor user experience.

The new tests are specified in https://github.com/mongodb/specifications/pull/1474.

### DIFF
--- a/data/index-management/searchIndexIgnoresReadWriteConcern.json
+++ b/data/index-management/searchIndexIgnoresReadWriteConcern.json
@@ -1,0 +1,252 @@
+{
+  "description": "search index operations ignore read and write concern",
+  "schemaVersion": "1.4",
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": false,
+        "uriOptions": {
+          "readConcernLevel": "local",
+          "w": 1
+        },
+        "observeEvents": [
+          "commandStartedEvent"
+        ]
+      }
+    },
+    {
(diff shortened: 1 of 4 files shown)


kkloberdanz (approved):
Looks good!

PR #1061, opened 2023-11-10
CXX-2288 Add forward headers for bsoncxx and mongocxx

eramongodb:
## Description

Resolves CXX-2288. Verified by [this patch](https://spruce.mongodb.com/version/654e4c92e3c33169458a6be7).

This PR initially proposes maximizing modularity by providing a forward header per-component_. If this is considered excessive, forward headers can be merged as requested, i.e. merging `mongocxx/events/(.*)-fwd.hpp` into `mongocxx/events/fwd.hpp`.

Requesting reviews keep a close eye on whether `BSONCXX_API` and `MONGOCXX_API` are correctly preserved when moved into forward headers, and that they are not accidentally added to classes that did not have them prior.

## Principles

There are several principles/rules that motivated/directed how changes in this PR were made. These are:

- Any public header `a.hpp` that contains a declaration of a scoped enumeration or class type _in namespace scope_ is given a forward header `a-fwd.hpp` containing a forward declaration of that enumeration or class type.
  - Public headers containing only declarations of deprecated entities are excluded. This only applies to `mongocxx/options/create_collection.hpp`.
- A forward header `a-fwd.hpp` is always included its corresponding `a.hpp`. This is both to physically encode the header dependency and to ensure the definition for class types in `a.hpp` is always consistent with its forward declaration in `a-fwd.hpp`.
- Symbol visibility for class types using `(BSONCXX|MONGOCXX)_API` is always declared in the forward 
header (when applicable). This guarantees consistent symbol visibility regardless whether the class declaration is imported via the forward or normal header. (A followup PR may relocate Doxygen documentation comments for enumeration/class types into the forward header.) The only instances of `(BSONCXX|MONGOCXX)_API` for class declarations in normal headers is now limited to member classes and deprecated classes.
- All friend declarations are fully qualified relative to the global scope to minimize any potential for confusion about what class is being declared as a friend. The addition of forward headers that include these changes are made in separate commits in this PR to make it easier to track the related changes. (Unfortunately, C++ does not provide a mechanism that enforces a friend declaration always refers to an existing declaration, rather than introducing a new declaration.)
- Although using-declarations (`using foo::bar;`), using-directives (`using namespace foo;`), type aliases (`using foo = bar;`), and typedefs in namespace scope could be moved into forward headers, this was avoided to keep the forward headers as simple as possible. Therefore, the forward headers contain no `using` or `typedef` declarations. Headers that _only_ declare a type alias or typedef are not given a forward header due to redundancy, e.g. `bsoncxx/document/view_or_value.hpp`.

Additionally:

- For convenience (and as requested by CXX-2288), a forward header for _all_ bsoncxx/mongocxx entities is provided as `bsoncxx/fwd.hpp` and `mongocxx/fwd.hpp`. These headers simply include all the `(.*)-fwd.hpp` headers.
- Ensured all class declarations have only a _single_ definitive declaration that specifies symbol visibility. This only applied to member classes such as `bsoncxx::document::view::iterator`.
- Reduced potential confusion (both for C++ name lookup and simple readability) when referring to (un/under)qualified names, e.g. `bsoncxx::document` vs. `bsoncxx::builder::document`.

## ClangFormat

The forward headers, identified as either `.../fwd.hpp` or `.../<name>-fwd.hpp`, are prioritized before normal headers. Ideally, system headers should be deprioritized to come after all normal headers (including the macro guard headers), but that change is deferred to avoid increasing the already large PR diff. The header sort is thus the following:

```
#include <system> // System

#include <bsoncxx/document/value-fwd.hpp> // Forward (new!)

#include <bsoncxx/document/view.hpp> // Driver

#include <bsoncxx/config/prelude.hpp> // Macro Guard

// ... (normal header contents)

#include <bsoncxx/config/postlude.hpp> // Macro Guard
```

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
@@ -17,6 +17,7 @@
 #include <cstddef>
 #include <cstdint>
 
+#include <bsoncxx/array/element-fwd.hpp>
 #include <bsoncxx/types/bson_value/view-fwd.hpp>
 
 #include <bsoncxx/document/element.hpp>
@@ -34,7 +35,7 @@ namespace array {
 /// interrogated by calling type() and a specific value can be extracted through
 /// get_X() accessors.
 ///
-class BSONCXX_API element : private document::element {
+class element : private document::element {
    public:
     element();
 
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
@@ -17,18 +17,14 @@
 #include <cstddef>
 #include <cstdint>
 
+#include <bsoncxx/types/bson_value/view-fwd.hpp>
+
 #include <bsoncxx/document/element.hpp>
 
 #include <bsoncxx/config/prelude.hpp>
 
 namespace bsoncxx {
 inline namespace v_noabi {
-namespace types {
-namespace bson_value {
-class view;
-}  // namespace bson_value
-}  // namespace types
-
 namespace array {
 
 ///
@@ -121,7 +117,6 @@ BSONCXX_API bool BSONCXX_CALL operator!=(const types::bson_value::view& v, const
 ///
 
 }  // namespace array
-
 }  // namespace v_noabi
 }  // namespace bsoncxx
 
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
@@ -18,6 +18,7 @@
 #include <cstdint>
 
 #include <bsoncxx/array/element-fwd.hpp>
+#include <bsoncxx/array/view-fwd.hpp>
 #include <bsoncxx/types/bson_value/view-fwd.hpp>
 
 #include <bsoncxx/document/element.hpp>
@@ -77,7 +78,7 @@ class element : private document::element {
     using document::element::raw;
 
    private:
-    friend class view;
+    friend class ::bsoncxx::v_noabi::array::view;
 
     BSONCXX_PRIVATE explicit element(const std::uint8_t* raw,
                                      std::uint32_t length,
--- a/.clang-format
+++ b/.clang-format
@@ -9,12 +9,14 @@ Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 IncludeBlocks: Regroup
 IncludeCategories:
-  - Regex:    'prelude\.(hpp|hh)' # preludes
-    Priority: 3
-  - Regex:    '<[[:alnum:]_.]+>' # system headers
+  - Regex: 'prelude\.(hpp|hh)' # preludes
+    Priority: 4
+  - Regex: '<[[:alnum:]_.]+>' # system headers
     Priority: 1
-  - Regex:    '.*' # driver headers
+  - Regex: '.*(-|\/)fwd\.(hpp|hh)' # forwarding headers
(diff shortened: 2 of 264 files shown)


src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
```diff
@@ -80,7 +78,7 @@ class BSONCXX_API element : private document::element {
     using document::element::raw;
 
    private:
-    friend class view;
+    friend class ::bsoncxx::v_noabi::array::view;
```
vector-of-bool, line 81: > C++ does not provide a mechanism that enforces a friend declaration always refers to an existing declaration, rather than introducing a new declaration

C++11 fixed this by allowing a the friend declaration to name an existing type by omitting the elaborated type tag, e.g.

```suggestion
    friend view;
```

If the name does not name type according to name lookup rules, it generates a hard error (instead of spontaneously generated an incomplete type). Refer to point no. 4 on [cppreference](https://en.cppreference.com/w/cpp/language/friend).

eramongodb: TIL! That is very good to know. Converted all friend class declarations into simple type specifier form to enforce existing declaration behavior.

However, for consistency, I've elected to keep the type specifier fully namespace-qualified to avoid possibility for confusion even with this change. e.g.:

```cpp
namespace mongocxx {
inline namespace v_noabi {
namespace options {

class client_session {
  // ...
private:
  friend client_session; // Intent: mongocxx::v_noabi::client_session
                         // Actual: this class (no warnings either...).
};

} // namespace options
} // namespace v_noabi
} // namespace mongocxx
```

There are several other variations of this potential mistake throughout the codebase due to how often we reuse the same identifier within different namespaces.

Full namespace qualification will also help disambiguate entities declared in different ABI namespaces once changes related to CXX-2745 are implemented (`v_noabi`, `v1`, etc.). If including the root namespace is too excessive (`::bsoncxx::`, `::mongocxx::`), then I can at least remove those.

kevinAlbs (approved):
LGTM with suggestion: update the Headers section of CONTRIBUTING.md to match the new expected header order. I had not considered CONTRIBUTING.md when reviewing related ABI PRs.

> Requesting reviews keep a close eye on whether `BSONCXX_API` and `MONGOCXX_API` are correctly preserved when moved into forward headers, and that they are not accidentally added to classes that did not have them prior.

LGTM. I assume classes like `bsoncxx::v_noabi::bsoncxx::view_or_value` did not, and do not include `BSONCXX_API` due to methods being inlined.

eramongodb:
> I assume classes like `bsoncxx::v_noabi::bsoncxx::view_or_value` did not, and do not include `BSONCXX_API` due to methods being inlined.

Correct. The `view_or_value` class template (and related aliases) do not require export macros due to the related functions being declared `BSONCXX_INLINE` (AFAIK redundantly, given we compile with hidden visibility by default for shared libraries). In other words, the member functions of `view_or_value<View, Value>` are part of the public API but _not_ part of the public ABI, even when `View` and/or `Value` are part of the public ABI. I think (intentionally or not) this is consistent with their intended usage as "forwarding" parameters for view and value types. (Note: the _layout_ of data members of `view_or_value` is still unavoidably part of the ABI due to its use by exported symbols, even if the related functions are not.)

eramongodb:
Identified more instances of (re)declarations of class types in the form of "elaborated type specifiers". Replaced all such instances with references to the already-declared class type instead, to ensure the use of (forward) headers containing the single definitive declaration for said class type. The sole exceptions are the definitions of the static local `category` variables in the `error_category()` and `server_error_category()` functions to disambiguate the class types with the same name as the functions they are used in.

Latest changes verified by [this patch](https://spruce.mongodb.com/version/655676109ccd4e82bcdaa5d4/).

vector-of-bool (approved):

PR #1062, opened 2023-11-16
[CXX-2625] Bring our very own string_view (plus: Some iterators + ranges backports)

vector-of-bool:
Refer: CXX-2625

This changeset replaces our usage of external string_view implementations with a single custom implementation, and includes several smaller supporting library components and general code cleanup. The commits have been carefully organized to represent the iteration process taken while developing this changeset, and *should* be reviewable in isolation, rather than requiring the entire 1.8k line change be reviewed in whole. For reader's benefit, the changes are summarized here as well, in commit order.

# Change Summaries

## Utility macro: `bsoncxx_returns(...)`

This macro is a shorthand for

```c++
noexcept(noexcept(...)) 
  -> decltype(...)
{ return ...; }
static_assert(true, "")
```

(The trailing `static_assert` forces usage to be followed by a semicolon, and prevents clang-format from being confused.)

It is used like this:

```c++
template <typename T>
auto some_func(T x)
  bsoncxx_returns(x.foo());
```

Besides being easier to read and write, this ensures that: The `noexcept`, `-> decltype` and `return` expressions are all equivalent and correct with respect to each other. Placing the return expression in the `noexcept` and `decltype` specifier causes semantically invalid code to introduce a substitution failure context rather than a hard error, which is especially helpful when doing SFINAE trickery.

## ~~`decay_copy` and~~ `rank`

~~The `decay_copy` invocable object is an implementation of the C++23 `auto(...)` expression, previously known as the exposition-only `decay-copy(...)` It decays its argument and invokes the appropriate move/copy constructor thereof in-situ. This is especially useful with `bsoncxx_returns`, in order to eliminate reference-returning operations.~~ Edit: `decay_copy` was removed as YAGNI. May come back someday, if useful.

The `rank<N>` struct template is simply a class template which inherits from `rank<N-1>`, except for `rank<0>`, which inherits from nothing. This is used in later changes to create ordering between ambiguous overloads using tag dispatch. A candidate conversion from `rank<10>` to `rank<2>` is "preferrable" to a conversion from `rank<10>` to `rank<1>`.

## Operator Utilities in `operators.hpp`

The code in `operators.hpp` is not strictly based on and C++ feature and is used to provide simpler implementation of operator overloading in the `string_view` implementation (and eventually the `optional<T>` implementation). Potentially this could be spread to other types which provide overloaded operators to eliminate existing boilerplate.

1. The `struct detail::is_equality_comparable<T, U>` trait template detects whether `const T&` and `const U&` are comparable using the `==` and `!=` operators.
2. `struct detail::equal_to` implements a callable object like `std::equal_to<void>`. If given to arguments `l` and `r` that satisfy `is_equality_comparable`, it returns `l == r`.
3. `class detail::equality_operators` is a mixin base class that provides ADL-only (hidden `friend)` implementations of `operator==(x, y)` and `operator!=(x, y)` if-and-only-if there exists a valid ADL-visible `tag_invoke(detail::equal_to{}, x, y)` implementation.
4. `class detail::strong_ordering` is based on the C++20 `std::strong_ordering` class. This changeset does not introduce the other ordering types, as this is all we need for `string_view`. To simulate C++ `inline` variables, the `[[gnu::weak]]` and `__declspec(selectany)` attributes are used for the `strong_ordering` constants.
5. `compare_three_way` implements the `std::compare_three_way` invocable object from C++20. This will be the first usage of `rank<N>`. Invoking this object as `compare_three_way{}(l, r)` does the first of the following:
  1. If `tag_invoke(compare_three_way{}, l, r)` is valid, returns that value as `strong_ordering`
  2. Otherwise, if `l < r` is valid *and* `l == r` is valid:
    1. If `l < r`, returns `strong_ordering::less`
    2. Otherwise, `l == r` return `strong_ordering::equal`
    3. Otherwise, returns `strong_ordering::greater`.
  3. Otherwise, the `operator()` is ill-formed with a substitution error.
6. `class detail::ordering_operators` is a mixin base class that provides all of `operator<`, `operator>`, `operator<=` and `operator>=` as ADL-only hidden `friend` functions based on `tag_invoke(compare_three_way{}, lhs, rhs)`.

## C++20 Iterators

To support some niceties of C++20/23 additions to `string_view`, a detour was made for some iterator (and ranges) backports.

1. `pointer_traits` and `to_address`. These are utilities simply for converting an iterator into a raw pointer to the referred-to object. `to_address_t` and `dereference_t` are not part of a standard, but are useful with `is_detected`.
2. `is_dereferencable` is a traits template detects an object to which we can apply unary operator `*` which returns a non-`void` expression.
3. `iter_value_t`, `iter_reference_t`, and `iter_difference_t` all come from C++20 to obtain the value type, reference type, and difference type of an iterator-like object, respectively.
4. `is_weakly_incrementable` is a trait template based on the `weakly_incrementable` C++20 concept, and the basis for detecting iterators.
5. `is_iterator` is a trait template that detects whether a type is usable as an iterator.
6. `contiguous_iterator_tag` is a C++20 iterators concept tag that extends `random_access_iterator_tag`.
7. `iterator_concept_t` is not part of C++20, but is useful for obtaining the iterator category. It is not possible to infer `contiguous_iterator_tag`, but it is extremely useful to know it if its available. Unless C++20 support is enabled, only raw pointers will return `contiguous_iterator_tag`.
8. `is_{input,forward,bidirection,random_access,contiguous}_iterator` are rough trait templates that detect an iterator type based on the result of `iterator_concept_t`. They do not implement the full iterator checks of the C++20 concepts.
9. `is_sentinel_for<S, I>` trait template detects whether `S` is a valid sentinel type for `I`.
10. `is_sized_sentinel_for<S, I>` extends `is_sentinel_for<S, I>` to require that taking the difference between `S` and `I` returns a value convertible to `iter_difference_t<I>`.

## Detour: Preprocessor Macros

This is a bit of an aside change to add some "probably useful" macros to clean up some conditional compilation and diagnostic control.

1. `bsoncxx_concat` concatenates tokens, and `bsoncxx_stringify` turns tokens into a string literal.
2. `bsoncxx_if_msvc`, `bsoncxx_if_gcc`, `bsoncxx_if_clang`, and `bsoncxx_if_gnu_like` are function-like macros that expand to their arguments if-and-only-if we are compiling for the respective compiler.
3. `bsoncxx_pragma(...)` is a function-like port of MSVC `__pragma`. It accepts a token soup, which is then evaluated as a preprocessor `#pragma` at the use site.
4. `bsoncxx_push_warnings()`/`bsoncxx_pop_warnings()` is equivalent to the appropriate compiler pragma to push/pop the diagnostics settings at the use site.
5. `bsoncxx_disable_warning(Spec)` disables warnings for a certain compiler (See macro code comment).

## A Little C++20 Ranges, as a Treat

A very small bit of C++20 ranges and the range-algorithms were backported, plus some additional utilities. This was finicky to get working on VS 2015, but doable. The following features are available in the `bsoncxx::detail` namespace:

1. The `begin()`, `end()`, `size()`, `ssize()`, and `data()` invocable objects.
2. `iterator_t`, `sentinel_t`, `range_size_t`, `range_difference_t`, `range_data_t`, `range_value_t`, and `range_concept_t` alias templates.
3. `is_range<R>` trait template detects `iterator_t<R>` and `sentinel_t<R>`
4. `is_contiguous_range<R>` is not strictly equivalent to the `contiguous_range` concept, but suites our needs. It requires a valid `data(R)` and `size(R)`.
5. A `subrange<>` class template for creating ranges from iterator pairs.
6. `unreachable_sentinel` for unbounded views.
7. Algorithms: `advance`, `next`, `equal`, `find`, `find_if`, and `search`. Also: `not_fn` and `default_searcher`.
8. Other non-standard bits: `make_reversed_view`, `equal_to_any_of`, and `equal_to_value`.

The algorithms differ from std-ranges in that the iterator-only overloads are omitted, instead all require a range and are bounds-checked. To use a "zero overhead" unchecked algorithms, use a a `subrange` with `unreachable_sentinel`.

## The Actual `string_view`

Finally, the actual `stdx::basic_string_view` class template. It builds upon all of the previous additions and allows us to stop using mnmlstc-core and Boost for our string-viewing needs.

The `basic_string_view` class template is not equal to the C++23 implementation, but provides most of the same functionality.

1. Constructible from:
   1. Nothing. Default-constructs to null.
   2. A pointer+size pair.
   3. An iterator+sentinel pair, from C++23. Requires that they be a sized sentinel and contiguous iterator.
   4. A contiguous range of the character type, from C++23. This allows construction from `std::vector<char>`, `std::array<char>`, etc. This is an `explicit` conversion.
   9. An implicit conversion from string-like ranges of the character type. This includes C++ `std::string`, as well as `std::string_view` itself. It detects "string-like" based on the presence of a `.c_str()` method.
   10. Construction from `nullptr_t` is explicitly deleted (from C++23).
2. All of `begin()`, `cbegin()`, `rbegin()`, and `crbegin()` are defined, as well as the corresponding `end()`s
3. `operator[]` and all non-throwing access functions that specify UB for out-of-bounds access include an assertion with a useful error message rather than a strict terminations, done using `_assert_inbounds()`. This method is defined inline so that assertion condition is visible to the inliner without requiring LTO, while the actual diagnostic+termination function is implemented out-of-line to reduce code size.
4. The C++20 additions of `starts_with` and `ends_with` are included.
5. C++23 `contains`.
6. All `find`-etc members are included.
7. The equality and ordering operators are implemented using the base class mixins from `operators.hpp` using `tag_invoke`. Only two functions make all six operators!


src/bsoncxx/test/CMakeLists.txt
```diff
@@ -93,6 +96,25 @@ if(ENABLE_MACRO_GUARD_TESTS)
...
+            bsoncxx_force_semicolon
+            bsoncxx_pragma
+            _bsoncxxPragma
+            bsoncxx_stringify
+            bsoncxx_stringify_impl
+            bsoncxx_if_msvc
+            bsoncxx_if_gcc
+            bsoncxx_if_clang
+            bsoncxx_if_gnu
+            bsoncxx_returns
```
eramongodb, line 117: Unless there is compelling reason otherwise, recommend using a consistent `UPPER_CASE` + `BSONCXX_`-prefix naming scheme for macros used in public headers.

eramongodb: Note: if we want to better distinguish public macros from private macros via their names, rather than solely by documentation, recommend using a `BSONCXX_PRIVATE_` prefix rather than leading underscores (we already have `BSONCXX_PRIVATE` to declare symbol visibility for non-public-API symbols).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/config/util.hpp
```diff
@@ -0,0 +1,125 @@
...
+#define bsoncxx_stringify_impl(...) #__VA_ARGS__
+
+/**
+ * @brief Token-paste two macro arguments, after macro expansion
+ */
+#define bsoncxx_concat(A, ...) bsoncxx_concat_impl(A, __VA_ARGS__)
+#define bsoncxx_concat_impl(A, ...) A##__VA_ARGS__
+
+/**
+ * @brief Expands to a _Pragma() preprocessor directive, after macro expansion
```
eramongodb, line 15: Unless we intend to support these macros as part of our public API, recommend either demoting these documentation comments to plain `//` (non-Doxygen) or adding `@internal`+`@endinternal` so they do not show up in the API reference documentation pages.

.clang-format
```diff
@@ -21,3 +21,5 @@ NamespaceIndentation: None
 SpaceBeforeAssignmentOperators: true
 Standard: Cpp11
 UseTab: Never
+WhitespaceSensitiveMacros:
```
eramongodb, line 24: I believe this requires bumping our required ClangFormat version from clang-format-7 to clang-format-11.

```
YAML:25:3: error: unknown key 'WhitespaceSensitiveMacros'
  - bsoncxx_pragma
  ^
Error reading .clang-format: Invalid argument
```

kevinAlbs (commented):
Impressive.

.clang-format
vector-of-bool: Removed as unnecessary.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
```diff
@@ -420,6 +444,33 @@ struct is_invocable : is_detected<invoke_result_t, Fun, Args...> {
...
+        return static_cast<decay_t<T>>(static_cast<T&&>(arg));
+    }
+};
+
+/**
+ * @brief Perform a decay-copy on the given value.
+ *
+ * Equivalent to the C++23 `auto()` expression.
+ */
+static constexpr _decay_copy_fn decay_copy;
```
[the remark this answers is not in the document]
vector-of-bool: Correct. Removed.

src/bsoncxx/test/CMakeLists.txt
vector-of-bool: I don't think there's a need to qualify private macros since we already `push_macro` and `pop_macro` them all before/after every source file, so there is no risk of them leaking out.

The same might be said about using `UPPER_CASE` for all macros, since we aren't going to collide with any names. I'm not _against_ `UPPER_CASE` macros, I just default to `lower_snake_case` for most things.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/operators.hpp
```diff
@@ -0,0 +1,168 @@
...
+    constexpr strong_ordering(std::nullptr_t) noexcept : strong_ordering(_construct{}, 0) {}
+
+    constexpr bool operator==(strong_ordering o) const noexcept {
+        return _c == o._c;
+    }
+    constexpr bool operator!=(strong_ordering o) const noexcept {
+        return !(*this == o);
+    }
+
+#define DEFOP(Op)                                               \
```
eramongodb, line 94: Missing macro push/pop + `BSONCXX_` prefix.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/iterator.hpp
```diff
@@ -0,0 +1,227 @@
...
+template <typename T>
+using to_address_t = decltype(to_address(std::declval<T&>()));
+
+/**
+ * @brief The result of applying unary operator* to the given object, if valid
+ */
+bsoncxx_push_warnings();
+bsoncxx_disable_warning(Clang("-Wvoid-ptr-dereference"));
+template <typename I>
+using dereference_t = decltype(*std::declval<I>());
```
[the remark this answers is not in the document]
vector-of-bool: Not here, but probably yes on `is_dereferencable`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/iterator.hpp
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * @brief Detect a type that can be dereferenced (like a pointer) and the result
+ * type is non-void
+ */
+template <typename I>
+struct is_dereferencable : conjunction<is_detected<dereference_t, I>,
+                                       // Clang supports dereferencing void*, and we can't detect
+                                       // that easily. Refuse if I is (cv-)void*
+                                       negation<std::is_void<remove_pointer_t<I>>>,
+                                       negation<std::is_void<detected_t<dereference_t, I>>>> {};
```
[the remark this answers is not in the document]
vector-of-bool: Possibly, but any other use of `dereference_t` (including the upcoming optional<T> changes) that could potentially trigger the warning, so I think it would be easier to just supress the warning in one location.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/iterator.hpp
```diff
@@ -0,0 +1,227 @@
...
+#include <iterator>
+#include <memory>
+#include <type_traits>
+#include <utility>
+#include <vector>
+#if defined(__has_include)
+#if __has_include(<version>)
+#include <version>
+#endif
+#endif
```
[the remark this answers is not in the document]
vector-of-bool: This one is an intentional exception to pull the stdlib tag if it is available. I will explain the reasoning with a comment for readers' benefit.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/ranges.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+#include <bsoncxx/stdx/iterator.hpp>
+#include <bsoncxx/stdx/operators.hpp>
+#include <bsoncxx/stdx/type_traits.hpp>
+
+#include <bsoncxx/config/prelude.hpp>
+
+namespace bsoncxx {
+inline namespace v_noabi {
+namespace detailx {
```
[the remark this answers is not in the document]
vector-of-bool: Unfortunately that will trigger the VS2015 bug being avoided, as the parent namespace `bsoncxx::detail` will contain the names that we want to hide from it.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,561 @@
...
+        self_type sub = this->substr((std::min)(pos, size()));
+        detail::subrange<iterator> found = detail::search(sub, infix);
+        if (detail::distance(found) != static_cast<difference_type>(infix.size())) {
+            return npos;
+        }
+        return _iter_to_pos(found.begin());
+    }
+
+    /**
+     * @brief Find the zero-based index of the right-box occurrence of the given substring
```
kevinAlbs, line 398:
```suggestion
     * @brief Find the zero-based index of the right-most occurrence of the given substring
```

src/bsoncxx/test/string_view.test.cpp
```diff
@@ -157,3 +157,48 @@ TEST_CASE("string_view: Overloading safety") {
...
+    pos = s1.find("bc1", 2);
+    CHECK(pos == 7);
+    pos = s1.find("", 4);
+    CHECK(pos == 4);
+    CHECK(s1.find("") == 0);
+    CHECK(s1.find("nowhere") == s1.npos);
+    CHECK(s1.rfind("nowhere") == s1.npos);
+    CHECK(s1.find("123") == 3);
+    CHECK(s1.rfind("123") == 9);
+    CHECK(s1.rfind("abc", 8) == 0);
```
kevinAlbs, line 177: There appears to be behavior differences with `std::string_view`.

```c++
std::string_view s1 = "abc123abc123";
CHECK(s1.rfind("abc", 8) == 0);
```

Results in:

```c++
/Users/kevin.albertson/review/mongo-cxx-driver-1062/src/bsoncxx/test/string_view.test.cpp:223: FAILED:
  CHECK( s1.rfind("abc", 8) == 0 )
with expansion:
  6 == 0
```

Suggest running this test side-by-side with `std::string_view` to compare behavior.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/ranges.hpp
```diff
@@ -0,0 +1,275 @@
...
+using range_reference_t = iter_reference_t<iterator_t<R>>;
+
+template <typename R>
+using range_concept_t = iterator_concept_t<iterator_t<R>>;
+
+/**
+ * @brief Trait detects if the given type is a range
+ */
+template <typename R>
+struct is_range : conjunction<is_detected<iterator_t, R>, is_detected<sentinel_t, R>> {};
```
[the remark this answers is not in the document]
kevinAlbs: I suggest removing support of constructing a `string_view` from a range.

I prefer having the option to remove the polyfills once the C++ driver requires C++17 language support with minimal user impact.

I expect users cannot currently construct a `string_view` from a range using the MNMLSTC. Example:
```c++
#include <ranges>
// Configure driver with CMAKE_CXX_STANDARD=20 and BSONCXX_POLY_USE_MNMLSTC=1
TEST_CASE("test string view from range") {
    std::vector<char> cs = {'a', 'b', 'c', 'd'};
    bsoncxx::stdx::string_view sv(cs);
    // Results in "no matching constructor"
}
```

Adding the range constructors may make it more difficult to remove the polyfills in the future.

kevinAlbs (changes requested):

src/bsoncxx/test/string_view.test.cpp
vector-of-bool: Additional tests and fixes added. Looks to be me misreading the spec and some off-by-one errors as well.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/ranges.hpp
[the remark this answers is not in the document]
vector-of-bool: Removed the C++20+ interfaces.

src/bsoncxx/test/CMakeLists.txt
vector-of-bool: pain. Macros updated.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,489 @@
...
+        return Name(self_type(cstr), pos);                                                        \
+    }                                                                                             \
+    BSONCXX_FORCE_SEMICOLON
+    DECL_FINDERS(find, 0);
+    DECL_FINDERS(rfind, npos);
+    DECL_FINDERS(find_first_of, 0);
+    DECL_FINDERS(find_last_of, npos);
+    DECL_FINDERS(find_first_not_of, 0);
+    DECL_FINDERS(find_last_not_of, npos);
+#pragma pop_macro("DECL_FINDERS")
```
eramongodb, line 427: Macro guards in config headers preceed the `pop_macro` with an `#undef`:

```cpp
#pragma push_macro("X")
#undef X
#define X
...
#undef X
#pragma pop_macro("X")
```

Recommend including the `#undef` here as well for consistency. Also applies to other instances of macro guards in this PR.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,489 @@
...
+
+    /**
+     * @brief Find the zero-based offset of the right-most occurrence of the given infix,
+     * starting with (and including) pos. If infix does not occur, returns npos.
+     */
+    bsoncxx_cxx14_constexpr size_type rfind(self_type infix, size_type pos = npos) const noexcept {
+        // Calc the endpos where searching should begin, which includes the infix size
+        const size_type endpos = pos != npos ? pos + infix.size() : pos;
+        self_type searched = this->substr(0, endpos);
+        bsoncxx::detail::reversed_t<self_type> found =
```
eramongodb, line 367: Implement this and functions below in terms of C++11 `<algorithm>` instead of C++20 Ranges.

e.g. this function is just a `std::search()` with reverse iterators:

```cpp
if (self->empty() && infix.empty()) {
  return 0u; // Special case that `std::search` doesn't handle.
}

const auto sub = self.substr((std::min)(pos, this->size()));
const auto riter = std::search(sub.rbegin(), sub.rend(), infix.rbegin(), infix.rend());

if (riter == sub.rend()) {
  return npos;
}

return static_cast<std::size_t>(
  std::distance(
    this->begin(),
    std::prev(
      riter.base(),
      static_cast<std::ptrdiff_t>(infix.size())
    )
  )
);
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/ranges.hpp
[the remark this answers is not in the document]
eramongodb: Most if not all of `algorithm.hpp`, `iterator.hpp`, and `ranges.hpp` should not be necessary and can be removed, as C++20 Ranges are not needed to implement a C++17-conforming `string_view`. See other comments suggesting use of C++11 `<algorithm>`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,489 @@
...
+        return self_type(_begin + pos, (std::min)(count, size() - pos));
+    }
+
+    /**
+     * @brief Compare two strings lexicographically
+     *
+     * @param other The "right hand" operand of the comparison
+     * @retval 0 If *this == other
+     * @retval n : n < 0 if *this is "less than" other.
+     * @retval n : n > 0 if *this is "greater than" other.
```
eramongodb, line 304: Doxygen doesn't seem to like repeated use of the `@retval` command with the same return value:

```
warning: return value 'n' of bsoncxx::v_noabi::stdx::basic_string_view::compare has multiple documentation sections
warning: return value 'n' of bsoncxx::v_noabi::stdx::basic_string_view::compare has multiple documentation sections
```

Suggest either:

```
@retval negative-value if `*this` is "less than" other.
@retval positive-value if `*this` is "less than" other.
```
Or using `@return` instead (no return value argument that demands uniqueness):
```
* @return `0` if `*this == other`.
* @return `n : n < 0` if `*this` is "less than" other.
* @return `n : n > 0` if `*this` is "greater than" other.
```

eramongodb (changes requested):

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
vector-of-bool: Does that `undef` really do anything if the macro definition is immediately popped?

eramongodb: On second thought, probably not. I thought it would trigger a warning if the macro being undef'd was not already defined (e.g. as a typo check), but that does not seem to be the case for either GCC or Clang in my local tests. The `pop_macro` doesn't seem to generate any warnings if the macro being popped is already defined either. The `#undef` preceeding a `#pragma pop_macro` doesn't seem to serve a useful function, so I'm in favor of it's omission going forward.

kevinAlbs (approved):

eramongodb (changes requested):

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,490 @@
...
+
+    /**
+     * @brief Implicit conversion from string-like ranges.
+     *
+     * Requires that `StringLike` is a non-array contiguous range with the same
+     * value type as this string view, and is a std::string-like value.
+     */
+    template <typename StringLike, _enable_string_conversion<StringLike> = 0>
+    constexpr basic_string_view(StringLike&& str) noexcept
+        : _begin(str.data()), _size(str.size()) {}
```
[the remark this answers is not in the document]
vector-of-bool: This inclusion is intentional to support compat with std::string+std::string_view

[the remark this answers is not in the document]
eramongodb: Good point, I overlooked support for `std::string::operator std::string_view()`.

Rather than constraining the template to _any_ string-like type with SFINAE, a simple dedicated overload for `std::string` should work just fine:

```cpp
/**
 * @brief Implicit conversion from std::string ranges.
 */
constexpr basic_string_view(const std::basic_string<Char, Traits>& str) noexcept
    : _begin(str.data()), _size(str.size()) {}
```

I'm not sure supporting an implicit conversion from `std::string_view` to `bsoncxx::stdx::string_view` is desirable, given they are distinct implementations of the same vocabulary type. (Despite our efforts, they're still unlikely to be perfectly substitutable with one another.) We may want to encourage downstream code converting between the two types to be explicit with conversions so it's clear which `string_view` they're dealing with.

If we _do_ still want to support it, I think we should support conversion _both_ ways in C++17 mode:

```cpp
#if /* C++17 */

constexpr basic_string_view(const std::basic_string_view<Char, Traits>& str) noexcept
    : _begin(str.data()), _size(str.size()) {}

operator std::basic_string_view<Char, Traits>() const noexcept {
    return std::basic_string_view<Char, Traits>(data(), size());
}

#endif
```

This would allow the two to be fully interchangeable/convertable with each other:

```cpp
std::string_view std_sv;
bsoncxx::stdx::string_view bsoncxx_sv;

bsoncxx_sv = std_sv; // OK: ctor (currently handled by range ctor).
std_sv = bsoncxx_sv; // OK: conversion operator (currently not supported).
```

Thoughts?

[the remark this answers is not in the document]
vector-of-bool: That may work.

In either case, supporting conversion to/from `std::string_view` may be problematic by ambiguating with the C++23 from-range converting constructors, since `bsoncxx::string_view` satisfies the constraints thereof. Time will tell.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
```diff
@@ -679,6 +674,7 @@ BSONCXX_INLINE bool operator==(const b_maxkey&, const b_maxkey&) {
 }  // namespace v_noabi
 }  // namespace bsoncxx
 
+BSONCXX_POP_WARNINGS();
```
eramongodb, line 677:
```suggestion
BSONCXX_POP_WARNINGS();

```
Spacing.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
+
+    /**
+     * @brief Default constructor. Constructs to an empty/null string view
+     */
+    constexpr basic_string_view() noexcept = default;
+    constexpr basic_string_view(const basic_string_view&) noexcept = default;
+    bsoncxx_cxx14_constexpr basic_string_view& operator=(const basic_string_view&) noexcept =
+        default;
+
+    /// Default copy/move/assign/destroy
```
eramongodb, line 87:
```suggestion

    constexpr basic_string_view(const basic_string_view&) noexcept = default;
    bsoncxx_cxx14_constexpr basic_string_view& operator=(const basic_string_view&) noexcept =
        default;
```
Fix(?) spacing by separating the copy special member functions from the default ctor + remove a stray comment.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
+    constexpr size_type max_size() const noexcept {
+        return static_cast<size_type>(std::numeric_limits<difference_type>::max());
+    }
+
+    /**
+     * @brief In-place modify the string_view to view N fewer characters from the beginning
+     *
+     * @param n The number of characters to remove from the beginning. Must be less than size()
+     */
+    bsoncxx_cxx14_constexpr void remove_prefix(size_type n) noexcept {
```
eramongodb, line 207:
```suggestion
    bsoncxx_cxx14_constexpr void remove_prefix(size_type n) {
```
Stray `noexcept`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,92 +14,490 @@
...
+     *
+     * @param dest The destination at which to write characters
+     * @param count The maximum number of characters to copy.
+     * @param pos The offset within the viewed string to begin copying from.
+     * @returns The number of characters that were copied to `dest`. The number
+     * of copied characters is always the lesser of `size()-pos` and `count`
+     *
+     * @throws std::out_of_range if pos > size()
+     */
+    bsoncxx_cxx14_constexpr size_type copy(pointer dest, size_type count, size_type pos = 0) const {
```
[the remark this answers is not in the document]
eramongodb:
```suggestion
    size_type copy(pointer dest, size_type count, size_type pos = 0) const {
```
Not yet addressed.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
+                          size_type count2) const {
+        return substr(pos1, count1).compare(basic_string_view(str, count2));
+    }
+
+    /**
+     * @brief Find the zero-based offset of the left-most occurrence of the given infix,
+     * starting with pos. If infix does not occur, returns npos.
+     */
+    bsoncxx_cxx14_constexpr size_type find(basic_string_view infix,
+                                           size_type pos = 0) const noexcept {
```
eramongodb, line 337:
```suggestion
    bsoncxx_cxx14_constexpr size_type find(basic_string_view infix, size_type pos = 0) const
        noexcept {
```
ClangFormat.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
+        }
+        return static_cast<size_type>(found - begin());
+    }
+
+    /**
+     * @brief Find the zero-based offset of the right-most occurrence of the given infix,
+     * starting with (and including) pos. If infix does not occur, returns npos.
+     */
+    bsoncxx_cxx14_constexpr size_type rfind(basic_string_view infix,
+                                            size_type pos = npos) const noexcept {
```
eramongodb, line 358:
```suggestion
    bsoncxx_cxx14_constexpr size_type rfind(basic_string_view infix, size_type pos = npos) const
        noexcept {
```
ClangFormat.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
+    constexpr size_type find_first_not_of(basic_string_view set, size_type pos = 0) const noexcept {
+        return _find_if(pos, [&](value_type chr) { return set.find(chr) == npos; });
+    }
+
+    /**
+     * @brief Find the zero-based index of the right-most occurrence of any character that
+     * is NOT a member of the given set of characters, starting at (and including) pos
+     */
+    constexpr size_type find_last_not_of(basic_string_view set,
+                                         size_type pos = npos) const noexcept {
```
eramongodb, line 401:
```suggestion
    constexpr size_type find_last_not_of(basic_string_view set, size_type pos = npos) const
        noexcept {
```
ClangFormat.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
```diff
@@ -14,101 +14,517 @@
...
-using ::boost::basic_string_view;
-using ::boost::string_view;
-
-}  // namespace stdx
-}  // namespace v_noabi
-}  // namespace bsoncxx
+#include <bsoncxx/config/prelude.hpp>
 
-#else
+#ifdef __has_include
```
eramongodb, line 30: I am reminded that this pattern is pre-C++17-compatible. Very nice. 👍

eramongodb (approved):
Thank you for your hard work on this PR. A couple suggestions remaining; otherwise, LGTM. 👍

kkloberdanz (approved):

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp
vector-of-bool: Newer clang-format wants to break this line differently. Pain.

PR #1070, opened 2023-12-12
CXX-2790 Redeclare mongocxx::v_noabi as a non-inline namespace

eramongodb:
## Brief

Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1066. Resolves CXX-2790. Verified by [this patch](https://spruce.mongodb.com/version/6578c428e3c33121617ea029/tasks).

As before, this PR _SHOULD NOT_ break source or binary compatibility. Verified by [this patch](https://spruce.mongodb.com/version/6578c70f2a60ed2936a084b8/tasks). No exceptions this time, although the [low severity warning](https://mciuploads.s3.amazonaws.com/mongo-cxx-driver/master/1b9b5a1b22ace73d7afb711822a357959b5c8b1a/6578c70f2a60ed2936a084b8/cxx_driver_abi_compatibility_patch_1b9b5a1b22ace73d7afb711822a357959b5c8b1a_6578c70f2a60ed2936a084b8_23_12_12_20_48_15/0/abi-compliance-check/noabi/compat_report.html#Type_Source_Problems_Low) for `system_clock::time_point` in the source compatibility report is curious.

## Description

Analogous to the changes in https://github.com/mongodb/mongo-cxx-driver/pull/1066 for the bsoncxx library, this PR applies the following changes to the mongocxx library:

* `inline namespace v_noabi` -> `namespace v_noabi`
* Redeclaration of `v_noabi` entities in the `mongocxx` root namespace via using-declarations.
* Substitute `mongocxx::X` -> `mongocxx::v_noabi::X` in ABI namespace code (not in test code).
* Removal of `v_noabi::` for private and test entities (not specific to an ABI version).
* `MONGOCXX_INLINE` removed from private and test files (do not document public API).

There are only two notable exceptions to the regular modification patterns above:

* `bsoncxx::stdx` is redeclared in `mongocxx::stdx` and `mongocxx::v_noabi::stdx` separately, with `mongocxx::v_noabi::stdx` being marked as deprecated. This is to reflect the fact that `mongocxx::stdx` is just a convenient, straightforward redeclaration of `bsoncxx::stdx` entities. As with all other referenced bsoncxx entities, those redeclared in `mongocxx::stdx` are neither owned or managed by mongocxx. It's the _bsoncxx_ library's responsibility to ensure ABI stability of the `stdx` entities being redeclared. Therefore, there is no reason to continue encouraging `mongocxx::v_noabi::stdx`: the entities declared within are not owned by `mongocxx`, thus have no direct relevance to the mongocxx library's ABI version.
* The error message strings in `mongocxx/exception/error_code.cpp` are left unchanged despite referencing root namespace declarations without ABI namespace (e.g. `mongocxx::client`) to preserve existing behavior. If it is not important to preserve this behavior, the messages can be changed as well.
* The mocked definition of `MONGOCXX_LIBMONGOC_SYMBOL` is now declared with `MONGOCXX_TEST_API` rather than `MONGOCXX_API` to better reflect it's intended usage.

Despite the greater diff volume due to more source code in mongocxx, the PR should otherwise be very similar to https://github.com/mongodb/mongo-cxx-driver/pull/1066.


kevinAlbs:
> If it is not important to preserve this behavior, the messages can be changed as well.

I expect there are no documented guarantees to preserve error messages. Though it would not surprise me if some error messages are relied on by consumers. Example: SERVER-41365 describes how the "ns not found" error message was relied on in drivers. I expect changing the namespace in the error messages of `error_category` is unlikely to meaningfully break consumer code (maybe could break tests?).

I slightly prefer to remove namespace from the error message. E.g. change `invalid use of default constructed or moved-from mongocxx::client object` to `invalid use of default constructed or moved-from client object`. If the ABI namespace is included in the message (e.g. `mongocxx::v_noabi::client`), I expect it may prevent reuse of the error codes in future ABI versions.

kevinAlbs (commented):
Looks good with additional suggested changes:

- Add ABI namespace to `@see mongocxx::estimated_document_count` and `@see mongocxx::count_documents` in collection.hpp.
- Add ABI namespace to `mongocxx::result::bulk_write` in `rewrap_many_datakey.hpp`
- ~~Replace remaining use of `mongocxx::stdx::optional` with `bsoncxx::v_noabi::stdx::optional`~~ (Disregard: only used in private implementation)

eramongodb:
Further discussion following feedback given above revealed references to bsoncxx entities in mongocxx _should_ include an ABI namespace qualifier (this should have been done in https://github.com/mongodb/mongo-cxx-driver/pull/1066). This is to ensure ABI stability of the mongocxx library when the bsoncxx library changes the ABI version of a root namespace redeclaration, which _shouldn't_ be an ABI breaking change for either the bsoncxx or mongocxx library. This also demonstrates correct, deliberate use of ABI namespace qualifiers in code that is designed for ABI stability (model for downstream users). This is all precluding the fact that `v_noabi` is not a stable ABI namespace, but there is no better time to test and establish patterns to use in the upcoming actual stable ABI namespaces than now.

Latest changes verified by [this patch](https://github.com/mongodb/mongo-cxx-driver/pull/1066).

kevinAlbs (approved):

vector-of-bool (approved):
Apologies for the long delay. LGTM.

PR #1082, opened 2024-01-16
CXX-2039 Add VS2019 support and migrate VS2017 tasks to windows-vsCurrent distro

eramongodb:
Resolves CXX-2039. Verified by [this patch](https://spruce.mongodb.com/version/65a6d7943066150d92e11de2).

This PR primarily consists of three steps:

* Move existing VS 2017 tasks from `windows-64-vs2017` to `windows-vsCurrent`.
* Rename VS 2017 fields/properties to accomodate multiple VS versions.
* Extend VS 2017 tasks to VS 2019 (simple duplication).

This should also address the spurious "cygheap base mismatch detected" failures on Evergreen due to BUILD-17907, which were frequently observed on the `windows-64-vs2017` distro (VS 2015 tasks thankfully seem to be unaffected?).

As part of this effort, several drive-by improvements are made to the EVG config and scripts:

* Use `CMAKE_GENERATOR_PLATFORM` or `-A` to specify the [generator platform](https://cmake.org/cmake/help/v3.15/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM) rather than the `Win64` generator suffix (which only exists for pre-CMake 3.1 compatibility and is not supported by the VS 2019 generator or newer). Documentation and examples were updated accordingly.
* Drop obsolete MSBuild path handlers (toolchain detection is handled by the "latest" CMake binary, and build system invocation is handled by the toolchain-independent `cmake --build` command).
* Ensure CMake generator and platform preferences are also communicated to example projects via CMake environment variables.

This PR initially proposes a simple duplication of the existing VS 2017 tasks with VS 2019 instead. If preferable, the set of tasks and coverage between the VS 2017 and VS 2019 tasks can be adjusted to reduce redundancy.

Note: this PR does not extend C++20 compile coverage to VS 2019, nor does it extend test coverage to VS 2022 (which is available on the new distro).

### DIFF
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -28,23 +28,6 @@ set -o pipefail
 : "${USE_SANITIZER_UBSAN:-}"
 : "${USE_STATIC_LIBS:-}"
 
-# Add MSBuild.exe to path.
-if [[ "${OSTYPE:?}" == "cygwin" ]]; then
-  case "${generator:-}" in
-  *2015*)
-    PATH="/cygdrive/c/cmake/bin:/cygdrive/c/Program Files (x86)/MSBuild/14.0/Bin:$PATH"
-    ;;
-  *2017*)
-    PATH="/cygdrive/c/cmake/bin:/cygdrive/c/Program Files (x86)/Microsoft Visual Studio/2017/Professional/MSBuild/15.0/Bin:$PATH"
-    ;;
-  *)
-    echo "missing explicit CMake Generator on Windows distro" 1>&2
-    exit 1
-    ;;
-  esac
-fi
-export PATH
-
 mongoc_prefix="$(pwd)/../mongoc"
 echo "mongoc_prefix=${mongoc_prefix:?}"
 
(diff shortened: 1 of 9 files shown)


kevinAlbs (approved):
LGTM with a suggestion for removing all but one VS 2017 integration tasks. Addressing the "cygheap base mismatch detected" failures, and use of newer CMake features is much appreciated.

> If preferable, the set of tasks and coverage between the VS 2017 and VS 2019 tasks can be adjusted to reduce redundancy.

I expect there is little value in running tests with multiple server versions with the C++ driver compiled on VS 2017. Testing compile seems valuable, but I expect there is little behavior change between the compiled versions.
I suggest only testing one server version:

```yml
# Only test VS 2017 with the latest server and auth. Testing other server versions is covered with newer VS tasks.
- name: integration-auth-vs2017-latest-single
  display_name: "Windows (VS 2017) Debug Latest Auth"
  run_on: windows-vsCurrent-large
  expansions:
      mongodb_version: "latest"
      <<: *integration_matrix_expansions_windows_vs2017
  <<: *integration_matrix_auth_tasks_single
```

PR #1086, opened 2024-01-23
CXX-2796 Restore external polyfill library and stdlib support

eramongodb:
## Description

This PR ~partially~ resolves CXX-2796. ~Full resolution pending https://github.com/mongodb/mongo-cxx-driver/pull/1075.~ (Merged.) Verified by [this patch](https://spruce.mongodb.com/version/65b023e5d1fe07852c1b96dd).

This PR restores polyfill library (pre-C++17) and stdlib (post-C++17) support, including the explicit setting of a `BSONCXX_POLY_USE_*` option regardless of the selected C++ standard. This avoids the pending r3.10.0 release from potentially causing source-breaking changes in downstream code that depends on expectations of the underlying type for `bsoncxx::stdx::string_view` or `bsoncxx::stdx::optional<T>` being consistent with a selected polyfill library.

## BSONCXX_USE_POLY_IMPLS

This PR adds a new CMake option `BSONCXX_POLY_USE_IMPLS` to opt-into using bsoncxx implementations for polyfills. This configuration is denoted as `i` in the ABI tag. The `ENABLE_BSONCXX_POLY_USE_IMPLS` option controls whether bsoncxx implementations are considered by default polyfill library selection (when no polyfill library is explicitly selected by the user).

`ENABLE_BSONCXX_POLY_USE_IMPLS` is currently `OFF` by default to avoid source-breaking changes in user code that depends on current default polyfill library selection behavior for pre-C++17 configurations. This option is expected to be set to `ON` by default in an upcoming major release as part of CXX-2797. Installation documentation is updated to recommend users set `BSONCXX_POLY_USE_IMPLS=ON` to avoid external library dependencies. The external polyfill library options are now **deprecated** in favor of bsoncxx implementations per CXX-2797. CMake deprecation warning messages are also added accordingly.

Note, `BSONCXX_POLY_USE_IMPLS=ON` is currently broken due to https://github.com/mongodb/mongo-cxx-driver/pull/1075 not yet being merged (there is no bsoncxx implementation of `std::optional`).

## Restoring Polyfill Support

This PR restores polyfill selection behavior and support to be consistent with the r3.9.0 release.

This was fairly straightforward for `string_view.hpp`, which, relative to r3.9.0 (sans using-decl refactor changes), simply appends the `BSONCXX_POLY_USE_IMPLS` case to the list of alternatives before the `#error` case:

```diff
  /* ... */ // (Mostly) unchanged relative to 3.9.0.

  #elif defined(BSONCXX_POLY_USE_STD)
  
  #include <string_view>
  
  namespace bsoncxx {
  namespace v_noabi {
  namespace stdx {
  
  using ::std::basic_string_view;
  using ::std::string_view;
  
  }  // namespace stdx
  }  // namespace v_noabi
  }  // namespace bsoncxx

+ #elif defined(BSONCXX_POLY_USE_IMPLS)
+ 
+ /* ... */ // bsoncxx implementation of std::string_view.
+ 
  #else
  #error "Cannot find a valid polyfill for string_view"
  #endif
```

This pattern is expected to be applied similarly to the `optional.hpp` header after https://github.com/mongodb/mongo-cxx-driver/pull/1075 is merged in a followup PR.

The `make_unique.hpp` header is different. Because it only provides non-exported _function templates_, its impact on source-compatibility is minimal (situations where our changes _can_ break source would be considered abnormal usage of these interfaces). Furthermore, r3.9.0 introduced a _C++20_ polyfill for `make_unique_for_overwrite<T>()` which is not covered by the usual C++17 external polyfill libraries. Because neither of these features are related to C++17-compatibility, this PR proposes using feature test macros for both `make_unique<T>()` (C++14) and `make_unique_for_overwrite<T>()` (C++20) to select bsoncxx vs. stdlib implementations, independent of the C++17 polyfill library configuration.

Restoration of r3.9.0 behavior can be verified by the ABI stability reports for both [polyfill](https://spruce.mongodb.com/task/cxx_driver_abi_stability_polyfill_display_ABI_Stability_Checks_(polyfill)_patch_77094d18bef6a14c09627dcb20662444b0f25c44_65b023e5d1fe07852c1b96dd_24_01_23_20_39_02/files?execution=0&sorts=STATUS%3AASC) and [stdlib](https://spruce.mongodb.com/task/cxx_driver_abi_stability_stdlib_display_ABI_Stability_Checks_(stdlib)_patch_77094d18bef6a14c09627dcb20662444b0f25c44_65b023e5d1fe07852c1b96dd_24_01_23_20_39_02/files?execution=0&sorts=STATUS%3AASC) configurations (compared to those in the latest commit for [polyfill](https://spruce.mongodb.com/task/cxx_driver_abi_stability_polyfill_display_ABI_Stability_Checks_(polyfill)_f9f48a10f54bb4b7059f39533369c0c84ede7f1f_24_01_22_20_42_22/files?execution=0&sorts=STATUS%3AASC) and [stdlib](https://spruce.mongodb.com/task/cxx_driver_abi_stability_stdlib_display_ABI_Stability_Checks_(stdlib)_f9f48a10f54bb4b7059f39533369c0c84ede7f1f_24_01_22_20_42_22/files?execution=0&sorts=STATUS%3AASC)).

## Miscellaneous

- Disable `string_view` tests for non-bsoncxx polyfills due to compatibility and conformance issues. We only need the tests to test _our_ implementations anyways (+ stdlib to verify test consistency).
- Update abi-compliance-check-test.sh to ignore detail namespaces in compatibility reports to reduce noise.
- Fix the abidiff-test.sh script so it actually uses common flags as intended. Removed the `--non-reachable-types` flag due to producing undesirable errors beyond what we want to test.


src/bsoncxx/CMakeLists.txt
```diff
@@ -46,6 +47,10 @@ elseif(BSONCXX_POLY_OPTIONS_SET EQUAL 0)
         # If we are in C++17 mode, use the C++17 versions
```
kevinAlbs, line 47: Update the `foreach(BSONCXX_POLY_OPTION` above to include `${BSONCXX_POLY_USE_IMPLS}`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp
```diff
@@ -84,6 +95,21 @@ template <typename T>
 struct make_unique_impl<T&&> {};
 
 }  // namespace detail
+}  // namespace stdx
+}  // namespace v_noabi
+}  // namespace bsoncxx
+
+#endif  // defined(BSONCXX_DETAIL_USE_IMPLS_MAKE_UNIQUE)
```
kevinAlbs, line 102:
```suggestion
#endif  // !defined(BSONCXX_DETAIL_USE_STD_MAKE_UNIQUE) || !defined(__cpp_lib_smart_ptr_for_overwrite)
```

src/bsoncxx/CMakeLists.txt
```diff
@@ -46,15 +47,23 @@ elseif(BSONCXX_POLY_OPTIONS_SET EQUAL 0)
...
         message(STATUS "Auto-configuring bsoncxx to use boost std library polyfills since C++17 is inactive and compiler is MSVC")
+        message(DEPRECATION "Default selection of external polyfill libraries is deprecated and will be removed in an upcoming major release. Set ENABLE_BSONCXX_POLY_USE_IMPLS=ON to enable recommended behavior.")
     else()
         # Otherwise, we are on a platform that can handle MNMLSTC
         set(BSONCXX_POLY_USE_MNMLSTC ON)
         message(STATUS "Auto-configuring bsoncxx to use MNMLSTC for polyfills since C++17 is inactive")
+        message(DEPRECATION "Default selection of external polyfill libraries is deprecated and will be removed in an upcoming major release. Set ENABLE_BSONCXX_POLY_USE_IMPLS=ON to enable recommended behavior.")
     endif()
+elseif(BSONCXX_POLY_USE_MNMLSTC OR BSONCXX_POLY_USE_BOOST)
+    message(DEPRECATION "Support for external polyfill libraries is deprecated and will be removed in an upcoming major release. Set BSONCXX_POLY_USE_IMPLS=ON (pre-C++17) or set CMAKE_CXX_STANDARD=17 (or newer) to select recommended behavior.")
```
kevinAlbs, line 66:
```suggestion
elseif(BSONCXX_POLY_USE_MNMLSTC OR BSONCXX_POLY_USE_BOOST OR BSONCXX_POLY_USE_STD_EXPERIMENTAL)
    message(DEPRECATION "Support for experimental and external polyfill libraries is deprecated and will be removed in an upcoming major release. Set BSONCXX_POLY_USE_IMPLS=ON (pre-C++17) or set CMAKE_CXX_STANDARD=17 (or newer) to select recommended behavior.")
endif()
```

Suggest also warning for the experimental polyfill.

kevinAlbs (approved):
LGTM with minor changes.

src/bsoncxx/CMakeLists.txt
eramongodb: The experimental std polyfill option currently has the following documentation:

> Note that this polyfill is _not_ recommended and is unsupported.

This line has been present [since April 2017](https://github.com/mongodb/mongo-cxx-driver/commit/612b31790c4d486a8124133355e0854c89ce8e48). I neglected to add a warning for it as I do not expect it is being used. I can still include it in the warnings if preferable.

On that note, do you think we could simply remove experimental std altogether, given it's been long-documented as unsupported?

kevinAlbs: > On that note, do you think we could simply remove experimental std altogether, given it's been long-documented as unsupported?

Yes. I expect `std::experimental` is OK to remove given it has been long-documented as unsupported. I also expect it is not being used. Though I do not think it is necessary in this PR. I agree with leaving out the warning.

eramongodb: Deferring removal to [CXX-2814](https://jira.mongodb.org/browse/CXX-2814).

eramongodb:
PR has been updated following merge of https://github.com/mongodb/mongo-cxx-driver/pull/1075. This PR now fully resolves CXX-2796.

Latest changes verified by [this patch](https://spruce.mongodb.com/version/65cce2dda4cf475fb82a79c9).

Notable changes:

- Restore polyfill selection for `bsoncxx::stdx::optional`

Use of mnmlstc/core and Boost equivalents as polyfills is restored to r3.9.0 behavior (sans experimental stdlib, removed by https://github.com/mongodb/mongo-cxx-driver/commit/8e1c47ab21070a807add81d758388e99344b8699).

For better consistency with other polyfills, the Boost polyfill was updated to _conditionally_ include a `std::in_place_t` equivalent (this is new), as the [minimum Boost version of 1.56.0](https://github.com/mongodb/mongo-cxx-driver/blob/08e79165fef8766a362c9c53c406ae7a000ec51d/cmake/BsoncxxUtil.cmake#L152) doesn't provide it. `boost::in_place_init_t` is a 1.63.0+ feature. Alternatively, we could bump the minimum Boost version to 1.63.0 and remove the condition.

- Only enable `optional<T>` tests for bsoncxx and stdlib polyfills

Same as what was done for `string_view`. We're only interested in testing the quality of the bsoncxx implementation against the stdlib implementations.

- Relaxed Type Trait Assertions for `optional<T>`

In addition to improving the error messages for type trait equality assertions on failure, when testing against stdlib implementations, the equality assertions are relaxed to assert _implication_ instead. This is to account for deviations in stdlib behavior in the negative cases (e.g. https://godbolt.org/z/P5ab6W46W). Instead of `Trait<T> == Trait<optional<T>>`, we instead test `Trait<T> -> Trait<optional<T>>` (implication). Equality is still used when testing the bsoncxx implementation.

- scoped_bson_t ctors

Rather than conditional inclusion of potentially-ambiguous overloads based on polyfill configuration, explicit overloads for all view/value/view_or_value arguments are defined instead to address the potential ambiguity issues.

- Drive-by: improve abidiff test script

Some changes were made to reduce inclusion of unrelated symbols in the diff report.

- Drive-by: CCACHE_BASEDIR on Windows

Overlooked that config-time CMake paths must be Windows-based.

- Drive-by: ENABLE_TESTING as CMake option

The `ENABLE_TESTING` normal variable is now a cache variable that can be toggled via CMake GUI, ccmake, etc.

kevinAlbs (approved):
LGTM. Nicely done. The test improvements are appreciated.

vector-of-bool:
I'm not sure what the point of all of my work was if we're just bringing back all the old polyfill support anyway. Why did I do anything at all if it just adds another configuration that we need to test? @kevinAlbs @eramongodb

kevinAlbs:
> if we're just bringing back all the old polyfill support anyway.

@vector-of-bool I expect restoring the polyfills prevents possible source breaking changes. Quoting from a slack conversation with @eramongodb (which I should have included in the ticket):

> ```c++
> // User expectation: BSONCXX_POLY_USE_BOOST=ON -> bsoncxx::stdx::string_view is boost::string_view.
> // Same assumption applies to other polyfill library configurations.
> void user::impl(boost::string_view sv) { /* ... */ }
> 
> void user::code() {
>   auto sv = bsoncxx::stdx::string_view("abc");
>   user::impl(sv); // error: breaking change in 3.10 (without CXX-2796 or with CXX-2697).
> }
> ```
> We do not have much documentation concerning our expectations of how users use our polyfills except a line about std::experimental being discouraged and unsupported (despite us having EVG tasks for it), and this line (despite us not supporting ABI stability 
> > *DO NOT* change your project's polyfill if you need to create a stable binary interface.
>
> So the user expectation above is justified even if not desirable by us.

The removal of the other polyfills (CXX-2797) is marked for the next major release (4.0.0). That said, I think think the bsoncxx polyfills provide value in the 3.x releases. Users on pre C++17 compilers can use the bsoncxx polyfills to avoid using Boost / MNMLSTC.

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/private/libbson.cpp
```diff
@@ -31,20 +31,41 @@ void doc_to_bson_t(const bsoncxx::v_noabi::document::view& doc, bson_t* bson) {
...
+    this->init_from_static(bsoncxx::document::view_or_value(std::move(doc)));
+}
+
+scoped_bson_t::scoped_bson_t(bsoncxx::stdx::optional<bsoncxx::document::view_or_value> doc) {
     if (doc) {
-        _doc = std::move(doc);
-        doc_to_bson_t(*_doc, &_bson);
+        this->init_from_static(std::move(*doc));
     }
 }
```
vector-of-bool, line 63: Is there an explanation on why these constructors needed to be revived? I'm guessing that one or more of the optional impls doesn't "do the right thing"?

docs/content/mongocxx-v3/installation/linux.md
```diff
@@ -9,30 +9,55 @@ title = "Linux"
...
-     C++17.
-
-Most users should be fine sticking with the default. However, if you
-have an existing application which makes heavy use of one of the
-available libraries, you may prefer to build the mongocxx driver
-against the same library.
-
-**DO NOT** change your project's polyfill if you need to create a
-stable binary interface.
+The mongocxx driver uses C++17 features `std::optional` and `std::string_view`.
```
vector-of-bool, line 12: Could this section -- repeated between `{linux,macos,windows}.md` -- be refactored into a separate page?

CMakeLists.txt
```diff
@@ -157,6 +155,9 @@ option(BUILD_SHARED_LIBS_WITH_STATIC_MONGOC
     OFF
 )
 
+# Allow the user to opt into using bsoncxx implementations for C++17 polyfills by default.
+option(ENABLE_BSONCXX_POLY_USE_IMPLS "Enable using bsoncxx implementations of C++17 polyfills for pre-C++17 configurations by default" OFF)
```
vector-of-bool, line 159: I'm iffy on this CMake setting name. It is unclear what it means compared to `USE_BOOST`, `USE_STD`, `USE_MNMLSTC` etc. I'm unfortunately not sure on what a better name might be.

vector-of-bool (approved):
Sorry for the delay, and apologies for the brief existential dread. LGTM, with only a few minor comments/questions.

docs/content/mongocxx-v3/installation/linux.md
eramongodb: Good idea. Added a new `mongocxx-v3/polyfill-selection` page that is linked to from the per-OS installation pages.

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/private/libbson.cpp
eramongodb: Correct. Rather than dealing with inconsistent `optional<T>` converting constructor behavior across configurations, settled on making it definitively unambiguous on all implementations for all expected view/value/view_or_value/optional<view_or_value> arguments expected to be given.

docs/content/mongocxx-v3/polyfill-selection.md
```diff
@@ -0,0 +1,59 @@
++++
+date = "2024-02-20T00:00:00-00:00"
+title = "Choosing a C++17 Polyfill"
+[menu.main]
+  identifier = "mongocxx3-polyfill-selection"
+  parent = "mongocxx3"
+  weight = 8
++++
```
eramongodb, line 8: I must admit I do not understand how the value of some of these fields are supposed to be set (date and weight in particular), so simply copied the template from similar files.

PR #1087, opened 2024-01-24
Allow ccache to reuse results across build directories

eramongodb:
The CXX Driver equivalent of changes in https://github.com/mongodb/mongo-c-driver/pull/1522. Verified by [this patch](https://spruce.mongodb.com/version/65b1705357e85a2a383b69fd).

The only caveat is that the CXX Driver does not have its own ccache detect-and-use routine a la CCache.cmake in the C Driver, so each instance of setting the ccache environment variables is preceeded by a test for `ccache` being present.

For consistency, other instances of the ccache test have been changed to use the same `command -V` method, which prints the path to the ccache binary when present in an informative manner (e.g. `ccache is /usr/bin/ccache`) in lieu of the explicit message that was being used prior.

### DIFF
--- a/.evergreen/abi-stability-setup.sh
+++ b/.evergreen/abi-stability-setup.sh
@@ -62,7 +62,7 @@ if command -V ccache 2>/dev/null; then
 
   # Allow reuse of ccache compilation results between different build directories.
   export CCACHE_BASEDIR CCACHE_NOHASHDIR
-  CCACHE_BASEDIR="$(pwd)"
+  CCACHE_BASEDIR="$(pwd)/mongo-cxx-driver"
   CCACHE_NOHASHDIR=1
 fi
 
--- a/.evergreen/abi-stability-setup.sh
+++ b/.evergreen/abi-stability-setup.sh
@@ -56,10 +56,14 @@ else
 fi
 
 # Use ccache if available.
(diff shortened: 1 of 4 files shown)


.evergreen/abi-stability-setup.sh
```diff
@@ -56,10 +56,14 @@ else
...
-  export CMAKE_C_COMPILER_LAUNCHER CMAKE_CXX_COMPILER_LAUNCHER
-  CMAKE_C_COMPILER_LAUNCHER="ccache"
-  CMAKE_CXX_COMPILER_LAUNCHER="ccache"
+if command -V ccache 2>/dev/null; then
+  export CMAKE_C_COMPILER_LAUNCHER=ccache
+  export CMAKE_CXX_COMPILER_LAUNCHER=ccache
+
+  # Allow reuse of ccache compilation results between different build directories.
+  export CCACHE_BASEDIR CCACHE_NOHASHDIR
+  CCACHE_BASEDIR="$(pwd)"
```
kevinAlbs, line 65:
```suggestion
  CCACHE_BASEDIR="$(pwd)/mongo-cxx-driver"
```

To refer to source directory path.

kevinAlbs (approved):
LGTM with one path fix.

PR #1089, opened 2024-01-26
CXX-2814 Remove BSONCXX_POLY_USE_STD_EXPERIMENTAL from polyfill options

eramongodb:
This PR removes the `BSONCXX_POLY_USE_STD_EXPERIMENTAL` config option along with all related code and documentation given it's been explicitly documented as both _not_-recommend and unsupported [since February 2018 in r3.2.0](https://github.com/mongodb/mongo-cxx-driver/blob/r3.2.0/docs/content/mongocxx-v3/installation.md?plain=1#L81) (committed in [April 2017](https://github.com/mongodb/mongo-cxx-driver/commit/612b31790c4d486a8124133355e0854c89ce8e48)), which should be more than enough time for users to have responded accordingly. Verified by [this patch](https://spruce.mongodb.com/version/65b4347c0ae606f30b388d3a).


cmake/BsoncxxUtil.cmake
```diff
@@ -63,14 +63,12 @@ function(bsoncxx_add_library TARGET OUTPUT_NAME LINK_TYPE)
                 set(polyfill "m")
             elseif(BSONCXX_POLY_USE_BOOST)
                 set(polyfill "b")
-            elseif(BSONCXX_POLY_USE_STD_EXPERIMENTAL)
```
kevinAlbs, line 66: Remove `# - 'x' for experimental standard library.` above.

kevinAlbs (approved):
LGTM with a comment change.

PR #1090, opened 2024-01-29
Pin test Docker image to Alpine 3.19

kevinAlbs:
# Summary

Pin test Docker image to Alpine 3.19

This is a follow-up to https://github.com/mongodb/mongo-cxx-driver/pull/1088

# Background & Motivation

Motivated by this [slack thread](https://mongodb.slack.com/archives/GMY98JS86/p1706316771569039).

Using `alpine:3.19` may be preferable to `alpine:3` to prevent unexpected test failures from Alpine image updates. Pinning to a specific version enables intentional upgrade of the image.

Tested on an Ubuntu 22.04 spawn host:

```
cd extras/docker/alpine3.19
yes | docker system prune -a
make nocachebuild
make test
# Prints `THE alpine3.19 IMAGE WORKS!`
```

Attempting to test locally on my arm64 macOS resulted in an error building libmongocrypt fixed in MONGOCRYPT-601. MONGOCRYPT-601 is not yet released.


kkloberdanz (approved):
While I think that there is merit to always test against the latest Alpine image in CI, i.e., using `alpine3` as the base image, I don't have a strong opinion either way, so I'm fine with this too.

vector-of-bool (approved):

PR #1092, opened 2024-02-13
CXX-2832 Restore `VERSION_CURRENT` in release tarball

kevinAlbs:
# Summary

Restore the VERSION_CURRENT file in the tarball produced by the `dist` target.

# Background & Motivation

The 3.8.1 release tarball includes the file `build/VERSION_CURRENT`. The 3.9.0 release tarball (currently) does not include a `build/VERSION_CURRENT` file. This requires users to specify the version when building.

The missing `VERSION_CURRENT` file was discovered investigating CDRIVER-4834. It appears to be the result of changes in https://github.com/mongodb/mongo-cxx-driver/pull/1020.

Building the `dist` target on b133ac58aba263fc270901d67dba9fe0db3de229 results in the `VERSION_CURRENT` file being omitted:

```bash
#!/usr/bin/env bash

# Clear build directory
rm -rf build/*
cd build

# Make dist tarball similar to the release process.
echo "3.9.0-pre" > VERSION_CURRENT
cmake \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_PREFIX_PATH=/Users/kevin.albertson/code/c-bootstrap/install/mongo-c-driver-1.25.1 \
    -DENABLE_UNINSTALL=ON \
    ..
cmake --build . --target dist

# Untar dist tarball.
tar xf mongo-cxx-driver-r3.9.0-pre.tar.gz

# Check for VERSION_CURRENT
if [[ -f mongo-cxx-driver-r3.9.0-pre/build/VERSION_CURRENT ]]; then
    echo "VERSION_CURRENT found with contents $(cat mongo-cxx-driver-r3.9.0-pre/build/VERSION_CURRENT) Test passed"
else
    echo "VERSION_CURRENT not found. Test failed"
    exit 1
fi

# Test passes on 3.8.1. Test fails on commit b133ac58.
```

This PR changes the path to `VERSION_CURRENT` from absolute to relative to address this error:
```
make[3]: *** No rule to make target `Users/kevin.albertson/code/tasks/mongo-cxx-driver-restore-VERSION_CURRENT/build/VERSION_CURRENT', needed by `mongo-cxx-driver-r3.9.0-pre.tar.gz'.  Stop.`
```

The `distcheck` target has been updated to ensure `VERSION_CURRENT` is present.

Once this PR is merged, I plan to update the 3.9.0 release tarball to include the `VERSION_CURRENT` file.


cmake/make_dist/MakeDistCheck.cmake
```diff
@@ -22,6 +22,11 @@ function (RUN_DIST_CHECK PACKAGE_PREFIX EXT)
       ERROR_MSG "Command to untar ${tarball} failed."
    )
 
+   # Ensure a VERSION_CURRENT file is present.
+   if(NOT EXISTS ${PACKAGE_PREFIX}/build/VERSION_CURRENT)
+      message (FATAL_ERROR "Expected tarball to include `build/VERSION_CURRENT` file but did not.")
```
eramongodb, line 27:
```suggestion
      message (FATAL_ERROR "Expected tarball to contain a `build/VERSION_CURRENT` file, but it does not")
```
Wording tweak suggestion.

cmake/make_dist/MakeDistCheck.cmake
```diff
@@ -22,6 +22,11 @@ function (RUN_DIST_CHECK PACKAGE_PREFIX EXT)
       ERROR_MSG "Command to untar ${tarball} failed."
    )
 
+   # Ensure a VERSION_CURRENT file is present.
+   if(NOT EXISTS ${PACKAGE_PREFIX}/build/VERSION_CURRENT)
```
eramongodb, line 26: This check should be guarded by the same `"${CMAKE_BINARY_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}/build"` condition used to determine if the VERSION_CURRENT file should be included in the tarball. Otherwise, distcheck fails when `${CMAKE_BINARY_DIR}` is not `./build`.

Suggest defining a variable (i.e. `MONGOCXX_INCLUDE_VERSION_FILE_IN_DIST`) to track this condition:

```cmake
set(MONGOCXX_INCLUDE_VERSION_FILE_IN_DIST OFF)
set(MONGOCXX_CURRENT_VERSION_FILE "")

# ...

        if("${CMAKE_BINARY_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}/build")
            set(MONGOCXX_INCLUDE_VERSION_FILE_IN_DIST ON)
            set(MONGOCXX_CURRENT_VERSION_FILE build/VERSION_CURRENT)
        endif ()

# ...
```
```cmake
# Ensure a VERSION_CURRENT file is present.
if(MONGOCXX_INCLUDE_VERSION_FILE_IN_DIST AND "${CMAKE_BINARY_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}/build")
   if(NOT EXISTS ${PACKAGE_PREFIX}/build/VERSION_CURRENT)
      message (FATAL_ERROR "Expected tarball to include `build/VERSION_CURRENT` file but did not.")
   endif ()
endif()
```

eramongodb: Fixed suggestion to omit redundant condition.

kevinAlbs: Thank you for the suggestion. Updated as suggested.

kevinAlbs (commented):
Filed CXX-2834 to propose including a VERSION_CURRENT file in the source. That may help prevent the possibility of building a tarball without a VERSION_CURRENT file.

CMakeLists.txt
```diff
@@ -238,13 +238,20 @@ set(CMAKE_MODULE_PATH
 include(GNUInstallDirs)
 include(ParseVersion)
 
+set(MONGOCXX_INCLUDE_VERSION_FILE_IN_DIST OFF)
 set(MONGOCXX_CURRENT_VERSION_FILE "")
 
 if(BUILD_VERSION STREQUAL "0.0.0")
     if(EXISTS ${CMAKE_BINARY_DIR}/VERSION_CURRENT)
         file(STRINGS ${CMAKE_BINARY_DIR}/VERSION_CURRENT BUILD_VERSION)
```
eramongodb, line 246: Should this `file()` be moved into the `"${CMAKE_BINARY_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}/build"` branch below?

eramongodb (approved):

CMakeLists.txt
kevinAlbs: I expect "no" to preserve the existing behavior of preferring the `${CMAKE_BINARY_DIR}/VERSION_CURRENT`. CXX-2834 has been updated to propose simplifying how the version is obtained.

kevinAlbs:
Tarball on https://github.com/mongodb/mongo-cxx-driver/releases/tag/r3.9.0 has been updated.

PR #1102, opened 2024-02-28
CXX-2967 Document the Stable ABI Compatibility Policy

eramongodb:
Resolves CXX-2967. This is a documentation-only PR.

Running `hugo server` in the `docs` directory to live-preview the changes in a web browser is recommended.

This PR aims to expand and clarify our API and ABI compatibility policies. This will serve as a reference for all stable ABI changes going forward. This should be established before we set the ABI version number to its first integer value and commit to ABI stability. The API compatibility policy is updated accordingly.

Notable points include:

* Explicit definitions of what constitutes the "public API" and "stable ABI" for which we are committed to communicating compatibility for with the API and ABI version numbers respectively (see quotes from SemVer and Drepper).
  * Exclude most build system properties from the public API (e.g. library and package filenames) while making an effort to keep sensible ones included (e.g. `CMAKE_CXX_STANDARD`).
  * Include the soname as part of the stable ABI (e.g. library filename), except on Windows (where the soname convention doesn't apply).
  * Exclude entities and symbols explicitly documented as "private", "experimental", etc. so features can be added and developed without immediately committing to stability (or so they can be "soft-removed" via documentation without physically breaking compatibility).
* Describe the expected evolution of the API and ABI, including deprecation and removal, and when a given change requires updating the API and ABI version numbers.
* Document the existence and purpose of root namespace redeclarations w.r.t. ABI namespaces as well as the ABI namespace directory structure of public headers.
* Improve tips for using CMake package config and pkg-config files, strongly recommending use of CMake package config files, as well as discouraging manual configuration (although not outright declaring it unsupported).
* Improve MSVC-only documentation to cover package files and static libraries.

The stable ABI compatibility policy should be compatible with shared library versioning policies for mainstream package distributions. Both the ABI and API stability policy should be consistent with established practice for the CXX Driver, but refined to clearly set boundaries on what we consider in-scope and out-of-scope w.r.t. stability (such as build system properties).

### DIFF
--- a/docs/content/mongocxx-v3/api-abi-versioning.md
+++ b/docs/content/mongocxx-v3/api-abi-versioning.md
@@ -113,7 +113,7 @@ cmake -S example -B example-build -D bsoncxx_ROOT="$INSTALLDIR"
 Although the CXX Driver also provides pkg-config files, users are strongly encouraged to use CMake package config files instead.
 
 The name of the pkg-config file for bsoncxx libraries is also controlled by `BSONCXX_OUTPUT_NAME` and follow the pattern `lib<basename>.pc`.
-Because the CXX Driver does not install pkg-config files to typical pkg-config system directories, the `PKG_CONFIG_PATH` environment variable may need to be used to augment pkg-config's search path, e.g.:
+The `PKG_CONFIG_PATH` environment variable may need to be used to augment pkg-config's search path, e.g.:
 
 ```bash
 bsoncxx_cflags="$(PKG_CONFIG_PATH="$INSTALLDIR" pkg-config --cflags "libbsoncxx >= 1.2.3")"
@@ -237,7 +237,7 @@ ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/
 
 ### Static Libraries (MSVC Only)
 
-The name of static libraries use the same pattern as regular (non-MSVC) static library behavior described above, but uses `static` as `<abi-version-number>` and the library output name as-if it were the basename.
+The name of static libraries use the same pattern as MSVC shared library behavior described above, but uses `static` as the `<abi-version-number>` and the library output name as-if it were the basename.
 
 For example, the following install prefix directory contains the expected shared _and_ static library files for a bsoncxx library (the bin directory is included to account for `.dll` files):
 
--- a/docs/content/mongocxx-v3/api-abi-versioning.md
+++ b/docs/content/mongocxx-v3/api-abi-versioning.md
@@ -186,9 +186,9 @@ ${CMAKE_INSTALL_PREFIX}/
 
 The CMake package config files are the same as the regular (non-MSVC) shared library behavior described above.
 
-**NOTE:** CMake automatically handles selection of libraries according to build type, but does not enforce build type consistency when only one build type is installed. Installing _both_ debug and release configurations on Windows highly recommended for this reason. This does not extend to _all_ build configuration parameters.
+**NOTE:** CMake automatically handles selection of libraries according to build type, but does not enforce build type consistency when only one build type is installed. Installing _both_ Debug and Release configurations on Windows is highly recommended for this reason. (This note only applies to the build type configuration parameter).
 
-The name of the pkg-config file for CXX Driver libraries is the same as the regular (non-MSVC) shared library behavior described above.
+The name of the pkg-config file for CXX Driver libraries is the same as that for the regular (non-MSVC) shared library behavior described above.
 However, when `ENABLE_ABI_TAG_IN_PKGCONFIG_FILENAMES=ON`, the library output name is used as-if it were the basename for pkg-config files.
 For example, to obtain flags for the shared library `bsoncxx-v_noabi-rhs-x64-v142-md.dll`, the pkg-config command may look as follows when `ENABLE_ABI_TAG_IN_PKGCONFIG_FILENAMES=OFF` (the default):
 
@@ -202,7 +202,7 @@ or as follows when `ENABLE_ABI_TAG_IN_PKGCONFIG_FILENAMES=ON`:
 pkg-config --cflags "libbsoncxx-v_noabi-rhs-x64-v142-md"
 ```
 
-Setting `ENABLE_ABI_TAG_IN_PKGCONFIG_FILENAMES=ON` is recommended when expected parallel installation of the bsoncxx library with different build configurations (e.g. Debug vs. Release).
+Setting `ENABLE_ABI_TAG_IN_PKGCONFIG_FILENAMES=ON` is recommended when parallel installation of CXX Driver libraries with different build configurations (e.g. Debug vs. Release) is expected.
 However, using CMake package config files instead is most recommended.
 
 ### Static Libraries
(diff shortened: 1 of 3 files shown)


docs/content/mongocxx-v3/policies/api-versioning.md
```diff
@@ -0,0 +1,94 @@
...
+* CMake configuration variables and corresponding behavior (with some exceptions: see below).
+* The name, location, and contents of:
+  * CMake source files (e.g. under `${PROJECT_SOURCE_DIR}`).
+  * CMake binary files (e.g. under `${PROJECT_BINARY_DIR}`).
+  * Installed library files (e.g. under `{CMAKE_INSTALL_LIBDIR}`).
+  * Installed package files (e.g. under `{CMAKE_INSTALL_LIBDIR}/pkgconfig`).
+
+However, build system configuration variables that directly impact the public API are considered part of the public API.
+
+For example, C++17 polyfills declared in the `bsoncxx::stdx` namespace depend on configuration variables that determine the polyfill library used, e.g. `CMAKE_CXX_STANDARD`, `BSONCXX_POLY_USE_STD`, etc..
```
rcsanchez97, line 64:
```suggestion
For example, C++17 polyfills declared in the `bsoncxx::stdx` namespace depend on configuration variables that determine the polyfill library used, e.g. `CMAKE_CXX_STANDARD`, `BSONCXX_POLY_USE_STD`, etc.
```

rcsanchez97 (commented):
I've made a few suggested typo/grammar/wording fixes.

As far as distro packaging, I don't see anything here but upside. This looks really solid and should give us a nice, clean, and maintainable way to manage the API/ABI which will make the packaging part work more smoothly.

docs/content/mongocxx-v3/api-abi-versioning.md
```diff
@@ -6,122 +6,252 @@ title = "API and ABI versioning"
...
 
-## Sonames and symlinks
+**NOTE:** The C Driver is a required dependency of CXX Driver. The CMake package config files for the C Driver may also need to be imported using the same (or similar) approach to satisfy this dependency.
 
-*Note that below examples are given for libmongocxx, but also apply to libbsoncxx*
+Although the CXX Driver also provides pkg-config files, users are strongly encouraged to use CMake package config files instead.
 
-*DSO = Dynamic Shared Object, to use Ulrich Drepper's terminology*
+The name of the pkg-config file for bsoncxx libraries is also controlled by `BSONCXX_OUTPUT_NAME` and follow the pattern `lib<basename>.pc`.
+Because the CXX Driver does not install pkg-config files to typical pkg-config system directories, the `PKG_CONFIG_PATH` environment variable may need to be used to augment pkg-config's search path, e.g.:
```
kevinAlbs, line 116:
```suggestion
The `PKG_CONFIG_PATH` environment variable may need to be used to augment pkg-config's search path, e.g.:
```

If the `CMAKE_INSTALL_PREFIX` override is disabled (with the recommended `MONGOCXX_OVERRIDE_DEFAULT_INSTALL_PREFIX=OFF`), I expect the package config file is in a default searched location:

```bash
cmake \
    -D CMAKE_CXX_STANDARD=17 \
    -D MONGOCXX_OVERRIDE_DEFAULT_INSTALL_PREFIX=OFF \
    -S. -B cmake-build
sudo cmake --build cmake-build --target install # Install /usr/local/lib/pkgconfig/libbsoncxx.pc
pkg-config --libs --cflags libbsoncxx # Locates libbsoncxx.pc.
```

docs/content/mongocxx-v3/api-abi-versioning.md
```diff
@@ -6,122 +6,252 @@ title = "API and ABI versioning"
...
+```
 
-// release 2 where the feature name is renamed
-MONGOCXX_DEPRECATED void do_thing();
-void do_stuff();
+### Static Libraries (MSVC Only)
 
-// release 3 where the original feature name is removed
-void do_stuff();
+The name of static libraries use the same pattern as regular (non-MSVC) static library behavior described above, but uses `static` as `<abi-version-number>` and the library output name as-if it were the basename.
```
kevinAlbs, line 240:
```suggestion
The name of static libraries use the same pattern as shared MSVC library behavior described above, but uses `static` as `<abi-version-number>` and the library output name as-if it were the basename.
```

kevinAlbs (commented):
I think this information will be very helpful to both users and C++ driver developers.

docs/content/mongocxx-v3/policies/api-versioning.md
```diff
@@ -0,0 +1,94 @@
...
+A _public header file_ is installed to the include directory for any given CMake build configuration. Any other header file is a _private header file_.
+
+A library's _root namespace_ is declared in the global namespace scope, e.g. `bsoncxx`.
+All entities owned by the library are declared within the library's root namespace.
+Preprocessor macros owned by the library are prefixed with the library name instead, e.g. `BSONCXX_EXAMPLE_MACRO`.
+
+The _public API_ is the set of _documented_ entities that are declared or defined in a public header file within the library's root namespace, with the following exceptions:
+
+* The entity is explicitly documented as private or not intended for public use.
+* The entity is declared in a `detail` namespace.
```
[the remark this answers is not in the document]
eramongodb: Oh, right, that exists. 🙃 

I certainly don't think it's _intended_ to be part of the public API, but of course, given it is in a public header, we can't entirely rule out the possibility that someone's using it. We could attempt to remove it in a minor release and if anyone complains, re-add it in a patch release.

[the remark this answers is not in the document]
kevinAlbs: Filed CXX-2978 to consider future removal.

kevinAlbs (approved):

rcsanchez97 (approved):

eramongodb:
Added documentation concerning how we handle symbol visibility (hidden by default; only explicitly exported symbols are exported).

Importantly, changed the note regarding ABI symbol behavior such that behavior _is_ considered part of the stable ABI. This is to ensure consistency across shared libraries with the same ABI version number (substitutability). However, the bounds of behavioral compatibility is still defined by the documented behavior of public API entities that (indirectly) use the exported symbol. (We don't need to commit to preserving compatibility of behavior that is not observable via the public API.)

docs/content/mongocxx-v3/policies/abi-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+
+The ABI version number is bumped whenever there is a backward incompatible change to the stable ABI as defined below.
+
+**IMPORTANT:** Only the stability of the stable ABI is communicated by the ABI version number. Backward (in)compatible changes to the unstable ABI are not communicated by the ABI version number.
+
+**NOTE:** The ABI versioning policy is only applicable to _shared libraries_. It does not apply to static libraries.
+
+## Binary Compatibility
+
+An _ABI namespace_ is declared within the library's root namespace (see [Source Compatibility]({{< relref "api-versioning#source-compatibility" >}})), prefixed with the letter `v` and followed by either an integer or `_noabi`.
```
rachel-mack, line 33: Where is this supposed to link to?

eramongodb: It links to the new "Source Compatibility" header in `api-versioning.md`: https://github.com/mongodb/mongo-cxx-driver/pull/1102/files#diff-c5a6c43199b7f312d661e08521043dc28fc1341a58ade10b1402a74fcb0453c0R26

vector-of-bool (approved):
Apologies for the lengthy delay. Good work!

PR #1131, opened 2024-05-06
Update README.md

undefinedfunction:
The links in README.md were outdated. Updated them to what I think are the correct replacements.


kevinAlbs (approved):
Thank you @undefinedfunction.

PR #1134, opened 2024-05-08
CXX-3021 Add SBOM Lite and update release instructions

eramongodb:
Resolves CXX-3021. No CI automation is included in this PR. Instead, the release instructions include commands documenting how to update the SBOM Lite file.

Three dependencies are identified:

- C Driver (via FetchContent)
- mnmlstc/core (via FetchContent)
- Catch2 (bundled)

No attempt is made to conditionally specify SBOM Lite according to the build configuration (specifically concerning mnmlstc/core), if such a pattern is even possible with PURL.

To better assist with specifying/documenting dependencies of this kind, all FetchContent-related routines are moved as-is into `Fetch*` modules under the `cmake/` subdirectory which are then referenced by `etc/purls.txt`. Note this required moving up the `set(CMAKE_MODULE_PATH ...)` higher in the root CMakeLists.txt file.


etc/purls.txt
```diff
@@ -0,0 +1,15 @@
...
+# re-generate the SBOM JSON file!
+
+# libbson and libmongoc are obtained via cmake/FetchMongoC.cmake.
+pkg:github/mongodb/mongo-c-driver@v1.25.0
+
+# mnmlstc/core is obtained via cmake/FetchMnmlstcCore.cmake.
+pkg:github/mnmlstc/core@v1.1.0
+
+# Catch2 is bundled as src/third_party/catch/include/catch.hpp.
+pkg:github/catchorg/Catch2/@v2.13.7#single_include/catch2/catch.hpp
```
kevinAlbs, line 15: Suggest removing Catch2 due to being a test dependency.

From [Centralized Vulnerability Management README](https://docs.google.com/document/d/18DZhrMvdxlFJJS7KaZB1gBNIKispX1W4rlTD-16yGI8/edit):

>  Test- and build-time dependencies (such as the compiler) do not need to be declared.

> We only need to report and fix CVEs in dependencies we actually ship, so reducing any false-positives results in less auditing and updating overhead.

kevinAlbs (approved):

PR #1136, opened 2024-05-14
CXX-3010 sign and upload dist tarball during release

eramongodb:
Resolves CXX-3010.

This PR adds a new `garasign_dist_file.sh` script invoked by `make_release.py` after generating the distribution file that creates a corresponding `.asc` file that may be used to verify the tarball as done at the end of the script. This `.asc` file is then uploaded alongside the tarball during GitHub release draft creation. A new `~/.secrets/garasign-creds.txt` is added to the release instructions by which Artifactory and Garasign credentials may be provided to the script (matching existing pattern for obtaining/providing credentials for GitHub and JIRA).


.evergreen/garasign_dist_file.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Used by make_release.py.
+# See: https://docs.devprod.prod.corp.mongodb.com/release-tools-container-images/garasign/garasign_signing/
+
+set -o errexit
+set -o pipefail
+
+: "${1:?"missing dist_file as first argument"}"
+
+if ! command -v podman >/dev/null; then
```
kevinAlbs, line 11: I (selfishly) request adding `docker` as a fallback:

```sh
declare launcher
if command -v podman >/dev/null; then
  launcher=podman
elif command -v docker >/dev/null; then
  launcher=docker
else
  echo "podman or docker is required to sign distribution tarball" 1>&2
fi
```

At present, I am unable to install `podman`:
```sh
podman: This software does not run on macOS versions older than Ventura.
Error: podman: An unsatisfied requirement failed this build.
```

I expect that is a short-term issue (I can upgrade), but this help me to release in the short-term.

.evergreen/garasign_dist_file.sh
kevinAlbs, line 1: Suggest relocating file to `etc` directory since script is not run in Evergreen.

etc/make_release.py
```diff
@@ -650,7 +654,9 @@ def create_github_release_draft(gh_repo,
     gh_release = gh_repo.create_git_release(tag=release_tag, name=release_name,
```
kevinAlbs, line 654: Suggest modifying the `generate_release_notes` function to document signature verification:

```python
    footer = textwrap.dedent("""
    ## Feedback
    To report a bug or request a feature, please open a ticket in the MongoDB issue management tool Jira:

    - [Create an account](https://jira.mongodb.org) and login.
    - Navigate to the [CXX project](https://jira.mongodb.org/browse/CXX)
    - Click `Create`.
                             
    ## Signature verification
    Use `gpg` to verify the signature. The public key for `cpp-driver` is available on https://pgp.mongodb.com/.
    """).lstrip()
```



This suggestion is intended to satisfy the requirement noted in DRIVERS-2895:

> Drivers MUST provide documentation for users to verify release artifacts if they wish

.evergreen/garasign_dist_file.sh
```diff
@@ -0,0 +1,63 @@
...
+  -e "PLUGIN_COMMANDS=${plugin_commands[*]:?}" \
+  --rm \
+  -v "$(pwd):$(pwd)" \
+  -w "$(pwd)" \
+  artifactory.corp.mongodb.com/release-tools-container-registry-local/garasign-gpg
+
+# This is the key used to sign C++ Driver release artifacts.
+# This should match the key downloaded from pgp.mongodb.com below, otherwise
+# verification will fail due to no matching (trusted) public key.
+key=DC7F679B8A34DD606C1E54CAC4FC994D21532195
```
kevinAlbs, line 58: Confirming: is including `key` intended for documentation only (currently unused)? If so, suggest putting in a comment:

```sh
# Public key ID is DC7F679B8A34DD606C1E54CAC4FC994D21532195
```

kevinAlbs (commented):
I like the added step for verifying the signature. That may prevent accidental signing with another key.

.evergreen/garasign_dist_file.sh
eramongodb: Artifact of prior iterations that used `gpg --verify` instead of `gpgv`. Removed.

.evergreen/garasign_dist_file.sh
eramongodb: Added support for a `GARASIGN_LAUNCHER` env var that may be defined as a means to select an alternative to `podman`.

kevinAlbs (approved):

PR #1139, opened 2024-05-21
CXX-3008 add instructions to examine Snyk reports during release

eramongodb:
Related to CXX-3008. Additionally updates `etc/requirements.txt` with output of `pip freeze` to ensure dependencies are pinned, which should resolve some open Snyk issues.

### DIFF
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -21,7 +21,7 @@ by changes in the new release.
 
 ## Check Coverity
 
-Ensure there are no new, unexpected, or high severity issues on Coverity.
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
LGTM

> Additionally updates `etc/requirements.txt` with output of `pip freeze` to ensure dependencies are pinned, which should resolve some open Snyk issues.

I expect issues reported from `etc/requirements.txt` do not need to be resolved since they are not required by users of the shipped product. Pinning seems like an improvement nonetheless.

PR #1142, opened 2024-06-03
CXX-3043 Address Coverity warnings for Ro3/Ro5 violations

eramongodb:
Resolves CXX-3043. Verified by [this patch](https://spruce.mongodb.com/version/665dfca13ef01a0007f9530f).

This PR _explicitly_ defaults or deletes special member functions (SMFs) which are currently _implicitly_ defaulted or deleted, in accordance with the Rule of All or Nothing. This addresses multiple Coverity issues as well as a few `-Wdeprecated-copy-with-*` Clang warnings (depending on implicit default of copy SMFs given a user-declared dtor, copy ctor, or copy assign op).

This PR should NOT change either the API or the ABI. This PR should NOT change the behavior at either compile-time or runtime, with exactly one exception: `index_model`'s explicitly deleted copy assignment operator is now defaulted. It is unclear why this function was deleted when all other SMFs are (out-of-line) defaulted. This change is not expected to break user code.

Some additional notes worth mentioning:

- `= default` within class definition does not require `(BSONCXX|MONGOCXX)_INLINE` per `VISIBILITY_INLINES_HIDDEN=ON`, as class member functions defined within the class definition are implicitly `inline` (even with `= default`), so all instances of such redundant inline macro usage are removed for consistency.
- The quoted bug with [inline dtor default definition behavior for VS2015](https://github.com/mongodb/mongo-cxx-driver/commit/025ebb2df0a8c8ee2e100dce55a78aae7b4481b4) in `concatenate.hpp` is no longer an issue (AFAICT).
- Removed redundant explicit `noexcept` for defaulted/deleted SMFs. This is only applicable to SFMs that are defaulted _on first declaration_, thus SMFs defaulted out-of-line still require explicit `noexcept`.
- Deleted SMFs that are unnecessary/unused internally (i.e. impl classes).
- Inconsistencies in inline vs. out-of-line SMF defaults are expected to be resolved in the near future via v1 namespace interfaces. This PR favors avoiding the introduction of new ABI symbols; if consistency is preferred, new (explicitly defaulted) SMFs can be defined out-of-line instead (this can also be done in a followup PR).

### DIFF
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/events/heartbeat_started_event.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/events/heartbeat_started_event.hpp
@@ -42,6 +42,13 @@ class heartbeat_started_event {
     ///
     ~heartbeat_started_event();
 
+    heartbeat_started_event(heartbeat_started_event&&) = default;
+    MONGOCXX_INLINE heartbeat_started_event& operator=(heartbeat_started_event&&) noexcept =
+        default;
+
+    heartbeat_started_event(const heartbeat_started_event&) = default;
+    heartbeat_started_event& operator=(const heartbeat_started_event&) = default;
+
     ///
     /// Returns the host name.
     ///
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/events/heartbeat_started_event.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/events/heartbeat_started_event.hpp
@@ -43,8 +43,7 @@ class heartbeat_started_event {
     ~heartbeat_started_event();
 
     heartbeat_started_event(heartbeat_started_event&&) = default;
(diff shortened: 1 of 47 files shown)


src/mongocxx/include/mongocxx/v_noabi/mongocxx/events/heartbeat_started_event.hpp
```diff
@@ -42,6 +42,13 @@ class heartbeat_started_event {
     ///
     ~heartbeat_started_event();
 
+    heartbeat_started_event(heartbeat_started_event&&) = default;
+    MONGOCXX_INLINE heartbeat_started_event& operator=(heartbeat_started_event&&) noexcept =
```
kevinAlbs, line 46:
```suggestion
   heartbeat_started_event& operator=(heartbeat_started_event&&) noexcept =
```

kevinAlbs (approved):
LGTM with one comment addressed.

PR #1147, opened 2024-06-14
CXX-2840 Docs example request: Projection with aggregation expressions

laurelxiang:
[evergreen](https://spruce.mongodb.com/version/666c7eef283b1a00075c6a70/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

[ticket](https://jira.mongodb.org/browse/CXX-2840)

Added docs example for projection with aggregation. Expanded server version check within the documentation_examples file and moved it to the top of the file for easier use. Fixed a broken link in the examples/README.

### DIFF
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -210,7 +210,10 @@ static bool is_replica_set(const mongocxx::client& client) {
     return static_cast<bool>(reply.view()["setName"]);
 }
 
-static bool version_at_least(mongocxx::database& db, int minimum_major, int minimum_minor, int minimum_patch) {
+static bool version_at_least(mongocxx::database& db,
+                             int minimum_major,
+                             int minimum_minor,
+                             int minimum_patch) {
     using bsoncxx::builder::basic::kvp;
     using bsoncxx::builder::basic::make_document;
 
@@ -227,11 +230,9 @@ static bool version_at_least(mongocxx::database& db, int minimum_major, int mini
         }
         if (split == 0) {
             major_string += i;
-        }
-        else if (split == 1) {
+        } else if (split == 1) {
             minor_string += i;
-        }
-        else if (split == 2) {
+        } else if (split == 2) {
             patch_string += i;
         }
     }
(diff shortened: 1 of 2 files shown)


examples/mongocxx/mongodb.com/documentation_examples.cpp
```diff
@@ -187,6 +210,40 @@ static bool is_replica_set(const mongocxx::client& client) {
...
+            minor_string += i;
+        } else if (split == 2) {
+            patch_string += i;
+        }
+    }
+    int server_major = std::stoi(major_string);
+    int server_minor = std::stoi(minor_string);
+    int server_patch = std::stoi(patch_string);
+
+    return server_major >= minimum_major && server_minor >= minimum_minor &&
```
kkloberdanz, line 243: I believe this function has a bug in it. Suppose we are running server version 7.0.0, and we are setting a minimum version of 4.4.0. The return statement as written would do the following comparisons below. We can see that the return will be `false` because even though the major version is successfully checked to be greater, the minor version of the 7.0.0 version is 0, which is less than 4, and therefore the entire statement will return false.
```
return (7 >= 4) && (0 >= 4) && (0 >= 0);
```

kkloberdanz (changes requested):

kkloberdanz (changes requested):
I believe that this function still has the same bug when performing the semantic version check. For example, I put together s small C++ program that uses the logic from this function, but takes a string for the server version so you can experiment with different scenarios. When I set the server version to 7.0.0 and the minimum version to 4.4.0, the minor version check fails, which we see as an assertion failure in the main function below. Perhaps try using something like this to experiment with how to get this logic working and try it with a few different versions to verify that it works.

```
#include <iostream>
#include <string>
#include <cassert>

static bool version_at_least(const std::string &version,
                             int minimum_major,
                             int minimum_minor,
                             int minimum_patch) {
    std::string major_string;
    std::string minor_string;
    std::string patch_string;
    int split = 0;
    for (auto i : version) {
        if (i == '.') {
            split += 1;
            continue;
        }
        if (split == 0) {
            major_string += i;
        } else if (split == 1) {
            minor_string += i;
        } else if (split == 2) {
            patch_string += i;
        }
    }
    int server_major = std::stoi(major_string);
    int server_minor = std::stoi(minor_string);
    int server_patch = std::stoi(patch_string);

    if (server_major < minimum_major) {
        return false;
    } else if (server_minor < minimum_minor) {
        return false;
    } else if (server_patch < minimum_patch) {
        return false;
    }

    return true;
}

int main(void) {
    assert(version_at_least("7.0.0", 4, 4, 0));
}
```

kkloberdanz (approved):
Looks good!

PR #1148, opened 2024-06-14
CXX-3008 Add post-release instructions for Silk, Snyk, and patch release tags

eramongodb:
Adds additional steps following creation of a new release branch `releases/vX.Y` to ensure Silk and Snyk are aware of, and continue analyzing, the release branch.

Silk commands defer to scripts in the C Driver introduced by https://github.com/mongodb/mongo-c-driver/pull/1619 to minimize redundancy. The CXX Driver will ideally adopt and reuse the same scripts as those used by libmongocrypt and the C Driver.

Snyk commands use the changes proposed in https://github.com/mongodb/libmongocrypt/pull/820 as reference. Note: Snyk does not appear to be able to detect the mnmlstc/core library despite `--detection-depth`, so a [Snyk support ticket](https://support.snyk.io/hc/en-us/requests/new) may be necessary.

Additionally, a new "merge back into `master`" branch is added to release instructions to ensure patch release tags are reachable from the latest commit in `master`. This enables support for accurate `git describe --tags` results as is used by Coverity nightly snapshot labels, abi-compliance-checker HTML compatibility reports, and eventually the `calc_release_version.py` script (significant simplifications incoming). This step is not expected to affect the current implementation of `calc_release_version.py`: it should continue to return the correct latest tag as intended.

### DIFF
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -230,6 +230,10 @@ The following credentials are required. Ask for these from a team member if nece
   GRS_CONFIG_USER1_USERNAME=<username>
   GRS_CONFIG_USER1_PASSWORD=<password>
   ```
+- Snyk credentials. Save these to `~/.secrets/snyk-creds.txt`:
+  ```bash
+  SNYK_API_TOKEN=<token>
+  ```
 
 Run the release script with the git tag created above as an argument and
 `--dry-run` to test for unexpected errors.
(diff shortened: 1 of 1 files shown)


eramongodb:
Snyk support ticket for mnmlstc/core detection support [has been submitted](https://support.snyk.io/hc/en-us/requests/83046).

etc/releasing.md
```diff
@@ -310,6 +314,41 @@ git push --set-upstream origin releases/v1.2
 
 The new branch should be continuously tested on Evergreen. Update the "Display Name" and "Branch Name" of the [mongo-cxx-driver-latest-release Evergreen project](https://spruce.mongodb.com/project/mongo-cxx-driver-latest-release/settings/general) to refer to the new release branch.
 
+## Update Silk and Snyk with new branch if necessary
+
+After creating the new minor release branch in the prior step, update Silk and Snyk to trach the new release branch.
+
+For Silk, see the [silk-create-asset-group.sh script](https://github.com/mongodb/libmongocrypt/blob/master/etc/silk-create-asset-group.sh) in the C Driver as reference for the commands to run against the Silk API. Use `mongo-cxx-driver` as the name and prefix in place of `mongo-c-driver` accordingly.
```
kevinAlbs, line 321: Step and PR description refers to "C driver" but links to libmongocrypt. If C driver is intended, suggest updating:

```suggestion
For Silk, use the [create-silk-asset-group.py script](https://github.com/mongodb/mongo-c-driver/blob/master/tools/create-silk-asset-group.py) in the C Driver to create a new Silk asset group.
```

etc/releasing.md
```diff
@@ -230,6 +230,10 @@ The following credentials are required. Ask for these from a team member if nece
   GRS_CONFIG_USER1_USERNAME=<username>
   GRS_CONFIG_USER1_PASSWORD=<password>
   ```
+- Snyk credentials. Save these to `~/.secrets/snyk-creds.txt`:
+  ```bash
+  SNYK_API_TOKEN=<token>
+  ```
```
kevinAlbs, line 236: Suggest adding the `SILK_CLIENT_ID` and `SILK_CLIENT_SECRET` needed for "Update Silk" steps below.

etc/releasing.md
eramongodb: Good catch.

etc/releasing.md
eramongodb: Done.

kevinAlbs (approved):

vector-of-bool (approved):
LGTM

PR #1150, opened 2024-06-18
CXX-2742 Consider addition of empty() to mongocxx::bulk_write

laurelxiang:
[CXX-2742](https://jira.mongodb.org/browse/CXX-2742)
Consider addition of empty() to mongocxx::bulk_write 

[evergreen](https://spruce.mongodb.com/version/6671b7c166038200070e2eef/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

Added empty() to mongocxx::bulk_write and added a test for this call.

### DIFF
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
@@ -58,6 +58,8 @@ class bulk_write {
     ///
     ~bulk_write();
 
+    bool empty();
+
     ///
     /// Appends a single write to the bulk write operation. The write operation's contents are
     /// copied into the bulk operation completely, so there is no dependency between the life of an
@@ -104,6 +106,8 @@ class bulk_write {
 
     bool _created_from_collection;
     std::unique_ptr<impl> _impl;
+
+    bool is_empty = true;
 };
 
 }  // namespace v_noabi
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
@@ -61,9 +61,9 @@ class bulk_write {
     ///
     /// Checks if a bulk write operation is empty.
     ///
-    /// @return A boolean that returns true if the bulk write operation is empty.
+    /// @return A boolean indicating if the bulk write operation is empty.
     ///
-    bool empty();
+    bool empty() const noexcept;
 
     ///
     /// Appends a single write to the bulk write operation. The write operation's contents are
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
@@ -58,6 +58,11 @@ class bulk_write {
     ///
     ~bulk_write();
 
+    ///
+    /// Checks if a bulk write operation is empty.
+    ///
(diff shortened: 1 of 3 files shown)


src/mongocxx/test/bulk_write.cpp
```diff
@@ -351,4 +352,15 @@ TEST_CASE("passing write operations to append calls corresponding C function", "
...
 }
+
+TEST_CASE("calling empty on a bulk write before and after appending", "[bulk_write]") {
+    instance::current();
+    mongocxx::client client{mongocxx::uri{}, test_util::add_test_server_api()};
+    auto bw = client["db"]["coll"].create_bulk_write();
+
+    REQUIRE(bw.empty());
+    bw.append(model::insert_one(make_document(kvp("_id", 1))));
+    bw.execute();
```
kkloberdanz, line 363: It might be nice to have a check after the append and before executing, since what I understand as the main use case being to check if a bulk write is not empty before executing it.
```suggestion
    bw.append(model::insert_one(make_document(kvp("_id", 1))));
    REQUIRE_FALSE(bw.empty());
    bw.execute();
```

kkloberdanz (changes requested):

kkloberdanz:
I think things look good 👍 Since this will affect the public API, I think it would be best if we get one more reviewer on this. I'll add Kevin to the list of reviewers.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
```diff
@@ -58,6 +58,13 @@ class bulk_write {
     ///
     ~bulk_write();
 
+    ///
+    /// Checks if a bulk write operation is empty.
+    ///
+    /// @return A boolean that returns true if the bulk write operation is empty.
+    ///
+    bool empty();
```
kevinAlbs, line 66:
```suggestion
    bool empty() const noexcept;
```

Use `noexcept` since no exceptions are thrown. Agrees with existing patterns in public methods.

Use `const` since object is not modified.

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/bulk_write.cpp
```diff
@@ -40,6 +40,10 @@ bulk_write& bulk_write::operator=(bulk_write&&) noexcept = default;
 
 bulk_write::~bulk_write() = default;
 
+bool bulk_write::empty() {
```
kevinAlbs, line 43:
```suggestion
bool bulk_write::empty() const noexcept {
```

kevinAlbs (approved):
LGTM with addition of `const noexcept`. When merging, use an imperative for the commit message. Example: `Add empty() to mongocxx::bulk_write`

See [CONTRIBUTING.md > Commit Messages](https://github.com/mongodb/mongo-cxx-driver/blob/6495c3a933bc97a68c6613c23e111f93046587f0/CONTRIBUTING.md#commit-messages)

kkloberdanz (approved):

src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
```diff
@@ -58,6 +58,13 @@ class bulk_write {
     ///
     ~bulk_write();
 
+    ///
+    /// Checks if a bulk write operation is empty.
+    ///
+    /// @return A boolean that returns true if the bulk write operation is empty.
```
kevinAlbs, line 64:
```suggestion
    /// @return A boolean indicating if the bulk write operation is empty.
```

Suggest tweaking wording to avoid "boolean that returns".

PR #1152, opened 2024-06-21
CXX-1843: Provide a cleaner API for working with pool entries

Julia-Garland:
Overload operator[] for obtaining the underlying database of a pool entry.

### DIFF
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/pool.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/pool.hpp
@@ -17,11 +17,12 @@
 #include <functional>
 #include <memory>
 
-#include <mongocxx/client-fwd.hpp>
 #include <mongocxx/options/auto_encryption-fwd.hpp>
 #include <mongocxx/pool-fwd.hpp>
 
 #include <bsoncxx/stdx/optional.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/database.hpp>
 #include <mongocxx/options/pool.hpp>
 #include <mongocxx/stdx.hpp>
 #include <mongocxx/uri.hpp>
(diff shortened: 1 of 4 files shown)


src/mongocxx/include/mongocxx/v_noabi/mongocxx/pool.hpp
```diff
@@ -96,6 +96,11 @@ class pool {
         /// Return true if this entry has a client acquired from the pool.
         explicit operator bool() const noexcept;
 
+        // Allows the pool_entry["db_name"] syntax to be used to access a database within the entry's underlying client.
+        mongocxx::v_noabi::database operator[](bsoncxx::v_noabi::string::view_or_value name) const&;
```
kevinAlbs, line 100:
```suggestion
        MONGOCXX_INLINE mongocxx::v_noabi::database operator[](bsoncxx::v_noabi::string::view_or_value name) const&;
```

Suggest marking function as inline. This is consistent with `mongocxx::client::operator[]`. Marking as inline may [enable compiler optimizations](https://learn.microsoft.com/en-us/cpp/cpp/inline-functions-cpp?view=msvc-170#when-to-use-inline-functions).

Marking the function inline requires the function definition be present in this file so the compiler can access the function definition at the compile step:

```c++
MONGOCXX_INLINE mongocxx::v_noabi::database pool::entry::operator[](
    bsoncxx::v_noabi::string::view_or_value name) const& {
    return (**this)[name];
}
```

Consequently, this appears to require definitions of `mongocxx::database` and `mongocxx::client`. Add includes:
```c++
#include <mongocxx/client.hpp>
#include <mongocxx/database.hpp>
```

And remove the forwarding header `#include <mongocxx/client-fwd.hpp>`.

kevinAlbs (changes requested):

Julia-Garland:
When I make these changes and remove the function definition from pool.cpp, the test cases run properly but when I run the pool example (examples/pool.cpp) after the three threads are run it gives the message `terminate called recursively \\ terminate called recursively \\ Aborted (core dumped)` which I isolated to `for (auto&& runner : threads) { runner.join(); }` (lines 72-74).

The same issue happens when I revert to the original code, explicitly dereferencing client. Although now it also sometimes says` terminate called after throwing an instance of 'mongocxx::v_noabi::bulk_write_exception`.

kevinAlbs (approved):

kkloberdanz (approved):

PR #1166, opened 2024-07-03
CXX-2745 apply precursor changes to Doxygen and CMake configs

eramongodb:
## Summary

Precursor changes related to CXX-2745 concerning Doxygen and CMake configs. Verified by [this patch](https://spruce.mongodb.com/version/66859525255dcf0007546b9b).

## Doxygen

In an effort to better align the Doxygen API documentation with our API compatibility policy, `SHOW_INCLUDE_FILES` is set to `NO`: we only want to reveal the presence of certain `#include` directives only when we deliberately support transitive inclusion via the documented header.

Similarly, `EXTRACT_LOCAL_CLASSES` is set to `NO`: we do not want to reveal details of internal and private implementation classes in the Doxygen API documentation. Only classes whose declarations or definitions are present in public API header files should be included.

Lastly, `INCLUDE_PATHS` is updated to be consistent with the install target include paths for bsoncxx and mongocxx by including the include prefix to permit `<bsoncxx/v<abi>/bsoncxx/header.hpp>` include patterns.

## CMake

To support the upcoming addition of `v1` files, some redundancy in CMake file lists is addressed by making use of `file(GLOB_RECURSE)` (already being used for `install()` commands) and a dedicated `(bsoncxx|mongocxx)_sources_v_noabi` variable.

Additionally, the macro guard tests are updated to explicitly exclude the prelude/postlude headers themselves to account for their eventual presence under multiple directories.


kevinAlbs (approved):

eramongodb:
Additionally sets `CLASS_GRAPH = NO` as it is observed to expose private inheritance relationships (e.g. `array::element` -> `document::element`). These graphs will be further complicated by upcoming v_noabi -> v1 private inheritance.

PR #1170, opened 2024-07-17
CXX-3064 Fix use of Doxygen @relates, @relatesalso, and groups

eramongodb:
## Summary

Resolves CXX-3064. Related to CXX-2827.

This PR is a parallel companion to https://github.com/mongodb/mongo-cxx-driver/pull/1169. Whereas https://github.com/mongodb/mongo-cxx-driver/pull/1169 focuses on the deployment process and tooling, this PR focuses on various fixes and improvements to the contents of the current API docs itself.

## The "Incompleteness" Problem

The Doxyfile sets `EXTRACT_ALL = NO`. This enables warnings for undocumented entities. However, this _also_ means directories and namespaces MUST be explicitly documented in order to properly generate their corresponding pages. This is the reason behind so many pages for directories and namespaces being completely devoid of useful information or missing a navigatable page entirely.

For example, this is the page for the `bsoncxx/v_noabi/bsoncxx` directory (note: subdirectory information is completely absent despite having pages):

image

Undocumented headers do not generate a page at all (hence their absence in directory pages as above):

image

Undocumented namespaces similarly do not generate a page at all:

image

This means the "array" in "bsoncxx > v_noabi > array > element" is not navigatable/interactable (among many similar cases of "dead" links):

image

These factors all contribute to a feeling of "incompleteness" in the quality of the API docs.

This PR lays the foundations to address all of the issues described above. However, the addition of documentation of directories, namespaces, and files are deferred to a followup PR due to the volume of changes required.

A preview of upcoming changes:

image

## Root Level Forward Headers

This PR adds a `bsoncxx/fwd.hpp` and `mongocxx/fwd.hpp` at the library root level, _not_ to be confused with `bsoncxx/v_noabi/bsoncxx/fwd.hpp` and `mongocxx/v_noabi/mongocxx/fwd.hpp` added by https://github.com/mongodb/mongo-cxx-driver/pull/1061. These headers are _NOT_ expected to be included or even includeable (`<bsoncxx/v_noabi/bsoncxx>` is prioritized before `<bsoncxx>` by include directory search paths, so these new `fwd.hpp` headers will never be found). Nevertheless, just in case, they are defined to behave equivalently to their `v_noabi` counterparts should they be included anyways.

These headers may be renamed to something else if `fwd.hpp` is confusing for their purpose.

The purpose of these headers is to provide a definitive location to add Doxygen documentation for root-level _directories_ and _namespaces_.

```
src/bsoncxx/include/ # Similarly for mongocxx.
└── bsoncxx/
    ├── fwd.hpp      # Documents bsoncxx/ and bsoncxx::.
    ├── v_noabi/bsoncxx/
    │   ├── fwd.hpp  # Documents bsoncxx/v_noabi and bsoncxx::v_noabi.
    │   └── ...
    └── v1/          # CXX-2745
        ├── fwd.hpp  # Documents bsoncxx/v1 and bsoncxx::v1.
        └── ...
```

https://github.com/mongodb/mongo-cxx-driver/pull/1039#issuecomment-1773264817 added some of these docs to the `prelude.hpp` headers, as forward headers were not yet implemented at the time. These are relocated to their corresponding forward headers.

The addition of documentation of directories, namespaces, and files are deferred to a followup PR due to the volume of changes required.

## Sorted Documentation

This PR proposes setting various `SORT_*` options to `YES`. This is primarily to automate the (alphabetical) grouping of related functions without requiring use of Doxygen groups. This will secondarily discourage writing documentation sensitive to in-source-declaration-order (e.g. "see above/below") allowing for greater freedom of in-source ordering without impacting API documentation results.

## `@relates`, `@relatesalso`, and Groups

After much pain and debugging, some rules were derived:

1. `@relatesalso` is _not_ equivalent to `@see`.
1. `@relatesalso` is not inherited by group members (`@relates` is).
2. Member group comments should not be their own doc comment block.
3. Documentation distributed to group members generally comes _before_ the `@{`, not after, except when `DISTRIBUTE_GROUP_DOC` applies instead (reuse doc of first group member for the rest).

This PR addresses all known violations of these rules by applying the following patterns (and generally avoiding use of groups when able):

```cpp
class example {
public:
  ///
  /// @relates example <!-- Inherited by group members. -->
  ///
  /// Common documentation for group members.
  ///
  /// @{ <!-- Asymmetric block for group marker. -->

  friend void group_member(int i);
  friend void group_member(double d);

  /// @} <!-- Asymmetric block for group marker. -->
  ///

  void unrelated(); // Asymmetry of the group marker comments avoid
                    // leaking the group documentation into the
                    // next entity when used in class definition scope.
};

///
/// Common documentation for group members.
///
/// @{

/// @relatesalso example <!-- Per group member. -->
inline void group_member(example e, int i);

/// @relatesalso example <!-- Per group member. -->
inline void group_member(example e, double d);

/// @} <!-- Asymmetric block for group marker. -->
///
```

It is important to use `@relatesalso` for non-member functions (especially operators) to ensure their declaration is included in the documentation of the enclosing namespace. Otherwise, they are only documented in the related class page. This is especially important for correctly documenting root namespace redeclarations in CXX-2745.

Use of per-member documentation should be avoided whenever possible to avoid the complexities of merging group-and-member documentations (`@param` commands in particular are a problem). Copying relevant documentation per member (e.g. `mongocxx::v_noabi::hint::operator==` overloads) is preferable to twisting contents to accomodate all group members.

## BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR

This (and its mongocxx counterpart) are added to facilitate workarounds for the very unfortunate Doxygen using-declaration problem described in https://github.com/mongodb/mongo-cxx-driver/pull/1066. These workarounds will come in a followup PR, but as a preview of their intended usage:

```cpp
namespace bsoncxx {
namespace array {

#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)

/// @ref v_noabi::array::element
class element {};

#else

using v_noabi::array::element;

#endif // defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)

} // namespace array
} // namespace bsoncxx
```

A preview of the results:

image

This will be discussed in greater detail in the relevant followup PR.

## STRIP_FROM_PATH = ON

Old:

image

New:

image

## Miscellaneous

- Set `BSONCXX_INLINE=inline` (+ mongocxx) to correctly communicate inline functions to Doxygen.
- Added ClangFormat `CommentPragmas` to avoid breaking long Doxygen commands by wrapping (e.g. `@copydoc`).
- Added a missing `@internal` in `util.hpp`.
- Fixed links to [Working with BSON](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/working-with-bson/#basic-builder).

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -2213,11 +2213,11 @@ INCLUDE_FILE_PATTERNS  =
 PREDEFINED             = BSONCXX_API= \
                          BSONCXX_CALL= \
                          BSONCXX_DEPRECATED \
-                         BSONCXX_INLINE= \
+                         BSONCXX_INLINE=inline \
                          MONGOCXX_API= \
                          MONGOCXX_CALL= \
-                         MONGOCXX_DEPRECATED \
-                         MONGOCXX_INLINE=
+                         MONGOCXX_DEPRECATED= \
+                         MONGOCXX_INLINE=inline
 
 # If the MACRO_EXPANSION and EXPAND_ONLY_PREDEF tags are set to YES then this
 # tag can be used to specify a list of macro names that should be expanded. The
--- a/Doxyfile
+++ b/Doxyfile
@@ -2214,10 +2214,12 @@ PREDEFINED             = BSONCXX_API= \
                          BSONCXX_CALL= \
                          BSONCXX_DEPRECATED \
                          BSONCXX_INLINE=inline \
+                         BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR= \
                          MONGOCXX_API= \
                          MONGOCXX_CALL= \
                          MONGOCXX_DEPRECATED= \
-                         MONGOCXX_INLINE=inline
+                         MONGOCXX_INLINE=inline \
+                         MONGOCXX_PRIVATE_DOXYGEN_PREPROCESSOR=
 
 # If the MACRO_EXPANSION and EXPAND_ONLY_PREDEF tags are set to YES then this
 # tag can be used to specify a list of macro names that should be expanded. The
--- a/Doxyfile
+++ b/Doxyfile
@@ -639,7 +639,7 @@ SORT_MEMBER_DOCS       = YES
 # this will also influence the order of the classes in the class list.
 # The default value is: NO.
 
-SORT_BRIEF_DOCS        = NO
+SORT_BRIEF_DOCS        = YES
 
 # If the SORT_MEMBERS_CTORS_1ST tag is set to YES then doxygen will sort the
 # (brief and detailed) documentation of class members so that constructors and
@@ -651,14 +651,14 @@ SORT_BRIEF_DOCS        = NO
 # detailed member documentation.
 # The default value is: NO.
 
-SORT_MEMBERS_CTORS_1ST = NO
+SORT_MEMBERS_CTORS_1ST = YES
 
 # If the SORT_GROUP_NAMES tag is set to YES then doxygen will sort the hierarchy
 # of group names into alphabetical order. If set to NO the group names will
 # appear in their defined order.
 # The default value is: NO.
 
-SORT_GROUP_NAMES       = NO
+SORT_GROUP_NAMES       = YES
 
 # If the SORT_BY_SCOPE_NAME tag is set to YES, the class list will be sorted by
 # fully-qualified names, including namespaces. If set to NO, the class list will
@@ -668,7 +668,7 @@ SORT_GROUP_NAMES       = NO
 # list.
 # The default value is: NO.
 
-SORT_BY_SCOPE_NAME     = NO
+SORT_BY_SCOPE_NAME     = YES
 
 # If the STRICT_PROTO_MATCHING option is enabled and doxygen fails to do proper
 # type resolution of all parameters of a function it will reject a match between
--- a/Doxyfile
+++ b/Doxyfile
@@ -160,7 +160,8 @@ FULL_PATH_NAMES        = YES
 # will be relative from the directory where doxygen is started.
 # This tag requires that the tag FULL_PATH_NAMES is set to YES.
 
-STRIP_FROM_PATH        =
+STRIP_FROM_PATH        = src/bsoncxx/include \
+                         src/mongocxx/include
 
 # The STRIP_FROM_INC_PATH tag can be used to strip a user-defined part of the
 # path mentioned in the documentation of a class, which tells the reader which
--- a/src/bsoncxx/include/bsoncxx/fwd.hpp
+++ b/src/bsoncxx/include/bsoncxx/fwd.hpp
@@ -1,0 +1,17 @@
+// Copyright 2009-present MongoDB, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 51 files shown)


src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
```diff
@@ -657,19 +657,172 @@ struct b_maxkey {
...
 /// free function comparator for b_maxkey
 ///
-/// @relatesalso b_maxkey
+/// @relatesalso bsoncxx::v_noabi::types::b_maxkey
 ///
 BSONCXX_INLINE bool operator==(const b_maxkey&, const b_maxkey&) {
     return true;
 }
 
+#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)
```
kevinAlbs, line 666: Instead of using `BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR`, consider defining the functions and removing the use of the `BSONCXX_ENUM` X-macro:

```c++
///
/// @relatesalso bsoncxx::v_noabi::types::b_double
///
/// free function comparator for b_double
///
BSONCXX_INLINE bool operator!=(const b_double& lhs, const b_double& rhs) {
    return !(lhs == rhs);
}
// And others ...
```

That may help avoid inconsistency between the documentation and the used declarations (e.g. the documented types do not currently include `BSONCXX_INLINE`).

kevinAlbs (commented):
The documentation improvements are greatly appreciated.

src/bsoncxx/include/bsoncxx/fwd.hpp
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// THIS FILE IS NOT INTENDED TO BE INCLUDABLE.
+#include <bsoncxx/v_noabi/bsoncxx/fwd.hpp>
+// THIS FILE IS NOT INTENDED TO BE INCLUDABLE.
```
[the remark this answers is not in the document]
eramongodb: Added the error directive + clarified the purpose of the root-level `.cpp` files to validate that the root-level fwd.hpp is not being included.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
eramongodb: Good call. Done as suggested.

Doxyfile
```diff
@@ -638,7 +639,7 @@ SORT_MEMBER_DOCS       = YES
 # this will also influence the order of the classes in the class list.
 # The default value is: NO.
 
-SORT_BRIEF_DOCS        = NO
+SORT_BRIEF_DOCS        = YES
```
kevinAlbs, line 642: Suggest removing "below" from [mongocxx::v_noabi::mongocxx::model::write::type()](https://github.com/mongodb/mongo-cxx-driver/blob/edc22ae9072106cce7b1b1fd2cd1ae275c2a5d98/src/mongocxx/include/mongocxx/v_noabi/mongocxx/model/write.hpp#L90-L91) since documentation sorts below other methods.

kevinAlbs (approved):
LGTM with one additional comment.

Doxyfile
eramongodb: Removed the order-dependent language + `get_*()` precondition documentation (it is already documented by the `get_*()` functions).

PR #1172, opened 2024-07-19
CXX-3067 test for existing examples target before add_custom_target

eramongodb:
Resolves CXX-3067. Companion to https://github.com/mongodb/mongo-c-driver/pull/1675.

Avoid the C Driver imported via `add_subdirectory()`, which defines the `examples` target, from conflicting with CXX Driver's attempt to create an `examples` target. Otherwise the following error may be observed:
```
CMake Error at examples/CMakeLists.txt:33 (add_custom_target):
  add_custom_target cannot create target "examples" because another target
  with the same name already exists.  The existing target is a custom target
  created in source directory
  "_deps/mongo-c-driver-src/src/libbson".
  See documentation for policy CMP0002 for more details.
```

If a `tests` target is ever added to the CXX Driver, it will need the existence check as well.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -30,7 +30,10 @@ add_subdirectory(bsoncxx)
 add_subdirectory(mongocxx)
 add_subdirectory(projects)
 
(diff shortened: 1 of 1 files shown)


eramongodb:
No longer necessary due to prefixed target names by C Driver: https://github.com/mongodb/mongo-c-driver/pull/1675#issuecomment-2239859044

PR #1173, opened 2024-07-22
CXX-3069 document directories, namespaces, and generated headers

eramongodb:
## Summary

Partially resolves CXX-3069. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1170.

Adds Doxygen documentation for directories, namespaces, and generated headers. Docs for files and root namespace redeclarations are deferred to a followup PR.

This PR proposes organizing Doxygen documentation at the _bottom_ of header files in the following order:

```cpp
// License notice.

#pragma once

#include <headers> // Include directives and etc.

namespace bsoncxx { ... } // Declarations and definitions.

///
/// @file
/// Documentation for this file.
///

///
/// @dir path/to/dir
/// Documentation for relevant subdirectories.
///

///
/// @namespace ns::qualifier
/// Documentation for relevant namespaces.
///
```

This is to avoid their interference with general development workflows, as these particular sets of documentation is not expected to be changed often.

## Directories

This PR adds documentation of directories, which partially enables navigation via intermediate directories and the "File List" page.

image

More importantly, it triggers population of directory pages with subdirectory and file information (file docs are deferred to a followup PR).

image

Directories defer documentation to namespaces whenever able to avoid redundancy by referencing the namespace in its brief, since the directory structure mirrors the namespace structure.

To avoid ambiguity regarding which header within subdirectory to place directory documentation, all directory docs are collectively placed in the ABI directory's `fwd.hpp` file (with the exception of the library root directory itself).

Note: sometimes paths must be fully qualified relative to the project root directory or the library root directory for proper disambiguation by Doxygen, but this usually isn't necessary.

## Namespaces

This PR adds documentation of namespaces (including root namespaces; redeclaration docs are deferred to a followup PR), which enables navigation via intermediate namespaces.

image

This triggers population of namespace pages with nested namespaces.

image

Similar to namespaces, all namespace docs are collectively placed in the ABI directory's `fwd.hpp` file to avoid placement ambiguity.

Currently, due to having a 1-to-1 relationship, v_noabi namespace docs reuse root namespace docs via `@copydoc`. This may no longer be possible if namespace restructuring occurs in the future, but this is expected to be a far-future concern.

## Config Headers

This PR also adds documentation for generated headers under `config/` (i.e. `config.hpp`) in the form of Doxygen modules (aka "named groups"). This PR also documents headers already visible under the `config/` directory (i.e. `prelude.hpp`) for completeness.

image

image

This means macros such as `BSONCXX_POLY_USE_*` and `BSONCXX_VERSION_*` are _finally_ documented proper by the API doc pages! 🎉 

Headers which are not intended for use by consumers of the library are labeled with a `@warning For internal use only!`. Docs for such files are intentionally kept brief and sparse in detail to avoid encouraging users to use the provided entities despite the warning.

Due to generated headers "not existing" according to Doxygen due to docs being generated from files under `include/` only, their docs are defined in the ABI directory's `fwd.hpp` file alongside directory and namespace docs within a `#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)` block defining stubs for entities (primarily macros) provided by the relevant header files.

This has the effect of listing the generated header docs both in the `fwd.hpp` file page (where it is document) as well as the dedicated Module page (corresponding to the generated header). Some care and emphasis was required to make clear what file actually provides the documented entities.

## Transitive Includes

The `prelude.hpp` and `fwd.hpp` headers demonstrates how _supported_ transitive includes are intended to be explicitly documented following `SHOW_INCLUDE_FILES = NO` in https://github.com/mongodb/mongo-cxx-driver/pull/1166.

image

image

This is implemented using a `@par` command with a paragraph title and a `@li` list. The `@parblock` is not strictly necessary but is included regardless to clearly denote the expected scope of the page element (e.g. such that it does not unintentionally apply to subsequent paragraphs).

image

This pattern is expected to be used sparingly and only for header files where the transitive includes are clearly expected and deliberately supported (e.g. `view_or_value.hpp` transitively includes the corresponding `view.hpp` and `value.hpp` headers).

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/fwd.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/fwd.hpp
@@ -48,11 +48,371 @@
 #include <bsoncxx/view_or_value-fwd.hpp>
 
 ///
+/// @file
+/// Aggregate of all forward headers declaring entities in @ref bsoncxx::v_noabi.
+///
+/// @par "Includes" @parblock
+/// @li All header files under @ref src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx
+/// "bsoncxx/v_noabi/bsoncxx" whose filename ends with `-fwd.hpp`.
+/// @endparblock
+///
+
+///
+/// @dir bsoncxx/v_noabi
+/// The directory relative to which headers declaring entities in @ref bsoncxx::v_noabi are
+/// typically included.
+///
+/// For backward compatibility, unstable ABI headers may be included using the syntax
+/// `<bsoncxx/foo.hpp>`, which is equivalent to `<bsoncxx/v_noabi/bsoncxx/foo.hpp>`.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx
+/// Provides headers declaring entities in @ref bsoncxx::v_noabi.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx/array
+/// Provides headers declaring entities in @ref bsoncxx::v_noabi::array.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx/builder
+/// Provides headers declaring entities in @ref bsoncxx::v_noabi::builder.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx/builder/basic
+/// Provides headers declaring entities in @ref bsoncxx::v_noabi::builder::basic.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx/builder/stream
+/// Provides headers declaring entities in @ref bsoncxx::v_noabi::builder::stream.
+///
+
+///
+/// @dir bsoncxx/v_noabi/bsoncxx/config
+/// Provides headers related to bsoncxx library configuration.
+///
+/// @par "Generated Headers"
+///
+/// Generated headers are documented by the following pages:
+///
+/// @li @ref bsoncxx-v_noabi-bsoncxx-config-config-hpp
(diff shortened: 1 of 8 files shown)


src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/fwd.hpp
```diff
@@ -47,12 +47,372 @@
...
+/// @warning For internal use only!
+///
+/// This macro is not defined when the bsoncxx library is built as a shared library.
+///
+/// @note This macro is intended to be automatically defined by the CMake or pkg-config package
+/// configuration file used to import the bsoncxx library. Manually defining this macro is NOT
+/// recommended, but may be necessary depending on how the mongocxx library is being imported or
+/// packaged.
+///
+#define BSONCXX_STATIC
```
kevinAlbs, line 212: Remove `BSONCXX_STATIC`? The rendered documentation notes "These macros are provided by export.hpp". But `BSONCXX_STATIC` is applied [on the CMake target](https://github.com/mongodb/mongo-cxx-driver/blob/ae07bd2574ae7340f7a047f4dd99186c644a262b/cmake/BsoncxxUtil.cmake#L140), rather than defined in `export.hpp`.

Similar question applies to `MONGOCXX_STATIC`.

kevinAlbs (approved):
LGTM with one question.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/fwd.hpp
eramongodb: Good point. Removed as suggested.

PR #1177, opened 2024-07-24
CXX-2811: Mark Server version 3.6 as EOL from driver's PoV (Min server version as 4.0, minWireVersion as 7.0)

Julia-Garland:
### Summary
Requirements resolved by [CDRIVER-4815](https://jira.mongodb.org/browse/CDRIVER-4815). 

Removed non-spec tests that only ran on server version 3.6 or below and cleaned up if statements that ran on now always-true conditionals ex. `max_wire_version >= 5`.


data/crud/unified/bulkWrite-replaceOne-dots_and_dollars.json
```diff
@@ -120,80 +120,6 @@
         }
       ]
     },
-    {
-      "description": "Replacing document with top-level dotted key on pre-3.6 server yields server-side error",
```
kevinAlbs, line 124: Recommend keeping any specification tests that are still present in the specifications repo:
https://github.com/mongodb/specifications/blob/82be6f2203ee35a6c40af36e5233b632ef402058/source/crud/tests/unified/bulkWrite-replaceOne-dots_and_dollars.yml#L61

Even though this test is no longer expected to run (due to `maxServerVersion` being lower than 3.6), the test files are intended to be exact copies of what is in the spec repo. That enables easy updating.

Julia-Garland: It looks like _all_ the tests I removed are in the specifications repo. I think this ticket doesn't need any additional work to resolve so I can just abandon this PR.

kevinAlbs: Non-spec tests could be removed. Some tests check wire version ([example](https://github.com/mongodb/mongo-cxx-driver/blob/1baf35c085e61225f1d644a333bfa507ce8221dd/src/mongocxx/test/collection.cpp#L2198-L2215)). Tests only running when the maxWireVersion <= 6 could be removed. Wire version 6 [corresponds to server 3.6](https://github.com/mongodb/specifications/tree/master/source/wireversion-featurelist.md).

Though I expect it is not strictly necessary to remove tests. I'd suggest trying to remove those. If there is unexpected difficulty, then I think it is OK to skip.

Julia-Garland: That's the one test case that only runs for maxWireVersion <= 6, so simple enough to change!

data/crud/unified/bulkWrite-update-hint-clientError.json
```diff
@@ -281,4 +281,4 @@
       ]
     }
   ]
-}
+}
```
kevinAlbs, line 284: Suggest reverting whitespace changes to the spec test files, as it is not relevant to the PR. Can do with:

```bash
# Revert state of files to `master` branch:
git checkout master -- data/crud/unified
```

src/mongocxx/test/collection.cpp
```diff
@@ -2194,27 +2194,6 @@ TEST_CASE("CRUD functionality", "[driver::collection]") {
             REQUIRE(result->inserted_count() == 1);
         }
 
-        SECTION("fail if server has maxWireVersion < 5 and write has collation") {
```
kevinAlbs, line 2197: Suggest removing other code paths that expect to run when maxWireVersion is <= 6 (for server 3.6) ([example](https://github.com/mongodb/mongo-cxx-driver/blob/dc971edbc1b37b1d7a733238966fb99c97bf3201/src/mongocxx/test/collection.cpp#L1553-L1561)):

```c++
if (test_util::get_max_wire_version(mongodb_client) >= 5) {
    // do test for MongoDB >=3.4.
} else {
    // do test for MongoDB <3.4.
}
```

Can become:
```c++
// do test for MongoDB >=3.4.
```

kevinAlbs (approved):
LGTM with a revert to whitespace changes and additional dead code removal.

PR #1178, opened 2024-07-24
CXX-3072 Reduce Warnings - Code Consistency and Quality Improvements

eramongodb:
## Summary

Resolves CXX-3072. Verified by [this patch](https://spruce.mongodb.com/version/66a11d1ad1fabb000754f19c).

Similar to https://github.com/mongodb/mongo-c-driver/pull/1543 for the C Driver. Addresses 615 Clang warnings (83 unique warnings) with Clang 19.

## scoped_bson_value

A long overdue followup to https://github.com/mongodb/mongo-cxx-driver/pull/955:

> It may be worth eventually extracting this helper into a proper a proper component (probably under src/mongocxx/private?), but that work has been deferred for now.

This addresses `-Wunused-member-function` warnings in `options.cpp` due to unused functions with internal linkage (default ctor + `value_for_init`).

## C-Style Casts

A large part of this PR is devoted to addressing `-Wold-style-cast` warnings.

Most cases only needed a trivial change from `(T)expr` to `static_cast<T>(expr)`. Regarding numeric casts, range checks have not been added: those may be audited in the future when an assertion/contract methodology for the C++ Driver is implemented (directly reusing C Driver's `BSON_ASSERT()` does not seem ideal).

A notable change is the use of `const_cast` to workaround the `void*` parameter (lack of const) in `bsoncxx::v_noabi::types::bson_value::view`'s private member functions. To avoid unnecessarily breaking ABI compatibility, a `const_cast` is used instead of updating the parameter type, accompanied by a comment documenting the context and its validity (the const is recovered in the `_init` function + corresponding `convert_from_libbson()` overloads updated with the missing `const` for consistency).

## Miscellaneous

Other warnings addressed by this PR include:

- handling of semicolons (both removing redundant ones and adding some following macro use for consistency)
- missing case statements in switches on enumerations (in `error_code.cpp` in particular).
  - is the `write_concern::level::k_acknowledged` do-nothing fallthrough in `write_concer.cpp` correct?
- `#if defined(X)` instead of `#if X` when `X` is not guaranteed to be defined.
- `static` -> `inline` for function definitions in headers used by multiple components (where internal linkage does not seem sufficiently motivated).
- `NULL` -> `nullptr` even with C APIs (C idioms shouldn't leak into C++ code).
- `0` -> `nullptr` in pointer comparisons and initialization.

## Ignored Warnings

The following warnings were observed but deliberately unaddressed/ignored/silenced:

- `-Wdocumentation`: addressed by https://github.com/mongodb/mongo-cxx-driver/pull/1175.
- `-Walign-mismatch` in `bsoncxx/v_noabi/bsoncxx/private/stack.hh`: unclear how to resolve this warning.
- `-Wexit-time-destructors` for exception types: deferred to a future API major release.
- `-Wglobal-constructors` + `-Wexit-time-destructors` for `uri::k_default_uri`: deferred to a future API major release.
- `-Wundefined-func-template` for `soft_declval`: deferred to v1 migration PRs (CXX-2745).
- `-Wweak-vtables` for exception types: deferred to a future API major release.

### DIFF
--- a/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
+++ b/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
@@ -176,6 +176,8 @@ bsoncxx::v_noabi::document::value write_concern::to_document() const {
                     doc.append(kvp("w", *t));
                 }
                 break;
+
+            case write_concern::level::k_acknowledged:
             default:
                 break;
         }
--- a/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
+++ b/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
@@ -178,6 +178,9 @@ bsoncxx::v_noabi::document::value write_concern::to_document() const {
                 break;
 
             case write_concern::level::k_acknowledged:
+                // `ns.has_value()` implies an acknowledged write.
+                break;
+
             default:
                 break;
         }
--- a/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
+++ b/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
@@ -175,6 +175,7 @@ bsoncxx::v_noabi::document::value write_concern::to_document() const {
                 if (auto t = tag()) {
                     doc.append(kvp("w", *t));
                 }
+                break;
             default:
                 break;
         }
--- a/examples/mongocxx/mongodb.com/documentation_examples.cpp
+++ b/examples/mongocxx/mongodb.com/documentation_examples.cpp
@@ -41,7 +41,7 @@ void check_field(const T& document,
                  const char* field,
                  bool should_have,
                  int example_no,
-                 const char* example_type = NULL) {
+                 const char* example_type = nullptr) {
     std::string example_type_formatted = example_type ? example_type + std::string(" ") : "";
     if (should_have) {
         if (!document[field]) {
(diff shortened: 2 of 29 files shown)


src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
```diff
@@ -175,6 +175,9 @@ bsoncxx::v_noabi::document::value write_concern::to_document() const {
                 if (auto t = tag()) {
                     doc.append(kvp("w", *t));
                 }
+                break;
+
+            case write_concern::level::k_acknowledged:
```
Julia-Garland, line 180: > is the write_concern::level::k_acknowledged do-nothing fallthrough in write_concern.cpp correct?

I checked out the [API reference ](https://mongocxx.org/api/current/classmongocxx_1_1v__noabi_1_1write__concern.html#a756cc9e4f51467924887b2ceda9c8856)and the MongoDB docs, not sure myself.

Julia-Garland (approved):
Besides not being sure about your question on `write_concern::level::k_acknowledged` LGTM!

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/write_concern.cpp
kevinAlbs: I expect this case is unreachable. If this `else` block is entered, then `nodes()` did not return a value. That implies `acknowledge_level()` cannot be `k_acknowledged`.

I expect this could assert:

```c++
            case write_concern::level::k_acknowledged:
                // Acknowledged write concern is expected to return `nodes()` above.
                MONGOCXX_UNREACHABLE;
```

kevinAlbs (approved):
The warning reductions are appreciated. LGTM with a possible use of `BSONCXX_UNREACHABLE`.

PR #1174, opened 2024-07-23
CXX-3068 Revert API docs to their initial state + cap to Doxygen 1.9.1

eramongodb:
> [!WARNING]
> **DO NOT OPEN THE "FILES CHANGED" TAB!!!**
>
> ```
> 30945 files changed, 932833 insertions(+), 926358 deletions(-)
> ```
>
> This PR has a **HUGE** diff! Avoid opening the Files Changed tab or individual commits on GitHub. Otherwise, your tab/browser may lag greatly! Recommend reviewing changes locally instead.

## Summary

Resolves CXX-3068. Note the target branch is `gh-pages`, not `master`.

Reverts the state of every API subdirectory to their initial state + caps the latest Doxygen version used for generation to 1.9.1, matching the [current version in the Doxyfile](https://github.com/mongodb/mongo-cxx-driver/blob/ae07bd2574ae7340f7a047f4dd99186c644a262b/Doxyfile#L1).

Recommend verifying changes locally rather than on GitHub.

## Initial State

Prior to https://github.com/mongodb/mongo-cxx-driver/pull/1169 (CXX-3063), _all_ API docs were regenerated with every release in their totality regardless of the version of the Doxygen binary being used at the time or its compatibility with the Doxyfile configuration file for the given release. This resulted in the API docs being (re)generated with _thirteen_ different Doxygen versions over the course of their lifetime. The following table lists the Doxygen version used (when different from the version used to generate the docs prior), the date of the change, and the estimated C++ Driver version related to the change:

```
1.8.10  (2016-09-28) (r3.0.2)
1.8.9.1 (2016-12-13) (r3.1.0)
1.8.10  (2016-12-20) (r3.1.1)
1.8.13  (2017-07-12) (r3.1.2)
1.8.14  (2018-05-31) (r3.3.0)
1.8.13  (2018-10-01) (r3.4.0)
1.8.17  (2020-07-30) (r3.6.0)
1.8.20  (2020-11-03) (r3.6.1)
1.8.18  (2020-12-01) (r3.6.2)
1.8.20  (2021-06-03) (r3.6.4)
1.9.2   (2022-05-03) (r3.6.7)
1.8.17  (2022-09-30) (r3.7.0)
1.9.6   (2023-03-14) (r3.7.1)
1.9.1   (2023-06-06) (r3.7.2)
1.9.6   (2023-06-21) (r3.8.0)
1.9.8   (2023-10-05) (r3.8.1)
1.9.7   (2023-11-08) (r3.9.0)
[...]

PR #1180, opened 2024-07-26
CXX-3081 Keep API doc index pages up-to-date with each release

eramongodb:
## Summary

Resolves CXX-3081. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1169 and https://github.com/mongodb/mongo-cxx-driver/pull/1174.

This PR adds a new script which is invoked during API doc deployment which patches the index pages of all prior API docs to match that of the latest API doc. This permits each API doc to have an up-to-date list of C++ Driver releases, links to resources, and possibly other information (such as announcements).

## Index Page Contents

Regardless of Doxygen version and API doc version, the following patterns were observed:

- There is an `index.html` file at the top-level directory of each API doc directory corresponding to the "Main Page" for each API doc.
- The `index.html` file contains a `<div class="contents">...</div>` corresponding to main page contents.
- The title, headers, footers, and other elements specific to each API doc are _not_ members of the contents element.

Therefore, this PR proposes the "contents" element of each index page be directly substituted with that of the latest API doc with each deployment. This ensures a direct and targeted means of keeping the home page of old API docs up-to-date with the contents of `etc/apidocmenu.md` for each release while minimizing the risk of breaking old API doc pages due to Doxygen version inconsistencies/incompatibilities. Recommend examining the patch commit (not the format commit) in https://github.com/mongodb/mongo-cxx-driver/pull/1181 for a demonstration of expected ehavior.

## etc/patch-apidocs-index-pages.py

This script is now invoked as part of the `doxygen-deploy` command. An `APIDOCSPATH` environment variable is used to specify the `api/` directory under which API docs are located. This env var may be used when manually invoking the script for testing purposes.

This script is designed to fail fast, fail hard should it encounter any unexpected situations to avoid risk of unintentional modifications to unrelated files and avoid deploying broken pages.

The [Beautiful Soup](https://pypi.org/project/beautifulsoup4/) Python library is selected for purposes of formatting and editing HTML. In particular, its ability to easily [find an HTML element by class](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#searching-by-css-class) and [directly replace the contents of an HTML element](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#replace-with) perfectly fulfills the needs of this PR.

The script also ensures all index pages (including the latest) are formatted per `.prettify(formatter="html")`. This will make future patches to the index pages easier to inspect and review for correctness. Note: using the `html5` formatter was considered but dropped due to it applying more changes to the HTML than what appeared to be minimally necessary.

An example of the expected script output:

```
Patching API docs in: build/tmp-repo/Jfa6dNfcZ5/api
Finding API docs...
 - Skipping: mongocxx-3.11.0-example
Finding API docs... done.
 - Found 51 API docs: legacy-0.8.0 ... mongocxx-3.10.2
Searching for index pages...
Searching for index pages... done.
 - Found 51 index pages within 51 API docs.
 - Using mongocxx-3.10.2 as the latest API doc.
Formatting latest index page...
Formatting latest index page... done.
Extracting latest index page contents...
Extracting latest index page contents... done.
Replacing contents of 50 index pages...
Replacing contents of 50 index pages... done.
```

## Drive-By Improvements

- Replace `jira @ ...` with `jira==3.5.2` which contains the required patch in prior comment.
- Fixed links to the MongoDB C++ Driver documentation pages (under `cpp/cpp-driver`, not `cpp/drivers`; did this change at some point?).
- Clarify expectation of contents of `build/docs/api` in release instructions. `doxygen-deploy` depends on `doxygen-latest` so it is easier to simply start with a clean slate after reviewing the correctness of generated API doc.

### DIFF
--- a/etc/patch-apidocs-index-pages.py
+++ b/etc/patch-apidocs-index-pages.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python3
+
+# Copyright 2009-present MongoDB, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Patches the root-level index.html file in each API doc site to match the state of the latest API doc site.
+"""
+
+from packaging.version import Version, InvalidVersion
+from typing import List, Tuple
+
+import bs4
+import os
+
+
+def find_api_docs_path() -> str:
+    """
+    Return an absolute path to the directory containing the API docs.
+    """
+    api_docs_path: str | None = os.environ.get('APIDOCSPATH')
+    if not api_docs_path:
+        raise RuntimeError('APIDOCSPATH environment variable is not set!')
+
+    if not os.path.exists(api_docs_path):
+        raise RuntimeError('path to API docs does not exist!')
+
+    return os.path.abspath(api_docs_path)
+
+
+def find_api_docs(api_docs_path: str) -> List[str]:
+    """
+    Return a list of API doc directories by name.
+    """
+    api_docs: List[str] = []
+    for dir in os.scandir(api_docs_path):
+        if dir.is_dir() and not dir.is_symlink():
+            api_docs.append(dir.name)
+
+    # Sort by legacy vs. modern, then by SemvVer. Example:
+    #  - legacy-0.1.0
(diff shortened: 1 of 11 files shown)


etc/patch-apidocs-index-pages.py
```diff
@@ -0,0 +1,162 @@
...
+def find_api_docs(api_docs_path: str) -> List[str]:
+    """
+    Return a list of API doc directories by name.
+    """
+    api_docs: List[str] = []
+    for dir in os.scandir(api_docs_path):
+        if dir.is_dir() and not dir.is_symlink():
+            api_docs.append(dir.name)
+
+    # Sort by legacy vs. modern, then by SemvVer. Example:
```
kevinAlbs, line 51:
```suggestion
    # Sort by legacy vs. modern, then by SemVer. Example:
```

kevinAlbs (approved):
LGTM. The drive-by link fixes, and consistent formatting changes are also appreciated.

PR #1182, opened 2024-07-31
CXX-3069 document headers and root namespace redeclarations (bsoncxx)

eramongodb:
## Summary

Partially resolves CXX-3069. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1173. Verified by [this patch](https://spruce.mongodb.com/version/66aa9cfc71f13f0007e6682c/).

A followup PR will resolve CXX-3069 by applying the same pattern of changes in this PR to mongocxx.

A thorough audit of the documentation of interfaces will be addressed by CXX-3082.

## File Documentation

This PR follows the pattern used in https://github.com/mongodb/mongo-cxx-driver/pull/1173 and adds `@file` Doxygen commands at the _bottom_ of all public header files under `include/`. This is to avoid Doxygen-specific code from intruding in the code they are meant to document (+ they are not expected to require frequent maintenance).

```cpp
// License notice.

#pragma once

#include <headers> // Include directives and etc.

namespace bsoncxx { ... } // Declarations and definitions.

///
/// @file
/// Documentation for this file.
///

///
/// More Doxygen-specific comments pertaining to entities provided by this file.
///
```

The result of these efforts is the complete documentation of all files in the File List page (for bsoncxx; mongocxx soon to follow).

🎉 **This completes the trifecta of Doxygen API doc page navigation via namespaces, classes, and files.** 🎉 

Files list, before:

image

Files list, after:

image

File page, before: none.

File page, after:

image

> [!NOTE]
> File pages serve as a collection of references to _namespaces_ (and their immediate members) and _classes_ provided by the header. Accordingly, documentation for interfaces should always be primarily associated with the namespace or class, not the file it is declared/documented in. In short, `@file` documentation should be very brief.

## Root Namespace Redeclarations

As https://github.com/doxygen/doxygen/issues/3760 is yet to be resolved, Doxygen is completely blind to using-declarations. Therefore, great pains are taken to reproduce the documentation structure generated by `using namespace` prior to https://github.com/mongodb/mongo-cxx-driver/pull/1066 and https://github.com/mongodb/mongo-cxx-driver/pull/1070, as demonstrated by the [3.9.0 docs](https://mongocxx.org/api/mongocxx-3.9.0/namespaces.html):

image

This PR uses the `#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)` pattern implemented in https://github.com/mongodb/mongo-cxx-driver/pull/1173 to _actually_ "redeclare" equivalents to the root namespace using-declarations as class/function stubs, solely to generate doc pages that redirect their aliased counterparts:

```cpp
namespace bsoncxx {

using v_noabi::foo;

} // namespace bsoncxx

///
/// @file
/// Documentation for this file.
///

#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)

namespace bsoncxx {

/// @ref bsoncxx::v_noabi::foo
class foo {};

} // namespace bsoncxx

#endif // defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)
```

The result looks as follows (specifically the `bsoncxx::array` members at the top of the image):

image

This pattern has the unexpected benefit of explicitly displaying the redeclared entity as a clickable link in the brief description on the Namespace List page (instead of repeating the brief of the redeclared entity). This should hopefully make the state of redeclarations easier to follow than it would have otherwise been.

The Doxygen documentation for redeclarations _always_ qualify entities with the ABI namespace. This is to avoid Doxygen associating types with a root namespace redeclaration (Doxygen does not appear to be smart enough to resolve namespace scoping rules for parameter types):

```cpp
#if defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)

namespace bsoncxx {

/// @ref bsoncxx::v_noabi::bar
class bar {};

/// @ref bsoncxx::v_noabi::foo(bar b)
void foo(bar b);
//       ^~~ Should be `bsoncxx::v_noabi::bar`, not `bsoncxx::bar`!

} // namespace bsoncxx

#endif // defined(BSONCXX_PRIVATE_DOXYGEN_PREPROCESSOR)
```

This principle is similar to the one which motivated ABI namespace qualification of `bsoncxx::` entities in `mongocxx` to ensure ABI stability across redeclaration updates, as described in https://github.com/mongodb/mongo-cxx-driver/pull/1070#issuecomment-1856779596. A redeclaration changing from `v1 -> v2` must not inadvertently change the meaning of interfaces in ABI-specific namespaces in either code _or_ in documentation.

## Deprecation, Internal Use, and Deliberate No-Documentation

This PR takes this opportunity to document `bsoncxx/util/functor.hpp` as deprecated (consistent with the directory-level deprecation notice).

Several header files are documented as "for internal use only" in an effort to discourage users from directly depending on certain interfaces:

- `bsoncxx/builder/basic/impl.hpp`
- `bsoncxx/config/compiler.hpp`
- `bsoncxx/config/util.hpp`
- `bsoncxx/stdx/make_unique.hpp`
- `bsoncxx/stdx/operators.hpp`
- `bsoncxx/stdx/type_traits.hpp`

These headers have been updated to deliberately contain _no_ documentation of the provided entities in an effort to discourage their use by users. Doxygen-aware tools appear to have no issue parsing and treating `//` (double, not triple) comments as Doxygen documentation despite not being included in doc generation; therefore, existing documentation of internal interfaces have been "demoted" from `///` or `/**` to `//` to avoid their inclusion in generated docs.

For example, this is the entirety of the `type_traits.hpp` doc page:

image

On the other hand, given the significance of `std::string_view` and `std::optional<T>` polyfills in C++ Driver interfaces, these headers (and the polyfills) are given special documentation with a note regarding their relationship with polyfill library configuration. This documentation is expected to be expanded in a followup PR.

Some old/niche/undocumented features have been given the `Deliberately undocumented.` note to avoid their inclusion in generated docs.

## Doxygen Parsing Problems

The `operators.hpp` and `type_traits.hpp` headers were known to give Doxygen trouble, hence their explicit exclusion via the `EXCLUDE` Doxygen option in https://github.com/mongodb/mongo-cxx-driver/pull/1062. This PR reverts their exclusion and instead makes adjustments for consistency with other (internal use) header files and applying the `@cond DOXYGEN_DISABLE` pattern (first used in https://github.com/mongodb/mongo-cxx-driver/pull/1170) to address Doxygen warnings. These changes address three `Found ';' while parsing initializer list!` warnings and four `Detected potential recursive class relation` warnings.

## bsoncxx::v_noabi::error_code

As a drive-by improvement, identified and (partially) fixed misplaced documentation for enumerators of `bsoncxx::v_noabi::error_code`.

Before:

image

After:

image

Addressing the missing documentation for enumerators relating to BSON types and binary subtypes is deferred to a followup PR.

## Miscellaneous

Warnings-related fixes are a followup to https://github.com/mongodb/mongo-cxx-driver/pull/1178 that extends compilation output to the `test_bson` and `test_driver` targets.

- Drive-by formatting fix to `mongocxx/test/collection.cpp` following https://github.com/mongodb/mongo-cxx-driver/commit/d373213181aaa13e3cfdbe6fa4ce2f683400b9e2.
- Addressed 617 (8 unique) `-Wzero-as-null-pointer-constant` warnings by `operators.hpp` and `catch.hh`.
- Addressed 92 (69 unique) `-Wundefined-func-template` warnings emitted by use of `soft_declval` (replaced with equivalent `std::declval` expressions).
- Silenced 33 (8 unique) `-Wfloat-equal` warnings emitted by `optional.hpp`.
- Address a `-Wundef` warning in `bsoncxx/test/string_view.test.cpp`.
- Silenced a `-Wweak-vtables` warning in `optional.hpp` (we do not export symbols in stdx).
- Set `EXAMPLES = examples` in Doxyfile (setup for CXX-3082).

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -901,8 +901,7 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/operators.hpp \
-                         src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/type_traits.hpp
+EXCLUDE                =
 
 # The EXCLUDE_SYMLINKS tag can be used to select whether or not files or
 # directories that are symbolic links (a Unix file system feature) are excluded
--- a/Doxyfile
+++ b/Doxyfile
@@ -938,7 +938,7 @@ EXCLUDE_SYMBOLS        = bsoncxx::detail \
 # that contain example code fragments that are included (see the \include
 # command).
 
-EXAMPLE_PATH           = .
+EXAMPLE_PATH           = examples
 
 # If the value of the EXAMPLE_PATH tag contains directories, you can use the
 # EXAMPLE_PATTERNS tag to specify one or more wildcard pattern (like *.cpp and
--- a/examples/bsoncxx/CMakeLists.txt
+++ b/examples/bsoncxx/CMakeLists.txt
@@ -12,10 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include_directories(
-    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}/bsoncxx/v_noabi
-)
-
 set(BSONCXX_EXAMPLES
     builder_core.cpp
     builder_list.cpp
--- a/examples/mongocxx/CMakeLists.txt
+++ b/examples/mongocxx/CMakeLists.txt
@@ -12,11 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include_directories(
-    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}/bsoncxx/v_noabi
-    ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}/mongocxx/v_noabi
-)
-
 set(MONGOCXX_EXAMPLES
     aggregate.cpp
     automatic_client_side_field_level_encryption.cpp
--- a/examples/mongocxx/connect.cpp
+++ b/examples/mongocxx/connect.cpp
@@ -19,7 +19,6 @@
 #include <bsoncxx/builder/basic/document.hpp>
 #include <bsoncxx/builder/basic/kvp.hpp>
 #include <bsoncxx/json.hpp>
-#include <bsoncxx/stdx/make_unique.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/instance.hpp>
 #include <mongocxx/logger.hpp>
@@ -44,6 +43,11 @@ class logger final : public mongocxx::logger {
     std::ostream* const _stream;
 };
 
+// Use `std::make_unique` with C++14 and newer.
+std::unique_ptr<logger> make_logger() {
+    return std::unique_ptr<logger>(new logger(&std::cout));
+}
+
 }  // namespace
 
 int main(int argc, char* argv[]) {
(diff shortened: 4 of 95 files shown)


eramongodb (commented):
Latest changes verified by [this patch](https://spruce.mongodb.com/version/66b0fe93030f0100073f1cf7).

kevinAlbs (approved):

PR #1186, opened 2024-08-06
CXX-3087 Replace single-argument FetchContent_Populate with FetchContent_MakeAvailable

eramongodb:
Resolves CXX-3087. Verified by [this patch](https://spruce.mongodb.com/version/66b25d01d973d20007c4d59d).

Addresses the following warning emitted by CMake 3.30 and newer for `mongo-c-driver` and `EP_mnmlstc_core`:

```
CMake Warning (dev) at /snap/cmake/1409/share/cmake-3.30/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(mongo-c-driver) is deprecated, call
  FetchContent_MakeAvailable(mongo-c-driver) instead.  Policy CMP0169 can be
  set to OLD to allow FetchContent_Populate(mongo-c-driver) to be called
  directly for now, but the ability to call it with declared details will be
  removed completely in a future version.
Call Stack (most recent call first):
  cmake/FetchMongoC.cmake:24 (FetchContent_Populate)
  CMakeLists.txt:78 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This is caused by [CMP0169](https://cmake.org/cmake/help/latest/policy/CMP0169.html):

> Calling `FetchContent_Populate()` with a single argument (the name of a declared dependency) is deprecated. [...] The OLD behavior of this policy allows `FetchContent_Populate()` to be called with the name of a declared dependency. The NEW behavior halts with a fatal error in such cases.

This PR replaces such calls with `FetchContent_MakeAvailable()` accordingly.

This was straightforward for `mongo-c-driver` given it was (manually) implementing `add_subdirectory()` behavior, which is consistent with `FetchContent_MakeAvailable()`'s regular behavior, which internally calls `add_subdirectory()` on population.

`EP_mnmlstc_core` required a workaround due to the `SOURCE_SUBDIR` option only being supported in 3.18 and newer. Using `SOURCE_SUBDIR` to prevents the internal `add_subdirectory()` call is an officially supported pattern which is needed by FetchMnmlstcCore.cmake to prevent mnmlstc/core's configurations from leaking into the C Driver's configuration (see https://github.com/mongodb/mongo-cxx-driver/pull/1019 for details). This workaround can be removed once the minimum required CMake version is raised to 3.18 or newer.

### DIFF
--- a/cmake/FetchMnmlstcCore.cmake
+++ b/cmake/FetchMnmlstcCore.cmake
@@ -12,6 +12,7 @@ FetchContent_Declare(
     EP_mnmlstc_core
 
     SOURCE_DIR "${core-src}"
+    SOURCE_SUBDIR ""
     SUBBUILD_DIR "${core-subbuild}"
     BINARY_DIR "${core-build}"
     INSTALL_DIR "${core-install}"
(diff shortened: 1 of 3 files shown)


cmake/FetchMnmlstcCore.cmake
```diff
@@ -12,6 +12,7 @@ FetchContent_Declare(
     EP_mnmlstc_core
 
     SOURCE_DIR "${core-src}"
+    SOURCE_SUBDIR ""
```
kevinAlbs, line 15:
```suggestion
    SOURCE_SUBDIR "" # To prevent adding project to CMake.
```

kevinAlbs (approved):

PR #1187, opened 2024-08-06
Update CHANGELOG with 3.10.2 entries

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1160 which forgot to include CHANGELOG entries in the post-release changes PR due to release instructions only committing the patch entries to the release branch. Instructions are also updated accordingly.

### DIFF
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -673,17 +673,27 @@ Commit the changes to the `releases/vX.Y` branch and push the branch to the remo
 
 Checkout the `post-release-changes` branch.
 
-Add a section for the new patch release containing the same entries as in the release, e.g. following a `1.2.3` release:
+Sync the entries in the patch release section to be consistent with the entries on the release branch, e.g. following a `1.2.3` release:
 
 ```md
 ## 1.3.0 [Unreleased]
+
+<!-- Will contain entries for the next minor release. -->
+<!-- Ensure these entries are not removed during the sync. -->
+
+## 1.2.4 [Unreleased]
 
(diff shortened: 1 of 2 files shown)


etc/releasing.md
```diff
@@ -671,6 +671,20 @@ Add a section for the next patch release, e.g. following a `1.2.3` release:
 
 Commit the changes to the `releases/vX.Y` branch and push the branch to the remote repository (a PR is not required for this step).
 
+Checkout the `post-release-changes` branch.
+
+Add a section for the new patch release containing the same entries as in the release, e.g. following a `1.2.3` release:
+
+```md
+## 1.3.0 [Unreleased]
```
kevinAlbs, line 679:
```suggestion
## 1.2.4 [Unreleased]
```

kevinAlbs (approved):
LGTM with a version number fix.

etc/releasing.md
eramongodb: 1.3.0 is correct because this code block corresponds to the `CHANGELOG.md` on `master`, not the release branch (as in earlier instructions above).

Do we expect / want to maintain patch release notes on `master` prior to release? IIRC we avoid doing so to avoid potential cherry-pick conflicts with the release branch.

eramongodb: On further thought, I believe the expectation is yes (with caveats) given the Important note added in https://github.com/mongodb/mongo-cxx-driver/pull/1161 concerning shifting of entries from an unreleased patch section -> next unreleased minor section.

Revised the instructions accordingly.

PR #1192, opened 2024-08-16
MSVC build fix: include <string> instead of <string_view>

vorlac:
* Fixes some unrecognized std::string usage after the 17.11 update of VS2022, since the changes bring an STL update where the `<string_view>` header no longer includes `<xstring>` by default.

PR #1193, opened 2024-08-16
CXX-3089 Ensure <string> is included in all components using std::string

eramongodb:
Resolves CXX-3089. Supercedes https://github.com/mongodb/mongo-cxx-driver/pull/1192. Verified by [this patch](https://spruce.mongodb.com/version/66bf670db574900007456c24).

Requesting an additional review from @vorlac to validate these changes in their environment. (I am unable to add you to the list of reviewers.)

`<string_view>` is not specified to include `<string>`. The [STL changes](https://github.com/microsoft/STL/blame/b191409cec71aa837072a5d1ebafb041989c71dd/stl/inc/string_view#L14-L16) cited in https://github.com/mongodb/mongo-cxx-driver/pull/1192 exposed pre-existing Include What You Use (IWYU) violations in C++ Driver headers (use of `std::string` without including `<string>`).

This PR fixes all IWYU violations for `std::string` (only!) in library code (not tests, examples, etc.). To keep the diff small, the include is added to the "top-most" layer of each component:

- If a `.cpp` uses `std::string`, include `<string>` unless its header(s) already includes it.
- If a private header uses `std::string`, include `<string>` unless its public header includes it.
- If a public header uses `std::string`, include `<string>` unless a fundamentally dependent header includes it.

> [!NOTE]
> I am of the opinion that every file, regardless of component membership or hierarchy, should IWYU, but this PR is deliberately kept small and focused. Auditing and addressing standalone inclusion and IWYU issues can be deferred to a followup task/PR.

The only case of the "unless" in the last bullet point is with the BSON document builder headers. All the builders directly depend on `<bsoncxx/builder/core.hpp>`, thus the `<string>` is added only to `core.hpp` to cover for all the dependent builder headers.

Some additional non-`<string>` include directives were added to the index_view and search_index_view components to properly satisfy their component include hierarchy.

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/builder/core.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/builder/core.hpp
@@ -16,6 +16,7 @@
 
 #include <memory>
 #include <stdexcept>
+#include <string>
 #include <type_traits>
 
 #include <bsoncxx/builder/core-fwd.hpp>
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <cstring>
+#include <string>
 
 #include <bsoncxx/types-fwd.hpp>
 
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
@@ -16,6 +16,7 @@
 
 #include <iostream>
 #include <memory>
+#include <string>
 #include <vector>
 
 #include <bsoncxx/document/element-fwd.hpp>
--- a/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/element.cpp
+++ b/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/element.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <cstdlib>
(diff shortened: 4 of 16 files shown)


kevinAlbs (approved):

eramongodb:
> Also, if you just want to continue the conversation in the new issue you opened, feel free to just respond to this directly in there and I'll see it when I free up to confirm the patch for you in a few hours.

@vorlac Responding here as suggested in https://github.com/mongodb/mongo-cxx-driver/pull/1192#issuecomment-2293938232.

> What should I use to properly test as much coverage as possible for something like this? Is it sufficient to build all of the tests bundled in this repo, or would you recommend a larger open source 3rd party project that uses this library instead?

For purposes of this PR and https://github.com/mongodb/mongo-cxx-driver/pull/1192, it is enough to check if the build you were attempting prior is able to proceed without the initially reported error given the proposed changes, e.g.:

```
heartbeat_failed_event.hpp(56): error C2039: 'string': is not a member of 'std'
```

Ensuring proper and thorough IWYU conformance is our responsibility as maintainers of the C++ Driver (via coding guidelines, build-and-test coverage in CI, etc.). Validating that the proposed changes address your blocking issue will suffice for reviewing this PR. Auditing other potential violations of IWYU can be addressed in CXX-2367.

In general, however, IWYU can be enforced by always including a component's header file _first_ in its corresponding source file (or private implementation header file, which would then be included first by the corresponding source file) so that transitive includes can be minimized as much as possible. There are also static analysis tools for detecting IWYU violations, but none are currently being used for the C++ Driver. We may consider integrating one for CXX-2367.

vorlac:
Hey just letting you guys know I just double checked everything on my end and it's building fine. I built all possible CMake targets just to be safe, everything build without any warnings/errors (using `\W4` warning level). I tried to fit as much as I could in this screenshot in case you wanted to review or confirm anything on your end for the build/setup i'm working with.
image

Feel free to reach out if you want me to check anything else.

vorlac (approved):

PR #1194, opened 2024-08-21
CXX-3088 fix build with GCC 4.8.5

kevinAlbs:
# Summary

Fix build with GCC 4.8.5.

Verified with this [patch build](https://spruce.mongodb.com/version/66c6161c488c38000703c3e4).

# Description

Supporting build with GCC 4.8.5 is motivated by a request to build on a customer RHEL 7.9 system. See [slack thread](https://mongodb.slack.com/archives/C0328AQA3/p1723037627729639).

This PR resolves errors observed when building with RHEL 7 with GCC 4.8.5.

## Error: redeclaration differs in 'constexpr'

```
/root/mongo-cxx-driver/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp:571:45: error: redeclaration 'bsoncxx::v_noabi::stdx::basic_string_view<Char, Traits>::npos' differs in 'constexpr'
const std::size_t basic_string_view<C, Tr>::npos;
                                            ^
/root/mongo-cxx-driver/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/string_view.hpp:136:32: error: from previous declaration 'bsoncxx::v_noabi::stdx::basic_string_view<Char, Traits>::npos'
    static constexpr size_type npos = static_cast<size_type>(-1);
```

I expect this is a quirk in old GCC. See: https://stackoverflow.com/a/17074539/774658. This PR adds `constexpr` to match.

## Error: missing space

```
/root/mongo-cxx-driver/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/json.hpp:83:42: error: missing space between '""' and suffix identifier
BSONCXX_API document::value BSONCXX_CALL operator""_bson(const char* json, size_t len);
```

This PR adds a space for C++11 compliance.

C++11 [user-defined literals](https://timsong-cpp.github.io/cppwp/n3337/over.literal) note a space is required:

> ```
> float operator ""E(const char*);                    // error: ""E (with no intervening space)
>                                                     // is a single token
> ```

C++14 [user-defined literals](https://timsong-cpp.github.io/cppwp/n4140/over.literal) permit no space:

> ```
> float operator ""_e(const char*);                   // OK
> ```

## Ignoring unused parameters

Building the C driver resulted in errors due to `unused-parameter` warnings:

```
../../_deps/mongo-c-driver-src/src/libbson/src/bson/bson-atomic.h:202:28: error: unused parameter 'order' [-Werror=unused-parameter]
    static BSON_INLINE Type bson_atomic_##NamePart##_fetch (      
```

This was addressed in CDRIVER-5673. However, trying to upgrade the C driver to include the fix resulted in [test failures](https://spruce.mongodb.com/version/66c004af4dfaa40007985227) in In-Use Encryption tests due to "rangePreview" being removed. I expect the In-Use Encryption test updates will be addressed in CXX-3014. To limit scope of this PR, `-Wno-unused-parameter` is added to compile flags on RHEL 7 as a temporary workaround until the C driver is upgraded to include CDRIVER-5673.

### DIFF
--- a/.mci.yml
+++ b/.mci.yml
@@ -2221,3 +2221,13 @@ buildvariants:
       display_name: silk
       tasks:
         - name: .silk
+
+    - name: rhel79-compile
+      display_name: "RHEL 7.9 (gcc 4.8.5)"
+      expansions:
+          build_type: "Release"
+          BSONCXX_POLYFILL: impls
+      run_on:
+          - rhel79-small
+      tasks:
+          - name: compile_without_tests
--- a/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/json.cpp
+++ b/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/json.cpp
@@ -99,7 +99,7 @@ document::value BSONCXX_CALL from_json(stdx::string_view json) {
     return document::value{buf, length, bson_free_deleter};
 }
 
-document::value BSONCXX_CALL operator""_bson(const char* str, size_t len) {
+document::value BSONCXX_CALL operator"" _bson(const char* str, size_t len) {
     return from_json(stdx::string_view{str, len});
 }
 
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -159,6 +159,9 @@ linux*)
 
   if [[ "${distro_id:?}" != rhel7* ]]; then
     cxx_flags+=("-Wno-expansion-to-defined")
+  else
+    cc_flags+=("-Wno-unused-parameter") # TODO: remove once C driver is upgraded to include fix of CDRIVER-5673.
+    cxx_flags+=("-Wno-unused-parameter") # TODO: remove once C driver is upgraded to include fix of CDRIVER-5673.
   fi
   ;;
 *)
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -160,7 +160,6 @@ linux*)
   if [[ "${distro_id:?}" != rhel7* ]]; then
     cxx_flags+=("-Wno-expansion-to-defined")
   else
-    cc_flags+=("-Wno-unused-parameter") # TODO: remove once C driver is upgraded to include fix of CDRIVER-5673.
     cxx_flags+=("-Wno-unused-parameter") # TODO: remove once C driver is upgraded to include fix of CDRIVER-5673.
   fi
   ;;
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -155,7 +155,11 @@ darwin*)
   ;;
 linux*)
   cc_flags+=("${cc_flags_init[@]}")
-  cxx_flags+=("${cxx_flags_init[@]}" -Wno-expansion-to-defined -Wno-missing-field-initializers)
+  cxx_flags+=("${cxx_flags_init[@]}" -Wno-missing-field-initializers)
+
+  if [[ "${distro_id:?}" != rhel7* ]]; then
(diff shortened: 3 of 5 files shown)


src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/json.cpp
```diff
@@ -99,7 +99,7 @@ document::value BSONCXX_CALL from_json(stdx::string_view json) {
     return document::value{buf, length, bson_free_deleter};
 }
 
-document::value BSONCXX_CALL operator""_bson(const char* str, size_t len) {
+document::value BSONCXX_CALL operator"" _bson(const char* str, size_t len) {
```
eramongodb, line 102: > C++11 [user-defined literals](https://timsong-cpp.github.io/cppwp/n3337/over.literal) note a space is required:

[CWG 1473](https://cplusplus.github.io/CWG/issues/1473.html) was accepted as a defect report and retroactively applied to C++11 in April 2013:

image

https://github.com/mongodb/mongo-cxx-driver/pull/1178 neglected to account for older compiler releases which might not have implemented the DR.

From [CWG 2521](https://cplusplus.github.io/CWG/issues/2521.html):

> CWG feels that the ostensible significance of whitespace in this context is unfortunate.

I guess we'll have to deal with the `Wdeprecated-literal-operator` warnings in the interim.

.mci.yml
```diff
@@ -2221,3 +2221,13 @@ buildvariants:
       display_name: silk
       tasks:
         - name: .silk
+
+    - name: rhel79-compile
+      display_name: "RHEL 7.9 (gcc 4.8.5)"
+      expansions:
+          build_type: "Release"
```
eramongodb, line 2228: Why release?

eramongodb:
> Building the C driver resulted in errors due to unused-parameter warnings: [...]

Shouldn't [FetchMongoC.cmake](https://github.com/mongodb/mongo-cxx-driver/blob/55ad3447dbd46560eca6e99adfcf195ecd7c1c7a/cmake/FetchMongoC.cmake#L30) be preventing warnings-as-errors when building the C Driver from source?

.mci.yml
kevinAlbs: Along with `BSONCXX_POLYFILL: impls`, I would rather test options recommended for install. Release matches the build type [shown in install instructions](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/installation/linux/#step-3--configure-the-driver).

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/json.cpp
kevinAlbs: Thank you for the references. TIL about the "Defect reports" section on cppreference.

kevinAlbs (commented):
> > Building the C driver resulted in errors due to unused-parameter warnings: [...]
> 
> Shouldn't [FetchMongoC.cmake](https://github.com/mongodb/mongo-cxx-driver/blob/55ad3447dbd46560eca6e99adfcf195ecd7c1c7a/cmake/FetchMongoC.cmake#L31) be preventing warnings-as-errors when building the C Driver from source?

Good spot. The errors were from the C++ driver building C driver headers. I reverted changes to `cc_flags`, but kept the `cxx_flags.

Verified with this patch: https://spruce.mongodb.com/version/66c63bc0968aff00077889c3

eramongodb (approved):

PR #1198, opened 2024-08-26
Address task timeouts on VS 2015 distros due to KMS KMIP mock server

eramongodb:
Verified by [this patch](https://spruce.mongodb.com/version/66ccc85588f0d50007b1526b).

Addresses task timeouts on VS 2015 distros since [2024-07-22](https://spruce.mongodb.com/version/mongo_cxx_driver_ae07bd2574ae7340f7a047f4dd99186c644a262b/tasks) due to what appears to be an incompatibility with the Python version/environment and its relation to the cryptography package as required by the KMS KMIP mock server:

```
Traceback (most recent call last):
  File "kms_kmip_server.py", line 6, in <module>
    from kmip.services.server import KmipServer
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\kmip\services\server\__init__.py", line 16, in <module>
    from kmip.services.server.server import KmipServer
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\kmip\services\server\server.py", line 33, in <module>
    from kmip.services.server import engine
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\kmip\services\server\engine.py", line 47, in <module>
    from kmip.services.server.crypto import engine
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\kmip\services\server\crypto\__init__.py", line 17, in <module>
    from kmip.services.server.crypto.engine import CryptographyEngine
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\kmip\services\server\crypto\engine.py", line 19, in <module>
    from cryptography import exceptions as errors
  File "drivers-evergreen-tools\.evergreen\csfle\kmstlsvenv\lib\site-packages\cryptography\exceptions.py", line 9, in <module>
    from cryptography.hazmat.bindings._rust import exceptions as rust_exceptions
ImportError: DLL load failed while importing _rust: The specified procedure could not be found.
```

This led to a task timeout after 1 hour (the EVG config's max timeout) due to the `wait_for_kms_server` function's curl command lacking a max-time argument (no response -> wait forever?). 

A brief investigation into the root cause of the ImportError exception did not yield any obvious solutions or workarounds.

Accordingly, this PR applies the following changes to the EVG config and scripts:

- Add `--max-time 1` (one second) to the curl command to correctly reflect its intended usage. (The `wait_for_kms_server` can be updated to use `--max-time 1 --retry 60` instead if preferred over the existing loop. Not yet applied due to not being entirely confident that the combination of flags is actually correct.)
- Replace the old CXX-2628 workaround (which applied to all Windows distros; no longer applicable...?) by skipping CSFLE tests entirely on VS 2015 distros only instead.
- Drive-by improvement: use `find-ccache.sh` added to C Driver in https://github.com/mongodb/mongo-c-driver/commit/35f38806bba00d17b4d7f0479d0d213b97b49553.

### DIFF
--- a/.evergreen/abi-stability-setup.sh
+++ b/.evergreen/abi-stability-setup.sh
@@ -26,6 +26,13 @@ declare cmake_binary
 cmake_binary="$(find_cmake_latest)"
 command -V "${cmake_binary:?}"
 
+# Use ccache if available.
+if [[ -f "./mongoc/.evergreen/scripts/find-ccache.sh" ]]; then
+  # shellcheck source=/dev/null
+  . "./mongoc/.evergreen/scripts/find-ccache.sh"
+  find_ccache_and_export_vars "$(pwd)" || true
+fi
+
 # To use a different base commit, replace `--abbrev 0` with the intended commit.
 # Note: EVG treat all changes relative to the EVG base commit as staged changes!
 declare base current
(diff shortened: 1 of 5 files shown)


kevinAlbs (approved):
Test fix is much appreciated. LGTM

PR #1207, opened 2024-09-12
CXX-3102 Remove use of BSONCXX_ENUM in public headers

eramongodb:
Resolves CXX-3102.

The X macro headers under `bsoncxx/v_noabi/enum/` have long cause trouble with Doxygen documentation.

These troubles have worsened following https://github.com/mongodb/mongo-cxx-driver/pull/1182, especially in enumerations:

image

image

These are in addition to numerous functions and enumerators which are not properly documented, e.g. `k_need_element_type_k_*` and `k_cannot_append_*` error codes:

image

image

as well as ctor overloads for `bsoncxx::v_noabi::types::bson_value::value` and `bsoncxx::v_noabi::types::bson_value::view`.

This PR proposes replacing all use of `BSONCXX_ENUM` (specifically the X macro headers themselves) in favor of fully expanding and properly documenting each BSON type and binary subtype. The X macro headers may still be used in internal code which does not require API documentation for convenience.

The examples shown above are fixed to looks as follows:

image

image

image

image

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1726,7 +1726,7 @@ FULL_SIDEBAR           = NO
 # Minimum value: 0, maximum value: 20, default value: 4.
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-ENUM_VALUES_PER_LINE   = 4
+ENUM_VALUES_PER_LINE   = 1
 
 # If the treeview is enabled (see GENERATE_TREEVIEW) then this tag can be used
 # to set the initial width (in pixels) of the frame in which the tree is shown.
--- a/Doxyfile
+++ b/Doxyfile
@@ -237,7 +237,7 @@ QT_AUTOBRIEF           = NO
 # not recognized any more.
 # The default value is: NO.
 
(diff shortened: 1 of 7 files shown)


eramongodb:
Aside: `ENUM_VALUES_PER_LINE = 1` is responsible for the strictly columnar list of enumerators. This change can be reverted or tweaked if it is deemed too space-wasteful, e.g. with the default value of 4:

image

With the maximum value of 20:

image

Or disabled entirely (note: enumerator values are no longer displayed at all; unsure where the [SHOW_ENUM_VALUES](https://www.doxygen.nl/manual/config.html#cfg_show_enum_values) setting has gone?):

image

eramongodb:
Opted to use `ENUM_VALUES_PER_LINE = 0` in favor of `\showenumvalues` to generate enumerator values in description tables only for enumerations that most benefit from it.

e.g. applied to BSON types and subtypes:

image

but not applied to error codes:

image

kevinAlbs (approved):

PR #1210, opened 2024-09-18
clang-format: separate bsoncxx and mongocxx header groups

eramongodb:
A minor adjustment to clang-format rules which separates bsoncxx and mongocxx headers into their own groups such that the library headers are separated by an empty newline:

```cpp
#include <bsoncxx/a.hpp>
#include <bsoncxx/b.hpp>
#include <bsoncxx/c.hpp>
                           // <-- Insert an empty newline here.
#include <mongocxx/a.hpp>
#include <mongocxx/b.hpp>
#include <mongocxx/c.hpp>
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -12,16 +12,24 @@ IncludeBlocks: Regroup
 IncludeCategories:
   - Regex: '".*"' # relative headers
     Priority: 10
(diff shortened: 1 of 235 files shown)


kevinAlbs (approved):
Neat. I like the separation.

PR #1212, opened 2024-09-18
Reduce Warnings - implicit-fallthrough, switch-enum, extra-semi-stmt, unneeded-member-function

eramongodb:
Addresses 40 Clang warnings (35 unique).

> [!NOTE]
> The value of `-Wswitch-enum` is denoting code which may need to be updated given changes to a set of enumerators. The value of `default:` in addition to explicit enumerators is to handle representable values which may not be covered by the set of enumerators. Therefore, having both in the same switch statement is an acceptable pattern, although the `default:` should ideally not be necessary.


kevinAlbs (approved):
Warning fixes are much appreciated. LGTM

PR #1224, opened 2024-10-03
CXX-3082 Add examples for bsoncxx::v_noabi::builder::concatenate

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1208. Prompted by CXX-2291.

Overlooked the addition of examples of using `concatenate()`, which are currently only used in error handling examples for sub-document/sub-array exception recovery. Updates documentation of `concatenate()` overloads to avoid implying their use is specific to the stream builders.

As a drive-by improvement, simplifies creation of arrays for examples by using `make_array(args...)` instead of going through `from_json(R"({"v": [args...]})")["v"].get_array().value`. ("When all you have is a hammer, ...")

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/builder/concatenate.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/builder/concatenate.hpp
@@ -26,8 +26,10 @@ namespace v_noabi {
 namespace builder {
 
 ///
-/// Container to concatenate a document. Use it by constructing an instance with the
-/// document to be concatenated, and pass it into a document stream builder.
+/// Container to concatenate a document.
+///
+/// Use this with a document or array builder to merge an existing document's fields with that of
+/// the document or array being built.
 ///
 struct concatenate_doc {
     document::view_or_value doc;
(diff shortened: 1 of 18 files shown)


src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/builder/concatenate.hpp
```diff
@@ -26,8 +26,10 @@ namespace v_noabi {
 namespace builder {
 
 ///
-/// Container to concatenate a document. Use it by constructing an instance with the
-/// document to be concatenated, and pass it into a document stream builder.
+/// Container to concatenate a document.
+///
+/// Use this with a document or array builder to merge an existing document's fields with that of
+/// the document or array being built.
```
kevinAlbs, line 32:
```suggestion
/// Use this with a document builder to merge an existing document's fields with that of the
/// document being built.
```

I expect the concatenate only works with array+array or document+document:

```c++
using bsoncxx::builder::concatenate;
using bsoncxx::builder::basic::make_array;

auto arr = make_array(1, 2, 3);
auto doc = make_document(kvp("foo", "bar"));

// Cannot concatenate an array to a document builder
{
    bsoncxx::builder::basic::document builder;
    builder.append(concatenate(doc.view()));     // OK
    // builder.append(concatenate(arr.view()));  // Error: no matching member function
}

// Cannot concatenate a document to an array builder
{
    bsoncxx::builder::basic::array builder;
    // builder.append(concatenate(doc.view()));  // Error: no matching member function
    builder.append(concatenate(arr.view()));     // OK
}
```

Similar suggestion applies to other concatenate helpers.

kevinAlbs (commented):
I like the readability improvement of using `make_array`. Suggested rewording concatenate docs.

PR #1237, opened 2024-10-17
CXX-1201 remove extra cmake config

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-cxx-driver/pull/1190 to remove an extra unused block in `examples/CMakeLists.txt`.


eramongodb (approved):
Thank you for the catch. 👍

PR #1071, opened 2023-12-18
Fix broken links

kevinAlbs:
# Summary

Fix broken links in Hugo documentation.

This PR includes additional drive-by improvements:

- replace `http://mongocxx.org` with `https://mongocxx.org`
- replace `mongodb.github.io` with `mongocxx.org`

# Background & Motivation

https://github.com/mongodb/mongo-cxx-driver/pull/1039 replaced use of the inline namespace macros with `inline namespace v_noabi`. I expect this changed the Doxygen page links include the namespace:

Example:
- In 3.8.1: https://mongocxx.org/api/mongocxx-3.8.1/classbsoncxx_1_1oid.html
- In 3.9.0: https://mongocxx.org/api/mongocxx-3.9.0/classbsoncxx_1_1v__noabi_1_1oid.html

This PR is intended as a short-term solution to fix the broken links. Further investigation may be needed to restore links omitting the namespace.

https://github.com/mongodb/mongo-cxx-driver/pull/1026 changed the directory structure. I expect this resulted in broken links to GitHub source pages:

- In 3.8.1: https://github.com/mongodb/mongo-cxx-driver/blob/r3.8.1/src/mongocxx/database.hpp
- In 3.9.1: https://github.com/mongodb/mongo-cxx-driver/blob/r3.9.0/src/mongocxx/include/mongocxx/v_noabi/mongocxx/database.hpp

Broken links were checked by building docs with `cmake --build cmake-build --target hugo` and running [find_broken_links.py](https://gist.github.com/kevinAlbs/4ca2bfc2835a0df2cadb167282796f69).

PR #1238, opened 2024-10-22
CXX-3097 add patch-apidocs-current-redirects.py

eramongodb:
## Summary

Related to CXX-3097. See also https://github.com/mongodb/mongo-cxx-driver/pull/1239.

Adds a new patch script to the page deployment process which inserts redirect routines to API doc pages such that navigating to an `/api/current` page automatically redirects the user to the canonical `/api/mongocxx-<version>` page.

## patch-apidocs-current-redirects.py

The new patch script behaves in a similar way to the `patch-apidocs-index-pages.py` script introduced in https://github.com/mongodb/mongo-cxx-driver/pull/1180. The `APIDOCSPATH=/path/to/api` environment variable is used to describe the location of the `/api` directory, under which `mongocxx-<version>` API doc directories are present. The latest API doc directory is identified by sorting the directory names by version.

Unlike `patch-apidocs-index-pages.py`, BeautifulSoup4 is _not_ used to parse and modify the HTML contents due to its implicit attribute-sorting behavior generating far more diffs than necessary. No straightforward way to disable this behavior while preserving the existing formatting of the HTML files (without "prettifying" its contents) could be found. Therefore, simple line parsing and index-based line insertion is used instead. See https://github.com/mongodb/mongo-cxx-driver/pull/1239 for the results of the patch applied to the current latest API doc pages (for 3.11.0).

The large number of files which need to be processed (+1K) motivated the use of `concurrent.futures.ProcessPoolExecutor` to achieve reasonably fast runtime performance.

The new patch script is automatically invoked by `deploy-to-ghpages.pl` immediately after invocation of the `patch-apidocs-index-pages.py` script.

Release instructions are updated to include adding an entry to the sitemap index for the newly-deployed API doc pages and updating `<lastmod>` to communicate the update to search engines. See https://github.com/mongodb/mongo-cxx-driver/pull/1239 for details.

## Miscellaneous

The icon for the API Documentation page link is updated for visual consistency.

Before:

image

After:

image

Additionally the now-unused `/categories` and `/tags` directories are no longer generated by Hugo (per `disableKinds`) to avoid stepping over the new redirect pages introduced in https://github.com/mongodb/mongo-cxx-driver/pull/1239.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -180,7 +180,8 @@ FULL_PATH_NAMES        = YES
 # will be relative from the directory where doxygen is started.
 # This tag requires that the tag FULL_PATH_NAMES is set to YES.
 
-STRIP_FROM_PATH        = src/bsoncxx/include \
+STRIP_FROM_PATH        = . \
+                         src/bsoncxx/include \
                          src/mongocxx/include
 
 # The STRIP_FROM_INC_PATH tag can be used to strip a user-defined part of the
@@ -190,7 +191,8 @@ STRIP_FROM_PATH        = src/bsoncxx/include \
 # specify the list of include paths that are normally passed to the compiler
 # using the -I flag.
 
-STRIP_FROM_INC_PATH    = src/bsoncxx/include \
+STRIP_FROM_INC_PATH    = . \
+                         src/bsoncxx/include \
                          src/mongocxx/include
 
 # If the SHORT_NAMES tag is set to YES, doxygen will generate much shorter (but
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -18,7 +18,7 @@ publishdir = "../build/hugo"
 
 [[menu.main]]
     name = "API Documentation"
-    pre = "<i class='fa fa-file-text-o'></i>"
+    pre = "<i class='fa fa-book'></i>"
     weight = 90
     identifier = "apiDocs"
     url = "https://mongocxx.org/api/current"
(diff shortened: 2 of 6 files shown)


eramongodb:
Drive-by improvements while inspecting deploy script behavior:

- https://github.com/mongodb/mongo-cxx-driver/pull/1071 added a trailing `/` to the `homepage` variables in `theme.toml`, which leads to double-slashes in Hugo generated URLs (primarily concerns legacy pages, but also the root index.html file). Removes this trailing slash.
- Adds `.` to `STRIP_FROM_PATH` (+ `STRIP_FROM_INC_PATH` for consistency) to avoid leaking local directory info in anchor IDs in Doxygen generated files (primarily concerns the `apidocmenu.md` file).

kevinAlbs (approved):
LGTM with a comment fix. I like the use of the patch tag to prevent double patching.

mdbmes (approved):
LGTM, this looks like it should be a big improvement for usability.

In an ideal world I'd really want to find a more efficient way to do the redirect, and a way to avoid repeating the site's base URI in so many places; but given all the constraints we're under the JavaScript does seem best.

It's unclear to me why we need to track whether the patch has been applied using a separate token. Could the 'patch-tag' actually just be 'window.location.href.replace' (or the whole line containing it)?

eramongodb:
> It's unclear to me why we need to track whether the patch has been applied using a separate token. Could the 'patch-tag' actually just be 'window.location.href.replace' (or the whole line containing it)?

It could, but I opted to insert a "named" tag that identifies the origins of where the inserted code block comes from (not part of regular Doxygen output + mentions the name of the script which generated it) for better traceability (e.g. grepping, git log, etc.). The double-patch prevention check is just a bonus that helps with local development + possibly help reduce the consequences of incorrect patching during a release.

PR #1243, opened 2024-10-29
Fix polyfill selection for compile tasks + use large distros

eramongodb:
Verified by [this patch](https://spruce.mongodb.com/version/6720f2c1b48a5c00078003e1).

https://github.com/mongodb/mongo-cxx-driver/pull/1194 added the RHEL 7.9 distro to test against GCC 4.8.5 and the bsoncxx polyfill impls via `BSONCXX_POLYFILL: impls`. However, the `"compile"` EVG function doesn't know about this env var, as it is only used by the scan-build tasks. Therefore, the task has actually been [compiling against mnmlstc/core instead](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_rhel79_compile_compile_without_tests_761aeb891d96dd7317258a72ed79873b5d2c7ac0_24_10_23_14_08_44/0/task?bookmarks=0,1779&shareLine=192).

This PR extends the `BSONCXX_POLYFILL` pattern to the `"compile"` EVG function and `compile.sh` script for consistency, replacing the `USE_POLYFILL_BOOST` env var. The task now [compiles with bsoncxx impls](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_rhel79_compile_compile_without_tests_patch_761aeb891d96dd7317258a72ed79873b5d2c7ac0_6720f2c1b48a5c00078003e1_24_10_29_14_35_46/0/task?bookmarks=0,1703&shareLine=96) as intended. This revealed some compilation errors due to unused named parameters in `optional.hpp`, which this PR also addresses.

As a drive-by improvement, this PR also moves all compile/build/etc. tasks to large distro variants to reduce task execution times.

### DIFF
--- a/.evergreen/compile.sh
+++ b/.evergreen/compile.sh
@@ -15,6 +15,7 @@ set -o pipefail
 : "${build_type:?}"
 : "${distro_id:?}" # Required by find-cmake-latest.sh.
 
+: "${BSONCXX_POLYFILL:-}"
 : "${COMPILE_MACRO_GUARD_TESTS:-}"
 : "${ENABLE_CODE_COVERAGE:-}"
 : "${ENABLE_TESTS:-}"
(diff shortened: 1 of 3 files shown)


kevinAlbs (approved):
Thank you for the fix. LGTM

PR #1244, opened 2024-10-29
CXX-3126 Refactor EVG config to use config_generator

eramongodb:
# New

## Summary

Resolves CXX-3126. Verified by [this patch](https://spruce.mongodb.com/version/67214670d11baa000782c9c0). Reattempt of https://github.com/mongodb/mongo-cxx-driver/pull/1242.

Imports the EVG config generator from the C Driver and converts the current EVG config into generator components. Changes to task matrices are deliberately minimized. Auditing and improving the task matrices is outside the scope of this PR.

Once this PR is merged, the `.mci.yml` file may be deleted in favor of the new `.evergreen/config.yml` file (for consistency with the C Driver). They are equivalent given the changes in this PR.

## Astral uv

The projects managed by our team have seen a steady progression of Python tooling: a pip `requirements.txt` file -> virtual environments (`venv`) -> Poetry project configuration (`pyproject.toml`) -> Python binary management (`pyenv` in `./tools/python.sh`). This PR takes this opportunity to take another step by adopting a new Python tool which aims to supercede _all_ such tooling which came prior: [uv](https://docs.astral.sh/uv/).

> [!IMPORTANT]
> This PR does _not_ use uv in Evergreen and does _not_ add any scripts to assist with obtaining and using uv. Users should install uv themselves according to uv's [installation instructions](https://docs.astral.sh/uv/getting-started/installation/) as best appropriate for their local development environment. Exploring the use of uv in Evergreen scripts is outside the scope of this PR.

The shockingly impressive convenience of this tool is demonstrated by running the following command, whose only prerequisite is that `uv` is installed and available to use:

```bash
uv run --frozen .evergreen/config_generator/generate.py
```

> [!NOTE]
> The `--frozen` argument is needed to avoid updating the lock file when running `uv` commands.

Running this _single_ command accomplishes _all_ of the following steps:

- automatically [detects and selects](https://docs.astral.sh/uv/guides/install-python/#using-an-existing-python-installation) a suitable Python binary on the system if one already exists; otherwise, [automatically downloads](https://docs.astral.sh/uv/guides/install-python/#automatic-python-downloads) a suitable Python binary [for subsequent (re)use](https://docs.astral.sh/uv/concepts/cache/#cache-directory).
    - This supercedes Python binary management, such as with `pyenv`, system package managers, and manual installation, which is arguably the most frustrating part of maintaining reproducibility of Python scripts.
- automatically creates a [project virtual environment](https://docs.astral.sh/uv/concepts/projects/#running-scripts) (`.venv` by default) for (re)use by all subsequent Python package operations and script execution, using the Python binary selected in the earlier step.
    - This supercedes virtual environment management, such as with `venv` or `poetry`, but goes even further by properly handling Python binary version compatibility through (re)generation of the virtual environment [according to the requested Python version](https://docs.astral.sh/uv/reference/cli/#uv-venv) (via `-p`/`--python`).
- automatically resolves and installs script dependencies within the project virtual environment according to the [project configuration file](https://docs.astral.sh/uv/configuration/files/).
    - This supercedes project-level package requirement specifications, such as with `requirements.txt` or `pyproject.toml` (Poetry), but goes even further by supporting [resolution strategies](https://docs.astral.sh/uv/concepts/resolution/) to easily validate minimum (or maximum) dependency version requirements.

The package version requirements in this PR were obtained by initially using [inline script metadata](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies) and `--resolution lowest-direct` for requirement validation before copying the dependencies into the `pyproject.toml` file, demonstrating the power of built-in Python binary management and Python-version-aware virtual environment management:

```bash
# Inline script metadata -> automatic isolated virtual environment.
# Straightforward Python (minimum) version testing with `-p`/`--python`.
uv run -p 3.10 --resolution lowest-direct path/to/script.py
```
> [!NOTE]
> This behavior can be reproduced despite the presence of the project configuration file using a combination of `--no-project`, `--isolated`, and `--with`:
>
> ```bash
> run_args=(
>     --no-project # Avoid detecting project dependencies.
>     --isolated   # Avoid reusing existing virtual environments.
>     # Validate the following package version requirements are correct.
>     --resolution lowest-direct
>     # Package requirements to use only for the following command.
>     --with 'shrub-py>=3.3.1,pydantic>=2.0,packaging>=14.0'
> )
> uv run "${run_args[@]:?}" .evergreen/config_generator/generate.py
> ```

> [!NOTE]
> Inline script metadata is not used to specify dependencies due to isolated virtual environments interacting poorly with external tooling which expect a persistent virtual environment to be present. The dependencies are instead (redundantly) listed per associated script in the project configuration file, which may be used to generate a local and persistent virtual environment (`.venv` by default) using `uv sync --frozen` (or as part of a `uv run --frozen` command).

> [!NOTE]
> Although testing with `lowest` would be ideal, the `lowest-direct` resolution was used instead due to insufficiently-specified requirements in transitive dependencies leading to unresolvable failures. This PR avoids introducing [constraints or overrides](https://docs.astral.sh/uv/concepts/resolution/) to keep things simple.

> [!NOTE]
> The `clang_format.py` script still requires Python 2. A `requires-python` was added to enforce this requirement (note `uv` does not support Python 2):
> ```
> $ uv run etc/clang_format.py
> Reading inline script metadata from `etc/clang_format.py`
> error: No interpreter found for Python <3.0 in virtual environments, managed installations, or system path
> 
> $ uv run -p 2 etc/clang_format.py
> Reading inline script metadata from `etc/clang_format.py`
> error: Invalid version request: Python <3 is not supported but 2 was requested.
> ```
> Updating this script to work with Python 3 is deferred, but would make a good case for using [`uv tool`/`uvx`](https://docs.astral.sh/uv/guides/tools/), which supercedes `pipx`, e.g. `uvx clang-tools -t clang-format -i <version> -d build && ./build/clang-format --version`.

Despite uv not yet [supporting a stable API](https://docs.astral.sh/uv/reference/versioning/), I believe its powerful featureset (and already high popularity) makes it an very valuable tool which we can and should adopt in our projects. This PR hopes to be a leading test case for its eventual adoption by other projects as well.

## EVG Config Generator Adjustments

Some adjustments were required to support the C++ Driver EVG config. These adjustments include:

- Adding `.evergreen` to `sys.path` to permit relative imports of `config_generator` modules regardless of the current working directory.
- Adding missing distros to `distros.py` and support for `*-latest` distros.
- Adding VS 2019+ support to `distros.py` and extending compiler helpers to help specify corresponding CMake generators and platforms (e.g. `vs2019x64` -> `CMAKE_GENERATOR="Visual Studio 16 2019"` + `CMAKE_GENERATOR_PLATFORM="x64"`).
- Adding support for the `teardown_task_can_fail_task` field for task groups.
- Fixing pydantic 2.0 compatibility issues with [subclass field behavior](https://docs.pydantic.dev/latest/migration/) with `serialize_as_any=True`:
  >  In V1, we would always include all fields from the subclass instance. In V2, when we dump a model, we only include the fields that are defined on the annotated type of the field. This helps prevent some accidental security bugs.

## EVG Config Migration

> [!TIP]
> Reviewing this PR by-commit is recommended to more easily compare how individual functions and components are migrated from the old config into their generator component form.

Most functions are translated as-is into their component form with minimal changes. Functions commonly reused by components are given explicit parameters in their `call()` functions to help with consistency and reduce verbosity (e.g. `mongodb_version` -> `TOPOLOGY`, `polyfill` -> `BSONCXX_POLYFILL`, etc.).

Scripts under `.evergreen` which are invoked by the EVG config are relocated into the `.evergreen/scripts` directory, formatted, given executable permissions, and audited with `shellcheck` (excluding the packaging-related scripts). Scripts under `etc` are left in their current location.

Some lessons learned during the C Driver's migration are applied in this PR. Unlike with the C Driver's generator components, the C++ Driver's components minimize cross-component inclusion and reuse (with the exclusion of function components). Despite leading to some repetition and verbosity across components, this is done to improve separation-of-concerns of components and matrices. In particular, despite its large matrix and complicated generation routine, I believe the single `integration` component is nevertheless more straightforward and understandable than the layered `sasl` -> `cse` -> `asan`/`tsan` components in the C Driver. Unlike with the C Driver, sanitizer and valgrind tasks are grouped into seperate, completely independent components.

> [!NOTE]
> Although many tasks are grouped into a display task (e.g. auth-matrix, compile-only-matrix, etc.), this is not done for the integration matrix. This is to facilitate better filtering, selection, and sorting in Spruce, as such operations appear to be limited for members of a display task. Grouping by distro, by build type, and by server version/topology were all considered but reverted in favor of the current no-grouping state. This may be reconsidered if the Spruce UI is improved to better handle filtering/selecting/sorting members of display tasks: see [DEVPROD-12402](https://jira.mongodb.org/browse/DEVPROD-12402).

Some additional notes:

- Some ARM64 distros had batchtimes, which does not appear to be necessary per EVG distro guidelines. These batchtimes were therefore removed.
- Auth tests appear to be somewhat flaky, but it is unclear why. They appear to be passing for the moment.

## Extra Alignment

> [!IMPORTANT]
> Adding support for configuring the auto-downloaded C Driver with `ENABLE_EXTRA_ALIGNMENT=ON` was considered but rejected to avoid encouraging use of extra alignment; see also CDRIVER-2813 and MONGOCRYPT-725.

Validating correct migration revealed that extra alignment tasks were broken by https://github.com/mongodb/mongo-cxx-driver/pull/967. The `BSON_EXTRA_ALIGNMENT` has no affect on `fetch_c_driver_source` nor the subsequent call to the `compile` EVG function (despite misleadingly being set for the `compile` function). Extra alignment was therefore not being enabled and tested with auto-downloaded C Driver configurations, as the auto-downloaded C Driver disables extra alignment by default. All current extra alignment tasks use the auto-downloaded C Driver, therefore extra alignment current has zero test coverage.

Fixing the enabling of extra alignment (by reverting relevant tasks to using `install_c_driver` with `BSON_EXTRA_ALIGNMENT=ON` instead of `fetch_c_driver_source`) revealed compiler errors due to `-Werror` + `-Waligned-new` (enabled by `-Wall`):

```
src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp: In substitution of 'template<class ... Args, class> static std::unique_ptr<bsoncxx::v_noabi::builder::core::impl> bsoncxx::v_noabi::stdx::detail::make_unique_impl<bsoncxx::v_noabi::builder::core::impl>::make<Args ..., <template-parameter-1-2> >(std::true_type, Args&& ...) [with Args = {long unsigned int}; <template-parameter-1-2> = <missing>]':
src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp:134:48:   required by substitution of 'template<class T, class Impl, typename std::enable_if<std::is_array< <template-parameter-1-1> >::value, decltype ((Impl::make(std::integral_constant<bool, true>{}, declval<long unsigned int>()), void()))>::type* <anonymous> > std::unique_ptr<T> bsoncxx::v_noabi::stdx::make_unique(std::size_t) [with T = bsoncxx::v_noabi::builder::core::impl; Impl = bsoncxx::v_noabi::stdx::detail::make_unique_impl<bsoncxx::v_noabi::builder::core::impl>; typename std::enable_if<std::is_array< <template-parameter-1-1> >::value, decltype ((Impl::make(std::integral_constant<bool, true>{}, declval<long unsigned int>()), void()))>::type* <anonymous> = <missing>]'
src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/builder/core.cpp:259:45:   required from here
src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp:52:15: error: 'new' of type 'bsoncxx::v_noabi::builder::core::impl' with extended alignment 128 [-Werror=aligned-new=]
   52 |               typename = decltype(new T(std::declval<Args>()...))>
      |               ^~~~~~~~
src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp:52:15: note: uses 'void* operator new(std::size_t)', which does not have an alignment parameter
src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/stdx/make_unique.hpp:52:15: note: use '-faligned-new' to enable C++17 over-aligned new support
```

This appears to be due to https://github.com/mongodb/mongo-cxx-driver/pull/1049 which removed the `-Wno-aligned-new` flag "due to conflicts with Clang tasks + appearing to cause no trouble for existing tasks (not entirely clear what motivated its addition or if it still applies)". This PR being dated Nov 2023 likely explains why "no trouble" was observed following https://github.com/mongodb/mongo-cxx-driver/pull/967 on Aug 2023. The `-Wno-aligned-new` flag is restored with exclusions for Clang using the since-introduced `CC`/`CXX` env vars.


eramongodb:
Accidentally dropped extra alignment fixes for valgrind tasks during a rebase. Addressed and verified by [this patch](https://spruce.mongodb.com/version/67214e86fd12610007fb3e10). Aside: libmongocrypt is disabled for sanitizer tasks for more accurate current-config reproduction, but we probably want to "revert" this and extend sanitizer tasks to cover CSFLE code paths soon.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/6722473512fc8e0007233c5a).

.evergreen/scripts/compile-scan-build.sh
```diff
@@ -67,14 +67,18 @@ scan_build_flags=(
...
 impls) cmake_flags+=(-D "BSONCXX_POLY_USE_IMPLS=ON") ;;
 std) cmake_flags+=(-D "BSONCXX_POLY_USE_STD=ON") ;;
 esac
 
-echo "Configuring with CMake flags: ${cmake_flags[*]}"
+echo "Configuring with CMake flags:"
+for flag in "${cmake_flags[@]}"; do
+  echo " - ${flag:?}"
+done
+echo
```
vector-of-bool, line 81: Minor shorthand with `printf`:

```suggestion
echo "Configuring with CMake flags:"
printf " - %s\n" "${cmake_flags[@]}"
echo
```

kevinAlbs (approved):
LGTM. The config generation, use of display tasks, and fixes to extra alignment tasks are very much appreciated.

PR #1248, opened 2024-11-01
CXX-3160 Remove mongocxx::stdx

eramongodb:
Resolves CXX-3160. Verified by [this patch](https://spruce.mongodb.com/version/67251faab1427c000737467e).

The redeclaration of `bsoncxx::stdx` as `mongocxx::stdx` has caused ambiguities, confusion, and difficulties using `stdx` interfaces when both `bsoncxx` and `mongocxx` are present (e.g. CXX-1060). It is unclear what the original motivation for this pattern is (CXX-716), but given its dubious value and recurring troubles, this PR proposes to remove the namespace in v4 in favor of consistently using `bsoncxx::stdx` instead.

> [!NOTE]
> The 3.11.0 release documented the `mongocxx/stdx.hpp` header as "for internal use only" in https://github.com/mongodb/mongo-cxx-driver/pull/1184.


src/mongocxx/include/mongocxx/v_noabi/mongocxx/bulk_write.hpp
```diff
@@ -100,7 +100,7 @@ class bulk_write {
     /// @see
     /// - https://www.mongodb.com/docs/manual/core/bulk-write-operations/
     ///
-    MONGOCXX_ABI_EXPORT_CDECL(stdx::optional<result::bulk_write>) execute() const;
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::stdx::optional<result::bulk_write>) execute() const;
```
kevinAlbs, line 103:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v_noabi::stdx::optional<result::bulk_write>) execute() const;
```

Suggest using `v_noabi` namespace. Recalling https://github.com/mongodb/mongo-cxx-driver/pull/1066:

> entities within an ABI namespace MUST NOT be defined in terms of root namespace declarations

Suggestion applies to other (non-test) instances in this PR.

kevinAlbs (approved):

PR #1255, opened 2024-11-01
CXX-3157 Remove mongocxx::v_noabi::options::create_collection

eramongodb:
Resolves CXX-3157. Long overdue followup to CXX-1943. Despite being a larger API breaking change than others currently scheduled for v4, given the amount of time which has passed since its deprecation in https://github.com/mongodb/mongo-cxx-driver/commit/45f62832b3889aaf54b0a3c732c66bec01ab6468 (March 2020 for the 3.5.0 release) and plenty of alternative interfaces to achieve the same behavior, I believe this should be a safe removal.


vector-of-bool (approved):
LGTM

PR #1257, opened 2024-11-07
 CXX-2797 Remove support for external polyfill libraries 

eramongodb:
## Summary

Resolves CXX-2797. Verified by [this patch](https://spruce.mongodb.com/version/672ce1406512fc0008ef534e).

This PR is the long-awaited followup to https://github.com/mongodb/mongo-cxx-driver/pull/1062, https://github.com/mongodb/mongo-cxx-driver/pull/1075, and https://github.com/mongodb/mongo-cxx-driver/pull/1086 which fully removes support for external polyfill libraries, namely mnmlstc/core and Boost). `ENABLE_BSONCXX_POLY_USE_IMPLS=ON` is now implicit behavior, which selects bsoncxx impls for pre-C++17 and the standard library otherwise.

> [!NOTE]
> The `stdx::make_unique*` polyfills are not configuration-dependent as they use their own feature detection routines (following https://github.com/mongodb/mongo-cxx-driver/pull/1028 and https://github.com/mongodb/mongo-cxx-driver/pull/1086) and do not affect either the API (just a simple function template interface compatible with the stdlib) or ABI (no exported symbols), so they are not included or mentioned in these changes.

## CMake Configuration

All references to mnmlstc/core and Boost are removed from the CMake configuration, including related variables, options, and files. These include:

- `ENABLE_BSONCXX_POLY_USE_IMPLS` (now implicit behavior).
    - `BSONCXX_POLY_USE_IMPLS=ON` can be used to explicitly opt-out of using stdlib with C++17 or newer (hopefully no one needs to do this?).
    - `BSONCXX_POLY_USE_STD=ON` can be used to explicitly opt-out of using bsoncxx impls with pre-C++17 (probably pointless?).
- `BSONCXX_POLY_USE_MNMLSTC`
  - `BSONCXX_POLY_USE_SYSTEM_MNMLSTC`
  - `BSONCXX_POLY_MNMLSTC_DEPRECATED_INCLUDE_DIRS` (used for legacy CMake package config files; obsoleted by https://github.com/mongodb/mongo-cxx-driver/pull/1037)
  - `BSONCXX_POLY_MNMLSTC_PKGCONFIG_INCLUDE_DIRS` (used for pkg-config config files)
- `BSONCXX_POLY_USE_BOOST`
  - `BOOST_ROOT`
  - `BSONCXX_BOOST_PKG_DEP` (used for pkg-config config files)
- The `FetchMnmlstcCore.cmake` file.
- The `src/bsoncxx/third_party` directory and related files.

## Source Files

- The `BSONCXX_POLY_USE_*` preprocessor macros for mnmlstc/core and Boost are removed, including relevant conditional compilation (i.e. in stdx headers), macro guards, and documentation.
- Workaround for Boost and `noexcept`-ness in `bsoncxx/types/bson_value/view.cpp` is removed.
- The license notice for mnmlstc/core is removed from `THIRD-PARTY-NOTICES` as it is no longer used.
- Mentions of mnmlstc/core are removed from release instructions.
- Mentions of mnmlstc/core are removed from SSDLC files (purls, SBOMs, 3rd Party Vulnerabilities, etc.).

## Evergreen Config and Scripts

- All tasks which do not explicitly set a polyfill library now compile against either bsoncxx impls or stdlib according to the C++ standard (`CMAKE_CXX_STANDARD`).
- MacOS tasks now use bsoncxx impls instead of Boost.
- `compile.sh` no longer defaults to C++17 for VS 2017 or newer.
  - This makes MSVC tasks consistent with Linux/MacOS regarding polyfill selection behavior.
- `run-clang-tidy.sh` still compiles with C++17, but the note regarding mnmlstc/core is removed.
  - We may want to add a pre-C++17 equivalent to this task matrix so that bsoncxx impls configurations is also scanned.
- Example projects are now unconditionally compiled with VS 2017 on Windows regardless of the requested C++ standard.
  - We may want to eventually update these scripts so they use a consistent generator+platform as what is used to build the driver libraries (via `$generator` and `$platform`) and avoid redundant variables (e.g. `$example_*`).

## Packaging Files and Scripts

- The `debian-package-build-mnmlstc` task is removed.
- The `build_snapshot_rpm.sh` and `debian_package_build.sh` scripts remove all mentions of mnmlstc/core or Boost.
- The `mongo-cxx-driver.spec` and `spec.patch` files are deliberately left untouched; appropriate changes are deferred to a followup PR to properly sync these updates with the `debian/unstable` branch (CC @rcsanchez97).
  - `debian-package-build` [task failure](https://spruce.mongodb.com/task/mongo_cxx_driver_packaging_debian_package_build_patch_0421d9916f6ee8da5dfb31aac58a3ab29f2acd34_672ce1406512fc0008ef534e_24_11_07_15_48_17/logs?execution=0): `dpkg-checkbuilddeps: error: Unmet build dependencies: libboost-dev`. Attempts to resolved this error failed, likely due to requiring corresponding changes to the `debian/unstable` branch.

## Catch2 and VS 2017

Test cases for `stdx::optional<T>` were failing to compile on VS 2017 due to numerous errors of the form:

```
error C2039: 'value': is not a member of 'Catch::Detail::is_eq_comparable<LhsT,U,void>'
error C2039: 'value': is not a member of 'Catch::Detail::is_ne_comparable<LhsT,T,void>'
...
```

This appears to be specifically due to poor compatibility between Catch2's expression decomposition implementation (for purposes of expressive error message construction) and VS 2017 concerning friend function/operator overload templates, which are heavily used by `stdx::optional<T>` via `equality_operators` and `tag_invoke` in `operators.hpp`.

Because this appears to be specifically a Catch2 + VS 2017 compatibility issue rather than an implementation correctness issue, the optional test cases were refactored to avoid directly using comparison operators in its assertion expressions (see: new `*_VS2017` macros) to avoid expression decomposition.

Although the compiler errors only mention equality operators, relational comparison operators also appear to be affected by this issue. For simplicity and consistency, rather than narrowing down the exact types of comparison that trigger these errors (e.g. by comparison type, argument type, etc.), the workaround is applied to _all_ relational comparisons in optional test cases using pattern-matching find-and-replace.


kevinAlbs (commented):
Exciting!

Remove polyfill related Evergreen config. Changes may have been overwritten by https://github.com/mongodb/mongo-cxx-driver/pull/1244. (e.g. `packaging-debian-mnmlstc` is still present)

eramongodb:
> Changes may have been overwritten

Good catch, thank you. Reapplied the config changes in terms of the EVG config generator.

kevinAlbs (approved):

vector-of-bool (approved):
Huzzah!

rcsanchez97:
> - The mongo-cxx-driver.spec and spec.patch files are deliberately left untouched; appropriate changes are deferred to a followup PR to properly sync these updates with the debian/unstable branch (CC @rcsanchez97).
> - debian-package-build [task failure](https://spruce.mongodb.com/task/mongo_cxx_driver_packaging_debian_package_build_patch_0421d9916f6ee8da5dfb31aac58a3ab29f2acd34_672ce1406512fc0008ef534e_24_11_07_15_48_17/logs?execution=0): dpkg-checkbuilddeps: error: Unmet build dependencies: libboost-dev. Attempts to resolved this error failed, likely due to requiring corresponding changes to the debian/unstable branch.

Ack. I have it on my todo list to look at and fix both the RPM and DEB packaging tasks.

PR #1262, opened 2024-11-08
CXX-3093 Revert include-guard-conditioned macro guard in mongocxx 

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1202, where use of these "smart macro guards" conditioned on include guard macros was intended to be reverted (see: https://github.com/mongodb/mongo-cxx-driver/pull/1202#discussion_r1757495333) but neglected to apply the changes to mongocxx headers. Reverting now for consistency + may cherry-pick onto the r3.11 branch on the off-chance this _does_ somehow happen to break downstream code (e.g. due to missing postlude header include directives...? I really don't expect this to _actually_ break any downstream users, but nevertheless...)


eramongodb:
> IMO a cherry-pick seems unnecessary.

Will leave as-is then for the r3.11 branch.

PR #1266, opened 2024-11-11
CXX-3179 Set ENABLE_TESTS=OFF by default

eramongodb:
Resolves CXX-3179. Verified by [this patch](https://spruce.mongodb.com/version/673283e445a9270007fcea3c).

Changes the default value of `ENABLE_TESTS` from `ON` to `OFF`.

Tweaked descriptions of CMake options to better document how they are conditioned on `ENABLE_TESTS=ON` (e.g. `BUILD_TESTING=ON` alone does _not_ enable test targets). In particular, the difference between `BUILD_TESTING` (add test targets to "all") and `ENABLE_TESTS` (enable the test targets which may be added to "all") is clarified by the addition of a warning when `BUILD_TESTING=ON` + `ENABLE_TESTS=OFF` (which suggests a conflict of intent).

All instances of `ENABLE_TESTS=OFF` in Evergreen scripts (except in `spec.patch`) are removed in favor of using off-by-default behavior. All EVG tasks which require test targets or their corresponding build artifacts now explicitly set `ENABLE_TESTS=ON` (which atm always implies `BUILD_TESTING=ON`). In short, building test targets is now explicitly opt-in both in CMake configuration and in Evergreen scripts.

### DIFF
--- a/.evergreen/config_generator/components/atlas_search_indexes.py
+++ b/.evergreen/config_generator/components/atlas_search_indexes.py
@@ -55,7 +55,7 @@ def tasks():
             run_on=distro.name,
             commands=[
                 InstallCDriver.call(),
-                Compile.call(build_type='Debug'),
+                Compile.call(build_type='Debug', vars={'ENABLE_TESTS': 'ON'}),
                 TestSearchIndexHelpers.call(),
             ],
         )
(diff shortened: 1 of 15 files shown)


kevinAlbs (approved):
LGTM. I think the updated description for `ENABLE_TESTS` and new warning helps clarify the difference.

PR #1272, opened 2024-11-18
Use Artifactory instead of Amazon ECR to obtain Docker images

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1270. Verified by [this patch](https://spruce.mongodb.com/version/673b872168da510007e0d3ea).

Uses Artifactory for consistency with https://github.com/mongodb/mongo-c-driver/pull/1794 and https://github.com/mongodb/libmongocrypt/pull/911.


kevinAlbs (approved):

kevinAlbs:
I am unsure if using Artifactory is an acceptable long-term solution. Filed CXX-3182 (along with MONGOCRYPT-747 and CDRIVER-5814) to track removing workarounds once DEVPROD-12779 is resolved.

PR #1278, opened 2024-11-20
CXX-2813 document KMIP "delegated" option

kevinAlbs:
Document the KMIP "delegated" masterKey option. Fix rendering of documentation for `data_key::master_key`:

Before:
Before changes

After:
After changes


eramongodb (approved):
Good catch. LGTM.

PR #1281, opened 2024-11-20
Fix indentation level of CHANGELOG entries

eramongodb:
Unintentionally(?) detected by the `etc/make_release.py` script:

```
Error: Failed to find changelog contents for 4.0.0
```

### DIFF
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -411,9 +411,9 @@ Run the release script with the git tag created above as an argument and
 
 ```bash
 make_release_args=(
+    -r upstream
     --jira-creds-file ~/.secrets/jira-creds.txt
     --github-token-file ~/.secrets/github-token.txt
-    -r upstream # Only if mongodb/mongo-cxx-driver is not "origin".
 )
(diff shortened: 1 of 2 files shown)


etc/releasing.md
```diff
@@ -431,6 +437,7 @@ follows:
   distribution tarball.
 - If the script succeeded at creating the distribution tarball, pass it directly
   with `--dist-file ./build/mongo-cxx-driver-rX.Y.Z.tar.gz`.
+- If the only open JIRA ticket is the ticket tracking the release itself, pass `--allow-open-issues`.
```
kevinAlbs, line 440:
```suggestion
```

I expect this is not needed. An automation creates the "Release x.y.z" ticket when a fixVersion is added. The created ticket gets a `release` label (example CXX-3104). Tickets with the `release` label are [filtered out](https://github.com/mongodb/mongo-cxx-driver/blob/2a118824477fc5f21e01d8c873c5cab27b617b04/etc/make_release.py#L506).

I expect the "Release 4.0.0" ticket was created manually and did not get the label.

kevinAlbs (approved):
LGTM with a suggested step removed.

PR #1282, opened 2024-11-20
add release instruction for webhook

kevinAlbs:
To avoid excessive e-mails when updating the `debian/unstable` branch, the GitHub Webhook is updated to limit posting to selected branches. As a consequence, this requires updating the settings on each minor release.

Quoting https://wiki.corp.mongodb.com/display/INTX/Githook:

> The default behavior of this webhook is to post commit to all branches in a repo to JIRA.  If you want to limit the commit activity to a specific repository or set of repositories, you must do the following:
> - https://githook.mongodb.com/?branches=master --> Master only
> - https://githook.mongodb.com/?branches=master..legacy..someotherbranch --> Master, Legacy and Someotherbranch.  The syntax is strange (.. not &)

### DIFF
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -331,6 +331,16 @@ Commit the updates to `CHANGELOG.md`.
 git commit -m 'Update CHANGELOG for X.Y.Z'
 ```
 
+### Update GitHub Webhook
+
+For a non-patch release, update the [Github Webhook](https://wiki.corp.mongodb.com/display/INTX/Githook) to include the new branch.
+
+Navigate to the [Webhook Settings](https://github.com/mongodb/mongo-cxx-driver/settings/hooks)
+
+Click `Edit` on the hook for `https://githook.mongodb.com/`.
+
+Add the branch to the `Payload URL`.
+
 ### Pre-Release Changes PR
 
 Push the `pre-release-changes` branch to a fork repository and create a PR to merge `pre-release-changes` onto `master`:
--- a/etc/releasing.md
+++ b/etc/releasing.md
@@ -331,16 +331,6 @@ Commit the updates to `CHANGELOG.md`.
 git commit -m 'Update CHANGELOG for X.Y.Z'
 ```
 
-### Update GitHub Webhook
-
-For a non-patch release, update the [Github Webhook](https://wiki.corp.mongodb.com/display/INTX/Githook) to include the new branch.
-
-Navigate to the [Webhook Settings](https://github.com/mongodb/mongo-cxx-driver/settings/hooks)
-
-Click `Edit` on the hook for `https://githook.mongodb.com/`.
-
-Add the new release branch to the `Payload URL`. Remove unmaintained release branches.
-
 ### Pre-Release Changes PR
 
 Push the `pre-release-changes` branch to a fork repository and create a PR to merge `pre-release-changes` onto `master`:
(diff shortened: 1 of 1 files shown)


etc/releasing.md
```diff
@@ -331,6 +331,16 @@ Commit the updates to `CHANGELOG.md`.
 git commit -m 'Update CHANGELOG for X.Y.Z'
 ```
 
+### Update GitHub Webhook
```
eramongodb, line 334: Recommend moving this script down immediately before the "Update releases/stable" step to avoid updating the webhook before additional changes may be committed prior/during the release.

etc/releasing.md
```diff
@@ -331,6 +331,16 @@ Commit the updates to `CHANGELOG.md`.
...
+
+For a non-patch release, update the [Github Webhook](https://wiki.corp.mongodb.com/display/INTX/Githook) to include the new branch.
+
+Navigate to the [Webhook Settings](https://github.com/mongodb/mongo-cxx-driver/settings/hooks)
+
+Click `Edit` on the hook for `https://githook.mongodb.com/`.
+
+Add the branch to the `Payload URL`.
```
eramongodb, line 342: Recommend keeping only actively-maintained branches in the list, which I believe would currently mean the following:

```
https://githook.mongodb.com/?branches=master..releases/v3.11..releases/v4.0
```

eramongodb (changes requested):
Thank you for these updates.

kevinAlbs:
Reverted changes to the webhook to again apply to all branches. Recent deliveries with the `?branches=` argument receive HTTP 400 errors.

kevinAlbs:
> HTTP 400 errors.

I misdiagnosed the error. The HTTP 400 resulted from a commit to `gh-pages`. The webhook expectedly rejected the request due to being excluded in the list of branches (confirmed in [slack thread](https://mongodb.slack.com/archives/C05TFJN9LDB/p1732216594126699?thread_ts=1732205419.136939&cid=C05TFJN9LDB)).

To verify the `branches=` argument works, I temporarily changed the webhook to include `test-branch`, pushed a commit including "CXX-1114", and saw a [comment posted](https://jira.mongodb.org/browse/CXX-1114?focusedCommentId=6877611&focusedId=6877611&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6877611).

The webhook is updated to `https://githook.mongodb.com/?branches=master..releases/v3.11..releases/v4.0`.

kevinAlbs:
FYI @rcsanchez97 the updated web hook will prevent changes to the `debian/unstable` branch from posting to Jira tickets ([example](https://jira.mongodb.org/browse/CXX-2990?focusedCommentId=6481701&focusedId=6481701&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6481701)). I expect this is OK (and probably desirable). If OK, this change can similarly be applied to C and libmongocrypt.

etc/releasing.md
eramongodb: Poke, not yet resolved.

eramongodb (changes requested):

etc/releasing.md
kevinAlbs: Oops. Good catch. I remember relocating this section, but do not know how I lost the changes. Updated.

eramongodb (approved):

rcsanchez97 (approved):
@kevinAlbs I agree that all the extra commit notifications that result from merging commits into `debian/unstable` are unnecessary. It's too bad that there isn't a way to configure the receiver to simply discard notifications for the specified branch(es), but this approach works too.

LGTM.

PR #1284, opened 2024-11-21
Exclude the releases/stable tag from matching 'r*' tags in calc_releases_version.py

eramongodb:
Addresses EVG task failures due to the following error:

```
Exception: Expected 1 or 0 tags on HEAD, got: ['r4.0.0', 'releases/stable']
```

This appears to be due to `releases/stable` matching the `'r*'` tagname filter now that it is a tag rather than a branch. This PR adds a filter to remove the `releases/stable` tag from results. No other tags should start with `'r'`.


eramongodb:
Closed in favor of https://github.com/mongodb/mongo-cxx-driver/pull/1284.

PR #1294, opened 2024-12-03
CXX-3187 Extend ABI stability tasks to C++20 and C++23

eramongodb:
Resolves CXX-3187. Verified by [this patch](https://spruce.mongodb.com/version/674f476332d276000786090a/).

Extends the ABI Stability Checks matrix to include C++20 and C++23 for compilation coverage. This PR does not include extension of execution or test coverage.

Tasks were mostly migrated to the `rhel9-latest` distro to obtain Clang 17, which has the most C++23 feature coverage across compilers currently available on Evergreen (see: [Compiler Support](https://en.cppreference.com/w/cpp/compiler_support)). The `abidiff-*` tasks remain on `ubuntu2204` (which provide Clang 12) due to lack of `abidiff` availability (could not find an appropriate package). The latest comparable GCC compiler version appears to be GCC 13.

> [!NOTE]
> There should be no issues with mixing GCC (abi-compliance-checker) and Clang (C++ Driver libraries) in these circumstances due to the consistent use of libstdc++ for ABI consistency.

Switched from `--symbols-only` to `--portability` for compatibility with `nm` on RHEL.

Compilation output is now uploaded as `ABI Compatibility Setup: old.txt` and `ABI Compatibility Setup: new.txt` to help verify the build behaves as expected on success.

Tasks in the task group are explicitly listed rather than matched via tags to avoid task scheduling issues ([DEVPROD-13210](https://jira.mongodb.org/browse/DEVPROD-13210)).


eramongodb:
Investing why Clang 17 is not being selected/used as expected on RHEL 9 distros.

eramongodb:
It seems the default Clang installed on the system is Clang 12, not Clang 17 as indicated by the distro image visibility page. Explicitly qualifying with `-17` fixes the issue. Latest changes verified by [this patch](https://spruce.mongodb.com/version/674f4f61b2144300076a855b).

eramongodb:
On further inspection, the build output log files were being overwritten by subsequent tasks being executed due to using the same remote file path across tasks. Log files are now distinct across tasks as intended by adding `${task_id}` to the remote file path. The compiler version suffix is preserved for readability.

kevinAlbs (approved):
LGTM. Thank you for the added tasks.

PR #1300, opened 2024-12-10
CXX-1110 add CHECK_THROWS_WITH_CODE

eramongodb:
Resolves CXX-1110. Verified by [this patch](https://spruce.mongodb.com/version/675877e0c945f10007c344e2/) (containing the proposed changes in https://github.com/mongodb/mongo-cxx-driver/pull/1298).

Precursor changes to faciliate the improvement and addition of tests for bsoncxx and mongocxx error codes and error categories as part of CXX-834 and CXX-2745.

The two new macros `(CHECK|REQUIRE)_THROWS_WITH_CODE` mirror the existing Catch2 exception assertion macros `(CHECK|REQUIRE)_THROWS_WITH`. It is expected to be used as follows:

```cpp
CHECK_THROWS_WITH_CODE(throwing-expr, error-code-enumerator);
CHECK_THROWS_WITH_CODE(throwing-expr, error-code);
CHECK_THROWS_WITH_CODE(throwing-expr, error-condition);
```

The exception type must be or derive from `std::system_error` which provides the error code (via `.code()`) to be compared against. Otherwise, the test assertion fails via an always-false `*_THROWS_WITH(throw, std::system_error)` to reuse Catch's exception type-match error messages.

The option to define and use a custom Catch matcher for use with `*_THROWS_WITH` was rejected due to not correctly supporting Catch test exception propagation and not fully supporting customization of the resulting error messages due to unavoidable internal exception handler routines being executed within Catch. It was also comparatively unwieldy to write due to having to construct the custom matcher as an argument (its omission would just lead to the same `*_THROWS_WITH_CODE` interface as proposed by this PR):

```cpp
CHECK_THROWS_WITH(throwing-expr, Code(error-code));
```

Although Catch seems to support string-making a `std::error_code` (e.g. `std::errc::invalid_argument` -> `generic:22`), it doesn't seem to support `std::error_condition`, so a specialization is also added which uses the same format as error codes. This produces assertion messages like the following:

```cpp
// path/to/file:line: PASSED:
//   CHECK( ex.code() == (std::errc::invalid_argument)) )
// with expansion:
//   generic:22 == 22
CHECK_THROWS_WITH_CODE(..., std::errc::invalid_argument);

// path/to/file:line: PASSED:
//   CHECK( ex.code() == (std::make_error_code(std::errc::invalid_argument)) )
// with expansion:
//   generic:22 == generic:22
CHECK_THROWS_WITH_CODE(..., std::make_error_code(std::errc::invalid_argument));

// path/to/file:line: PASSED:
//   CHECK( ex.code() == (std::make_error_condition(std::errc::invalid_argument)) )
// with expansion:
//   generic:22 == generic:22
CHECK_THROWS_WITH_CODE(..., std::make_error_condition(std::errc::invalid_argument));
```

and in the event of an unexpected exception whose type does not provide `ex.code() -> std::error_code`:

```
// file:line: FAILED:
//   CHECK_THROWS_AS( throw, std::system_error )
// due to unexpected exception with message:
//   foo
CHECK_THROWS_WITH_CODE(throw std::runtime_error("foo"), std::errc::invalid_argument);
```

### DIFF
--- a/src/bsoncxx/test/catch.hh
+++ b/src/bsoncxx/test/catch.hh
@@ -27,9 +27,43 @@
 #include <catch2/catch_test_macros.hpp>  // TEST_CASE, SECTION, CHECK, etc.
 #include <catch2/catch_tostring.hpp>     // Catch::StringMaker
 
+#define THROWS_WITH_CODE_IMPL(_assertion, _expr, _code)                       \
+    if (1) {                                                                  \
+        try {                                                                 \
+            (void)(_expr);                                                    \
+            INFO("expected an exception to be thrown: " #_expr);              \
+            _assertion(false);                                                \
+        } catch (const Catch::TestFailureException&) {                        \
+            throw; /* Propagate Catch exceptions. */                          \
+        } catch (const Catch::TestSkipException&) {                           \
+            throw; /* Propagate Catch exceptions. */                          \
+        } catch (const std::system_error& ex) {                               \
+            using std::make_error_code;                                       \
+            (void)ex; /* Avoid unused variable warnings. */                   \
+            _assertion(ex.code() == (_code));                                 \
+        } catch (...) {                                                       \
+            BSONCXX_CONCAT(_assertion, _THROWS_AS)(throw, std::system_error); \
+        }                                                                     \
+    } else                                                                    \
+        ((void)0)
+
+#define CHECK_THROWS_WITH_CODE(_expr, _code) THROWS_WITH_CODE_IMPL(CHECK, _expr, _code)
+#define REQUIRE_THROWS_WITH_CODE(_expr, _code) THROWS_WITH_CODE_IMPL(REQUIRE, _expr, _code)
+
 namespace Catch {
 
-// Catch2 must be able to stringify documents, optionals, etc. if they're used in Catch2 macros.
+template <>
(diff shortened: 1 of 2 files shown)


src/bsoncxx/test/catch.hh
```diff
@@ -27,9 +27,43 @@
...
+        } catch (const Catch::TestFailureException&) {                        \
+            throw; /* Propagate Catch exceptions. */                          \
+        } catch (const Catch::TestSkipException&) {                           \
+            throw; /* Propagate Catch exceptions. */                          \
+        } catch (const std::system_error& ex) {                               \
+            using std::make_error_code;                                       \
+            (void)ex; /* Avoid unused variable warnings. */                   \
+            _assertion(ex.code() == (_code));                                 \
+        } catch (...) {                                                       \
+            BSONCXX_CONCAT(_assertion, _THROWS_AS)(throw, std::system_error); \
```
kevinAlbs, line 45:
```suggestion
            /* Unexpected. Use Catch _THROWS_AS to produce test failure */    \
            BSONCXX_CONCAT(_assertion, _THROWS_AS)(throw, std::system_error); \
```

Suggest clarifying the assert is always expected to fail.

kevinAlbs (approved):
Nice. LGTM

PR #1026, opened 2023-09-19
CXX-2753 Refactor directory structure to allow for multiple ABI namespaces

eramongodb:
## Description

Resolves CXX-2753 as part of CXX-1569. Verified by [this patch](https://spruce.mongodb.com/version/6509cf1ed1fe072ed8caa29a).

This PR significantly changes the directory structure of CXX Driver sources files for bsoncxx and mongocxx in preparation for the addition of C++ sources and headers corresponding to additional ABI namespaces (not just `v_noabi`). Those will be introduced in subsequent PRs.

The _only_ observable changes to contents of the install directory structure pre-PR vs. post-PR should be the legacy CMake package config files (to support multiple include directories) and the uninstall script (simply due to changes in ordering of generated commands). Any other changes should be considered bugs in this PR.

Both binary and source compatibility (should be 100%) is verified by [this patch](https://spruce.mongodb.com/version/6509e2282a60ed69ab8efe4c), which runs abi-complance-checker comparing the proposed PR changes relative to the latest state of the target branch (note: _not_ against the latest release of the CXX Driver, as is the case for the C Driver abi-compliance-check task).

## Directory Structure

In this PR description, "source directory" refers to files tracked by the Git repository (.cpp, .hpp, .in, etc.), "binary directory" refers to files that are generated by CMake during configuration and build, and "install directory" refers to files present in the install prefix after the CXX Driver is commanded to evaluate installation rules.

The pre-PR source directory structure for the bsoncxx and mongocxx libraries is as follows (only the minimal set of files necessary to explain the directory refactor are listed):

```
src
├── CMakeLists.txt
├── bsoncxx
│   ├── CMakeLists.txt
│   ├── cmake
│   │   ├── bsoncxx-config.cmake.in
│   │   └── ...
│   ├── config
│   │   ├── CMakeLists.txt
│   │   ├── config.hpp.in
│   │   ├── libbsoncxx.pc.in, libbsoncxx-static.pc.in
│   │   ├── private
│   │   │   ├── config.hh.in
│   │   │   └── prelude.hh, postlude.hh
│   │   └── prelude.hpp, postlude.hpp, ...
│   ├── private
│   │   └── libbson.hh, ...
│   ├── document
│   │   └── value.hpp, value.cpp, ...
│   ├── types.hpp, types.cpp, ...
│   ├── test
│   │   ├── CMakeLists.txt
│   │   └── ...
│   └── third_party
│       └── CMakeLists.txt
├── mongocxx
[...]

PR #1301, opened 2024-12-11
CXX-3199 Generate API documentation with generated headers

eramongodb:
## Summary

Addresses CXX-3199.

Refactors the `doxygen-current` and `doxygen-latest` targets to generate API documentation pages using a full set of headers, including those which are generated by CMake during configuration (e.g. `export.hpp`).

## Generated Headers

This PR effectively requires Doxygen be executed via the CMake targets `doxygen-current` (although it is really the ~`docs_source_directory`~ `doxygen-install-headers` target that is important). The ~`docs_source_directory`~ `doxygen-install-headers` target takes advantage of the symmetric directory layout implemented by https://github.com/mongodb/mongo-cxx-driver/pull/1026 and _raw-copies_ the `<lib>/include` and `<lib>/lib` directories into the same directory. This effectively merges the public and internal directory structure together into a single "install" directory which is parsed by Doxygen.

> [!IMPORTANT]
> The `build/doxygen` directory is now reserved for use by Doxygen so that this behavior is reliable regardless of the custom binary directory being used by the user.

> [!WARNING]
> This blindly copies the contents of the `include` and `lib` directories without any filters. If there are any stray `.hpp` header files in either directory, they will be blindly copied as well. This is not an issue for the `doxygen-latest` target which uses a temporary directory, but may cause some inconvenience for the `doxygen-current` target.

This permits Doxygen documentation comments to be embedded directly in the generated headers' input files, obviating the need to define "Group" pages to document non-existent header files. The headers can now be referenced directly (e.g. `@ref bsoncxx/v_noabi/bsoncxx/config/export.hpp`) and are listed under the Files page just like any other header.

To continue supporting the use of a custom Doxygen binary (to enforce an exact Doxygen version and corresponding generated output), the CMake [FindDoxygen](https://cmake.org/cmake/help/v3.15/module/FindDoxygen.html) module is used to permit customization via the `DOXYGEN_EXECUTABLE` variable (use as a result variable is deprecated in favor of `Doxygen::doxygen`, but this variable is still used for the internal call to `find_program()`).

The following command thus generates the current API doc pages for local preview:

```bash
# Custom binary if needed.
export DOXYGEN_BINARY=doxygen-1.12.0

# To avoid (re)downloading the C Driver.
export CMAKE_PREFIX_PATH=/path/to/c-driver/install

# Local repository.
cmake -S . -B <binary-dir>

# Generate API docs under build/docs/api/current.
cmake --build <binary-dir> --target doxygen-current
```

## Latest API Docs

The `generate-*` Perl scripts are completely removed in favor of a Bash script to ~~remove~~ reduce our toolchain dependency on Perl (still used for `deploy-to-ghpages.pl`). Following https://github.com/mongodb/mongo-cxx-driver/pull/1169, the `generate-latest-apidocs.pl` effectively lost its purpose as the "generate the latest instead of generating all" script, since we no longer regenerate old doc pages. The `generate-apidocs-from-tag.pl` was only being used to implement the behavior of `generate-latest-apidocs.pl` and no longer used for any other purpose.

To avoid continuing to maintain Perl, both scripts are replaced with a straightforward Bash script that preserves only the minimum required behaviors from the old scripts. A temporary directory is used to clone a clean copy of the repository and checkout the relevant release tag whose source files are to be documented. `sed` is used to directly patch the Doxyfile with relevant version information and output path. CMake is called to generate the headers (via `--target docs_source_directory`) before running Doxygen manually to avoid confusing the current CMake project (which invokes `--target doxygen-latest`) and the CMake project in the temporary directory being used to generate the API documentation.

```bash
# For generate-latest-apidocs.sh.
export DOXYGEN_BINARY=doxygen-1.12.0

# To avoid (re)downloading the C Driver.
export CMAKE_PREFIX_PATH=/path/to/c-driver/install

# Local repository.
cmake -S . -B build

# Generate API docs under build/docs/api/<latest>.
cmake --build build --target doxygen-latest
```


etc/generate-apidocs-from-tag.pl
```diff
@@ -1,175 +0,0 @@
...
-use utf8;
-use open qw/:std :utf8/;
-
-use Cwd qw/getcwd realpath/;
-use File::Temp qw/tempdir/;
-use List::Util qw/first/;
-
-# The required Doxygen version.
-# The generated results are sensitive to the release version.
-our $doxygen_version_required = "1.12.0";
```
mdbmes, line 14: I noticed a dangling reference to this variable in the docs, the "IMPORTANT" note in `releasing.md` should refer to the new shell variable instead.

etc/generate-latest-apidocs.sh
```diff
@@ -0,0 +1,64 @@
...
+apidocpath="${working_dir:?}/build/docs/api/mongocxx-${LATEST_VERSION:?}"
+
+# Use a temporary directory for the following operations.
+tmpdir="$(mktemp -d)"
+trap "rm -rf \"${tmpdir}\"" EXIT
+
+mkdir -p "${apidocpath:?}"
+
+# Use a clean copy of the repository.
+git clone -q -c advice.detachedHead=false -b "cxx-3199" . "${tmpdir}"
```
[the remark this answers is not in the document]
mdbmes: Oh, good to know. Thanks!

mdbmes (approved):

CMakeLists.txt
```diff
@@ -283,20 +283,41 @@ add_custom_target(hugo-deploy
...
 
+# Set `DOXYGEN_EXECUTABLE` to specify the full path to a custom Doxygen binary.
+find_package(Doxygen 1.12.0 EXACT QUIET)
+
 add_custom_target(doxygen-current
-    DEPENDS docs_dir_current
+    DEPENDS Doxygen::doxygen docs_source_directory docs_target_directory
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    COMMAND doxygen ${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile
+    COMMAND Doxygen::doxygen
```
kevinAlbs, line 314: Rather than using `find_package`, consider using a bash script like `generate-latest-apidocs.sh` to find and run Doxygen.

It may help make configuration more consistent. Both scripts can check `DOXYGEN_BINARY`. It may improve the error for mismatched Doxygen versions. When testing with Doxygen 1.11.0, I got this error from the `doxygen-current` target:

```
ninja: error: 'Doxygen::doxygen', needed by 'CMakeFiles/doxygen-current', missing and no known rule to make it
```

But the `doxygen-latest` target error includes the version:
```
doxygen version 1.11.0 does not equal 1.12.0
```

etc/generate-latest-apidocs.sh
```diff
@@ -0,0 +1,64 @@
...
+apidocpath="${working_dir:?}/build/docs/api/mongocxx-${LATEST_VERSION:?}"
+
+# Use a temporary directory for the following operations.
+tmpdir="$(mktemp -d)"
+trap "rm -rf \"${tmpdir}\"" EXIT
+
+mkdir -p "${apidocpath:?}"
+
+# Use a clean copy of the repository.
+git clone -q -c advice.detachedHead=false -b "v${LATEST_VERSION}" . "${tmpdir}"
```
kevinAlbs, line 45:
```suggestion
git clone -q -c advice.detachedHead=false -b "r${LATEST_VERSION}" . "${tmpdir}"
```

To match tag patterns `rX.Y.Z`.

etc/generate-latest-apidocs.sh
```diff
@@ -0,0 +1,64 @@
...
+
+cd "${tmpdir:?}"
+
+# Update the Doxyfile configuration file:
+#  - set OUTPUT_DIRECTORY to `build/docs/api/mongocxx-<version>`.
+#  - set PROJECT_NUMBER to `<version>`.
+sed -i \
```
kevinAlbs, line 52: Request adding special case for BSD sed so this script can be run on macOS:

```bash
sed_inplace_args=("-i")
if [ "$(uname -s | tr '[:upper:]' '[:lower:]')" == "darwin" ]; then
  # BSD sed expects suffix argument.
  sed_inplace_args+=("")
fi
# Update the Doxyfile configuration file:
#  - set OUTPUT_DIRECTORY to `build/docs/api/mongocxx-<version>`.
#  - set PROJECT_NUMBER to `<version>`.
sed "${sed_inplace_args[@]}" \
  -e "s|^OUTPUT_DIRECTORY\s*=\s*.*$|OUTPUT_DIRECTORY = ${apidocpath:?}|g" \
  -e "s|^PROJECT_NUMBER\s*=\s*.*$|PROJECT_NUMBER = ${LATEST_VERSION:?}|g" \
  Doxyfile
```

Fixes an observed error running locally: `sed: -e: No such file or directory`.

kevinAlbs (commented):
Apologies for the delayed review. Migrating perl scripts to bash is much appreciated.

etc/generate-latest-apidocs.sh
vector-of-bool: It looks like you can unconditionally add the suffix argument `-i ""` or `--in-place=`

vector-of-bool (approved):
LGTM

etc/generate-latest-apidocs.sh
eramongodb: Thank you. Fixed.

CMakeLists.txt
eramongodb: Extended the `generate-latest-apidoc.sh` script to support a `--current` flag. This flag is now used by the `doxygen-current` target. Behavior is the same as without `--current` except that the _current_ repository's `build` directory is used to configure, generate, and run Doxygen without additional patching/modification of the Doxyfile.

etc/generate-latest-apidocs.sh
eramongodb: Added a zero-length suffix/extension for conformance with both GNU and BSD `sed`.

etc/generate-latest-apidocs.sh
```diff
@@ -0,0 +1,86 @@
...
+if [[ ! -d build ]]; then
+  echo "missing build directory: this script must be run from the project root directory" 1>&2
+  exit 1
+fi
+
+# Validate Doxygen version.
+doxygen_version="$("${DOXYGEN_BINARY:?}" -v | perl -lne 'print $1 if m/^(\d+\.\d+\.\d+).*$/')"
+if [[ "${doxygen_version:-}" != "${DOXYGEN_VERSION_REQUIRED:?}" ]]; then
+  echo "${DOXYGEN_BINARY} version ${doxygen_version:-"unknown"} does not equal ${DOXYGEN_VERSION_REQUIRED:?}" 1>&2
+  exit 1
```
kevinAlbs, line 48: Requiring a version of Doxygen appears to cause failures in the [packaging tasks](https://spruce.mongodb.com/task/mongo_cxx_driver_packaging_packaging_debian_patch_5079fccebbae94bf5d3c53444acebaaed5ff8996_6762ed585a0e580007f6b323_24_12_18_15_42_21/logs?execution=0):

> doxygen version 1.9.8 does not equal 1.12.0

The Debian package appears to include docs built with the version of Doxygen available from apt. Consider reducing this to a warning for the `doxygen-current` target.

kevinAlbs (approved):
LGTM with a fix to the packaging tasks.

etc/generate-latest-apidocs.sh
eramongodb: @rcsanchez97 The [override_dh_auto_build](https://github.com/mongodb/mongo-cxx-driver/blob/debian/unstable/debian/rules#L30C1-L30C23) rule in the debian/unstable branch (along with other related scripts) may need to be updated to set `DOXYGEN_ANY_VERSION=1` when building the `doxygen-current` target. Alternatively they may be updated to avoid building API docs if that is permitted.

rcsanchez97: @eramongodb Thanks for the heads up. The solution was fairly trivial, but testing it was a bit of a different story (owing to the Debian packaging being on a separate branch). I figured out a way to inject the proposed change into a patch build and confirmed that it works ([here](https://spruce.mongodb.com/task/mongo_cxx_driver_packaging_packaging_debian_patch_f1ad6b63d33672b6875e096765ee782046b515a8_676afb9a00bb6d00077bef4f_24_12_24_18_21_15/logs?execution=0)).

After pushing the tested change (on the `debian/unstable` branch), I re-ran the task on the waterfall and it worked ([here](https://evergreen.mongodb.com/task/mongo_cxx_driver_packaging_packaging_debian_f1ad6b63d33672b6875e096765ee782046b515a8_24_12_20_18_05_45)).

PR #1302, opened 2024-12-12
CXX-3195 add 404 page with automatic redirection

eramongodb:
Resolves CXX-3195. Precursor changes to CXX-2745 to gracefully handle API doc pages which are no longer present in the latest API documentation, such as in https://github.com/mongodb/mongo-cxx-driver/pull/1301.

Implemented according to [GitHub Pages Docs](https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site) using a Hugo-generated custom 404 homepage as a template which was stripped down to its current state with site-wide contents preserved (e.g. content security policy, stylesheets, site visit tracker, etc.).

404 and redirection behavior has not yet been tested with GitHub Pages deployment. Behavior will be tested and validated after deployment with followup PRs if necessary.

GitHub Pages supports using a custom 404 error page located at `/404.html` relative to the root directory. This PR proposes a 404 page which waits 5 seconds before redirecting users (with page history preservation, unlike `/api/current` page redirects) to _either_ the current API docs homepage for URLs under `/api`, or to the website homepage for all other URLs. It is possible this behavior does not work as intended with GitHub Pages depending on how it implements the 404 error page (does it "redirect" to `/404.html`, thus `window.location` is always the 404 page, not the original link? does it "substitute" the contents of `/404.html` using the original link?). Regardless, the proposed 404 error page is expected to automatically redirect the user instead of the current status quo of GitHub Pages' default 404 page with no redirection.

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests">
+    <link rel="shortcut icon" href="https://mongocxx.org/img/favicon.png">
+
+    <title>Page Not Found</title>
+
+    <link rel="stylesheet" href="https://mongocxx.org/lib/bootstrap.css" type="text/css" />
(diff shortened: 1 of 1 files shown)


rcsanchez97 (approved):

eramongodb:
Confirmed 404 automatic redirection works as intended with the deployed pages.

PR #1304, opened 2024-12-16
Migrate ABI stability tasks to rhel95 distro (Clang 18)

eramongodb:
EVG task failures due to a distro image misconfiguration revealed that Clang 18 is available on the newly-available rhel95 distro. Addresses task failures by migrating to the rhel95 distro with Clang 18 for improved C++23 compilation coverage.

Additionally added some missing early return handlers to correctly report failures during setup.

### DIFF
--- a/.evergreen/scripts/abi-stability-setup.sh
+++ b/.evergreen/scripts/abi-stability-setup.sh
@@ -109,7 +109,7 @@ git -C mongo-cxx-driver reset --hard "${base:?}"
 
 # Install old (base) to install/old.
 echo "Building old libraries..."
-{
+(
   "${cmake_binary:?}" \
     -S mongo-cxx-driver \
     -B build/old \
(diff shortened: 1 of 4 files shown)


.evergreen/scripts/abi-stability-setup.sh
```diff
@@ -115,9 +115,9 @@ echo "Building old libraries..."
     -B build/old \
     -DCMAKE_INSTALL_PREFIX=install/old \
     -DBUILD_VERSION="${old_ver:?}-base" \
-    "${configure_flags[@]:?}"
-  "${cmake_binary:?}" --build build/old
-  "${cmake_binary:?}" --install build/old
+    "${configure_flags[@]:?}" || exit
```
kevinAlbs, line 118: I expect the updated blocks may need to be updated to `(` and `)` instead of `{` and `}`. The `exit` appears to exit the shell without running the `cat old.log 1>&2` command.

eramongodb: Good catch. Applied as suggested.

PR #1306, opened 2024-12-30
CXX-3198 modernize ClangFormat configuration options

eramongodb:
## Summary

Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1299.

This PR proposes modifying several format options in the `.clang-format` configuration file to modernize and (hopefully) better fit our development practices/preferences. See [ClangFormatStyleOptions](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html) (for 19.1.0, the version we currently use).

This PR is organized such that each option update is accompanied by corresponding changes to source code for comparison/reference. Particularly notable changes are to `ColumnLimit`, `AlignAfterOpenBracket`, and `QualifierAlignment`.

## ColumnLimit: 100 -> 120

For consistency with the [C Driver config](https://github.com/mongodb/mongo-c-driver/blob/c81fed7572ce60a51b49155ce9f3b269b4c3698e/.clang-format#L89) and for arguably greater benefit to our C++ codebase by reducing instances of awkward line-wrapping (particularly for templates).

## AlignAfterOpenBracket: Align -> AlwaysBreak

One of the major proposed changes in this PR. This is primarily motivated by the desire to remove (imo) excessive indentation of parameters and arguments, e.g.:

```cpp
// Before:
bsoncxx::document::value owner =
    bsoncxx::builder::basic::make_document(kvp("a", std::int32_t{1}),  // "$numberInt": "1"
                                           kvp("b", std::int64_t{2}),  // "$numberLong": "2"
                                           kvp("c", binary)  // "$binary": { "$base64": "dGhyZWU=", "subType": 00 }
    );
bsoncxx::document::view doc = owner.view();

// After:
bsoncxx::document::value owner = bsoncxx::builder::basic::make_document(
    kvp("a", std::int32_t{1}),  // "$numberInt": "1"
    kvp("b", std::int64_t{2}),  // "$numberLong": "2"
    kvp("c", binary)            // "$binary": { "$base64": "dGhyZWU=", "subType": 00 }
);
bsoncxx::document::view doc = owner.view();
```

```cpp
// Before:
constexpr friend bsoncxx::detail::strong_ordering tag_invoke(bsoncxx::detail::compare_three_way cmp,
                                                             basic_string_view left,
                                                             basic_string_view right) noexcept {
    return cmp(left.compare(right), 0);
}

// After:
constexpr friend bsoncxx::detail::strong_ordering tag_invoke(
    bsoncxx::detail::compare_three_way cmp, basic_string_view left, basic_string_view right) noexcept {
    return cmp(left.compare(right), 0);
}
```

There is a concern with this setting that a parameter list may be confused with the function body given the same indentation level, e.g.:

```cpp
bsoncxx::v_noabi::document::value database::run_command(
    client_session const& session,
    bsoncxx::v_noabi::document::view_or_value command) {
    return _run_command(&session, command);
}
```

Although `BlockIndent` rather than `AlwaysBreak` would be preferable to avoid this issue (by forcing the closing brace to be on its own line at the base indentation level), this setting does not behave consistently due to only applying to braced initializer lists and parenthesis per the associated note in documentation (e.g. does not apply to template parameters). If sufficiently confusing, we can consider adjusting `ContinuationIndentWidth` to either `2` (half-indentation) or `8` (double-indentation):

```cpp
// ContinuationIndentWidth: 2
bsoncxx::v_noabi::document::value database::run_command(
  client_session const& session,
  bsoncxx::v_noabi::document::view_or_value command) {
    return _run_command(&session, command);
}

// ContinuationIndentWidth: 8
bsoncxx::v_noabi::document::value database::run_command(
        client_session const& session,
        bsoncxx::v_noabi::document::view_or_value command) {
    return _run_command(&session, command);
}
```

## AllowAllParametersOfDeclarationOnNextLine: true -> false

This change ensures whenever a parameter/argument list is broken after the opening brace, each parameter/argument is given its own line. This is to encourage better wrapping and slightly reduce the parameter list vs. function body confusion described above.

```cpp
// Before:
constexpr friend bsoncxx::detail::strong_ordering tag_invoke(bsoncxx::detail::compare_three_way cmp,
                                                             basic_string_view left,
                                                             basic_string_view right) noexcept {
    return cmp(left.compare(right), 0);
}

// After:
constexpr friend bsoncxx::detail::strong_ordering
tag_invoke(bsoncxx::detail::compare_three_way cmp, basic_string_view left, basic_string_view right) noexcept {
    return cmp(left.compare(right), 0);
}
```

## QualifierAlignment: Left -> Right

This change "reverts" the fixes applied in https://github.com/mongodb/mongo-cxx-driver/pull/1299 by switching from "west `const`" style (`const T&`) to "east `const`" style (`T const&`). This is motivated by better consistency between syntax and semantics across C++ code, e.g.:

```cpp
using T = int*;

using A = T const; // int* const (OK: syntax matches semantics)
using B = const T; // int* const (!?: syntax does not match semantics)
```

Further motivations for this style is discussed in greater detail in [C++ Templates: The Complete Guide, 2nd Edition](http://www.tmplbook.com/), section xxxi: "Some Remarks About Programming Style". Note we've already let several instances of east `const` [leak into new code](https://github.com/mongodb/mongo-cxx-driver/pull/1299/commits/88c6741f79e8a7ee7cedad476430c59d7e6b9b80), which were (temporarly) reverted by https://github.com/mongodb/mongo-cxx-driver/pull/1299.

## SpacesBeforeTrailingComments: 2 -> 1

This change reduces the spaces count prior to `//` in closing namespace braces and similar cases (it is unclear why it was set to 2 to begin with).

```cpp
#if !defined(MACRO_GUARD)
#define MACRO_GUARD

namespace example {

...

} // namespace example
 ^

#endif // !defined(MACRO_GUARD)
      ^
```

## ShortNamespaceLines: 1 -> 0

This ensures the closing namespace brace is always accompanied by a trailing comment denoting the namespace being closed.

## KeepEmptyLines.AtStartOfFile: true -> false

This trims extra newlines at the beginning of a file.


kevinAlbs (approved):
LGTM. I support the ColumnLimit increase.

eramongodb:
@adriandole Due to GitHub not supporting fast-forward merges of PRs, the commit hash of the formatting commit can only be known _after_ the PR has been merged. A followup PR is therefore necessary to add the `.git-blame-ignore-revs` file (e.g. as done for the C Driver with https://github.com/mongodb/mongo-c-driver/pull/1546). I can add refs for both this and the prior PR after this PR is merged.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/types/bson_value/view.cpp
mdbmes, line 1: This "view::view(const view& rhs)" on line 55 didn't change. Might be another instance of the same confusion issue that commit 9a1f4893f8a28ab1b9414f879d8974b02c232c2e fixed one instance of.

mdbmes (approved):
I can't speak to the aesthetics of this change. It's subjective. I would make different choices, but I'm not aware of any stakeholder requirements that would support or refute these changes.

The 'const' movement seems especially risky; there's one fixup in here already, and I may have found another inconsistency in the interpretation of 'view' during review. The overall size of this diff is daunting: 789 files, 58kloc. I did my best to review it thoroughly for correctness.

I noticed some existing problems during the review: an unnecessary strlen and min, and more worrying a flawed RNG in the examples.

Using rand() like this even in a didactic context is troubling. (Maximum available entropy is going to be less than the key size even if rand() happens to have a secure entropy source) Additionally, change 0d649bf62c9 introduced a bug. The "% UINT8_MAX" pattern is not correct unless it's important to avoid 0xFF bytes for some reason I'm unaware of.

PR #1307, opened 2024-12-30
CXX-3198 support `clang-format-all.sh` on macOS

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-cxx-driver/pull/1299. Verified with this [patch build](https://spruce.mongodb.com/version/6772fc401713a80007c98ce4).

Use `grep` instead of `-regextype` to fix observed error running `etc/clang-format-all.sh` on macOS:
```
find: -regextype: unknown primary or operator
```

Also allow overriding the used `clang-format`. This is intended to ease my current set-up and can be reverted if desired. I installed LLVM 1.19 with `brew install llvm`, and it is not on a system path to avoid interfering with AppleClang.

### DIFF
--- a/etc/clang-format-all.sh
+++ b/etc/clang-format-all.sh
@@ -12,7 +12,8 @@
 set -o errexit
 set -o pipefail
 
-clang-format --version
+CLANG_FORMAT="${CLANG_FORMAT:-clang-format}"
+"$CLANG_FORMAT" --version
 
 source_dirs=(
   src
@@ -25,7 +26,7 @@ mapfile -t source_files < <(
 )
 
 if [[ -n "${DRYRUN:-}" ]]; then
-  clang-format --dry-run -Werror "${source_files[@]:?}"
+  "$CLANG_FORMAT" --dry-run -Werror "${source_files[@]:?}"
 else
-  clang-format --verbose -i "${source_files[@]:?}"
+  "$CLANG_FORMAT" --verbose -i "${source_files[@]:?}"
 fi
--- a/etc/clang-format-all.sh
+++ b/etc/clang-format-all.sh
@@ -12,8 +12,8 @@
 set -o errexit
 set -o pipefail
 
-CLANG_FORMAT="${CLANG_FORMAT:-clang-format}"
-"$CLANG_FORMAT" --version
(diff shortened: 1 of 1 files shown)


etc/clang-format-all.sh
```diff
@@ -12,7 +12,8 @@
 set -o errexit
 set -o pipefail
 
-clang-format --version
+CLANG_FORMAT="${CLANG_FORMAT:-clang-format}"
+"$CLANG_FORMAT" --version
```
eramongodb, line 16:
```suggestion
```suggestion
clang_format_binary="${CLANG_FORMAT_BINARY:-clang-format}"
"${clang_format_binary:?}" --version
```
For consistency with `DOXYGEN_BINARY`, `POETRY_PYTHON_BINARY` (C Driver), etc. + recommend using separate variables for input env var (UPPER_CASE) and local variable (snake_case) + use `${clang_format_binary:?}` to ensure not-null, not-empty when used below (e.g. catch typoes, etc.).

etc/clang-format-all.sh
```diff
@@ -21,11 +22,11 @@ source_dirs=(
 )
 
 mapfile -t source_files < <(
-  find "${source_dirs[@]:?}" -regextype posix-egrep -regex '.*\.(hpp|hh|cpp)'
+  find "${source_dirs[@]:?}" | grep -E '.*\.(hpp|hh|cpp)$'
```
eramongodb, line 25: Although I would prefer to use PCRE (`-P`) for consistency, IIRC it is not fully supported across all distros/versions, so I am fine with using `-E` here for now.

eramongodb (approved):
Minor suggestions; otherwise, LGTM.

etc/clang-format-all.sh
kevinAlbs: Unfortunately `-P` does not appear supported on macOS `grep`:

```
% grep -P
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
```

PR #1309, opened 2025-01-02
CXX-3201 remove Doxygen workarounds for using-declarations

eramongodb:
Resolves CXX-3201. An exciting followup to https://github.com/mongodb/mongo-cxx-driver/pull/1182 and https://github.com/mongodb/mongo-cxx-driver/pull/1184.

Following the release of [Doxygen 13.0.1](https://www.doxygen.nl/manual/changelog.html), we are finally able to remove the `*_PRIVATE_DOXYGEN_PREPROCESSOR` boilerplate to workaround the absence of Doxygen documentation pages for root namespace redeclarations.

> [!NOTE]
> The once-more absence of references to the original entity by redeclarations will be addressed separately by CXX-3202 to properly support disambiguating ABI namespaces.

The current class list page generated with 12.0.0 look as follows (using the workarounds):

image

The class list page generated with 13.0.1 look as follows:

image

Note that the class list once again resembles its prior state (e.g. for the 3.9.0 release) before the workarounds were implemented by https://github.com/mongodb/mongo-cxx-driver/pull/1182, where the documentation of the original entity is cloned for the redeclaration as-if it were its own distinct entity:

image

The combination of `SHOW_HEADERFILE` and `STRIP_FROM_INCLUDE_PATH` permits the class page to retain a hint regarding the true origins of a redeclaration in its `#include` directive even without CXX-3202:

image

Note that the class page is (once more) complete with documentation that clones the documentation of the original entity, unlike with the workarounds for Doxygen 1.12.0:

image

More importantly, unlike with old Doxygen documentation (e.g. for the 3.9.0 release), using-declarations of class members are now properly documented by all pages (both the class page of the original entity that using-declares members as well as the class page for the redeclaration). This notably affects the class page of `bsoncxx::v_noabi::array::element` the most, but also applies to various class and namespace pages.

This PR also removes documentation for `std` namespace entity specializations, moving them into a note in the corresponding enumeration documentation instead, to avoid the very-lengthy class template specializations from messing with the width of the class list page's first column:

image

The note looks as follows:

image


eramongodb:
Remaining uses of the `*_PRIVATE_DOXYGEN_PREPROCESSOR` macro are limited to the following:

- Forbidding inclusion of `doc.hpp` headers (for documentation only).
- Documentation of `config.hpp` macros (regardless which are conditionally defined).
- Documentation of `stdx` polyfills (regardless of the particular polyfill library configuration).

eramongodb:
After further inspection of generated pages, closed CXX-3202 as "Won't Do": the `SHOW_HEADERFILE` option embeds an `#include` directive at the top of the class page indicating the header file which provides the documented class. This header file path is sufficient to indicate the associate the ABI namespace (e.g. `<bsoncxx/v_noabi/...>`). All other symbols (functions, variables, typedefs, etc.) are either fully qualified (in their detailed description, e.g. `const std::error_category& bsoncxx::v_noabi::error_category()`) and/or have an associated ABI-namespaced type in its parameter list or return type which indicates the ABI namespace for which it is declared to be used with (e.g. `std::string bsoncxx::v_noabi::to_string(v_noabi::type rhs)`). Therefore, it is not necessary to manually embed references to the original entity in documented interfaces.

kevinAlbs (approved):
The efforts to fix Doxygen upstream and remove these workarounds are very much appreciated. LGTM!

PR #1312, opened 2025-01-06
Migrate most windows-64-vs2015 tasks to windows-vsCurrent distro

eramongodb:
Motivated by an attempt to resolve ongoing task failures on the (deprecated) `windows-64-vs2015` distro due to the [absence of a compatible Python version](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_compile_only_matrix_compile_only_vs2015_x64_debug_shared_2f53a561afa7ff06c12b7c6ec7a7f3bd6a5e135b_25_01_06_16_30_46/0/task?bookmarks=0,87&shareLine=44) after https://github.com/mongodb-labs/drivers-evergreen-tools/pull/567 dropped support for Python 3.8 (EOL).

This PR does _not_ succeed in resolving the original vs2015 distro issue, as unfortunately uv appears to have an [unresolved regression since 0.5.9](/astral-sh/uv/issues/10231) that is [preventing its use](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_compile_only_matrix_compile_only_vs2015_x64_debug_shared_patch_2f53a561afa7ff06c12b7c6ec7a7f3bd6a5e135b_677c43d65567600007fe82f9_25_01_06_20_57_59/0/task?bookmarks=0,209&shareLine=187) as a workaround, even after upgrading to 0.5.14.

Instead, the affected tasks are (mostly) migrated to the newer and recommended `windows-vsCurrent` distro, which provides VS 2017 and newer. The compile-only matrix still contains tasks with VS2015 compilation coverage (currently failing) and is pending updates to either Astral UV (fixing the reported segfault on Windows) or [DEVPROD-13875](https://jira.mongodb.org/browse/DEVPROD-13875), whichever comes first, due to refactoring the `compile.sh` script to `etc/calc_release_version.py` via uv rather than DET's `find-python3.sh`.

PR #1319, opened 2025-01-14
Manually invoke python3 on windows-64-vs2015

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1312.

Due to [DEVPROD-13875](https://jira.mongodb.org/browse/DEVPROD-13875), this PR proposes we resort to using a manual workaround for the C++ Driver instead until we drop VS 2015 coverage. The Python 3.8 binary which was being used prior to https://github.com/mongodb-labs/drivers-evergreen-tools/pull/567 is now manually invoked only on the `windows-64-vs2015` distro which requires it.

### DIFF
--- a/.evergreen/scripts/compile.sh
+++ b/.evergreen/scripts/compile.sh
@@ -83,9 +83,9 @@ esac
 
 # Create a VERSION_CURRENT file in the build directory to include in the dist tarball.
 if [[ "${distro_id:?}" == windows-64-vs2015* ]]; then
-  # DEVPROD-13875: Python 3.8+ is not available on this distro.
-  # Astral UV is not able to run properly on this OS+platform either.
-  # Manually fallback to Python 3.8 instead.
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/compile.sh
```diff
@@ -82,7 +82,14 @@ darwin* | linux*)
 esac
 
 # Create a VERSION_CURRENT file in the build directory to include in the dist tarball.
-PATH="${UV_INSTALL_DIR:?}:${PATH:-}" uv run --frozen python ./etc/calc_release_version.py >./build/VERSION_CURRENT
+if [[ "${distro_id:?}" == windows-64-vs2015* ]]; then
+  # DEVPROD-13875: Python 3.8+ is not available on this distro.
```
kevinAlbs, line 86:
```suggestion
  # DEVPROD-13875: Python 3.9+ is not available on this distro.
```

kevinAlbs (approved):
LGTM with a typo fix.

PR #1323, opened 2025-01-30
CXX-2745 Add advanced CMake options to toggle root namespace redeclarations

eramongodb:
An intermediate PR related to CXX-2745 following https://github.com/mongodb/mongo-cxx-driver/pull/1318.

Add advanced, for-development-only CMake options `BSONCXX_API_OVERRIDE_DEFAULT_ABI` and `MONGOCXX_API_OVERRIDE_DEFAULT_ABI` to permit toggling the ABI namespace that is used by root namespace redeclarations.

> [!IMPORTANT]
> These options are NOT intended for public use. This is strictly for development purposes. Enabling these options will emit CMake configuration warnings:
> ```
> CMake Warning at src/bsoncxx/CMakeLists.txt:29 (message):
>   BSONCXX_API_OVERRIDE_DEFAULT_ABI is an experimental feature
> CMake Warning at src/mongocxx/CMakeLists.txt:29 (message):
>   MONGOCXX_API_OVERRIDE_DEFAULT_ABI is an experimental feature
> ```
> The options are also marked as "advanced" so they do not show up in CMake GUIs by default.

When enabled, relevant root namespace redeclarations will use `v1` interfaces instead of `v_noabi` interfaces, e.g.:

```cpp
#if defined(BSONCXX_API_OVERRIDE_DEFAULT_ABI)
using v1::document::view; // New v1 equivalent or successor API.
#else
using v_noabi::document::view; // Current v_noabi API.
#endif
```

This permits manually verifying compilation and runtime compatibility of `v1` interfaces via root namespace redeclarations against existing code (e.g. `bsoncxx::document::view (bsoncxx::v1::document::view)`). This toggle will continue to be used throughout upcoming PRs to both document and control the set of v1 interfaces which are "ready for use". These options may be toggled individually for each library (although mongocxx should be only be enabled when bsoncxx is also enabled).

At the moment, this PR only applies this option to `stdx::string_view` and `stdx::optional<T>`. These are the only v1 interfaces which are currently "ready for use". This can be validated right now by building and running the C++ Driver with the `BSONCXX_API_OVERRIDE_DEFAULT_ABI` option enabled.

As more `v1` interfaces are introduced and `v_noabi` interfaces updated to support forward-compatibility with `v1` interfaces, the application of this configuration option will be incrementally extended to allow for reuse of existing (test) code for API and behavioral testing of `v1` interfaces. This macro may also be used to denote v_noabi interfaces which do _not_ have direct v1 counterparts (no corresponding `v1` redeclarations), e.g.:

```cpp
#if defined(BSONCXX_API_OVERRIDE_DEFAULT_ABI)
// No v1 API equivalent or successor.
#else
using v_noabi::builder::stream::document; // Current v_noabi API.
#endif
```

Conversely, new v1 interfaces, or those which substantially differ from their v_noabi equivalents, may introduce new redeclarations without v_noabi counterparts (reverse of the above).

Once v1 interfaces are sufficiently ready for use, this pattern will also permit users to opt-into using `v1` interfaces early and migrate from `v_noabi` to `v1`. This may be done either with a different CMake option designed to target user configuration rather than internal development, or by converting this CMake option from advanced, for-development-only into a user-facing CMake option.


eramongodb:
PR description is fixed/updated.

eramongodb:
Apologies, identified some stray local dev changes which were not properly refactored prior to PR creation. Give me a minute to address them.

eramongodb:
Ready for review.

src/mongocxx/lib/mongocxx/private/mongoc.hh
```diff
@@ -64,20 +431,20 @@ using log_set_handler_cdecl_t = void(MONGOCXX_ABI_CDECL*)(log_func_cdecl_t log_f
 
 extern MONGOCXX_ABI_EXPORT_TESTING mongocxx::test_util::mock<log_set_handler_cdecl_t>& log_set_handler;
 
-#if defined(__GNUC__) && (__GNUC__ >= 6) && !defined(__clang__)
-#pragma GCC diagnostic pop
-#endif
+SILENCE_IGNORED_ATTRIBUTES_END();
 
-#else
+#else // defined(MONGOCXX_TESTING) ^|v !defined(MONGOCXX_TESTING)
```
kevinAlbs, line 436: Comment fragment? I am not sure what `^|v` refers to.

src/bsoncxx/lib/CMakeLists.txt
```diff
@@ -81,19 +85,20 @@ endif()
...
+    bsoncxx/private/config/postlude.hh
+    bsoncxx/private/config/prelude.hh
+    bsoncxx/private/helpers.hh
+    bsoncxx/private/itoa.hh
+    bsoncxx/private/libbson.hh
+    bsoncxx/private/make_unique.hh
+    bsoncxx/private/stack.hh
+    bsoncxx/private/suppress_deprecation_warnings.hh
+    bsoncxx/test_util/export_for_testing.hh
     bsoncxx/v_noabi/bsoncxx/types/bson_value/private/value.hh
```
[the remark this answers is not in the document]
eramongodb: Good catch. I'd overlooked `private/` headers in subdirectories.

Headers were relocated/renamed accordingly:

- Private headers (not corresponding to any v_noabi component) are moved under `private/`.
- Internal headers (corresponding to an existing v_noabi component) are moved into the same directory as their implementation file (.cpp) and renamed for consistency (`tls.hh` <- `tls.cpp`).

src/mongocxx/lib/mongocxx/private/mongoc.hh
eramongodb: This is indicating the `#else` is changing the "scope" of the preprocessor condition from `defined(MONGOCXX_TESTING)` above (`^`) to `!defined(MONGOCXX_TESTING)` below (`v`) such that trailing comments for both `#if` and `#endif` have a consistent counterpart (greppable) without being misleading about the condition (negated). Combined, `^|v` is basically "☝️👇" ("see: above and below"). I can remove this style of comment if it is confusing.

I only just realized from your question that this style of comment is new. The only existing cases of this pattern are in the recently-committed `prelude.cpp` v1 implementation files.

https://github.com/mongodb/mongo-cxx-driver/blob/5d1930fd0ad41cfc0a04a33498a96ef04888f515/src/bsoncxx/lib/bsoncxx/v1/detail/postlude.cpp#L17

I do not remember where I picked up this pattern from. They may have used a different syntax than `^|v`, which I vaguely recall being an abbreviation of what I originally saw, e.g.:
```
#else // ^^^ cond ^^^ | vvv !cond vvv
```

kevinAlbs: Ah thank you for the explanation. I slightly prefer the more verbose `^^^ cond ^^^ | vvv !cond vvv` since I expect I would recognize it faster. But I am OK with the current style.

kevinAlbs (approved):

src/bsoncxx/lib/CMakeLists.txt
```diff
@@ -54,8 +58,8 @@ set(bsoncxx_sources "${bsoncxx_sources}" PARENT_SCOPE)
 # Generate private headers.
 if(1)
     configure_file(
-        bsoncxx/v_noabi/bsoncxx/config/private/config.hh.in
-        bsoncxx/v_noabi/bsoncxx/config/private/config.hh
+        bsoncxx/private/config/config.hh.in
+        bsoncxx/private/config/config.hh
     )
 endif()
```
vector-of-bool, line 64: Aside: What's up with the `if(1)` blocks all over the CMake code?

vector-of-bool (approved):
LGTM. I appreciate the progressive flattening of the deep directory structure.

src/bsoncxx/lib/CMakeLists.txt
eramongodb: It is a primitive workaround to make up for the lack of "scope blocks" in CMake without needing to resort to `function()` (overkill for these use cases) or comment-based begin+end delimiters. It is comparable to function-local block statements in C/C++ without any condition or loop.

That being said, this particular `if(1)` isn't doing anything useful anymore, so I've removed it in response.

Note: CMake 3.25 added [`block()`](https://cmake.org/cmake/help/v3.25/command/block.html) which would be a much better and more appropriate command to use than `if(1)` once the minimum CMake required version is raised accordingly.

.clang-format
```diff
@@ -156,13 +156,17 @@ IncludeCategories:
...
     Priority: 26
-  - Regex: 'mongocxx/v1/.*\.hh' # v1 private headers
+  - Regex: 'mongocxx/v1/.*\.hh' # v1 internal headers
     Priority: 27
-  - Regex: 'bsoncxx/config(/private)?/prelude\.(hpp|hh)' # v_noabi preludes
+  - Regex: 'bsoncxx/config/prelude\.(hpp|hh)' # v_noabi preludes
+    Priority: 62
+  - Regex: 'mongocxx/config/prelude\.(hpp|hh)' # v_noabi preludes
+    Priority: 63
+  - Regex: 'bsoncxx/private/.*' # private headers
```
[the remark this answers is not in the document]
eramongodb: The `Priority: 6*` categories look like they can be reordered, so updated as suggested.

However, in general, unfortunately no. The "first regex match" in the order they are listed is what determines the group and priority of the header. e.g. given:
```yaml
- Regex: '<[[:alnum:]_.]+>' # system headers
  Priority: 30
- Regex: 'bsoncxx/private/.*' # private headers
  Priority: 60
```
the `<bsoncxx/private/.*>` headers will never match the priority 60 "private headers" regex because they'll always match the priority 30 "system headers" regex first.

PR #1325, opened 2025-01-30
Re-exclude rhel79 from compile-only matrix in patch builds

eramongodb:
https://github.com/mongodb/mongo-cxx-driver/pull/1313/ incorrectly reverted a patch-only workaround introduced in https://github.com/mongodb/mongo-cxx-driver/pull/1313, which unintentionally reintroduced `compile-only-rhel79-release-shared-impls` task failure in GitHub PR patch builds.

### DIFF
--- a/.evergreen/config_generator/components/compile_only.py
+++ b/.evergreen/config_generator/components/compile_only.py
@@ -61,7 +61,10 @@ def generate_tasks():
             if any(pattern in distro_name for pattern in ['power8', 'zseries']):
                 patchable = False
 
-            # etc/calc_release_version.py: error: unknown option `--format=...'
+            # In etc/calc_release_version.py:
+            #   error: unknown option `format=...'
+            #   usage: git tag ...
+            #      or: ...
             if distro_name == 'rhel79':
                 patchable = False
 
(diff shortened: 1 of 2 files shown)


.evergreen/config_generator/components/compile_only.py
```diff
@@ -61,6 +61,10 @@ def generate_tasks():
             if any(pattern in distro_name for pattern in ['power8', 'zseries']):
                 patchable = False
 
+            # etc/calc_release_version.py: error: unknown option `--format=...'
```
kevinAlbs, line 64:
```suggestion
            # etc/calc_release_version.py: error: unknown option `format...'
```

To match [current message](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_compile_only_matrix_compile_only_rhel79_release_shared_impls_patch_dd0e48bdada5f3c3128791199352a4bb83dfb0c2_679bde77f496dd00075708ab_25_01_30_20_18_00/0/task?bookmarks=0,131&shareLine=67).

kevinAlbs (approved):
Oop. I missed that failing task on https://github.com/mongodb/mongo-cxx-driver/pull/1313. LGTM.

PR #1327, opened 2025-01-30
Attempt to address miscellaneous integration task errors on macos-14-arm64

eramongodb:
Preemptive changes to address macos-14-arm64 task failures observed during work on CXX-2665.

- The curl command to obtain the tag name of the latest release seems to spuriously fail due to unknown reasons specifically when executed on MacOS distros. Following https://github.com/mongodb/mongo-c-driver/pull/1521, a `VERSION_CURRENT` file is guaranteed for latest releases, so that file is detected and used instead when available.

- The `start-mongod.sh` script occasionally fails on MacOS distros due to permission errors. A call to `chmod -x` is added to ensure binaries under `MONGODB_BINARIES` are actually executable as they should be. An obsolete workaround for rhel9 distros is removed in favor of a simple fallback routine for legacy mongo shell binaries.

- A `-Wsign-conversion` warning blocks compilation of examples on MacOS distros when maintainer flags are enabled. The warning is addressed by an explicit cast to `std::size_t` from `std::int64_t`.

### DIFF
--- a/.evergreen/scripts/install-c-driver.sh
+++ b/.evergreen/scripts/install-c-driver.sh
@@ -44,7 +44,7 @@ else
   # RegEx pattern to match SemVer strings. See https://semver.org/.
   declare -r semver_regex="^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
   if echo "${mongoc_version}" | perl -ne "$(printf 'exit 1 unless /%s/' "${semver_regex}")"; then
-    # If $VERSION is already SemVer compliant, use as-is.
+    # If $mongoc_version is already SemVer compliant, use as-is.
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/install-c-driver.sh
```diff
@@ -38,29 +38,36 @@ curl -sS -o mongo-c-driver.tar.gz -L "https://api.github.com/repos/mongodb/mongo
...
-
-  # Append a prerelease tag, e.g. 1.1.0-pre+<version>.
-  build_version="$(printf "%s-pre+%s" "${build_version}" "${mongoc_version}")"
-
-  # Use the constructed prerelease build version when building the C driver.
-  echo "${build_version}" >|"${mongoc_dir}/VERSION_CURRENT"
+  # RegEx pattern to match SemVer strings. See https://semver.org/.
+  declare -r semver_regex="^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
+  if echo "${mongoc_version}" | perl -ne "$(printf 'exit 1 unless /%s/' "${semver_regex}")"; then
+    # If $VERSION is already SemVer compliant, use as-is.
```
kevinAlbs, line 47:
```suggestion
    # If $mongoc_version is already SemVer compliant, use as-is.
```

kevinAlbs (approved):

PR #1333, opened 2025-02-12
CXX-2745 fix missing include directives for SSL-related config macros

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1323 to address `atlas-search-indexes-matrix` task failures due to incorrect SSL configuration detection based on `MONGOCXX_ENABLE_SSL` and `MONGOC_ENABLE_SSL` macros which (incorrectly) may not be defined due to missing include directives, leading to false-positive "SSL support not available" exceptions.

### DIFF
--- a/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp
+++ b/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp
@@ -29,6 +29,8 @@
 #include <mongocxx/private/bson.hh>
 #include <mongocxx/private/client.hh>
 #include <mongocxx/private/client_session.hh>
+#include <mongocxx/private/config/config.hh>
+#include <mongocxx/private/mongoc.hh>
 #include <mongocxx/private/mongoc_error.hh>
 #include <mongocxx/private/pipeline.hh>
 #include <mongocxx/private/read_concern.hh>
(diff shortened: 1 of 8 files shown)


src/mongocxx/test/pool.cpp
```diff
@@ -21,6 +21,7 @@
 #include <mongocxx/options/tls.hpp>
 #include <mongocxx/pool.hpp>
 
+#include <mongocxx/private/config/config.hh>
```
kevinAlbs, line 24: Also add to src/mongocxx/test/client.cpp.

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp
kevinAlbs, line 1: Suggest replacing `#if defined(MONGOCXX_ENABLE_SSL) && defined(MONGOC_ENABLE_SSL)` checks to fail compile if the needed includes are missing. Example:

```c++
// in src/mongocxx/lib/mongocxx/private/config/config.hh.in
#cmakedefine MONGOCXX_ENABLE_SSL
#if defined(MONGOC_ENABLE_SSL) && defined(MONGOCXX_ENABLE_SSL)
#define MONGOCXX_SUPPORTS_SSL() 1
#else
#define MONGOCXX_SUPPORTS_SSL() 0
#endif
```

And update the check to:
```c++
#if MONGOCXX_SUPPORTS_SSL()
```

If the include is missing, results in a build error:
```bash
error: function-like macro 'MONGOCXX_SUPPORTS_SSL' is not defined
```

kevinAlbs (commented):
Good catch.

PR #1335, opened 2025-02-13
Sync augmented.sbom.json with remote version

eramongodb:
Resolves silk-check-augmented-sbom task failure due to mismatched timestamps.


eramongodb:
Investigating new requirements for SBOM generation and updates. Will post a new PR with changes that account for these new requirements.

PR #1351, opened 2025-03-13
CXX-3241 silence deprecation warnings for mongoc_read_prefs_(get|set)_hedge

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1889.

Does not address CXX-3241. This PR just silences the deprecation warnings emitted during builds due to CDRIVER-5916 in advance of CXX-3241 being resolved properly.


vector-of-bool (approved):
LGTM

PR #1356, opened 2025-03-20
CXX-3077 BSON Binary Vector accessor, sub_binary builder

ghost:
This is a basic implementation of CXX-3077, adding `bsoncxx` support for the [BSON Binary Vector](https://github.com/mongodb/specifications/blob/master/source/bson-binary-vector/bson-binary-vector.md) specification.

The main contribution in this PR is to add the `bsoncxx::vector` namespace, and especially the `bsoncxx::vector::accessor` template. A vector accessor can wrap any validated `b_binary` value as read-only, or a mutable accessor can be obtained during append.

General binary items can now be created and edited in-place during an append, using a new `sub_binary` basic builder.

Bumps the Libmongoc version to an unreleased pre-2.0.0 snapshot, to get access to new Libbson features: `bson_append_binary_uninit`, `bson_vector_*_const_view_init`, and `bson_vector_*_binary_data_length`. Validation for existing and newly allocated vectors has been delegated to Libbson, to reuse as much of its testing as possible. The actual vector accessor template in bsoncxx has been implemented in a way similar to Libbson's vector views, but without a direct dependency.

This implementation is currently less comprehensive than the Libbson implementation for BSON Binary Vector. No support is included for conversion between BSON arrays and BSON vectors, and there are no functions for bulk element conversion. Instead, a standard iterator interface allows access to individual elements or element ranges.

The spec includes JSON tests that are similar to bson_corpus. As with bsoncxx's approach to bson_corpus, this PR chooses to delegate those tests to the underlying Libbson implementation.


src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/impl.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+    using reference = std::int8_t&;
+    using const_reference = std::int8_t const&;
+    using iterator = std::int8_t*;
+    using const_iterator = std::int8_t const*;
+
+    static BSONCXX_ABI_EXPORT_CDECL(std::uint32_t) length_for_append(std::size_t element_count);
+    static BSONCXX_ABI_EXPORT_CDECL(header)
+    write_frame(std::uint8_t* binary_data, std::uint32_t binary_data_length, std::size_t element_count);
+    static BSONCXX_ABI_EXPORT_CDECL(view_data<formats::f_int8 const>) const_validate(types::b_binary const& binary);
```
eramongodb, line 80: The `impl` namespace is evading the [prohibited symbols](https://github.com/mongodb/mongo-cxx-driver/blob/22337913e5d9df873491e69612612dd7fda4f575/.evergreen/scripts/abi-prohibited-symbols-test.sh) check for these exported symbols: no symbol exported in the ABI should be declared in a `detail` namespace (converse: all exported ABI symbols should be declared in an ABI namespace).

The `length_for_append`, `write_frame`, and `const_validate` functions within these `format_traits<T>` specializations should be moved into the `v_noabi::vector::formats` namespace (possibly as members of the `f_*` classes):

```cpp
/// @brief Vector format for 32-bit floating point elements, packed least significant byte first.
struct f_float32 {
    static BSONCXX_ABI_EXPORT_CDECL(std::uint32_t) length_for_append(std::size_t element_count);
    static BSONCXX_ABI_EXPORT_CDECL(header)
    write_frame(std::uint8_t* binary_data, std::uint32_t binary_data_length, std::size_t element_count);
    static BSONCXX_ABI_EXPORT_CDECL(view_data<formats::f_int8 const>) const_validate(types::b_binary const& binary);
};
```

If made members of the `f_*` classes, the `bsoncxx::v_noabi::builder::basic::sub_binary::allocate` could then be implemented in terms of `Format` rather than `format_traits<Format>`, e.g.:

```cpp
template <typename Format, typename SFINAE = typename vector::impl::format_traits<Format>::value_type>
vector::view<Format> allocate(Format, std::size_t element_count) {
    std::uint32_t binary_data_length = Format::length_for_append(element_count);
    std::uint8_t* binary_data = allocate(binary_sub_type::k_vector, binary_data_length);
    return {{binary_data, binary_data_length, Format::write_frame(binary_data, binary_data_length, element_count)}};
}
```

This may necessitate moving the `vector::impl::header` and `vector::impl::view_data` types into the public API as well. Alternatively, the exported functions may be refactored to avoid needing/exposing these additional types.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/builder/core.cpp
```diff
@@ -686,6 +703,13 @@ core& core::close_array() {
     return *this;
 }
 
+core& core::close_binary() {
```
eramongodb, line 706: `core->close_binary()` without a `core->open_binary()` is odd. Should one be added for symmetry?

```cpp
template <typename T>
detail::requires_t<void, detail::is_invocable<T, sub_binary>> generic_append(core* core, T&& func) {
    core->open_binary(); // <-- ?
    detail::invoke(std::forward<T>(func), sub_binary(core));
    core->close_binary();
}
```

Then again, support for mismatched open/close functions only seems strictly necessary to extend support to the stream builder, which I do not think we are interested in doing.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
eramongodb, line 1: Missing a forward declaration header.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/iterators.hpp
eramongodb, line 1: Missing a forward declaration header.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
```diff
@@ -0,0 +1,274 @@
...
+   public:
+    /// @brief Construct a packed little-endian value from a float input in the local byte order.
+    /// @param value Floating point value to convert
+    float32(float value) noexcept {
+#if defined(_WIN32) ||                                                                                              \
+    (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)) || \
+    defined(__BYTE_ORDER) && defined(__LITTLE_ENDIAN) && (__BYTE_ORDER == __LITTLE_ENDIAN)
+        std::memcpy(bytes, &value, sizeof value);
+#elif (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)) || \
+    defined(__BYTE_ORDER) && defined(__BIG_ENDIAN) && (__BYTE_ORDER == __BIG_ENDIAN)
```
eramongodb, line 53: We should probably get around to defining our own [std::endian](https://en.cppreference.com/w/cpp/types/endian) at some point. 😅

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/formats.hpp
```diff
@@ -0,0 +1,48 @@
...
+} // namespace bsoncxx
+
+#include <bsoncxx/config/postlude.hpp>
+
+///
+/// @file
+/// Declares vector formats:
+/// @ref bsoncxx::v_noabi::vector::formats::f_float32
+/// @ref bsoncxx::v_noabi::vector::formats::f_int8
+/// @ref bsoncxx::v_noabi::vector::formats::f_packed_bit
```
eramongodb, line 47:
```suggestion
/// Declares entities in @ref bsoncxx::v_noabi::vector::formats.
```
For brevity + to fix formatting.

image

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/view.hpp
```diff
@@ -0,0 +1,281 @@
...
+#include <bsoncxx/config/prelude.hpp>
+
+namespace bsoncxx {
+namespace v_noabi {
+
+namespace builder {
+namespace basic {
+class sub_binary;
+} // namespace basic
+} // namespace builder
```
eramongodb, line 33:
```suggestion
```
Include `<bsoncxx/builder/basic/sub_binary-fwd.hpp>` instead.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/view.hpp
```diff
@@ -0,0 +1,281 @@
...
+        return *(byte_begin() + byte_difference_type(byte_size() - 1u));
+    }
+
+    /// @brief Obtain a reference to a numbered byte, with bounds checking
+    /// @param index Index in the range 0 to byte_size()-1 inclusive.
+    /// @return A byte reference
+    /// @throws std::out_of_range, if the index is outside the allowed range.
+    byte_reference byte_at(byte_count_type index) {
+        if (index >= byte_size()) {
+            throw std::out_of_range{"bsoncxx::vector::view::byte_at()"};
```
eramongodb, line 204: This is unusual. All exceptions thrown by the bsoncxx library (with the exception of standard library polyfills) have type `bsoncxx::v_noabi::exception`, [as documented](https://mongocxx.org/api/mongocxx-4.0.0/classbsoncxx_1_1v__noabi_1_1exception.html):

> Class representing any exceptions emitted from the bsoncxx library or its underlying implementation.

Consider `throw bsoncxx::v_noabi::exception{error_code::k_vector_out_of_range};` or similar instead (as is being done with `k_vector_too_large`).

src/bsoncxx/test/vector.cpp
```diff
@@ -0,0 +1,522 @@
...
+            REQUIRE(inner_back - outer_back == -inner_index);
+        }
+    }
+
+    Iterator iter_copy = begin;
+    REQUIRE(iter_copy == begin);
+    REQUIRE(iter_copy >= begin);
+    REQUIRE(iter_copy <= begin);
+    REQUIRE_FALSE(iter_copy != begin);
+    REQUIRE_FALSE(iter_copy < begin);
```
eramongodb, line 124:
```suggestion
    CHECK(iter_copy == begin);
    CHECK(iter_copy >= begin);
    CHECK(iter_copy <= begin);
    CHECK_FALSE(iter_copy != begin);
    CHECK_FALSE(iter_copy < begin);
```
Recommend using `CHECK` by default rather than `REQUIRE` so more assertions can be evaluated and reported as diagnostic context when able. `REQUIRE` can/should be reserved for preconditions where subsequent test code cannot be executed with well-defined behavior (i.e. `REQUIRE(ptr); CHECK(*ptr == 123);`).

eramongodb (changes requested):

eramongodb (changes requested):
There is currently a const-correctness problem for packed bit iterators:

```cpp
using bsoncxx::builder::basic::sub_binary;

namespace vector = bsoncxx::v_noabi::vector;

auto doc = make_document(kvp("example", [](sub_binary bin) {
    auto vec = bin.allocate(vector::formats::f_packed_bit(), 32);
    auto const& cvec = vec;

    for (auto  e :  vec) {  e = 0; } // OK: as expected.
    for (auto ce : cvec) { ce = 1; } // OK (!?)
}));
```

Expanding the types involved in `for (auto ce : cvec)`:

```cpp
// begin() const -> iterator
vector::iterators::packed_bit_element<std::uint8_t*> __begin = cvec.begin();

// operator*() const -> reference
vector::elements::packed_bit_element<std::uint8_t*> ce = *__begin;

// operator=(packed_bit_element const& v) const -> packed_bit_element const&
ce = 1;
```

To address this issue, member functions in `elements::packed_bit_*` should not be falsely const-qualified:

```cpp
packed_bit_element& operator=(value_type const& v);
```

The `reference` and `pointer` type aliases in `vector::iterators::packed_bit_*` classes should not be const-qualified:

```cpp
using reference = elements::packed_bit_element<Iterator>;
using pointer = elements::packed_bit_element<Iterator>*;
```

Finally, the `begin()` and `end()` functions (as well as their `byte_*()` counterparts) should have proper const and non-const overloads (+ `cbegin()` and `cend()` is optional but nice-to-have) which return `const_iterator` and `iterator` respectively:

```cpp
constexpr const_iterator begin() const noexcept;
constexpr const_iterator end() const noexcept;

constexpr iterator begin() noexcept;
constexpr iterator end() noexcept;

constexpr const_byte_iterator byte_begin() const noexcept;
constexpr const_byte_iterator byte_end() const noexcept;

constexpr byte_iterator byte_begin() noexcept;
constexpr byte_iterator byte_end() noexcept;
```

> [!NOTE]
> The other type aliases and member functions for `accessor<T>`, `iterators::packed_bit_*`, and `format_traits<T>` do not appear to inherit these issues and do not require any changes. 

The example code above should then correctly error for `ce = 1` due to the `Iterator` template parameter inheriting const-qualification from the const-qualified `vector::accessor<T>`:

```cpp
// begin() const -> const_iterator
vector::iterators::packed_bit_element<std::uint8_t const*> __begin = cvec.begin();

// operator*() const -> reference
vector::elements::packed_bit_element<std::uint8_t const*> ce = *__begin;

// operator=(packed_bit_element const& v) -> packed_bit_element&
ce = 1;
```

```
 error: read-only variable is not assignable
  113 |             *byte &= std::uint8_t(~mask);
      |             ~~~~~ ^
note: in instantiation of member function 'packed_bit_element<const unsigned char *>::operator=' requested here
  543 |         ce = 1; // error: good.
      |            ^
```

eramongodb:
> There is currently a const-correctness problem for packed bit iterators.

Apologies, disregard the feedback regarding const-correctness. Most of the behavior mentioned were incorrectly interpreted by me as const-correctness issues due to confusing `accessor<T>`'s span-like properties and `elements::packed_bit_*`'s proxy-reference-like properties. Upon further inspection, it is not wrong for them to support shallow-const access (i.e. [std::span<T>::begin() -> iterator](https://timsong-cpp.github.io/cppwp/n4950/span.iterators)) and const-qualified assignment operators (i.e. [std::vector<bool>::reference::operator](https://timsong-cpp.github.io/cppwp/n4950/vector.bool.pspc)).

Instead, I believe `accessor<T>` should support `cbegin()` + friends which return a `const_iterator` or `const_byte_reference`. The member functions of `elements::packed_bit_*` do not need their const-qualifications removed as I suggested above. The `reference and `pointer` type aliases in `iterators::packed_bit_*` however can still probably do without the const-qualifications.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
```diff
@@ -79,6 +79,7 @@ enum class binary_sub_type : std::uint8_t {
     k_encrypted = 0x06,         ///< Encrypted BSON value.
     k_column = 0x07,            ///< Compressed BSON column.
     k_sensitive = 0x08,         ///< Sensitive.
+    k_vector = 0x09,            ///< BSON Binary Vector specification.
```
eramongodb, line 82:
```suggestion
    k_vector = 0x09,            ///< BSON Binary Vector.
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/accessor.hpp
```diff
@@ -0,0 +1,343 @@
...
+        typename format_traits::const_byte_iterator,
+        typename format_traits::byte_iterator>::type;
+
+    /// @brief Construct a const Vector accessor by validating a bsoncxx::v_noabi::types::b_binary reference.
+    /// @param binary Non-owning reference to BSON binary data
+    /// @throws bsoncxx::v_noabi::exception with bsoncxx::v_noabi::error_code::k_invalid_vector, if validation fails.
+    ///
+    /// The Binary data is validated as a Vector of the templated Format. On success, an accessor is created which
+    /// references the same data as the bsoncxx::v_noabi::types::b_binary pointer.
+    accessor(types::b_binary const& binary) : _data((format::validate(binary), binary)) {}
```
eramongodb, line 163: Suggest adding a method to support easily convert an `accessor<T>` to `accessor<T const>`.

An iterator constructor via `.cbegin()` and `.cend()` might be easiest to implement, which can take advantage of the fact that `accessor<T>::const_iterator` and `accessor<T const>::iterator` will always be the same type, e.g.:

```cpp
auto  vec = sb.allocate(f_int8_t(), ...);
auto cvec = accessor<f_int8_t const>(vec.cbegin(), vec.cend());
```

Optionally, a converting constructor could also be added for convenience:

```cpp
auto  vec = sb.allocate(f_int8_t(), ...);
auto cvec = accessor<f_int8_t const>(vec);
```

Alternatively, instead of constructors (which may require non-trivial template parameter constraints), an `accessor<T>::as_const() -> accessor<T const>` would also be acceptable:

```cpp
auto  vec = sb.allocate(f_int8_t(), ...);
auto cvec = vec.as_const();
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/accessor.hpp
```diff
@@ -0,0 +1,343 @@
...
+    /// Byte iterator type
+    ///
+    /// @hideinitializer
+    /// For example: std::uint8_t*, std::uint8_t const*, bsoncxx::v_noabi::vector::iterators::packed_bit_byte
+    using byte_iterator = typename std::conditional<
+        std::is_const<Format>::value,
+        typename format_traits::const_byte_iterator,
+        typename format_traits::byte_iterator>::type;
+
+    /// @brief Construct a const Vector accessor by validating a bsoncxx::v_noabi::types::b_binary reference.
```
eramongodb, line 157:
```suggestion
    /// @brief Construct a Vector accessor by validating a bsoncxx::v_noabi::types::b_binary reference.
```
Not necessarily const?

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
```diff
@@ -0,0 +1,265 @@
...
+   private:
+    friend class iterators::packed_bit_byte<Iterator>;
+
+    constexpr packed_bit_byte(Iterator byte_iter, uint8_t byte_mask) noexcept : byte(byte_iter), mask(byte_mask) {}
+
+    Iterator byte;
+    std::uint8_t mask;
+};
+
+/// packed_bit_byte is Swappable even when it's not an lvalue reference
```
eramongodb, line 246:
```suggestion
/// Swap the referenced values for `a` and `b`.
///
/// @note `packed_bit_byte<Iterator>` is a proxy reference and behaves like an lvalue reference.
```
Minor wording suggestions to clarify the expected behavior of `packed_bit_byte<Iterator>` as a "proxy reference" type.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/vector.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <bsoncxx/exception/exception.hpp>
+
+#include <bson/bson.h>
+
+namespace bsoncxx {
+namespace v_noabi {
+namespace vector {
+namespace detail {
+
+enum element_type : std::uint8_t {
```
eramongodb, line 32:
```suggestion
// Equivalent to bson_vector_element_type_t.
enum element_type : std::uint8_t {
```
A reference to help maintain C and C++ driver consistency.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/vector.cpp
```diff
@@ -0,0 +1,119 @@
...
+    signed_integer = 0,
+    unsigned_integer = 1,
+    floating_point = 2,
+};
+
+enum element_size : std::uint8_t {
```
eramongodb, line 38:
```suggestion
// Equivalent to bson_vector_element_size_t.
enum element_size : std::uint8_t {
```
A reference to help maintain C and C++ driver consistency.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/detail.hpp
```diff
@@ -0,0 +1,167 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <array>
```
eramongodb, line 17:
```suggestion
#pragma once

#include <bsoncxx/vector/detail-fwd.hpp>

//

#include <array>
```

eramongodb (approved):
Minor suggestions remaining; otherwise, LGTM!

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/accessor.hpp
eramongodb: I see. If that's the case, I think I'm most in favor of the `.as_const()` approach, which would avoid forcing the user to specify the target type.

kevinAlbs (commented):
Apologies for delayed review.

kevinAlbs (approved):

PR #1363, opened 2025-03-31
Address gethostname failure due to missing WinSock initialization

eramongodb:
Addresses the following `stderr` output by the `test_bson` test binary on Windows:

```
unable to get hostname: 0
```

This message is emitted by a call to `gethostname` in [`_bson_context_get_hostname`](https://github.com/mongodb/mongo-c-driver/blob/7a05b03ac246cbbe317c5a304b02a7b00c5bd499/src/libbson/src/bson/bson-context.c#L88) by the following stack trace:

```
bsoncxx::v_noabi::oid::oid()
  - bson_oid_init(bson_oid_t *oid, bson_context_t *context)
  - bson_context_get_default()
  - _bson_context_init_default()
  - _bson_context_init(bson_context_t *context, bson_context_flags_t flags)
  - _bson_context_init_random (bson_context_t *context, bool init_seq)
  - _bson_context_get_hostname (char out[HOST_NAME_MAX])
  - gethostname (char *name, int namelen)
```

Patching the mongoc library to print the value of [`WSAGetLastError`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsagetlasterror) (rather than `errno`) reveals an error code value of [`WSANOTINITIALIZED`](https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2#WSANOTINITIALISED). This reveals that `gethostname`, a function provided by the WinSock library, is being invoked before a successful call to [`WSAStartup`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup). This call is usually handled by [`mongoc_init`](https://github.com/mongodb/mongo-c-driver/blob/7a05b03ac246cbbe317c5a304b02a7b00c5bd499/src/libmongoc/src/mongoc/mongoc-init.c#L130). However, `test_bson` does not use the `mongoc` or `mongocxx` library. Therefore, it does not call `mongoc_init` or `WSAStartup` before calling `bson_oid_init`.

This PR proposes a straightforward workaround to manually invoke `WSAStartup` prior to executing the test suite on Windows. Filed CDRIVER-5957 to investigate a more thorough solution to this problem (e.g. document that `mongoc_init` or `WSAStartup` must have been invoked first? add internal calls to `WSAStartup`/`WSACleanup` that wrap the `gethostname` on Windows?).

### DIFF
--- a/src/bsoncxx/test/catch.cpp
+++ b/src/bsoncxx/test/catch.cpp
@@ -18,11 +18,41 @@
 
 #include <bsoncxx/v1/config/export.hpp>
 
+#include <cstdlib>
+#include <iostream>
 #include <system_error>
+
+#include <bsoncxx/private/bson.hh> // <winsock.h> via <bson/bson-compat.h>
 
 #include <catch2/catch_session.hpp>
 
+// Ensure the Winsock DLL is initialized prior to calling `gethostname` in `bsoncxx::oid::oid()`:
+//  - bson_oid_init -> bson_context_get_default -> ... -> _bson_context_init_random -> gethostname.
+struct WSAGuard {
+    ~WSAGuard() {
+#if defined(_WIN32)
+        (void)WSACleanup();
+#endif
+    }
+
+    WSAGuard(WSAGuard&&) = delete;
+    WSAGuard& operator=(WSAGuard&) = delete;
+    WSAGuard(WSAGuard const&) = delete;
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM. Nice investigation.

> Filed [CDRIVER-5957](https://jira.mongodb.org/browse/CDRIVER-5957)

Commented with idea for alternative.

PR #1370, opened 2025-04-03
Temporarily relocate sbom and valgrind from rhel80 to rhel8.8

eramongodb:
Unblocks patch builds executing the sbom and valgrind tasks due to a regression in the system packages installed on the rhel80 distro (currently being investigated). The next highest-host-availabity compatible distro appears to be `rhel8.8`. The valgrind task names are manually left unchanged from `rhel80` to avoid breaking task history with this workaround.

### DIFF
--- a/.evergreen/config_generator/etc/distros.py
+++ b/.evergreen/config_generator/etc/distros.py
@@ -60,6 +60,8 @@ MACOS_ARM64_DISTROS = [
 ]
 
 RHEL_DISTROS = [
+    *ls_distro(name='rhel8.8', os='rhel', os_type='linux', os_ver='8.9'),
+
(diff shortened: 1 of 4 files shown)


.evergreen/config_generator/etc/distros.py
```diff
@@ -60,6 +60,8 @@ def ls_distro(name, **kwargs):
 ]
 
 RHEL_DISTROS = [
+    *ls_distro(name='rhel8.8', os='rhel', os_type='linux', os_ver='8.9'),
```
kevinAlbs, line 63:
```suggestion
    *ls_distro(name='rhel8.8', os='rhel', os_type='linux', os_ver='8.8'),
```

kevinAlbs (approved):

PR #1372, opened 2025-04-03
Remove client arg from test helpers

kevinAlbs:
# Summary

Remove `client` arg from test helpers that cache results.

Verified with this patch build: https://spruce.mongodb.com/version/67ee96310b754c000700e66c

# Motivation

When working on CXX-3209 I was surprised to discover calling `test_util::get_max_wire_version` on a differently versioned `mongocryptd` returned the same value as the running `mongod`:

```c++
// Check maxWireVersion of mongocryptd:
mongocxx::client mongocryptd_client(uri("mongodb://localhost:27020"));
auto mwv = test_util::get_max_wire_version(mongocryptd_client); // same as the running mongod?!
```

The cause appeared to be due to cached results in test helpers:
```c++
// These frequently used network calls are cached to avoid bottlenecks during tests.
document::value get_is_master(client const& client) {
    static auto reply = client["admin"].run_command(make_document(kvp("isMaster", 1)));
    return reply;
}
```

Later calls ignore the `client` argument. To avoid surprising readers, the `client` argument is removed from these calls. The calls always use a client constructed with the URI "mongodb://localhost:27017". Overloads accepting a client can be added later as needed.

### DIFF
--- a/src/mongocxx/test/client_helpers.cpp
+++ b/src/mongocxx/test/client_helpers.cpp
@@ -57,7 +57,7 @@ namespace {
 // These frequently used network calls are cached to avoid bottlenecks during tests.
 document::value get_is_master() {
     static auto reply = []() {
-        auto client = mongocxx::client(mongocxx::uri());
+        auto client = mongocxx::client{mongocxx::uri{}, test_util::add_test_server_api()};
         return client["admin"].run_command(make_document(kvp("isMaster", 1)));
     }();
     return reply;
@@ -65,7 +65,7 @@ document::value get_is_master() {
 
 document::value get_server_status() {
     static auto status = []() {
-        auto client = mongocxx::client(mongocxx::uri());
+        auto client = mongocxx::client{mongocxx::uri{}, test_util::add_test_server_api()};
         return client["admin"].run_command(make_document(kvp("serverStatus", 1)));
     }();
     return status;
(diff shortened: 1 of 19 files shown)


src/mongocxx/test/client_helpers.cpp
```diff
@@ -262,18 +271,23 @@ std::int32_t get_max_wire_version(client const& client) {
...
+    auto output = get_server_status();
     return bsoncxx::string::to_string(output.view()["version"].get_string().value);
 }
 
-document::value get_server_params(client const& client) {
-    auto reply = client["admin"].run_command(make_document(kvp("getParameter", "*")));
+document::value get_server_params() {
+    // Cache reply.
+    static auto reply = []() {
+        auto client = mongocxx::client(mongocxx::uri());
```
eramongodb, line 282:
```suggestion
        auto client = mongocxx::client{mongocxx::uri{}, test_util::add_test_server_api()};
```
This seems to be what is causing versioned API task failures.

eramongodb (changes requested):

src/mongocxx/test/client_helpers.cpp
kevinAlbs: Oop. Thank you.

eramongodb (approved):

PR #1162, opened 2024-06-25
CXX-2828 QE Range Protocol V2

adriandole:
Addresses CXX-2828 and CXX-3014

Addresses several issues encountered attempting to run Range V2 tests:

- Add a Debian 12 variant to test 8.0+ servers. No tasks were running a new enough server version to test Range V2.
- Address compile issues encountered with GCC 12.
- Apply fix to run QE/CSFLE tests (https://github.com/mongodb/mongo-cxx-driver/pull/1162/commits/710193709e7891ebd79dc152a139020ddf900b5b). Tests were inadvertently skipped due to recent test runner changes (https://github.com/mongodb/mongo-cxx-driver/commit/fb2698102a2ddcfa0e3a120c39a8672a20e83753).

Addresses issues encountered upgrading to C driver 1.28.0:
- Add `-Wno-error=maybe-uninitialized` to RHEL 7.9. Intended to ignore this [RHEL 7.9 (gcc 4.8.5) build task](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_rhel79_compile_compile_without_tests_patch_87fe3eb5bc55d3789f7f0177adf5a5d95f1a1732_66fbbb8a2304e40007dce7ee_24_10_01_09_06_19/0/task?bookmarks=0,794) error:

  ```
  ../../../mongo-c-driver-src/src/libmongoc/src/mongoc/mongoc-uri.c:489:15: error: 'tmp' may be used uninitialized in this function [-Werror=maybe-uninitialized]
      } else if (*str) {
                 ^
  ../../../mongo-c-driver-src/src/libmongoc/src/mongoc/mongoc-uri.c:1419:16: note: 'tmp' was declared here
      const char *tmp;
                  ^
  ```
  This appears to be a false positive. The `tmp` on line 1419 is a different variable from the `tmp` in scope on 489. `-Wno-error=maybe-uninitialized` was added to the RHEL 7.9 distro.
- Filed CXX-3112 to investigate a test failure with the legacy transactions test on sharded cluster tasks on 4.2. Appears unrelated to QE Range.

PR #1373, opened 2025-04-04
CXX-3266 avoid inheriting BUILD_VERSION by auto-downloaded C Driver

eramongodb:
Resolves CXX-3266.

> [!IMPORTANT]
> This PR is only concerned with a C Driver project whose CMake targets are imported via `add_subdirectory()` instead of `find_package()`. This may either be an auto-downloaded C Driver (via `FetchMongoC.cmake`) or a manual call to `add_subdirectory()` by a higher-level CMake project.

The API version numbers embedded in the C Driver v2 install paths revealed suspicious behavior during local development and testing for CXX-3101. Given `LIBMONGOC_DOWNLOAD_VERSION` (+ related variables) set to `2.0.0` and using the [latest commit](fe7bfd3599c0dd7e1f6b9aacbef362c193f36c37) (targeting a 4.1.0 release), the following install directory structure was observed (the API version for C Driver libraries is `0.0.0`?!):

```
<prefix>
├── include
│   ├── bson-0.0.0
│   ├── mongoc-0.0.0
│   └── ...
└── lib
    ├── cmake
    │   ├── bson-0.0.0
    │   ├── mongoc-0.0.0
    │   └── ...
    ├── pkgconfig
    │   ├── bson0.pc
    │   ├── mongoc0.pc
    │   └── ...
    ├── libbson0.so.0.0.0
    ├── libmongoc0.so.0.0.0
    └── ...
```

This led down a rabbit hole of tracking down the root cause of this issue. Investigation revealed this bug is actually present since C++ Driver 3.9.0, which introduced auto-download support in https://github.com/mongodb/mongo-cxx-driver/pull/967 ([CXX-2695](https://jira.mongodb.org/browse/CXX-2695)). However, it was not observed due to the 3.9.0 release auto-downloading C Driver 1.25.0, which temporarily disabled support for `BUILD_VERSION` as a config option in https://github.com/mongodb/mongo-c-driver/pull/1382 ([CDRIVER-4777](https://jira.mongodb.org/browse/CDRIVER-4777)). Support was restored in 1.26.0 by https://github.com/mongodb/mongo-c-driver/pull/1462 ([CDRIVER-4763](https://jira.mongodb.org/browse/CDRIVER-4763)), but the C++ Driver continued to use 1.25.0 until https://github.com/mongodb/mongo-cxx-driver/pull/1162 ([CXX-2828](https://jira.mongodb.org/browse/CXX-2828)) in the 3.11.0 release.

> [!IMPORTANT]
> This bug is present since C++ Driver 3.9.0 when the auto-downloaded C Driver was introduced.

> [!IMPORTANT]
> This bug is present since C Driver 1.14.0 when `BUILD_VERSION` was introduced, _except_ for 1.25.0 when support for `BUILD_VERSION` as a CMake config option was temporarily removed, which also happened to be the version auto-downloaded by C++ Driver 3.9.0.

The problem is twofold. The C Driver may incorrectly inherit `BUILD_VERSION` during initial configuration:

```
# Initial configuration with BUILD_VERSION=5.0.0.
$ cmake -S . -B build -D BUILD_VERSION=5.0.0

# Auto-downloaded C Driver incorrectly inherited BUILD_VERSION.
$ grep -Rho --include 'bson-version.h' '#define BSON_.*VERSION (.*' build
#define BSON_MAJOR_VERSION (5)
#define BSON_MINOR_VERSION (0)
#define BSON_MICRO_VERSION (0)
#define BSON_PRERELEASE_VERSION ()
#define BSON_VERSION (2.0.0)
```

or on subsequent reconfiguration using an existing CMake cache:

```
# Initial configuration without BUILD_VERSION.
$ cmake -S . -B build

# Auto-downloaded C Driver computed its version correctly (for now...).
$ grep -Rho --include 'bson-version.h' '#define BSON_.*VERSION (.*' build
#define BSON_MAJOR_VERSION (2)
#define BSON_MINOR_VERSION (0)
#define BSON_MICRO_VERSION (0)
#define BSON_PRERELEASE_VERSION ()
#define BSON_VERSION (2.0.0)

# Reconfigure the project without making any changes.
$ cmake -S . -B build

# Auto-downloaded C Driver incorrectly inherited BUILD_VERSION.
$ grep -Rho --include 'bson-version.h' '#define BSON_.*VERSION (.*' build
#define BSON_MAJOR_VERSION (0)
#define BSON_MINOR_VERSION (0)
#define BSON_MICRO_VERSION (0)
#define BSON_PRERELEASE_VERSION ()
#define BSON_VERSION (2.0.0)
```

Note also the confusing mix of version numbers, which depend on whether the computation was derived from the `BUILD_VERSION` cache variable (via `BuildVersion.cmake`) or from the C Driver's own copy of the `VERSION_CURRENT` file (via `LoadVersion.cmake`).

This problem is further exacerbated by the following conditions:

- configuration output suggests the correct C Driver version is being computed and used: emits `libmongoc version (from VERSION_CURRENT file): 2.0.0` despite bad version numbers.
- requiring a specific configuration which selects auto-downloaded C Driver with either `BUILD_VERSION` set or reconfiguration step to trigger reuse of the `BUILD_VERSION` cache variable.
- a [deliberate(?)](https://github.com/mongodb/mongo-cxx-driver/blob/fe7bfd3599c0dd7e1f6b9aacbef362c193f36c37/CMakeLists.txt#L65) lack of version checks when C Driver targets are imported via `add_subdirectory()`.
- the [absence](https://github.com/mongodb/mongo-cxx-driver/blob/fe7bfd3599c0dd7e1f6b9aacbef362c193f36c37/src/bsoncxx/CMakeLists.txt#L86) of API version checks in CMake dependency requirements for downstream projects (no `<version>` included in call to `find_dependency(bson-1.0)` in `bsoncxx-config.cmake`).
- the absence of API version numbers in directory or file names (to make the discrepancy more immediately obvious on inspection of the install directory contents).
- the absence of auto-downloaded C Driver installation integrity checks as part of the C++ Driver's EVG configuration.
- most of our EVG config having migrated to using a pre-installed C Driver with `find_package()` (for CSFLE support) rather than the auto-downloaded C Driver (only used by benchmarks, macro guard tests, mongohouse tests, scan-build tests, uninstall checks, and versioned API tests).

Most of these issues will be addressed in a followup PR as part of CXX-3101.

The changes in PR are planned to be cherry-picked onto the v4.0 and v3.11 release branches.

### DIFF
--- a/cmake/FetchMongoC.cmake
+++ b/cmake/FetchMongoC.cmake
@@ -3,7 +3,7 @@
 include(FetchContent)
 
 function(fetch_mongoc)
-    message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... end")
+    message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... begin")
 
     set(fetch_args "")
     if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.25.0")
--- a/cmake/FetchMongoC.cmake
+++ b/cmake/FetchMongoC.cmake
@@ -23,11 +23,7 @@ function(fetch_mongoc)
 
     if(NOT mongo-c-driver_POPULATED)
         # Must ensure BUILD_VERSION is not inherited either as a normal variable or as a cache variable.
-        unset(OLD_BUILD_VERSION)
-        if(DEFINED BUILD_VERSION)
-            set(OLD_BUILD_VERSION ${BUILD_VERSION})
-            unset(BUILD_VERSION)
-        endif()
+        unset(BUILD_VERSION)
         unset(OLD_CACHE_BUILD_VERSION)
         if(DEFINED CACHE{BUILD_VERSION})
             set(OLD_CACHE_BUILD_VERSION $CACHE{BUILD_VERSION})
@@ -42,10 +38,7 @@ function(fetch_mongoc)
 
         FetchContent_MakeAvailable(mongo-c-driver)
 
-        # Restore prior value of BUILD_VERSION only when they were previously set.
-        if(DEFINED OLD_BUILD_VERSION)
-            set(BUILD_VERSION ${OLD_BUILD_VERSION})
-        endif()
+        # Restore prior value of BUILD_VERSION cache variable only if was previously set.
         if(DEFINED OLD_CACHE_BUILD_VERSION)
             set(BUILD_VERSION ${OLD_CACHE_BUILD_VERSION} CACHE STRING "Library version (for both bsoncxx and mongocxx)")
         endif()
--- a/cmake/FetchMongoC.cmake
+++ b/cmake/FetchMongoC.cmake
@@ -14,7 +14,7 @@ function(fetch_mongoc)
     FetchContent_Declare(
         mongo-c-driver
         GIT_REPOSITORY https://github.com/mongodb/mongo-c-driver.git
-        GIT_TAG ${MONGOC_DOWNLOAD_VERSION}
+        GIT_TAG ${LIBMONGOC_DOWNLOAD_VERSION}
 
         ${fetch_args}
     )
--- a/cmake/FetchMongoC.cmake
+++ b/cmake/FetchMongoC.cmake
@@ -2,43 +2,56 @@
 
 include(FetchContent)
 
-message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... begin")
+function(fetch_mongoc)
+    message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... end")
 
-set(fetch_args "")
-if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.25.0")
-    set(fetch_args "SYSTEM")
-endif()
+    set(fetch_args "")
+    if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.25.0")
+        list(APPEND fetch_args "SYSTEM")
+    endif()
 
-# Declare mongo-c-driver as a dependency
-FetchContent_Declare(
-    mongo-c-driver
-    GIT_REPOSITORY https://github.com/mongodb/mongo-c-driver.git
-    GIT_TAG ${LIBMONGOC_DOWNLOAD_VERSION}
+    # Declare mongo-c-driver as a dependency
+    FetchContent_Declare(
+        mongo-c-driver
+        GIT_REPOSITORY https://github.com/mongodb/mongo-c-driver.git
+        GIT_TAG ${MONGOC_DOWNLOAD_VERSION}
 
-    ${fetch_args}
-)
+        ${fetch_args}
+    )
 
-FetchContent_GetProperties(mongo-c-driver)
+    FetchContent_GetProperties(mongo-c-driver)
 
-if(NOT mongo-c-driver_POPULATED)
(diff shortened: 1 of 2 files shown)


cmake/FetchMongoC.cmake
```diff
@@ -2,43 +2,56 @@
...
+        set(BUILD_TESTING OFF)
+        string(REPLACE " -Werror" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+        string(REPLACE " -Werror" "" CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+
+        FetchContent_MakeAvailable(mongo-c-driver)
+
+        # Restore prior value of BUILD_VERSION only when they were previously set.
+        if(DEFINED OLD_BUILD_VERSION)
+            set(BUILD_VERSION ${OLD_BUILD_VERSION})
+        endif()
```
kevinAlbs, line 48:
```suggestion
        # Restore prior value of cached BUILD_VERSION only when previously set.
```

IIUC restoring the normal variable is not needed, since the `unset(BUILD_VERSION)` will not impact the parent scope.

cmake/FetchMongoC.cmake
```diff
@@ -2,43 +2,56 @@
...
-    FetchContent_MakeAvailable(mongo-c-driver)
-    set(CMAKE_CXX_FLAGS ${OLD_CMAKE_CXX_FLAGS})
-    set(CMAKE_C_FLAGS ${OLD_CMAKE_C_FLAGS})
-    set(ENABLE_TESTS ${OLD_ENABLE_TESTS})
-    set(BUILD_TESTING ${OLD_BUILD_TESTING})
-endif()
-
-message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... end")
+function(fetch_mongoc)
+    message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... end")
```
kevinAlbs, line 6:
```suggestion
    message(STATUS "Download and configure C driver version ${LIBMONGOC_DOWNLOAD_VERSION} ... begin")
```

kevinAlbs (approved):
LGTM with minor comments. Thank you for investigating and fixing. I like the use of a function scope to isolate the normal CMake variables.

cmake/FetchMongoC.cmake
```diff
@@ -2,43 +2,56 @@
...
+    set(fetch_args "")
+    if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.25.0")
+        list(APPEND fetch_args "SYSTEM")
+    endif()
+
+    # Declare mongo-c-driver as a dependency
+    FetchContent_Declare(
+        mongo-c-driver
+        GIT_REPOSITORY https://github.com/mongodb/mongo-c-driver.git
+        GIT_TAG ${MONGOC_DOWNLOAD_VERSION}
```
kevinAlbs, line 17:
```suggestion
        GIT_TAG ${LIBMONGOC_DOWNLOAD_VERSION}
```

This may fix failing tasks that seem to be using the auto-downloaded C driver. [Example](https://spruce.mongodb.com/task/mongo_cxx_driver_versioned_api_matrix_versioned_api_two_accepted_macos_14_arm64_debug_shared_patch_50cda18f9233e15bb0daf90715bc586db7650499_67f406375617eb000769ff23_25_04_07_17_07_08/logs?execution=0).

vector-of-bool (approved):
LGTM (+1 to Kevin's comments)

PR #1374, opened 2025-04-07
CXX-3204 add URI setter for `server_selection_try_once`

kevinAlbs:
# Summary

- Add method to `mongocxx::uri` to set `serverSelectionTryOnce` URI option.

# Background & Motivation

Intended to help simplify a a [workaround](https://github.com/10gen/mongo/blob/e758d87c23d7ffc70bb767e618692063dccaa826/src/mongo/db/modules/enterprise/src/streams/exec/mongocxx_utils.cpp#L72-L110) in Atlas Stream Processing. The workaround applies `serverSelectionTryOnce=false` to a URI if the option is not already specified. I expect this new setter can be used in [streams after constructing the mongocxx::uri](https://github.com/10gen/mongo/blob/e758d87c23d7ffc70bb767e618692063dccaa826/src/mongo/db/modules/enterprise/src/streams/exec/mongocxx_utils.cpp#L332) like:

```c++
auto uri = std::make_unique<mongocxx::uri>(uri);
// Set default "serverSelectionTryOnce=false" if not included in URI:
if (!uri->server_selection_try_once()) {
    uri->server_selection_try_once(false);
}
```

This is the first setter on `mongocxx::uri`. If acceptable, this pattern can be extended in the future to add setters for more URI options to resolve CXX-2146.

## Rejected alternatives

A new option to `mongocxx::options::client` was considered but rejected. Server selection spec [notes](https://github.com/mongodb/specifications/blob/ce35696db1c8da0e15a067b8022c853f1d6d1292/source/server-selection/server-selection.md#serverselectiontryonce):

> Multi-threaded drivers MUST NOT provide this mode.

But `mongocxx::options::client` is applied to both `mongocxx::client` and `mongocxx::pool` constructors.

The spec also notes:

> Conflicting usages of the URI option and the symbol is an error.

A URI setter avoids possible disagreement.

### DIFF
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/uri.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/uri.hpp
@@ -299,6 +299,14 @@ class uri {
     server_selection_try_once() const;
 
     ///
+    /// Sets the value of the option "serverSelectionTryOnce" in the uri.
+    ///
+    /// @param val The new value to apply to as "serverSelectionTryOnce".
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(void)
+    server_selection_try_once(bool val);
+
+    ///
     /// Returns the value of the option "socketTimeoutMS" if present in the uri.
     ///
     /// @return An optional std::int32_t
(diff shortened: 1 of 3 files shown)


vector-of-bool (approved):
LGTM

PR #1380, opened 2025-04-09
Use majority concern for tailable cursor tests

eramongodb:
Attempt to address the following [flaky test failure](https://spruce.mongodb.com/task/mongo_cxx_driver_integration_matrix_macos_integration_macos_14_arm64_debug_shared_cxx17_csfle_8.0_sharded_patch_a50c2f6018a804e11c03e8aafb2439c0e7122637_67f6cd788236020007393486_25_04_09_19_41_47/logs?execution=0):

```
collection.cpp:2413: FAILED: iter != cursor.end() for: {?} != {?} with 1 message: 'k_tailable_await'
```

Applies majority concern in a manner similar to https://github.com/mongodb/mongo-cxx-driver/pull/1375.

### DIFF
--- a/src/mongocxx/test/collection.cpp
+++ b/src/mongocxx/test/collection.cpp
@@ -2340,12 +2340,26 @@ TEST_CASE("Cursor iteration", "[collection][cursor]") {
     client mongodb_client{uri{}, test_util::add_test_server_api()};
     database db = mongodb_client["collection_cursor_iteration"];
 
-    auto capped_name = std::string("mongo_cxx_driver_capped");
-    collection coll = db[capped_name];
+    // Drop and (re)create the capped collection with majority concern.
+    collection coll = [&db] {
+        write_concern wc_majority;
+        wc_majority.acknowledge_level(write_concern::level::k_majority);
 
-    // Drop and (re)create the capped collection.
-    coll.drop();
-    db.create_collection(capped_name, make_document(kvp("capped", true), kvp("size", 1024 * 1024)));
+        read_concern rc_majority;
+        rc_majority.acknowledge_level(read_concern::level::k_majority);
+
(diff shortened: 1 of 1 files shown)


src/mongocxx/test/collection.cpp
```diff
@@ -2340,12 +2340,26 @@ TEST_CASE("Cursor iteration", "[collection][cursor]") {
...
+        auto ret = db[capped_name];
+
+        ret.drop();
+        ret = db.create_collection(capped_name, make_document(kvp("capped", true), kvp("size", 1024 * 1024)));
+
+        ret.write_concern(wc_majority);
+        ret.read_concern(rc_majority);
+
+        return ret;
+    }(); // IILE
```
kevinAlbs, line 2362:
```suggestion
    }(); // IIFE
```

Assuming this is referring to "Immediately Invoked Function Expression".

kevinAlbs (approved):

src/mongocxx/test/collection.cpp
eramongodb: Whoops, included this out of habit. It stands for "Immediately Invoked _Lambda_ Expression", since it is a "lambda expression" in C++ rather than a "function expression" as in JavaScript which popularized the term.

eramongodb:
It seems these changes were insufficient, as flaky task failures are [still occurring](https://parsley.mongodb.com/evergreen/mongo_cxx_driver_integration_matrix_macos_integration_macos_14_arm64_debug_shared_cxx11_csfle_latest_sharded_patch_f1c276f7e26b29063d2ec31bc85d6627fd07d38a_67fe7b1ab560be0007788818_25_04_15_15_28_28/0/task?bookmarks=0%2C1675&selectedLineRange=L1655&shareLine=1655).

PR #1381, opened 2025-04-11
CXX-3253 reintroduce VS 2015 task coverage

eramongodb:
Resolves CXX-3253. Followup to https://github.com/mongodb/mongo-c-driver/pull/1991.

### DIFF
--- a/.evergreen/config_generator/components/compile_only.py
+++ b/.evergreen/config_generator/components/compile_only.py
@@ -80,6 +80,7 @@ def tasks():
                 Compile.call(
                     build_type=build_type,
                     compiler=compiler,
+                    vars=compile_vars,
                 )
             ]
 
--- a/.evergreen/config_generator/components/compile_only.py
+++ b/.evergreen/config_generator/components/compile_only.py
@@ -9,8 +9,6 @@ from shrub.v3.evg_build_variant import BuildVariant
 from shrub.v3.evg_command import KeyValueParam, expansions_update
 from shrub.v3.evg_task import EvgTask, EvgTaskRef
 
-from itertools import product
-
 
 TAG = 'compile-only'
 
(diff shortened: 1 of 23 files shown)


eramongodb:
Converting to draft due to discovery of overlooked EVG configuration errors. Will fix and re-request reviews when ready.

eramongodb:
Identified `compile_vars` was [not being passed](https://github.com/mongodb/mongo-cxx-driver/blob/f1c276f7e26b29063d2ec31bc85d6627fd07d38a/.evergreen/config_generator/components/compile_only.py#L60) to `Compile.call()` as it should be. This means the `REQUIRED_CXX_STANDARD` and `USE_STATIC_LIBS` environment variables were not being applied as they should.

While addressing this bug, took the opportunity to extend compiler coverage to better match that of the C Driver's [std-matrix](https://github.com/mongodb/mongo-c-driver/blob/74c039a1ebbb6cee4c127e379501ee3d66ea7dd6/.evergreen/config_generator/components/c_std_compile.py) tasks, including coverage with `/std:c++latest` (which _is_ supported on EVG distros unlike `/std:clatest`). Rather than separate tasks for shared and static libraries, the `BUILD_SHARED_AND_STATIC_LIBS` option is set to `ON` instead.

Both of these updates revealed several issues which are addressed by this PR, as described below.

---

Identified a few more references to C Driver hedged reads API in test code that were missed in https://github.com/mongodb/mongo-cxx-driver/issues/1351. Applied suppressions accordingly.

---

The [false-positive `-Wmaybe-uninitialized`](https://github.com/mongodb/mongo-cxx-driver/blob/f1c276f7e26b29063d2ec31bc85d6627fd07d38a/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/options/transaction.hh#L109) warning for `optional<T>` made a return with GCC 8 and newer. Although it was meant to be suppressed by https://github.com/mongodb/mongo-cxx-driver/pull/968, the same false-positive warnings were (inconsistently) observed in other functions returning an `optional<T>` as well, e.g.:

```
uri.cpp: In member function 'std::optional<int> mongocxx::v_noabi::uri::wait_queue_timeout_ms() const':
uri.cpp:152:17: error: '<anonymous>' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         return {};
                 ^
```

To avoid this issue for good, changed all instances of the problematic `return {};` pattern to return `bsoncxx::v_noabi::stdx::nullopt;` instead.

---

Extending compiler coverage to GCC 4.8 (on rhel76) revealed the following compilation error:

```
error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]
  BSONCXX_PRIVATE_WARNINGS_DISABLE(GNU("-Wignored-attributes"));
```

This required adding the `defined(__GNUC__) && (__GNUC__ >= 6) && !defined(__clang__)` condition to avoid suppressing an unsupported warning flag for GCC versions older than GCC 6.

---

Following https://github.com/mongodb/mongo-cxx-driver/pull/1329, there should have been no more alignment-related issues in the C++ Driver. However, when removing [an overlooked `-Wno-aligned-new`](https://github.com/mongodb/mongo-cxx-driver/blob/f1c276f7e26b29063d2ec31bc85d6627fd07d38a/.evergreen/scripts/compile.sh#L158) flag in the compile script, more instances of this warning were identified with Clang 13 (on rhel90), e.g.:

```
error: passing 4-byte aligned argument to 8-byte aligned parameter 1 of 'bson_get_data' may result in an unaligned pointer access [-Werror,-Walign-mismatch]
    return {bson_get_data(failure), failure->len};
                          ^
```

Further investigation revealed a common pattern for all instances of this warning: the use of `auto` to deduce a `bson_t const*`. This suggests `auto` is also susceptible to the "attributes are ignored" issue which affects template parameters as described in https://github.com/mongodb/mongo-cxx-driver/pull/1329. Changing all instances of such `auto` variables to explicit `bson_t` addressed these warnings.

kevinAlbs (approved):
The compile matrix improvements (easier visibility to which compiler version is tested) are very much appreciated. LGTM

eramongodb:
Deferred attempts to resolve rhel76 task failure to https://github.com/mongodb/mongo-cxx-driver/pull/1384.

vector-of-bool (approved):
LGTM

PR #1388, opened 2025-04-21
CXX-3002 update release instructions to support signed release tags

eramongodb:
This PR attempts to implement the "signed release tag" requirement of the new Repository and Commit Security policy. This is primarily accomplished by the new `garasign_release_tag.sh` script, which mirrors the existing `garasign_dist_file.sh` script.

The `garasign-git` Artifactory image provides an environment where `git` commands may be executed using the same Release Signing Key provided by `garasign-gpg`, which is currently used by `garasign_dist_file.sh` to sign the release tarball. This allows us to create a GPG-signed release tag using `git tag --sign` in a manner which satisfies the policy requirements while (hopefully) minimizing disruption to the release process or regular development. (We do not want to impose any GPG key management overhead if possible.)

> [!NOTE]
> Unfortunately, it seems the `user.name` and `user.email` Git config options must be manually set for the command to succeed despite the information being present in the signing key specified via `--local-user <KeyID>`.

The `git tag` command is preceeded by a `gpg --list-keys <KeyID>` to validate the key we intend to use is indeed provided by the `garasign-git` environment. The expected output of the `garasign-git` command looks as follows (the "Success!" is from the implicit `gpgloader` command preceeding evaluation of `PLUGIN_COMMANDS`):

```
Success! Check "/root/.gnupggrs/keysinfo.txt" for info about the available keys.

gpg: checking the trustdb
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
pub   rsa4096 2024-04-30 [SC]
      DC7F679B8A34DD606C1E54CAC4FC994D21532195
uid           [ultimate] MongoDB C++ Release Signing Key <packaging@mongodb.com>
```

The `git tag` command is followed by a sanity-check that the newly-created git tag is indeed signed as it should be with the GPG key that is provided by https://pgp.mongodb.com/. To avoid interference with local keyrings, the `GNUPGHOME` environment variable is used to direct `git verify-tag` to use the temporary (otherwise empty) keyring.

Mirroring usage of the existing `garasign_dist_file.sh` script, the new `garasign_release_tag.sh` script is invoked by `./etc/make_release.py`. This means `make_release.py` now handles the creation of the (signed) release tag rather than the user. The release instructions have been updated accordingly.

---

This PR also contains the following drive-by improvements/fixes:

- Update the `garasign-gpg` image before use (as in [silkbomb](https://github.com/mongodb/mongo-cxx-driver/blob/f6fef68477e3f75a946a6d8180c2c59d12307b56/etc/releasing.md#sbom-lite) commands).
- Fix the detection of C Driver v2 headers (from https://github.com/mongodb/mongo-cxx-driver/pull/1379) to account for embedded API version numbers in the include path.

### DIFF
--- a/etc/garasign_dist_file.sh
+++ b/etc/garasign_dist_file.sh
@@ -52,6 +52,9 @@ plugin_commands=(
   artifactory.corp.mongodb.com/release-tools-container-registry-local/garasign-gpg
 
 # Validate the signature file works as intended.
-keyring="$(mktemp)"
-curl -sS https://pgp.mongodb.com/cpp-driver.pub | gpg -q --no-default-keyring --keyring "${keyring:?}" --import -
-gpgv --keyring "${keyring:?}" "${dist_file_signed:?}" "${dist_file:?}"
+(
+  GNUPGHOME="$(mktemp -d)"
+  export GNUPGHOME
+  curl -sS https://pgp.mongodb.com/cpp-driver.pub | gpg -q --no-default-keyring --import -
+  gpgv "${dist_file_signed:?}" "${dist_file:?}"
+)
--- a/etc/garasign_dist_file.sh
+++ b/etc/garasign_dist_file.sh
@@ -38,7 +38,7 @@ dist_file_signed="${dist_file:?}.asc"
 "${launcher:?}" login --password-stdin --username "${ARTIFACTORY_USER:?}" artifactory.corp.mongodb.com <<<"${ARTIFACTORY_PASSWORD:?}"
 
 # Ensure latest version of Garasign is being used.
-podman pull artifactory.corp.mongodb.com/release-tools-container-registry-local/garasign-gpg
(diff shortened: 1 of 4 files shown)


eramongodb:
Note: this PR proposes using "MongoDB C++ Release Signing Key <packaging@mongodb.com>" as the identity of the release tag creator, which matches the userid and email of the signing key itself. We can make accomodations to support using the developer's username+email (consistent with their usual git identity), but I do not think this is strictly necessary atm (+ the _commit_ to which the signed release tag points should be signed using the developer's usual git identity anyways).

kevinAlbs (approved):
LGTM with a small fix.

Moving the tagging steps into scripts seems like a bonus simplification to the release process.

PR #1389, opened 2025-04-22
Use Jira Personal Access Token (PAT)

kevinAlbs:
Update release scripts and instructions to use a Jira Personal Access Token.

# Motivation
The access token for the Jira service account used in `make_release.py` expired:
```
$ python ./etc/make_release.py --dry-run r4.1.0
[...]
response text = oauth_problem=token_expired
```

This token was for the service account created in [OFFICEIT-32460](https://jira.mongodb.org/browse/OFFICEIT-32460). Rather than requesting a new token, this PR proposes switching to use a Jira Personal Access Token (PAT). Using a PAT was recommended [in slack](https://mongodb.slack.com/archives/C0UK2L7RC/p1745325102967929?thread_ts=1745259700.741349&cid=C0UK2L7RC) and appears to have several advantages: the releasing developer can create the PAT, set automatic expiration, and revoke. The PAT is created with the same permissions as the developer.

### DIFF
--- a/etc/make_release.py
+++ b/etc/make_release.py
@@ -51,7 +51,6 @@ import click # pip install Click
 from git import Repo # pip install GitPython
 from github import Github # pip install PyGithub
 from jira import JIRA # pip install jira
-import oauthlib.oauth1
 
 if sys.version_info < (3, 0, 0):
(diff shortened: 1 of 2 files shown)


eramongodb (approved):
Nice improvement!

PR #1391, opened 2025-04-22
Use looseversion instead of deprecated distutils

eramongodb:
This PR addresses the following warning observed when running the `make_release.py` script:

```
etc/make_release.py:42: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.version import LooseVersion
```

This is a followup to https://github.com/mongodb/mongo-cxx-driver/pull/1280 and https://github.com/mongodb/mongo-cxx-driver/pull/1244, which applied a `<3.12` Python version restriction to workaround the aforementioned removal of the `distutils` package. This PR proposes using the [looseversion](https://pypi.org/project/looseversion/) package instead to obtain `LooseVersion`. This package was already added to the list of `make_release.py` dependencies in https://github.com/mongodb/mongo-cxx-driver/pull/1244 but was left unused (possibly overlooked.

This permits the removal of the Python 3.12 version restriction, which in turn revealed the following warning:

```
etc/make_release.py:593: SyntaxWarning: invalid escape sequence '\w'
  token_re = re.compile('^(?:Token - )?(?P<tok>\w{40}).*$')
```

This is apparently due to 3.12 [promoting](https://docs.python.org/3/reference/lexical_analysis.html#index-24) unrecognized escape sequences to a SyntaxWarning (and eventually a SyntaxError). I believe a simple addition of an escape `\` -> `\\` is sufficient in this case.

These changes (should) allow for compatibility with Python 3.10 through Python 3.13. ("Should" as the full script is not yet tested without `--dry-run`.)


etc/make_release.py
```diff
@@ -621,7 +621,7 @@ def read_github_creds(github_token_file):
     Read the GitHub token from the specified file and return it as a string.
     """
 
-    token_re = re.compile('^(?:Token - )?(?P<tok>\w{40}).*$')
+    token_re = re.compile('^(?:Token - )?(?P<tok>\\w{40}).*$')
```
kevinAlbs, line 624:
```suggestion
    token_re = re.compile(r'^(?:Token - )?(?P<tok>\w{40}).*$')
```

Consider using a raw string to treat the backslash as a literal character. Raw strings appear frequently in https://docs.python.org/3/library/re.html.

kevinAlbs (approved):

PR #1399, opened 2025-05-05
Address -Wsign-conversion warnings with Clang

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1356, where a `-Wsign-conversion` Clang warning due to the following expression was overlooked:

```cpp
0xFFu << (-element_end.bit & 7u)
```

`element_end.bit` is a `std::uint8_t`. The unary negation operator implicitly promotes the result to an `int`, whose signedness is inconsistent with `7u` (warning: `signed & unsigned`). The literal is made a signed `int` for consistency. This should not impact the behavior of the overall expression ([tested exhaustively](https://godbolt.org/z/M3Tvfe68h)). (Updated link: static assert `unsigned << signed -> unsigned` (after promotion) + test `UINT8_MAX` as well.)


kevinAlbs (approved):
LGTM. The exhaustive test to double check is appreciated.

PR #1402, opened 2025-05-13
CXX-2745 add immortal.hh and type_traits.hh

eramongodb:
Related to CXX-2745. Precursor changes to CXX-3232.

Adds two new internal bsoncxx components which are used by upcoming v1 interfaces.

---

The `bsoncxx::immortal<T>` class is used to define error category objects in a manner that avoids generating exit-time destructors (hence, "immortal") as can be diagnosed by the `-Wexit-time-destructors` Clang warning.

The pattern will look as follows:

```cpp
std::error_category const& example() {
    class type final : public std::error_category { ... };

    // The error category object instance. This is never destroyed!
    static bsoncxx::immortal<type> const instance;

    return instance.value();
}
```

---

Several new type traits are introduce to ensure consistency and correctness of v1 interfaces via static assertions. These type traits are derived from patterns observed in v_noabi interfaces and are enforced by v1 interfaces in a more rigorous manner.

The new type traits are:

- `is_semitrivial<T>`:
    - `T` is trivial to copy, move, and destroy ("like an `int`").
    - Trivial default constructibility is excluded (hence "semi") due to most types requiring non-trivial construction (to enforce class invariants) even when other special member functions are all trivial (invariants are entirely enforced by initialization).
- `is_semiregular<T>` and `is_regular<T>`:
    - `T` has value semantics ("like an `int` or `std::string`") and is (optionally) equality-comparable.
    - `is_semitrivial<T>` implies `is_semiregular<T>`.
- `is_nothrow_moveable<T>`:
    - `T` does not throw on destruction, move construction, or move assignment ("like a `std::unique_ptr<T>`").
    - This is meant to roughly approximate the requirements for [`is_nothrow_relocatable<T>` in C++26](https://wg21.link/p2786r13#is_nothrow_relocatable-trait) without misleadingly implying actual nothrow relocatability (which is complicated to accurately specify). This type trait is primarily to catch missing `noexcept` specifiers and highlight immovable types.
    - `is_semitrivial<T>` implies `is_nothrow_moveable<T>`.

All class types in bsoncxx and mongocxx are expected to (eventually) satisfy _semiregular_ and _nothrow moveable_ at a minimum to ensure consistent user-friendly behavior (e.g. `bsoncxx::v_noabi::document::value` is notably [not semiregular](https://jira.mongodb.org/browse/CXX-939)). Some types may additionally support semitrivialty. Some types may  additionally support regularity (equality comparison). If there are any circumstances where nothrow moveability is deliberately violated (potentially-throwing move operations or immovability), we should expect such instances to be highlighted with a negated static assertion ("we do something unusual here!") such as for `mongocxx::v_noabi::instance` (an immovable class type).

> [!IMPORTANT]
> Move constructors and move assignment operators which may allocate memory are still declared `noexcept`. `std::bad_alloc` exceptions are _**NOT**_ accounted for by nothrow moveability, as we do not support recovering from out-of-memory situations in either the C++ or C Driver libraries.

### DIFF
--- a/src/bsoncxx/lib/bsoncxx/private/immortal.hh
+++ b/src/bsoncxx/lib/bsoncxx/private/immortal.hh
@@ -20,6 +20,19 @@
 
 namespace bsoncxx {
 
+//
+// Used to construct an object whose destructor is never invoked (hence, "immortal").
+//
+// This is primarily to avoid generating exit-time destructors for error category objects:
+//
+// ```cpp
+// std::error_category const& name() noexcept {
+//     class type final : public std::error_category { ... };
+//     static bsoncxx::immortal<type> const instance;
+//     return instance.value();
+// }
+// ```
+//
 template <typename T>
 class immortal {
    private:
--- a/src/bsoncxx/lib/bsoncxx/private/immortal.hh
+++ b/src/bsoncxx/lib/bsoncxx/private/immortal.hh
@@ -46,7 +46,7 @@ class immortal {
     immortal& operator=(immortal const&) = delete;
 
     template <typename... Args>
-    immortal(Args&&... args) {
+    explicit immortal(Args&&... args) {
         new (_storage) T(BSONCXX_PRIVATE_FWD(args)...);
     }
 
--- a/src/bsoncxx/lib/bsoncxx/private/immortal.hh
+++ b/src/bsoncxx/lib/bsoncxx/private/immortal.hh
@@ -1,0 +1,49 @@
+// Copyright 2009-present MongoDB, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <bsoncxx/v1/detail/macros.hpp>
+
+#include <new>
+
+namespace bsoncxx {
+
+template <typename T>
+class immortal {
+   private:
+    alignas(T) unsigned char _storage[sizeof(T)];
+
+   public:
+    ~immortal() = default;
+    immortal(immortal&&) = delete;
+    immortal& operator=(immortal&&) = delete;
+    immortal(immortal const&) = delete;
+    immortal& operator=(immortal const&) = delete;
+
+    template <typename... Args>
+    immortal(Args&&... args) {
+        new (_storage) T(BSONCXX_PRIVATE_FWD(args)...);
+    }
+
+    T const& value() const {
+        return *reinterpret_cast<T const*>(_storage);
+    }
+
+    T& value() {
+        return *reinterpret_cast<T*>(_storage);
+    }
+};
+
+} // namespace bsoncxx
--- a/src/bsoncxx/lib/bsoncxx/private/type_traits.hh
+++ b/src/bsoncxx/lib/bsoncxx/private/type_traits.hh
@@ -1,0 +1,78 @@
+// Copyright 2009-present MongoDB, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <bsoncxx/v1/detail/type_traits.hpp>
+
+namespace bsoncxx {
+
+template <typename LHS, typename RHS>
+using is_assignable_from_expr = decltype(std::declval<LHS>() = std::declval<RHS>());
+
+// Approximates the `assignable_from` concept in C++20.
+template <typename LHS, typename RHS>
+struct is_assignable_from : detail::conjunction<
+                                std::is_lvalue_reference<LHS>,
+                                std::is_same<detail::detected_t<is_assignable_from_expr, LHS, RHS>, LHS>> {};
+
+// Equivalent to `is_assignable_from` but also requires noexceptness.
+template <typename LHS, typename RHS>
+struct is_nothrow_assignable_from
+    : detail::bool_constant<
+          is_assignable_from<LHS, RHS>::value && noexcept(std::declval<LHS>() = std::declval<RHS>())> {};
+
+// Approximates the `movable` concept in C++20.
+template <typename T>
+struct is_movable : detail::conjunction<
+                        std::is_object<T>,
+                        std::is_move_constructible<T>,
+                        is_assignable_from<T&, T>,
+                        detail::is_swappable<T>> {};
+
+// Equivalent to `is_moveable` but additionally requires noexceptness.
+template <typename T>
+struct is_nothrow_moveable : detail::conjunction<
+                                 std::is_object<T>,
+                                 std::is_nothrow_move_constructible<T>,
+                                 is_nothrow_assignable_from<T&, T>,
+                                 detail::is_nothrow_swappable<T>> {};
+
+// Approximates the `copyable` concept in C++20.
+template <typename T>
+struct is_copyable : detail::conjunction<
+                         std::is_copy_constructible<T>,
+                         is_movable<T>,
+                         is_assignable_from<T&, T&>,
+                         is_assignable_from<T&, T const&>,
+                         is_assignable_from<T&, T const>> {};
+
+// Approximates the `semiregular` concept in C++20.
+template <typename T>
+struct is_semiregular : detail::conjunction<is_copyable<T>, std::is_default_constructible<T>> {};
+
+// Approximates the `regular` concept in C++20.
+template <typename T>
+struct is_regular : detail::conjunction<is_semiregular<T>, detail::is_equality_comparable<T>> {};
+
+// Equivalent to `is_trivial` without requiring trivial default constructibility.
+template <typename T>
+struct is_semitrivial : detail::conjunction<
+                            std::is_trivially_destructible<T>,
+                            std::is_trivially_move_constructible<T>,
+                            std::is_trivially_move_assignable<T>,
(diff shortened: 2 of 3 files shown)


src/bsoncxx/lib/bsoncxx/private/type_traits.hh
```diff
@@ -0,0 +1,78 @@
...
+struct is_regular : detail::conjunction<is_semiregular<T>, detail::is_equality_comparable<T>> {};
+
+// Equivalent to `is_trivial` without requiring trivial default constructibility.
+template <typename T>
+struct is_semitrivial : detail::conjunction<
+                            std::is_trivially_destructible<T>,
+                            std::is_trivially_move_constructible<T>,
+                            std::is_trivially_move_assignable<T>,
+                            std::is_trivially_copy_constructible<T>,
+                            std::is_trivially_copy_assignable<T>> {};
```
vector-of-bool, line 76: Is "semitrivial" a (proposed) standard trait? I've not heard the term, but it is reasonably useful.

src/bsoncxx/lib/bsoncxx/private/immortal.hh
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include <bsoncxx/v1/detail/macros.hpp>
+
+#include <new>
+
+namespace bsoncxx {
+
+template <typename T>
+class immortal {
```
vector-of-bool, line 24: Add a doc comment on what this class template does any why it might want to be used, and how it should not be used (e.g. declaring a non-`static` `immortal<T>` is potentially bad).

src/bsoncxx/lib/bsoncxx/private/immortal.hh
```diff
@@ -0,0 +1,49 @@
...
+        new (_storage) T(BSONCXX_PRIVATE_FWD(args)...);
+    }
+
+    T const& value() const {
+        return *reinterpret_cast<T const*>(_storage);
+    }
+
+    T& value() {
+        return *reinterpret_cast<T*>(_storage);
+    }
```
vector-of-bool, line 46: Add (conditional) `noexcept` here. Possibly an `explicit` on the constructor?

vector-of-bool (approved):
Sorry to delay. Only minor comments, but otherwise looks good.

src/bsoncxx/lib/bsoncxx/private/immortal.hh
eramongodb: I'd meant to discuss this in the upcoming v1 ABI namespace declarations PR, but will initiate the discussion here since you mentioned it here. (CC @kevinAlbs.)

---

The v1 ABI interface intends to implement "Policy Statement E - Minimal `noexcept` (MIN)" as specified by [P3005R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3005r0.pdf) (section 4.1.1.1) going forward in a deliberate break from consistency with existing interfaces.

Quoting the policy verbatim:

> Only those functions with a sanctioned need (e.g., practical application of the `noexcept` operator) to be marked `noexcept` shall be marked unconditionally `noexcept`. Specifically, move constructors, move assignment operators, and the `swap` function have all been shown to be used in algorithms where significant algorithmic improvements can be made when these functions can be determined, at compile time, to not throw. No other functions shall be marked `noexcept` absent an update to this policy.

This policy can be summarized as:

> Use `noexcept` only for:
>
> - move constructors,
> - move assignment operator overloads, and
> - `swap` overloads
>
> when they are guaranteed to not throw.

The reason for this is to avoid burdening code reviews and future maintainability of the codebase with `noexcept` correctness evaluations to support hypothetical use cases which are likely irrelevant to actual usage of the C++ Driver libraries. Given the expected needs and usage of the C++ Driver, I believe using `noexcept` beyond the limited known-and-well-established cases listed above is an unnecessary complexity which will continue to challenge code reviews now and going forward should we adopt an alternative policy.

> [!NOTE]
> As an aside, we also effectively implement "No Lakos Rule on C Compatibility (NLC)" (section 4.1.1.2) due to the C Driver not supporting exceptions (e.g. the [log handler callback](https://github.com/mongodb/mongo-cxx-driver/blob/079af6c4a04f67c8f96cb100a77cf1e25eeabe9e/src/mongocxx/include/mongocxx/v_noabi/mongocxx/logger.hpp#L79)), but we haven't yet propagated this requirement into the public API [for APM callbacks](https://github.com/mongodb/mongo-cxx-driver/blob/079af6c4a04f67c8f96cb100a77cf1e25eeabe9e/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/options/apm.hh#L34) yet, whose function types _should_ be marked `noexcept`, but `std::function<T>` [deficiencies](https://wg21.link/p2548r6) unfortunately limit our available options here.

For example, the suggestion to apply `noexcept` to `immortal<T>::value()` here is consistent with both "Policy Statement B - Basic Maximal `noexcept` (MAX)" (aka "use `noexcept` to document and enforce an in-contract nothrow guarantee") and "Lakos Rule (LAK)" (aka "use `noexcept` given a wide-contract nothrow guarantee"). (We do not support allocator-awareness, so NLM is irrelevant.) The reason I propose we avoid either of these policies going forward (in a deliberate break in consistency with prior/existing interfaces, which is an understandably inconsistent mix of both over time) is:

- MAX: I believe it is challenging to review and maintain that the codebase truly does not throw _any_ exceptions when in contract (even discounting `std::bad_alloc`, which we do not support). The existing `v_noabi` interfaces are under-documented regarding whether a given function may throw an exception, and there are many functions whose implementations _should_ be throwing an exception for erroneous runtime conditions (i.e. missing return value checks) but are not. A codebase which diligently ensured all nothrow functions are consistently denoted with `noexcept` from the start that also fully accounts for thorough error handling might be able to get away with this policy; I do not think the C++ Driver is such a codebase. I believe adopting the MAX policy will risk too much opportunity for over-eager and (arguably) incorrect application of `noexcept` that may result in a poor user experience (irrecoverable termination as opposed to an unexpected but recoverable exception).
- LAK: for similar reasons, I believe this is unnecessarily challenging to review and maintain for our codebase. This is arguably even _more_ complicated than MAX due to requiring a thorough understanding and documentation of narrow vs. wide contracts. This is difficult to apply to the C++ Driver due to the lack of thorough precondition assertions, validation, or documentation (let alone exceptions). I believe it will be difficult to consistently and correctly apply LAK for all but the most trivial functions as the codebase continues to evolve and missing runtime checks continue to be identified and addressed.
- Allow Conditional `noexcept` (CNN): I believe this is overkill for anything beyond internal highly-generic utilities (e.g. `detail/compare.hpp`) and does not apply to most of the codebase (including `immortal<T>`; who/when/why will anyone care whether its ctor or `.value()` is `noexcept` or not?).

Therefore, I propose we adopt the MIN `noexcept` policy going forward for the C++ Driver. It is the easiest to identify, review, and maintain (limited to nothrow moves and nothrow swap) and is the most forgiving and permissive policy w.r.t. respecting API backward compatibility guarantees that will allow us to continue improving the state of our error-handling and precondition assertion behavior and documentation with minimal resistance or premature commitment to excessive nothrow guarantees. Under typical circumstances, we should not impose or deal with the cognitive burden of noexcept-termination on ourselves or our users beyond the bare minimum (move+swap and C compat).

Please let me know if you have any thoughts or opinions against adopting this proposal.

eramongodb:
Re-requesting reviews to solicit feedback regarding the adoption of [a new noexcept policy](https://github.com/mongodb/mongo-cxx-driver/pull/1402#discussion_r2096510603) going forward.

src/bsoncxx/lib/bsoncxx/private/immortal.hh
kevinAlbs: Thank you for the detailed write-up. I very much support limiting `noexcept` to the most practical value. I expect applying `noexcept` widely adds complexity and risks incorrectly marking a function that really throws.

> noexcept can be limited to the bare minimum (move+swap and C compat).

As I understand it, the reasoning to use `noexcept` in move+swap and C Compat differ (but I agree with both):
- move+swap: use `noexcept` for practical performance improvement.
- C Compat: use `noexcept` on C++ functions called by C to ensure defined behavior. (i.e. terminate if the C++ function unexpectedly throws).

kevinAlbs (approved):

src/bsoncxx/lib/bsoncxx/private/immortal.hh
eramongodb: > move+swap: use noexcept for practical performance improvement.

Correct, but specifically/importantly, for practical performance improvement due to _known_ use of the `noexcept` operator. Paraphrasing P3005R0 (page 13):

> Nearly 15 years of experience with `noexcept` [by the committee] has revealed that applying `noexcept` superfluously [...] can be highly detrimental to writing safe and reliable software. Therefore, [we] must scrupulously reconsider when `noexcept` is truly needed, which is when a function must — not just theoretically but in practice — have the `noexcept` operator applied to it (i.e., in a generic context) to enable an algorithmically better approach to be taken by a templated function. Up to now, only move construction and move assignment are used in this way by generic functions in the Standard Library itself, while swap is similar enough to potentially be useful in algorithms from other libraries. No other functions in any library have ever been known to demonstrate such value. Hence, until an existence proof can be demonstrated [to us], the argument can be made that no other functions shall be candidates for being marked as unconditionally `noexcept`.

---

> C compat: use `noexcept` on C++ functions called by C to ensure defined behavior (i.e. terminate if the C++ function unexpectedly throws).

Correct, as described by "No Lakos Rule on C Compatibility (NLC)" on page 14:

> That a C++ library function, such as those in the `<atomics>` API, intended to mirror a C library function could never throw was perfectly obvious. Hence, the original LEWG policy on the use of noexcept included an exemption for library functions designed for compatibility with C code. [...] A library function designed for compatibility with C code that does not throw in contract may always be marked as unconditionally `noexcept`, even when it has a narrow contract.

In our case, as described in https://github.com/mongodb/mongo-cxx-driver/pull/1015, rather than the (re)declaration of a _C++ equivalent_ of a C library function, we are dealing with _callback functions_ passed to the C library (mongoc) which may potentially throw an exception. It is undefined behavior for an exception to be thrown from within such a callback, as the C library is not designed to handle any exceptions being thrown from within it (that is, all callbacks should effectively behave like a C function). Therefore, any such callback function should be unconditionally `noexcept` as a matter of correctness and safety, ensuring what would otherwise be undefined behavior is instead well-defined program termination. (This is _technically_ not NLC because we are not adopting LAK, but it's useful to refer to it as such anyways.)

vector-of-bool (approved):

src/bsoncxx/lib/bsoncxx/private/immortal.hh
eramongodb: > I'd recommend adding it somewhere permanent (e.g. in the repo or in some internal docs) for future reference.

SGTM. Will include it in the updates to the [coding guidelines](https://github.com/mongodb/mongo-cxx-driver/blob/master/etc/coding_guidelines.md) document.

PR #1406, opened 2025-05-19
CXX-3290 fix static pkgconfig requirements

kevinAlbs:


examples/projects/mongocxx/pkg-config/static/build.sh
```diff
@@ -16,6 +16,14 @@ set -o pipefail
     exit 1
   )
 
+# Sanity-check that static libmongoc is required. Regression test for CXX-3290.
+(pkg-config --print-requires libmongocxx-static | grep -- mongoc2-static || (
+  (
+    echo "Expected mongoc2-static to be required" >&2
+    exit 1
+  )
+))
```
eramongodb, line 25:
```suggestion
if ! pkg-config --print-requires libmongocxx-static | grep -q "mongoc2-static"; then
  echo "Expected mongoc2-static to be required" >&2
  exit 1
fi
```
Lot of subshells; this can be reduced to a straightforward conditional.

Use `-q` to silence grep matches + per grep manpages:

> If no `FILE` is given [...] nonrecursive searches read standard input.

eramongodb (changes requested):

examples/projects/mongocxx/pkg-config/static/build.sh
kevinAlbs: Thank you for the suggestion. Updated.

eramongodb (approved):

vector-of-bool (approved):
LGTM

PR #1403, opened 2025-05-13
CXX-2745 refactor test directory structure by ABI namespace

eramongodb:
Related to CXX-2745. Precursor changes to CXX-3233.

Refactors the test directory structure and components to support the introduction of v1 test components by better distinguishing tests by ABI namespace. Additionally applies precursor changes to improve the state of undesirable transitive include dependencies due to headers such as `<bsoncxx/test/catch.hh>`.

As a drive-by improvement, Catch2 is updated to the latest release (v3.8.1). This addresses some Catch2-internal compiler warnings as well as [a bug](/catchorg/Catch2/issues/2944) which may affect v1 tests on Windows due to use of `StringMaker<std::chrono::time_point<Clock>>` for `v1::types::b_date` stringification (as part of CXX-3233).

---

The test directory components are refactored to mirror the structure of the library components which they test (with an exception for the v_noabi directory which is flattened):

```
src/bsoncxx/
├── lib/bsoncxx/
│   ├── private/
│   │   ├── make_unique.hh
│   │   └── ...
│   ├── v1/
│   │   ├── document/
│   │   │   ├── view.cpp
│   │   │   └── ...
│   │   └── ...
│   └── v_noabi/bsoncxx/
│       ├── document/
│       │   ├── view.cpp
│       │   └── ...
│       └── ...
└── test/
    ├── private/
    │   ├── make_unique.test.cpp
    │   └── ...
    ├── v1/
    │   ├── document/
    │   │   ├── view.cpp
    │   │   └── ...
    │   └── ...
    └── v_noabi/
        ├── document/
        │   ├── view.cpp
        │   └── ...
        └── ...
```

Test-specific components reside outside of ABI directories to avoid associating them with library components, e.g. `<bsoncxx/test/catch.hh>`, `<mongocxx/test/spec/util.hh>`, etc.

---

[...]

PR #1407, opened 2025-05-21
CXX-3294 Relocate all remaining Manual Doc Page examples into the examples directory

eramongodb:
Resolves CXX-3294. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1403, which broke downstream Manual Doc pages due to several example code blocks that still remained under the `test` directory (these are identified by `// (Start|Begin|End) <name>` comments). All such remaining code blocks are copied into equivalent files under the `examples/mongocxx/mongodb.com` directory for consistency and to inform future maintainers about their stability requirement by downstream Manual Doc pages. Although the comments under the `test` directory are removed, the code itself remains to preserve code coverage as part of the test suite.

Catch2 assertions such as `CHECK(cond)` are converted into `if (!cond) { throw std::logic_error{...}; }` for consistency with other example code. The rest of the code block contents and structure is otherwise preserved as much as possible with their prior form as `SECTION()` blocks.

As a drive-by fix/improvement, the requirement that the replica set against which the examples are run is named `repl0` (with `replicaSet=repl0`) is removed. AFAIK no such expectation or requirement is necessary for example code.


src/mongocxx/test/v_noabi/change_streams.cpp
```diff
@@ -457,20 +457,17 @@ TEST_CASE("Documentation Examples", "[change_stream]") {
     }};
 
     SECTION("Example 1") {
```
kevinAlbs, line 459: >  the code itself remains to preserve code coverage as part of the test suite.

Suggest removing the doc examples from tests to avoid going out-of-sync. I expect the main purpose is documentation rather than testing. And the examples are [run on most Evergreen configurations](https://github.com/mongodb/mongo-cxx-driver/blob/079af6c4a04f67c8f96cb100a77cf1e25eeabe9e/.evergreen/scripts/test.sh#L314-L320).

examples/mongocxx/mongodb.com/change_streams_examples.cpp
```diff
@@ -0,0 +1,218 @@
...
+        change_streams_example_3,
+        change_streams_example_4,
+    };
+
+    for (auto const& example : examples) {
+        mongocxx::pool pool{mongocxx::uri{}};
+
+        mongocxx::client const conn{mongocxx::uri{}};
+
+        auto const db = conn["documentation_examples"];
```
kevinAlbs, line 184:
```suggestion
```

Unused.

examples/mongocxx/mongodb.com/change_streams_examples.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    std::function<void(mongocxx::collection)> const examples[] = {
+        change_streams_example_1,
+        change_streams_example_2,
+        change_streams_example_3,
+        change_streams_example_4,
+    };
+
+    for (auto const& example : examples) {
+        mongocxx::pool pool{mongocxx::uri{}};
```
kevinAlbs, line 180:
```suggestion
    mongocxx::pool pool{mongocxx::uri{}};
    for (auto const& example : examples) {
```

Suggest creating pool outside of loop. Pool destruction signals and waits for monitoring threads to shut down. Timing `cmake --build cmake-build --target run-examples-mongocxx-mongodb.com-change_streams_examples` locally showed a decrease of 2.207s => 1.156s.

kevinAlbs (approved):
LGTM with minor suggestions.

PR #1409, opened 2025-05-27
CXX-3296 fix build if `GCC/GNU/Clang/MSVC` macro defined

kevinAlbs:
# Summary

Fix build if `GCC/GNU/Clang/MSVC` macros are already defined.

Tested with this [patch build](https://spruce.mongodb.com/patch/6835f4d420bcc600073f5976).

# Background

https://www.mongodb.com/community/forums/t/error-compiling-some-c-classes-with-driver-v4-0/321847 reports a build error due to the macro `GCC` being defined. I can reproduce the error by adding `#define GCC 1` before including C++ driver headers:
```
/Users/kevin.albertson/code/cxx-bootstrap/install/mongo-cxx-driver-r4.1.0/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp:185:9: error: use of undeclared identifier 'BSONCXX_PRIVATE_WARNINGS_DISABLE_IMPL_FOR_1'
        BSONCXX_PRIVATE_WARNINGS_DISABLE(GCC("-Wmaybe-uninitialized"));
        ^
```

This PR changes the `BSONCXX_PRIVATE_WARNINGS_DISABLE*` macros to avoid assuming `GCC/Clang/MSVC` are not already defined:

```c++
// Before:
BSONCXX_PRIVATE_WARNINGS_DISABLE(GCC("-Wmaybe-uninitialized"));
// After:
BSONCXX_PRIVATE_WARNINGS_DISABLE_FOR_GCC("-Wmaybe-uninitialized");
```

This PR assumes the `BSONCXX_PRIVATE_*` macros are OK to rename in a non-major API release. The macros are in a public header, but documented as private (refer: [API Versioning](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/api-abi-versioning/api-versioning/#source-compatibility)).


eramongodb (commented):
Was adding macro selection suffixes (`Clang`, `GCC`, `GNU`, and `MSVC`) to the list of guarded macros considered?

```cpp
// #pragma push_macro("GCC")
// #pragma undef GCC
#include <bsoncxx/v1/detail/prelude.hpp>

... // Public header code.

// #pragma pop_macro("GCC")
#include <bsoncxx/v1/detail/postlude.hpp>
```

I do not think we reasonably expect any platforms or third party libraries to define any of these macros (as they are usually prefixed with at least one underscore `_`) in such a way that guarding them within our public headers would change library behavior. I would consider any users building the C++ Driver libraries with such macros defined to be out-of-scope (they should be user-defining such macros in _their_ code, not ours).

kevinAlbs:
> Was adding macro selection suffixes (`Clang`, `GCC`, `GNU`, and `MSVC`) to the list of guarded macros considered?

Thank you for the suggestion. That seems to fit the purpose of the macro guards and is a smaller change. Updated.

eramongodb (approved):

vector-of-bool (approved):
LGTM

PR #1416, opened 2025-06-12
Update bsoncxx validation examples for CDRIVER-5710 and CDRIVER-6017

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/2026 which addressed CDRIVER-5710 and CDRIVER-6017. The changes in https://github.com/mongodb/mongo-c-driver/commit/d184525513e5bf21f30810c00d6b0094aac8b00c affect a few in the C++ Driver which demonstrate use of `bsoncxx::v_noabi::validate()` including the expected error offsets. Because the offsets are now changed/fixed, these examples need to be updated to run successfully.

This PR bumps the C Driver version on Evergreen to 2.0.2 for testing only. It does not bump the minimum required C Driver version for downstream users. We can also raise the minimum required C Driver version for downstream users if that is preferable (CXX-3302), but I do not think it is necessary at this time.

Alternatively we can defer applying this PR until the minimum required C Driver version for downstream users is bumped instead (e.g. when C Driver 2.1.0 is released).

### DIFF
--- a/.evergreen/config_generator/components/funcs/install_c_driver.py
+++ b/.evergreen/config_generator/components/funcs/install_c_driver.py
@@ -14,7 +14,7 @@ from typing import Mapping
 # Only MONGOC_DOWNLOAD_VERSION needs to be updated when pinning to an unreleased commit.
 # If pinning to an unreleased commit, create a "Blocked" JIRA ticket with
 # a "depends on" link to the appropriate C Driver version release ticket.
-MONGOC_VERSION_MINIMUM = '2.0.0'
+MONGOC_VERSION_MINIMUM = '2.0.2'
 
 
 class InstallCDriver(Function):
--- a/.evergreen/generated_configs/functions.yml
+++ b/.evergreen/generated_configs/functions.yml
@@ -443,7 +443,7 @@ functions:
       type: setup
       params:
         updates:
-          - { key: mongoc_version_minimum, value: 2.0.0 }
+          - { key: mongoc_version_minimum, value: 2.0.2 }
     - command: subprocess.exec
       type: setup
       params:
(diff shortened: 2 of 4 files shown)


.evergreen/config_generator/components/funcs/install_c_driver.py
```diff
@@ -14,7 +14,7 @@
 # Only MONGOC_DOWNLOAD_VERSION needs to be updated when pinning to an unreleased commit.
 # If pinning to an unreleased commit, create a "Blocked" JIRA ticket with
 # a "depends on" link to the appropriate C Driver version release ticket.
-MONGOC_VERSION_MINIMUM = '2.0.0'
+MONGOC_VERSION_MINIMUM = '2.0.2'
```
kevinAlbs, line 17: > We can also raise the minimum required C Driver version for downstream users if that is preferable ([CXX-3302](https://jira.mongodb.org/browse/CXX-3302)), but I do not think it is necessary at this time.

I am open to debate, but I slightly prefer bumping the minimum required version to 2.0.2.

I see three distinct versions:
1. `MONGOC_VERSION_MINIMUM` is the tested version in CI.
2. `MONGOC_DOWNLOAD_VERSION` is the automatically downloaded version.
3. `MONGOC_REQUIRED_VERSION` is the required version.

I expect these do not strictly need to match, as is proposed in this PR. However, only bumping tested+downloaded means tests fail with the minimum required C driver. That may be unlikely to impact users, but bumping all to 2.0.2 avoids the possibility. The other benefit: upgrading to C++ driver 4.1.0 (with minimum C driver 2.0.2) is guaranteed to have the bug fix of CDRIVER-6017.

eramongodb: Done. Included a reference to CXX-3302 for tracking purposes.

PR #1425, opened 2025-07-16
Update release instructions to use Amazon ECR instead of Artifactory

eramongodb:
Resolves CXX-3265. Applies changes similar to those in https://github.com/mongodb/mongo-c-driver/pull/2058.

- Release instructions are updated to authenticate with and use Amazon ECR instead of Artifactory.
- Dockerfiles under `extras/docker` are updated to support setting a `DEFAULT_SEARCH_REGISTRY` environment variable (pass to `make`) which is forwarded as a Docker `default_search_registry` argument, similar to `default_search_registry` for the Earthfile in https://github.com/mongodb/mongo-c-driver/pull/2058. (Note: Earthfile syntax slightly differs from Dockerfile syntax.)

The `generate.py` script and related files has been drive-by improved for better consistency (e.g. trailing newlines) and with inline script metadata to support `uv run`.


extras/docker/alpine3.19/Dockerfile
```diff
@@ -60,8 +63,10 @@ RUN wget https://github.com/mongodb/mongo-cxx-driver/archive/refs/tags/r${MONGOC
...
     cmake --install mongocxx-build --config RelWithDebInfo --prefix /opt/mongocxx
 
-FROM artifactory.corp.mongodb.com/dockerhub/library/alpine:3.19
+FROM ${default_search_registry}/library/alpine:3.19
 
 RUN apk update && apk upgrade && apk add --no-cache openssl3 libstdc++ libc6-compat
 
 COPY --from=builder /opt/mongocxx /usr/local
+
+RUN true
```
kevinAlbs, line 72: Remove stray `RUN true`?

extras/docker/generate.py
```diff
@@ -54,14 +64,14 @@ def render_template(template_filename, context):
         "redhat-ubi-9.4": "microdnf upgrade -y && microdnf install -y openssl",
     }
     post_install_commands = {
-        "alpine3.19": "",
-        "bookworm": "\nRUN ldconfig\n",
-        "noble": "\nRUN ldconfig\n",
+        "alpine3.19": "\nRUN true",
```
kevinAlbs, line 67:
```suggestion
        "alpine3.19": "",
```

Remove unnecessary `RUN true`? I expect these changes were for newline consistency. Since a newline is later appended, I expect this can be empty.

.evergreen/scripts/sbom.sh
```diff
@@ -6,6 +6,7 @@ set -o pipefail
 : "${ARTIFACTORY_USER:?}"
 : "${ARTIFACTORY_PASSWORD:?}"
```
kevinAlbs: No longer used:

```suggestion
```

.evergreen/config_generator/components/sbom.py
```diff
@@ -23,31 +29,52 @@ class CustomCommand(BuiltInCommand):
...
+                ],
+                script='aws ecr get-login-password --region us-east-1 | podman login --username AWS --password-stdin 901841024863.dkr.ecr.us-east-1.amazonaws.com',
+            ),
+        ],
         bash_exec(
             command_type=EvgCommandType.TEST,
             working_dir='mongo-cxx-driver',
             include_expansions_in_env=[
                 'ARTIFACTORY_PASSWORD',
                 'ARTIFACTORY_USER',
```
kevinAlbs: No longer used:

```suggestion
```

kevinAlbs (approved):
LGTM with minor comments.

extras/docker/generate.py
eramongodb: Annoyingly, if the expansion is empty (prior value of `""`), the Dockerfile for `alpine3.19` ends up with a trailing newline:

```diff
  COPY --from=builder /opt/mongocxx /usr/local
+ (empty line)
  (EOF newline)
```

I could not find a way to coerce Jinja2 to avoid expanding `{{ post_install_commands }}` without an extra newline given a value of `""` or `None`. For typical Jinja2 expansions, one would use `%` (e.g. `{% var %}`), but that doesn't seem to apply to `{{ var }}` expansions... 😕

PR #1428, opened 2025-07-21
Use uvx for CMake and Server Toolchain for Ninja

eramongodb:
Updates the Evergreen configuration and scripts to use `uv` to obtain CMake binaries. This removes the need for the `find-cmake-old.sh`, `find-cmake-latest.sh` (C Driver), and `find-cmake-version.sh` (C Driver) helper scripts in the C++ Driver.

Also updates the EVG configuration to unconditionally use the Ninja generator instead of Unix Makefiles on non-Windows distros. However, instead of using `uv`, the Server Toolchain is used instead. The reasons for this are explained below.

This PR is intended to be a "simpler" preview of a followup PR which will apply similar changes to the C Driver.

---

First, this PR updates the `uv-installer.sh` script to obtain the current latest release of `uv`, bumping the `uv` version from [0.5.14](https://github.com/astral-sh/uv/releases/tag/0.5.14) (Jan 2025) to ~~[0.8.0](https://github.com/astral-sh/uv/releases/tag/0.8.0)~~ [0.8.2](https://github.com/astral-sh/uv/releases/tag/0.8.2) (July 2025). ~~Related checksums continue to be manually applied to the `uv-installer.sh` script.~~ A new `patch-uv-installer.sh` script is used to automatically patch the installer script to enable and verify checksums. ~~To assist with future updates, a Bash script which may be used to obtain the list of checksums is now documented as a comment within the installer script, copied here for reference:~~

```bash
for checksum in $(curl -sSL https://github.com/astral-sh/uv/releases/download/0.8.0/dist-manifest.json | jq -r '.releases[0].artifacts.[] | select(startswith("uv-") and (endswith(".zip.sha256") or endswith(".tar.gz.sha256")))'); do
  curl -sSL "https://github.com/astral-sh/uv/releases/download/0.8.0/${checksum}"
done
```

Update: the `download_checksums` helper function in `patch-uv-installer.sh` can be used to obtain the list of checksums for a given release.

> [!NOTE]
> Downloading the checksums at runtime would be convenient, but would also defeat the point of the checksum validation as a guard against corrupt or malicious downloads since the checksum values would come from the same potentially-corrupted or potentially-malicious source as the artifacts to be validated.

The `uv.lock` file is also updated with latest package upgrades by running `uv lock -U`.

---

The bulk of this PR is updating the Evergreen configuration and scripts to use `cmake` binaries provided by `uv` instead of the system-provided binaries (inconsistent availability and versions) or manual download strategies (i.e. via `find-cmake-latest.sh`). This is expected to improve the reliability, consistency, and ease of build tool acquisition and usage across EVG distros when the same pattern can be applied to other Python-packaged tools going forward.

Update: for compatibility with uv 0.8.1 and newer, which implemented stronger protections against bypassing ephemeral environments in [astral-sh/uv#14790](/astral-sh/uv/pull/14790) (which is exactly what the `uv run ... "command -v <name>"` pattern was doing), this PR now proposes using `uv tool install` with the `UV_TOOL_DIR` and `UV_TOOL_BIN_DIR` environment variables set to local directories. This should not be a significant performance penalty, as the uv _cache_ directory would still be used to populate the custom tool directories. However, these environment variables unfortunately require the `C:\path` form on Windows instead of `/cygdrive/c/path` (otherwise, no diagnostic or error (?!); instead, ignored in favor of the user's local bin directory 😢).

~~Rather than `uv tool install cmake` (which may modify the user's environment, e.g. by installing binaries to `$HOME/.local/bin`), this PR proposes using `uvx cmake` instead (`uvx` is a convenient short-hand for `uv tool run`). Unlike `uv tool install`, `uvx` does not modify the user's environment. When a full path to the binary is required (e.g. for CMake compatibility tests or as an argument to other scripts/tools), the lengthy `uv run --no-project --isolated --with cmake bash -c "command -v cmake"` command is required. There unfortunately does not yet seem to be a short-hand `uv tool` command to obtain a path to the uv-managed executable tool. Note that `uvx bash -c "..."` does not work due to `uvx` interpreting `bash` as a Python package.~~

> [!NOTE]
> Concerning flags to `uv run`:
>
> * `--no-project` prevents discovery and installation (`uv sync`) of a project before running the command, but may still reuse a virtual environment that is active or found in the current or parent directories.
> * `--isolated` prevents reuse of any virtual environment by always using an isolated virtual environment, but may discover and install (`uv sync`) a project within that isolated virtual environment before running the command.
>
> Both flags are required to ensure an isolated virtual environment _without_ any project detection and installation is used. However, neither flag prevent use of uv's thread-safe, append-only dependency cache, which is how the same `<name>` binary can be safely reused by direct reference as obtained via `command -v <name>` (unless someone unsafely modifies the cache via `uv cache` or direct filesystem operations).
>
> Although `uvx` has an `--isolated` flag, it is currently redundant pending resolution of [astral-uv/uv#7186](/astral-sh/uv/issues/7186). For now, the `uvx` command always uses an isolated virtual environment.

---

This PR initially attempted to use the same approach to obtain the `ninja` binary. Obtaining `ninja` via `uv` is simple; however, this method was greatly frustrated by unexpected integration issues. After much effort, I've given up using `uv`-provided Ninja binaries; instead, this PR proposes unconditionally using the binary provided by Server Toolchain v4, which is available on all our currently-used EVG distros.

The first complication is that there is no `CMAKE_MAKE_PROGRAM` _environment variable_ to go alongside the `CMAKE_GENERATOR` environment variable. This means the path to the Ninja binary (which requires the lengthy `uv run` command described above) is almost always necessary and has to be explicitly passed as a CLI argument during the CMake configuration step via `-D CMAKE_MAKE_PROGRAM=<path/to/ninja>`, which is _not_ automatically propogated to subprocess or other tools (e.g. `distcheck` target, `compile-libmongocrypt.sh` script, etc.). This ultimately ends up forcing the addition of the `ninja` binary's parent directory to the `PATH` environment variable for reliable and consistent behavior.

The second and far worse complication is a strange bug which only seems to manifest when all of the following conditions are met:

- OS: RHEL
- CMake version 3.28 or newer (as obtained via `uv tool`)
- Ninja (as obtained via `uv tool`; version?)
- CXX: `clang++` (version?)
- CMake Generator: Ninja
- C++ Standard: 20 or newer

Given these conditions, CMake configuration fails with the following unexpected error for CMake 3.29 or newer:

```
CMake Error at FindPackageHandleStandardArgs.cmake:227 (message):
  Could NOT find Threads (missing: Threads_FOUND)
```

and with the following unexpected error for CMake 3.28:

```
/bin/sh: line 1: CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS-NOTFOUND: command not found
```

I am completely baffled by this issue. My best guess is that there may be an issue with how either CMake or Ninja are being packaged for Python, as system-installed equivalents (for CMake or Ninja) do not appear to exhibit this issue.  I could not manage to work out a way to patch this erroneous behavior that didn't end up becoming a variation of this terrible special-casing code across various EVG scripts:

```bash
ninja_binary="$(uv run --no-project --isolated --with ninja bash -c "command -v ninja")"

# Strange issue only on RHEL...
if [[ "${distro_id:?}" =~ rhel ]]; then
  : # ... when using PyPI-provided CMake (3.28 or newer), Clang, and Ninja...
  if [[ "${CXX:-}" =~ clang\+\+ && "${generator:-"Ninja"}" == Ninja ]]; then
    : # ... and when using C++20 or newer...
    if [[ "${REQUIRED_CXX_STANDARD:-}" -ge 20 ]]; then
      # ... results in `Could NOT find Threads (missing: Threads_FOUND)` (???).
      # Use Ninja provided by MongoDB Toolchain v4 as a workaround.
      ninja_binary=/opt/mongodbtoolchain/v4/bin/ninja
    fi
  fi
fi

command -v "${ninja_binary:?}"
```

Due to these two issues, I decided to unconditionally use the [MongoDB Server Toolchain](https://github.com/10gen/toolchain-builder/blob/master/mciproject.yml), as provided by the `/opt/mongodbtoolchain/v4/bin` path ~on all our currently-used non-Windows distros~ on RHEL distros only:

```bash
[[ "${distro_id:?}" == rhel* ]] && PATH="${PATH:-}:/opt/mongodbtoolchain/v4/bin" || uv tool install -q ninja
```

The v4 toolchain is chosen due to currently having better distro availability than v5 (e.g. it is missing on the `rhel7-latest` distro). In spite of [prior efforts](https://jira.mongodb.org/browse/CDRIVER-4157) to avoid depending on the MongoDB Toolchain (due to Server Team not supporting any downstream users beyond themselves), I believe switching from Unix Makefiles to Ninja is nevertheless worth the effort. We are already (knowingly) depending on the Server Toolchain for other tasks (e.g. [sanitizers](https://github.com/mongodb/mongo-cxx-driver/blob/bccac0ea4b5346908176c3a32dca84152f03530c/.evergreen/config_generator/components/sanitizers.py#L35-L36) and [ccache](https://github.com/mongodb/mongo-c-driver/blob/017cf329d7c9bf705cfa1d88eca15e884d13f7dc/.evergreen/scripts/find-ccache.sh#L38-L43)) despite this lack of downstream support guarantee. I am in favor of actively maintaining compatibility with the Server Toolchain ourselves (as we've already been doing) rather than attempting workarounds like those described above. We can revisit obtaining `ninja` via `uv` in followup PRs (i.e. on Windows distros to finally have ccache-friendly MSVC builds).

> [!NOTE]
> The path to the Server Toolchain is _appended_ rather than _prepended_ to `PATH` to avoid having higher precedence when invoking compilers (e.g. `g++`, `clang++`, etc.) and other binaries. It is okay if we end up using a system `ninja` (such as on Ubuntu distros): we just need to guarantee the availability of _a_ `ninja` binary for use as the CMake generator without needing to also use the `CMAKE_MAKE_PROGRAM` configuration variable.

---

Miscellaneous drive-by fixes and improvements include:

- `setup_group_can_fail_task` (incorrect) -> `setup_task_can_fail_task` (correct) in the `abi-stability` task group definition.
- Use [PEP 0440](https://peps.python.org/pep-0440/) syntax to specify CMake binary versions to use in `cmake-compat` tasks (e.g. `cmake~=3.0` for "latest v3 minor release", `cmake~=3.15.0` for "latest v3.15 patch release", etc.).
- Removal of extranneous commands in `setup` EVG function, which have been redundant/obsolete/unused information for a while now.
- Move the "uninstall" build command out of `uninstall_check*` scripts to avoid the need to handle paths within the scripts themselves (in particular the Windows batch script).
- Improve MSBuild parallelism by using `UseMultiToolTask` and `EnforceProcessCountAcrossBuilds` (VS 2019 16.3 and newer) and ~~`/maxcpucount`~~ `CMAKE_BUILD_PARALLEL_LEVEL` ([CMake 3.26 and newer](https://gitlab.kitware.com/cmake/cmake/-/commit/48db261e69c37cd9504c27dcee952d651f7ac43b)) in `compile.sh`.
- Extend ccache and CMake parallelism to building libmongocrypt in `install-c-driver.sh`.

### DIFF
--- a/.evergreen/config_generator/components/funcs/install_uv.py
+++ b/.evergreen/config_generator/components/funcs/install_uv.py
@@ -15,7 +15,7 @@ class InstallUV(Function):
                 set -o errexit
                 set -o pipefail
 
-                version="0.8.3"
+                version="0.8.6"
 
                 if [[ ! -n "${MONGO_CXX_DRIVER_CACHE_DIR}" ]]; then
                     echo "MONGO_CXX_DRIVER_CACHE_DIR is not defined!" 1>&2
@@ -25,10 +25,12 @@ class InstallUV(Function):
                 uv_install_dir="${MONGO_CXX_DRIVER_CACHE_DIR}/uv-$version"
                 mkdir -p "$uv_install_dir"
 
-                if ! command -v "$uv_install_dir/uv" 2>/dev/null; then
+                # Install if the binary is missing or the incorrect version.
+                if ! (command -v "$uv_install_dir/uv" >/dev/null && "$uv_install_dir/uv" --version 2>/dev/null | grep "$version"); then
                     script="$(mktemp)"
                     cp -f mongo-cxx-driver/.evergreen/scripts/uv-installer.sh "$script"
                     chmod +x "$script"
+                    # Always patch the install script so it validates checksums.
                     (
                       . mongo-cxx-driver/.evergreen/scripts/patch-uv-installer.sh
                       patch_uv_installer "$script" "$version"
--- a/.evergreen/generated_configs/functions.yml
+++ b/.evergreen/generated_configs/functions.yml
@@ -454,7 +454,7 @@ functions:
             set -o errexit
             set -o pipefail
 
-            version="0.8.3"
+            version="0.8.6"
 
             if [[ ! -n "${MONGO_CXX_DRIVER_CACHE_DIR}" ]]; then
                 echo "MONGO_CXX_DRIVER_CACHE_DIR is not defined!" 1>&2
@@ -464,10 +464,12 @@ functions:
             uv_install_dir="${MONGO_CXX_DRIVER_CACHE_DIR}/uv-$version"
             mkdir -p "$uv_install_dir"
 
-            if ! command -v "$uv_install_dir/uv" 2>/dev/null; then
+            # Install if the binary is missing or the incorrect version.
+            if ! (command -v "$uv_install_dir/uv" >/dev/null && "$uv_install_dir/uv" --version 2>/dev/null | grep "$version"); then
                 script="$(mktemp)"
                 cp -f mongo-cxx-driver/.evergreen/scripts/uv-installer.sh "$script"
                 chmod +x "$script"
+                # Always patch the install script so it validates checksums.
                 (
                   . mongo-cxx-driver/.evergreen/scripts/patch-uv-installer.sh
                   patch_uv_installer "$script" "$version"
--- a/.evergreen/scripts/patch-uv-installer.sh
+++ b/.evergreen/scripts/patch-uv-installer.sh
@@ -34,24 +34,24 @@ patch_uv_installer() {
 
     # The output of the `download_checksums` helper function.
     checksums=(
-        9ebfe9f3b51187932ef97270b689da48261acacadd6ea7018d2cc62719c86ffe  uv-aarch64-apple-darwin.tar.gz
-        6e0692b817c5d6cfddad13ad177e866e36d95e8d32b4a296a49d937fdcda18d3 *uv-aarch64-pc-windows-msvc.zip
-        e82b5a3eb19e5087a6ea92800b0402f60378bd395e3483acd0b46124128ab71f  uv-aarch64-unknown-linux-gnu.tar.gz
-        05707835aadb8d0dfacf64a50dc31eb77fb11d1773b13ddd78d2404d43be3b7f  uv-aarch64-unknown-linux-musl.tar.gz
-        63f09a167470e40f5a7afce7ba726fa31d700042107215a8f46572a617b2ba64  uv-arm-unknown-linux-musleabihf.tar.gz
-        ca3ab5110a248febf3af56ca8e417add5a164228d8861d2f6439d6c77654ad24  uv-armv7-unknown-linux-gnueabihf.tar.gz
-        79b9f2f8abf64b93aaf48f87b159bc34e73b92428c6424a9e633f481449bc8be  uv-armv7-unknown-linux-musleabihf.tar.gz
-        5d272849a94b7ad36711f336d745e08ed3732042fc51f5c7f28bfc4e95463615 *uv-i686-pc-windows-msvc.zip
-        5ca9dcdd8aba4c59266536c00453f3815ccf4852a1bef78f181f4ae43726e742  uv-i686-unknown-linux-gnu.tar.gz
-        7a7eb903b673c0a0fa8282e12d2130fb63c1f0e574b1b8eb24ac63503db5db9e  uv-i686-unknown-linux-musl.tar.gz
-        ad94079c0aa352730e5656246ecb8fd6cb5c7c1bbcc100954020f32128d6b0de  uv-powerpc64-unknown-linux-gnu.tar.gz
-        27ccd00f3c617b59d6977e48612dc5fe49d15c1181bf995863350e78002e54ac  uv-powerpc64le-unknown-linux-gnu.tar.gz
-        897257393ca612058bb22691fab3a45480d3f7fc1df5bb968696576d7d98d6c1  uv-riscv64gc-unknown-linux-gnu.tar.gz
-        d148aebfc019aa14c686713fb0a391bb700290ecdf2903c2fdab0009182d880b  uv-s390x-unknown-linux-gnu.tar.gz
-        77eac9622f76ad89a8c59b31a96277aa61eb290d2949c69ab2061076471aeda2  uv-x86_64-apple-darwin.tar.gz
-        4ca84e28b08f48255f95156c5987d61a5e4c51a43372708bc6d84e994eeb7bdb *uv-x86_64-pc-windows-msvc.zip
-        427c27ed5f87bf91aa045cf459ea34d348ed6377c62c3c054f1b4046b2f83fe2  uv-x86_64-unknown-linux-gnu.tar.gz
-        bfbe34caae328355b720ad0f667be99990b1b93721c8cb726e4a502796eb6d68  uv-x86_64-unknown-linux-musl.tar.gz
+        15269226c753f01137b5a35c79e59ab46d8aab25a242641fdc4003a6d0a831ff  uv-aarch64-apple-darwin.tar.gz
+        a2891f1f1c56e717115579da655951007e2e5e498535b473d9f7cbffe7369e1a *uv-aarch64-pc-windows-msvc.zip
+        6fd314ca589788265ff99ec754bd2fa2a5d090ef592ddbbe8ded6b141615a491  uv-aarch64-unknown-linux-gnu.tar.gz
+        d78076c7e0dfcd3580736c11e009ef856bd13015f00406e3ded1fc895e2104ba  uv-aarch64-unknown-linux-musl.tar.gz
+        87d33a579cc2ee986e3163bb5de1791c6052ef8716d5fb8d8bf73aa2e5f1bf66  uv-arm-unknown-linux-musleabihf.tar.gz
+        4139d74dfe3192bc71dacd3cf0cdf6a006d121db2c4d841e08f369761baaf00e  uv-armv7-unknown-linux-gnueabihf.tar.gz
+        5d4d117cebddc7dbdbf6e1410c65868f113a75700acac77eed28dfdc0411c113  uv-armv7-unknown-linux-musleabihf.tar.gz
+        4eaa185b61f9cfe73ab7534de7282e51cc0f6bf47361429bdcbb6a5f3264f6e4 *uv-i686-pc-windows-msvc.zip
+        7fb9324cfb0c57a9b9145e73598794c7bc4df01f50730bd926d4ab10d4fc59ff  uv-i686-unknown-linux-gnu.tar.gz
+        3c2e2d69e8da093df76ffffa9758669c33ae47624f73d06ec2a83a363f17fbd4  uv-i686-unknown-linux-musl.tar.gz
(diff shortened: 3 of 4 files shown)


.evergreen/scripts/uninstall_check.sh
```diff
@@ -5,20 +5,8 @@ set -o errexit # Exit the script with error if any of the commands fail
 BUILD_DIR="$(pwd)/build"
 INSTALL_DIR="$BUILD_DIR/install"
 
-if [[ "${distro_id:?}" == rhel* ]]; then
-  LIB_DIR="lib64"
-else
-  LIB_DIR="lib"
-fi
-
 touch "$INSTALL_DIR/$LIB_DIR/canary.txt"
```
kevinAlbs:
```suggestion
```

Redundant with `touch "$p/canary.txt"` in `UninstallCheck`.

.evergreen/scripts/install-c-driver.sh
```diff
@@ -37,39 +37,41 @@ mkdir "${mongoc_dir}"
...
-fi
-
-if [[ "${OSTYPE}" == darwin* ]]; then
-  # MacOS does not have nproc.
-  nproc() {
-    sysctl -n hw.logicalcpu
-  }
-fi
+"${cmake_binary:?}" --version | head -n 1
+echo "ninja version: $(ninja --version)"
```
kevinAlbs: This may need to be special cased for macOS to fix macOS tasks ([example](https://spruce.mongodb.com/task/mongo_cxx_driver_compile_only_matrix_compile_only_macos_14_cxx11_debug_patch_bccac0ea4b5346908176c3a32dca84152f03530c_687feee41ce1270007a9a7b0_25_07_22_20_04_54/logs?execution=0)): `ninja: command not found`.

.evergreen/scripts/install-c-driver.sh
```diff
@@ -37,39 +37,41 @@ mkdir "${mongoc_dir}"
...
-  } >/dev/null
-fi
-
-if [[ "${OSTYPE}" == darwin* ]]; then
-  # MacOS does not have nproc.
-  nproc() {
-    sysctl -n hw.logicalcpu
-  }
-fi
+"${cmake_binary:?}" --version | head -n 1
```
kevinAlbs: Minor: but consider adding `| head -n 1` to the other cmake `--version` invocations for shorter logs.

kevinAlbs (commented):
The efforts to move towards uv and Ninja are much appreciated. LGTM with a fix to macOS builds.

Aside: I would like to avoid manually patching `uv-installer.sh`, but I do not see a clear alternative.

.evergreen/scripts/install-c-driver.sh
eramongodb: Whoops. I completely overlooked MacOS in my ninja availability survey. 🤦

This motivated a compromise between a fully-special-cased exemption for the FindThreads bug mentioned above and "avoid it entirely" by special-casing the RHEL distro only, simplified to a one-liner:

```bash
[[ "${distro_id:?}" == rhel* ]] && PATH="${PATH:-}:/opt/mongodbtoolchain/v4/bin" || uv tool install -q ninja
```

eramongodb (commented):
Updated the PR to include a new `patch-uv-installer.sh` script which automatically patches the `uv-installer.sh` script to enable checksum validation. The list of checksums to be validated with is embedded in the `patch-uv-installer.sh` script along with a `download_checksums` helper function which may be used to update the list of checksums.

Updated the PR to use `uv tool install` as designed and intended rather than the `uv run --isolated ...` pattern initially proposed. This is due to uv 0.8.1 and newer implementing stronger protections against bypassing ephemeral environments in [astral-sh/uv#14790](/astral-sh/uv/pull/14790), which breaks the `uv run --isolated ...` pattern, as the directory no longer exists after the command as completed execution.

Instead, the `UV_TOOL_DIR` and `UV_TOOL_BIN_DIR` environment variables are used to install the required binaries into (task-)local directories. This should not be a significant performance problem, since the (user-level) uv cache will still be utilized to populate the local tool directories. This also further reduced the need for `*_binary` Bash variables, although it slightly complicated the sequence of commands to obtain preferred build tools due to Windows path handling.

kevinAlbs (approved):
LGTM. I like the newly added patching script, and expect that will ease future upgrades of uv.

.evergreen/scripts/test.sh
```diff
@@ -100,11 +100,22 @@ fi
...
+  UV_TOOL_DIR="$(pwd)/uv-tool"
+  UV_TOOL_BIN_DIR="$(pwd)/uv-bin"
+fi
+PATH="${UV_TOOL_BIN_DIR:?}:${UV_INSTALL_DIR:?}:${PATH:-}"
+uv tool install -q cmake
+[[ "${distro_id:?}" == rhel* ]] && PATH="${PATH:-}:/opt/mongodbtoolchain/v4/bin" || uv tool install -q ninja
+PATH="${PATH:-}:/opt/mongodbtoolchain/v4/bin" # For ninja and llvm-symbolizer.
+
+cmake --version | head -n 1
+echo "ninja version: $(ninja --version)"
```
vector-of-bool: This particular incantation is repeated several times. Could it be consolidated into a single script file to be dot-sourced?

etc/make_release.py
```diff
@@ -449,13 +449,12 @@ def build_distribution(release_tag, release_version, c_driver_dir, quiet, skip_d
         click.echo('Clear ./build with "git clean -xdf ./build"', err=True)
         return None
 
-    run_shell_script('. .evergreen/scripts/find-cmake-old.sh;'
-                     'cd build;'
+    run_shell_script('cd build;'
                      'echo ' + release_version + ' > VERSION_CURRENT;'
-                     '${CMAKE} -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON '
+                     'uvx cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON '
```
vector-of-bool: Is this `cmake` version pinned?

eramongodb: No, it is not. In this case, the `make_release.py` script is being run manually by the user performing release instructions. However, your question made me realize this should really be made a dependency of the `make_release` group in `pyproject.toml` instead. Updated accordingly. `cmake` is now pinned to `4.0.3` by the lockfile (for the `make_release` group).

.evergreen/scripts/test.sh
eramongodb: SGTM. Added an `install-build-tools.sh` which unconditionally obtains `cmake` and `ninja`, but in a way which does not require the `distro_id` variable to detect RHEL, and also in a way that does not import _all_ of MongoDB Toolchain's binaries by adding them to `PATH` (instead using a simple symlink to "install" only `ninja` into the uv tool bin directory).

PR #1412, opened 2025-05-28
CXX-3232 add bsoncxx v1 declarations

eramongodb:
Resolves CXX-3232. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1318, https://github.com/mongodb/mongo-cxx-driver/pull/1401, and https://github.com/mongodb/mongo-cxx-driver/pull/1402.

This is 2 out of an estimated 7 major PRs which in total are expected to resolve CXX-2745 and CXX-3320.

- [x] Top-Level Dependencies (CXX-3239)
- [ ] bsoncxx v1 Interfaces (CXX-3232) (<-- you are here)
- [ ] bsoncxx v1 Implementations (CXX-3233)
- [ ] bsoncxx v_noabi Refactor (CXX-3234)
- [ ] mongocxx v1 Interfaces (CXX-3236)
- [ ] mongocxx v1 Implementations (CXX-3237)
- [ ] mongocxx v_noabi Refactor (CXX-3238)

This PR introduces the very first set of v1 ABI exported symbols. 🎉 

image

... only because MSVC complains otherwise. The rest will come as planned in CXX-3233 (impls).

---

Due to the volume of changes relative to the equivalent v_noabi interfaces, the bulk of relative changelog (v_noabi -> v1) and rationale for design decisions are documented inline as GitHub comments (coincidentally, exactly 100 comments in total).

In summary, notable changes include:

- General Notes
	- Nearly all preconditions are now documented and well-defined to instead throw an exception or yield an "invalid" state.
	- Narrowing casts and discarded return values are now thoroughly accounted for as new error codes.
	- All potentially-thrown exceptions are now documented and include the list of error codes which may be thrown.
- ~~`v1::error`~~ `v1::<class>::errc`
    - All exceptions thrown by library components have a corresponding error code enumeration declared in ~~`v1::error`~~ `v1::<class>::errc` with corresponding error category `v1::<class>::error_category`.
    - Error conditions `source_errc` and `type_errc` are provided as the primary user-facing error code query mechanism when component-specific error codes are not necessary.
- `v1::types::b_*`
    - All types are now consistently non-aggregate for list-initialization consistency.
    - All "single underlying value" types now consistently support implicit conversion to said value.
    - Multi-argument constructors are no longer explicit (always favored over `value`).
- `v1::types::view`
    - This type is now semitrivial (no user-provided non-default special member functions).
    - Conversion to this type is now favored over conversion to `value` when applicable.
- `v1::types::value`
[...]

PR #1430, opened 2025-07-28
CXX-3233 add bsoncxx v1 implementations and tests

eramongodb:
Resolves CXX-3233. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1412.

This is 3 out of an estimated 7 major PRs which in total are expected to resolve CXX-2745.

- [x] Top-Level Dependencies (CXX-3239)
- [x] bsoncxx v1 Interfaces (CXX-3232)
- [ ] bsoncxx v1 Implementations (CXX-3233) (<-- you are here)
- [ ] bsoncxx v_noabi Refactor (CXX-3234)
- [ ] mongocxx v1 Interfaces (CXX-3236)
- [ ] mongocxx v1 Implementations (CXX-3237)
- [ ] mongocxx v_noabi Refactor (CXX-3238)

This PR provides implementations and tests for the v1 API introduced by CXX-3232 in https://github.com/mongodb/mongo-cxx-driver/pull/1412.

> [!IMPORTANT]
> This PR deliberately leaves the v_noabi API interfaces, implementations, and tests unchanged to facilitate direct, side-by-side comparisons between v1 and v_noabi. Redundancy and duplication between v1 and v_noabi will be addressed by the bsoncxx v_noabi refactor (CXX-3234).

---

In summary, notable changes (relative to the v_noabi API's implementation and tests) include:

- Smaller ABI footprint: 136 (unique) exported v1 symbols vs. 250 (approximately equivalent) exported v_noabi symbols.
    - Note: v1 is still missing equivalents for `builder::*`, `validate::*`, and `vector::*`, which bring the v_noabi symbol count up to 325.
    - The biggest ABI footprint savings come from:
        - Nearly all `array::*` function defined in terms of `document::*` except `v1::array::view::find(std::uint32_t) const`.
        - Defining `types::value` constructors in terms of `b_<type>` overloads only.
        - Inlined definitions of comparison operator overloads and many other trivial functions.
- "Internal" component headers `lib/**/*.hh`.
    - A pattern that is more common for the mongocxx v_noabi implementations, but unlike mongocxx, v1 keeps them lightweight (definitions in the `.cpp` to minimize recompilations).
    - Internal-only API within `internal` classes instead of `friend` declarations to access "private" interfaces or data.
    - The public static member functions within `internal` classes declare the exact set of private data being exposed for use by dependent components (including tests).
    - Invocations of specific `internal` public static member functions explicitly documents-as-code the dependents that require corresponding private data.
- An error category is consistently defined as a static local immortal object within the error category function.
    - The name of the category is the fully-qualified name of the associated class (e.g. `bsoncxx::v1::decimal128`) or the error code enumeration otherwise (e.g. `bsoncxx::v1::source_errc`).
    - Every error code defines an `equivalent()` mapping to the `source_errc` and `type_errc` error condition enumerations.
- More applications of the new `BSONCXX_V1_TYPES_XMACRO` X-macro pattern in place of `#include <bsoncxx/enums/type.hpp>` X-macro headers.
- Implementation of the "inline PIMPL" idiom referenced in CXX-3232 (placement new of `impl` object into reserved data member storage rather than dynamic allocation).
- A new internal and undocumented "invalid" state for `v1::document::element` used to implement the alternative approach to CXX-2120. Note the deliberate lack of any mention of this state in the public API: it is meant to be entirely unobservable except in service of corresponding exception error messages.
- "Test" component headers (`test/**/*.hh`) defining Catch stringification for tested entities.
    - Alternative to the include-everything `bsoncxx/test/catch.hh` header to minimize recompilations.
    - Open to alternative design proposals which are not susceptible to ODR violation of `StringMaker<T>` specializations. :melting_face:
    - `bsoncxx::test::stringify()` as equivalent to `Catch::Detail::Stringify()` (deduces `T` in call to `StringMaker<T>::convert()`).
    - New `"stringify"` test cases for Catch expression decomposition behavior (implicit "to string" test coverage).
    - Catch stringification validated by running tests with the [`--success`](https://catch2-temp.readthedocs.io/en/latest/command-line.html#showing-results-for-successful-tests) flag.
- Catch test suite quality improvements.
    - Nearly 100% test coverage of bsoncxx v1 API (note: don't expect this for mongocxx...), including copy+move semantics (note: _do_ expect this for mongocxx) to avoid problems like [this](https://github.com/mongodb/mongo-cxx-driver/commit/691c5a792cdfdc2613c994bf2fee8bb39bfa4de5#diff-2ed2b62b994de7481fb4518f799a73e8a352ea547bf6c80fd9772607d212d4c3R25-R28), [this](https://github.com/mongodb/mongo-cxx-driver/commit/fb2d4f7da3ec07f8fd33d4f4e5d36535a47b5c06#diff-c6ba7931024b6662604989ea07905632a4925981316d05d7b9de7e266e60474bR269-R280), etc.
        - Especially important for inline PIMPL classes and `v1::document::value` to ensure correct `impl` and custom deleter behavior.
        - `CHECK_NOFAIL` used to document behavior which may vary across standard library implementations (surprising!).
    -  `[namespace][to][component]` for better [test case filtering](https://catch2-temp.readthedocs.io/en/latest/command-line.html#specifying-which-tests-to-run).
        - `[error]` tag for testing error-code-specific behavior (e.g. mapping codes to conditions).
        - `[test]` tag for test-suite-specific behavior (e.g. Catch stringification).
    - All documented exception-throwing code paths are explicitly tested with `CHECK_THROWS_WITH_CODE` (excluding `v1::oid::oid()`...).
    - Static assertions of important compile-time properties (e.g. overload resolution behavior).
        - Defined in test components instead of library components to avoid impacting library-only build performance.

> [!NOTE]
> To avoid v1 test components depending on the v_noabi API (no v1 library or test component includes v_noabi headers), BSON data are hand-written as binary data (e.g. `std::uint8_t data[] = {...}`) instead of using the `bsoncxx::v_noabi::builder::*` API. These may eventually be substituted with a v1 builder API, but I think there is also merit to keeping them independent of any builder API, such that the test cases are not dependent on unrelated behavior and minimizes cross-component test behavior dependencies. :thinking:


src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
```diff
@@ -30,6 +42,29 @@ static_assert(
     is_nothrow_moveable<view::const_iterator>::value,
     "bsoncxx::v1::array::view::const_iterator must be nothrow moveable");
 
+view::const_iterator view::find(std::uint32_t i) const {
+    if (!_view) {
+        return this->cend();
+    }
```
eramongodb, line 48: This check is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/view.cpp#L111)). This is a consequence of the new, well-defined "invalid" state of an array view.

src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
```diff
@@ -30,6 +42,29 @@ static_assert(
...
+
+    if (!bson_init_static(&bson, _view.data(), _view.length())) {
+        throw v1::exception{code::invalid_data};
+    }
+
+    bson_iter_t iter;
+
+    if (!bson_iter_init_find_w_len(&iter, &bson, key.c_str(), static_cast<int>(key.length()))) {
+        return this->end();
+    }
```
eramongodb: This is slightly different to v_noabi's `bson_init_static` -> `bson_iter_init` -> `bson_iter_init_find` [implementation](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/view.cpp#L117-L127), but should still be equivalent in observable behavior.

src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
```diff
@@ -30,6 +42,29 @@ static_assert(
...
+
+    return const_iterator::internal::make_const_iterator(
+        _view.data(), _view.length(), bson_iter_offset(&iter), bson_iter_key_len(&iter));
+}
```
eramongodb: Invokes [this](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/include/bsoncxx/v1/document/view.hpp#L369) private constructor, which is [public](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/view.hpp#L165) in v_noabi.

src/bsoncxx/lib/bsoncxx/v1/decimal128.cpp
```diff
@@ -16,13 +16,117 @@
...
 static_assert(is_semitrivial<decimal128>::value, "bsoncxx::v1::decimal128 must be semitrivial");
 
+decimal128::decimal128(v1::stdx::string_view sv) {
+    if (sv.empty()) {
+        throw v1::exception{code::empty_string};
+    }
+
+    if (sv.size() > std::size_t{INT_MAX}) {
+        throw v1::exception{code::invalid_string_length};
+    }
```
eramongodb, line 45: These checks are [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/decimal128.cpp#L26).

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
     is_nothrow_moveable<view::const_iterator>::value,
     "bsoncxx::v1::document::view::const_iterator must be nothrow moveable");
 
+namespace {
+
+constexpr std::uint8_t k_default_view[5] = {5u, 0u, 0u, 0u, 0u};
```
eramongodb: This is the special empty document representation that is used to avoid `v1::document::value` allocation (with a no-op deleter).

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+
+} // namespace
+
+view::view() : view{k_default_view} {}
+
+view::view(std::uint8_t const* data, std::size_t length) : view{data} {
+    if (length < _empty_length || length < this->size()) {
+        throw v1::exception{v1::document::view::errc::invalid_length};
+    }
```
eramongodb, line 58: These checks are [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp#L147). This is a consequence of the new "embedded" length representation (as computed by `this->size()`).

This is the only time the `length < _empty_length` comparison is necessary. Beyond this point, it is a class invariant that `data` is either null or points to a storage region that is large enough to compute `this->size()` without undefined behavior.

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+}
+
+view::const_iterator view::cbegin() const {
+    if (!this->operator bool()) {
+        return this->cend();
+    }
```
eramongodb, line 64: This check (and the one in `find()` below) is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp#L84). This is a consequence of the new, well-defined "invalid" state of a document view.

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+    if (!bson_iter_init_from_data(&iter, _data, this->size())) {
+        throw v1::exception{code::invalid_data};
+    }
+
+    if (!bson_iter_next(&iter)) {
+        return this->cend();
+    }
+
+    return const_iterator::internal::make_const_iterator(
+        _data, this->size(), bson_iter_offset(&iter), bson_iter_key_len(&iter));
```
eramongodb: Invokes [this](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/include/bsoncxx/v1/document/view.hpp#L369) private constructor, which is [public](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp#L165) in v_noabi.

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+}
+
+view::const_iterator view::find(v1::stdx::string_view key) const {
+    if (!this->operator bool()) {
+        return this->cend();
+    }
+
+    if (key.size() >= std::size_t{INT_MAX}) {
+        return this->cend();
+    }
```
eramongodb, line 91: This check is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp#L111). This guards the cast to `int` in the later call to `bson_iter_init_find_w_len` (passing `-1` [does not help](https://github.com/mongodb/mongo-c-driver/blob/d3165e022ae9155289e1910c72c393d035f75658/src/libbson/src/bson/bson-iter.c#L304)).

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+
+    // Support null as equivalent to empty.
+    if (!key.data()) {
+        key = "";
+    }
```
eramongodb, line 96: Preserves [v_noabi behavior](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp#L120-L131).

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+                using condition = v1::type_errc;
+
+                auto const type = static_cast<condition>(ec.value());
+
+                switch (static_cast<code>(v)) {
+                    case code::invalid_length:
+                        return type == condition::invalid_argument;
+
+                    case code::invalid_data:
+                        return type == condition::runtime_error;
```
eramongodb, line 164: As a reminder, `type_errc::invalid_argument` is meant to only apply to immediate arguments given to the invoked function from which the exception may be thrown. Here, because `invalid_data` is only thrown when `bson_*` functions report failure (not by checking the value of an immediate function parameter), it is categorized as a `type_errc::runtime_error`. For this purpose, data members of `this` are not considered to be "arguments" to the given member function. This rationale applies to all other v1 error codes.

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+
+    return instance.value();
+}
+
+view::const_iterator& view::const_iterator::operator++() {
+    if (!_element) {
+        return *this;
+    }
+
+    auto iter_opt = to_bson_iter(_element);
```
eramongodb, line 186: This is using ADL to invoke `bsoncxx::v1::element::view::to_bson_iter(view const&)` in `lib/bsoncxx/v1/element/view.hh`.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+    enum : std::size_t {
+        _padding_size = sizeof(view::_storage)       // Total reserved.
+                        - sizeof(void*)              // _raw
+                        - 3u * sizeof(std::uint32_t) // _length, _offset, and _keylen.
+                        - 1u                         // _is_valid (final byte).
+                        - 0u,
+    };
+
+    // `_padding_size == 3` given `sizeof(void*) == 8`.
+    static_assert(_padding_size < sizeof(view::_storage), "sizeof(impl) must not exceed reserved storage size");
```
eramongodb, line 52: We do not support any platforms where `sizeof(void*)` or `alignof(void*)` exceed `8`, although they may equal `4` when targeting certain 32-bit platforms.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+    std::uint32_t _length = {};
+    std::uint32_t _offset = {};
+    std::uint32_t _keylen = {};
+
+    BSONCXX_PRIVATE_WARNINGS_PUSH();
+    BSONCXX_PRIVATE_WARNINGS_DISABLE(GNU("-Wunused"));
+    unsigned char _padding[_padding_size]; // Reserved.
+    BSONCXX_PRIVATE_WARNINGS_POP();
+
+    bool _is_valid = {}; // Last byte.
```
eramongodb: This representation is completely identical to the one for `v_noabi::document::element` (before CXX-2120), but alignment and reserved storage requirements grants us some extra space (even on 64-bit platforms) to make use of an extra `bool _is_valid` data member in place of the `v1::stdx::optional<v1::stdx::string_view>` as implemented by CXX-2120. This similarly applies to `v1::types::value::impl`, although the extra space is currently unused.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+    v1::types::id type_id() const {
+        auto const iter = this->iter();
+        return static_cast<v1::types::id>(bson_iter_type(&iter));
+    }
+
+    v1::types::id type_id_unchecked() const {
+        if (auto const iter_opt = this->iter_unchecked()) {
+            return static_cast<v1::types::id>(bson_iter_type(&*iter_opt));
+        }
+        return v1::types::id{}; // BSON_TYPE_EOD
```
eramongodb, line 114: This value is never exposed to the user. This is just to simplify calls to `type_id_unchecked()` such that they do not have to individually special-case the invalid state.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+        throw v1::exception{code::invalid_data};
+    }
+
+    v1::types::view type_view() const {
+        this->check();
+        return this->type_view_unchecked();
+    }
+
+    v1::types::view type_view_unchecked() const {
+        if (auto opt = v1::types::view::internal::make(_raw, _length, _offset, _keylen)) {
```
eramongodb, line 139: This invokes the internal function `bsoncxx::v1::types::view::internal::make(bson_value_t const& v)` in `lib/bsoncxx/v1/types/view.cpp`.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+            return *opt;
+        }
+        throw v1::exception{code::invalid_data};
+    }
+
+    v1::types::value type_value() const {
+        this->check();
+        if (auto opt = v1::types::value::internal::make(_raw, _length, _offset, _keylen)) {
```
eramongodb, line 147: This invokes the internal function `bsoncxx::v1::types::value::internal::make(std::uint8_t const* raw, std::uint32_t length, std::uint32_t offset, std::uint32_t keylen)` in `lib/bsoncxx/v1/types/value.cpp`.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+        throw v1::exception{code::invalid_data};
+    }
+};
+
+void view::impl::check() const {
+    static_assert(is_semiregular<impl>::value, "bsoncxx::v1::element::view::impl must be semiregular");
+    static_assert(is_semitrivial<impl>::value, "bsoncxx::v1::element::view::impl must be semitrivial");
+
+    static_assert(sizeof(view::_storage) >= sizeof(view::impl), "insufficient size");
+    static_assert(alignof(view) >= alignof(view::impl), "insufficient alignment");
```
eramongodb, line 171: Unfortunately, static assertions for `impl` must be moved outside of the (incomplete) class definition into at least one of its out-of-line member function definitions (to allow access of private data members). This similarly applies to `v1::types::value::impl`, where its out-of-line (non-trivial) destructor definition is used instead.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+
+    if (!this->is_valid()) {
+        bson_iter_t iter;
+        if (_raw && bson_iter_init_from_data_at_offset(&iter, _raw, _length, _offset, _keylen)) {
+            std::string msg;
+            msg += "last known element key \"";
+            msg += bson_iter_key(&iter);
+            msg += '"';
+            throw v1::exception{code::invalid_view, msg};
+        } else {
```
eramongodb, line 181: This is how the alternative approach to CXX-2120 is implemented: the existing BSON data pointed to by `_raw` for the invalid `element` constructed via `impl::to_invalid()`. See the "key" section of the "exceptions" test case in `test/v1/element/view.cpp` for a demonstration of its expected behavior.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+view::~view() = default;
+
+view::view(view const& other) noexcept {
+    new (internal::impl(this)) impl{internal::impl(other)};
+}
+
+view& view::operator=(view const& other) noexcept {
+    *internal::impl(this) = internal::impl(other);
+    return *this;
+}
```
eramongodb: Ownership semantics are implemented in terms of the `impl` class. As noted in the prior PR and by the static assertions above, the `impl` class is semitrivial (hence straightforward copies) even though `view` itself is not (due to the opaque `impl` class).

src/bsoncxx/lib/bsoncxx/v1/element/view.hh
```diff
@@ -0,0 +1,53 @@
...
+        std::uint8_t const* raw,
+        std::uint32_t length,
+        std::uint32_t offset,
+        std::uint32_t keylen,
+        bool is_valid = true);
+
+    static v1::stdx::optional<bson_iter_t> to_bson_iter(view const& v);
+
+   private:
+    friend view;
```
eramongodb: The only `friend` declarations in `v1` are by `internal` classes to permit private member access by their parent class. After all, the `internal` class is intended to be merely an (internal only) extension of the parent class' API.

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
```diff
@@ -16,13 +16,172 @@
...
+
+        WSAGuard() {
+            WSADATA wsaData;
+            if (WSAStartup((MAKEWORD(2, 2)), &wsaData) != 0) {
+                throw v1::exception{
+                    WSAGetLastError(), std::system_category(), "WSAStartup() failed in bsoncxx::v1::oid::oid()"};
+            }
+        }
+#endif
+    };
```
eramongodb: This `WSAGuard` is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp#L26) and is a slight modification to [the one in test_bson](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/test/catch.cpp#L31) as introduced by https://github.com/mongodb/mongo-cxx-driver/pull/1363 to silence diagnostic messages. This test-only workaround should be obsoleted once the v_noabi refactor applies this initialization to `v_noabi::oid` (but without allow the exception to propagate to users, which would be an API breaking change).

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
```diff
@@ -16,13 +16,172 @@
...
+        throw v1::exception{code::invalid_length};
+    }
+
+    std::memcpy(_bytes.data(), bytes, _bytes.size());
+}
+
+oid::oid(v1::stdx::string_view str) {
+    if (str.empty()) {
+        throw v1::exception{code::empty_string};
+    }
```
eramongodb, line 82: This check is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp#L33), which may lead to library-level undefined behavior when invoking [`bson_oid_is_valid`](https://github.com/mongodb/mongo-c-driver/blob/d3165e022ae9155289e1910c72c393d035f75658/src/libbson/src/bson/bson-oid.c#L229).

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
```diff
@@ -16,13 +16,172 @@
...
+
+    bson_oid_t oid;
+    ((void)WSAGuard{}, bson_oid_init(&oid, nullptr));
+    std::memcpy(_bytes.data(), oid.bytes, sizeof(oid.bytes));
+}
+
+oid::oid(std::uint8_t const* bytes, std::size_t len) {
+    if (!bytes) {
+        throw v1::exception{code::null_bytes_ptr};
+    }
```
eramongodb, line 70: This check is [not present in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp#L42), which may lead to library-level undefined behavior when invoking `memcpy()`.

src/bsoncxx/lib/bsoncxx/v1/types/value.cpp
```diff
@@ -12,19 +12,470 @@
...
+
+value& value::operator=(value const& other) {
+    *internal::impl(this) = internal::impl(other);
+    return *this;
+}
+
+namespace {
+
+// For backward compatibility, do not prematurely truncate strings in bsoncxx API. Instead, defer handling of potential
+// embedded null bytes to the bson library.
```
eramongodb, line 162: This comment documents an observation that v_noabi's [`make_copy_for_libbson`](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/private/convert.hh#L31) (which in hindsight should really be moved out of `private` and into `v_noabi` along with a few more current `private` headers) uses `memcpy` rather than `str(n)cpy` to create owning BSON strings. The v1 implementation preserves this behavior, avoiding truncation and deferring null byte validation to the bson library.

src/bsoncxx/lib/bsoncxx/v1/types/value.cpp
```diff
@@ -12,19 +12,470 @@
...
+    switch (v.type_id()) {
+        BSONCXX_V1_TYPES_XMACRO(X)
+
+        default:
+            throw v1::exception{code::invalid_type};
+    }
+#pragma pop_macro("X")
+}
+
+// BSONCXX_V1_TYPES_XMACRO: update below.
```
eramongodb, line 203: The overloads below implement the behavior currently handled by v_noabi's `convert_to_libbson()` in [convert.hh](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/private/convert.hh), ~~but without exposing `bson_value_t` to dependent components~~.

src/bsoncxx/lib/bsoncxx/v1/types/value.cpp
```diff
@@ -12,19 +12,470 @@
...
+
+value::value(v1::types::b_double const v) : value{} {
+    internal::impl(this)->t() = BSON_TYPE_DOUBLE;
+    internal::impl(this)->v().v_double = v;
+}
+
+value::value(v1::types::b_string const v) : value{} {
+    if (v.value.size() > UINT32_MAX) {
+        throw v1::exception{code::invalid_length_u32};
+    }
```
eramongodb, line 213: These `UINT32_MAX` checks are [not in v_noabi](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/private/convert.hh#L38) and guard the subsequent casts to `std::uint32_t`.

src/bsoncxx/lib/bsoncxx/v1/types/view.cpp
```diff
@@ -34,6 +53,199 @@ BSONCXX_V1_TYPES_XMACRO(X)
...
+    if (!data) {
+        return {};
+    }
+
+    return to_sv(data, std::strlen(data));
+}
+
+} // namespace
+
+view view::internal::make(bson_value_t const& v) {
```
eramongodb, line 152: This implements the behavior currently handled by v_noabi's `convert_from_libbson()` in [convert.hh](https://github.com/mongodb/mongo-cxx-driver/blob/ad622035ec9c9ab352b42449d5fcbf426e8782e9/src/bsoncxx/lib/bsoncxx/private/convert.hh), ~~but without exposing `bson_value_t` to dependent components~~.

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
 
+using code = v1::element::view::errc;
+
 static_assert(is_semiregular<view>::value, "bsoncxx::v1::element::view must be semiregular");
 static_assert(is_nothrow_moveable<view>::value, "bsoncxx::v1::element::view must be nothrow moveable");
 
+class alignas(BSONCXX_PRIVATE_MAX_ALIGN_T) view::impl {
+   private:
+    enum : std::size_t {
+        _padding_size = sizeof(view::_storage)       // Total reserved.
```
kevinAlbs, line 44: Confirming: I expect the size of `_storage` is part of the public ABI. If there is a future need to add storage, am I correct that `_storage` could be updated to point to a larger allocated object without changing ABI?

src/bsoncxx/lib/bsoncxx/v1/element/view.hh
kevinAlbs: Consider moving the private static methods of `element::view::internal` to `element::view::impl`:

```c++
// In bsoncxx::v1::element::view::impl:

static view::impl const& to_impl(view const& self) {
    return *reinterpret_cast<view::impl const*>(self._storage.data());
}

static view::impl const* to_impl(view const* self) {
    return reinterpret_cast<view::impl const*>(self->_storage.data());
}

static view::impl* to_impl(view* self) {
    return reinterpret_cast<view::impl*>(self->_storage.data());
}
```

That appears to allow removing the `friend` declaration. IIUC the `internal` classes are intended as a controlled internal view into private API from other components. In this case, no other components need to (or can) call these private static methods.

Comment similarly applies to `types::value::internal`.

src/bsoncxx/test/v1/array/value.cpp
```diff
@@ -0,0 +1,579 @@
...
+        CHECK(v == value{v});
+
+        CHECK_FALSE(v == invalid);
+        CHECK(v != invalid);
+        CHECK(v.begin() == invalid.begin());
+        CHECK(v.end() == invalid.end());
+    }
+
+    SECTION("valid") {
+        std::uint8_t const xdoc[] = {12, 0, 0, 0, 16, 'x', '\0', 1, 0, 0, 0, 0}; // { 'x': 1 }
```
kevinAlbs: Consider using an array-like document, like `{ '0': 1 }`, which can be used to test `v.find(0)` succeeds.

src/bsoncxx/test/v1/array/value.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+        value doc{ptr};
+
+        REQUIRE(static_cast<bool>(doc));
+
+        CHECK(bsoncxx::test::stringify(doc) == R"([1])");
+    }
+
+    SECTION("two") {
+        // {"x": 1, "y": 2}
```
kevinAlbs:
```suggestion
        // {"0": 1, "1": 2}
```

src/bsoncxx/test/v1/array/value.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+        value doc{ptr};
+
+        REQUIRE(static_cast<bool>(doc));
+
+        CHECK(bsoncxx::test::stringify(doc) == R"([1, 2])");
+    }
+
+    SECTION("three") {
+        // {"x": 1, "y": 2, "z": 3}
```
kevinAlbs:
```suggestion
        // {"0": 1, "1": 2, "2": 3}
```

src/bsoncxx/test/v1/array/value.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+            auto const expr = [&](std::size_t length) {
+                return value{make_ptr().release(), length, std::ref(deleter)};
+            };
+
+            CHECK_THROWS_WITH_CODE(expr(0u), code::invalid_length);
+            CHECK_THROWS_WITH_CODE(expr(sizeof(data) - 1u), code::invalid_length);
+
+            CHECK(get_deleter<deleter_type>(expr(sizeof(data))).has_value());
+            CHECK(get_deleter<deleter_type>(expr(sizeof(data) + 1u)).has_value());
```
kevinAlbs: I expect passing a greater length violates the precondition:

> size of the storage region pointed to by `data` must be greater than or equal to `length`.

Though I expect this has no observable effect (constructor does not try to read past), consider always adding an extra byte in `make_ptr` to meet the precondition:

```c++
    auto const make_ptr = [&data]() {
        size_t const res_len = sizeof(data) + 1; // Add extra byte to meet preconditions in tests of larger lengths.
        auto res = std::unique_ptr<std::uint8_t[]>(new std::uint8_t[res_len]);
        std::memcpy(res.get(), data, sizeof(data));
        return res;
    };
```

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -0,0 +1,594 @@
...
+        CHECK_THROWS_WITH_CODE(value{v}, code::invalid_type);
+    }
+
+    SECTION("invalid_length_u32") {
+#pragma push_macro("X")
+#undef X
+#define X(_name, _value) value{b_##_name{}},
+
+        value values[] = {BSONCXX_V1_TYPES_XMACRO(X)};
+#pragma pop_macro("X")
```
kevinAlbs:
```suggestion
```

Unused.

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -0,0 +1,594 @@
...
+    }
+}
+
+TEST_CASE("b_types", "[bsoncxx][v1][types][value]") {
+    bsoncxx::v1::oid const o{"507f1f77bcf86cd799439011"};
+    std::uint8_t const data[] = {12, 0, 0, 0, 16, 'x', '\0', 1, 0, 0, 0, 0}; // {"x": 1}
+    bsoncxx::v1::document::view const doc{data};
+    bsoncxx::v1::array::view const arr{data};
+
+    // BSONCXX_V1_TYPES_XMACRO: update below.
```
kevinAlbs, line 268: What do the `BSONCXX_V1_TYPES_XMACRO: update` comments mean? Does it mean the surrounded block could be rewritten in terms of `BSONCXX_V1_TYPES_XMACRO`?

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+// BSONCXX_V1_TYPES_XMACRO: update above
+
+} // namespace btype_vs_view
+
+namespace btype_vs_value {
+
+template <typename BType, typename... Args>
+struct overload {
+    // Prerequisite: Args is not BType or view.
```
kevinAlbs: Confirming: should this prerequisite be instead checking `value` (not `view`)?

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -0,0 +1,594 @@
...
+        return {};
+    }
+
+    template <typename... Ts>
+    using non_list_expr = decltype((fn)(std::declval<Ts>()...));
+
+    template <typename... Ts>
+    using list_expr = decltype((fn)({std::declval<Ts>()...}));
+
+    // `fn({})` is always ambiguous.
```
kevinAlbs, line 443: Is this ambiguity needed to be preserved for stability? Or are these tests validating the current behavior? If these tests are only validating the current behavior, consider adding a comment to say this behavior is subject to change (I would otherwise assume the tested behavior is the behavior we want to preserve).

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
eramongodb: > Am I correct that `_storage` could be updated to point to a larger allocated object without changing ABI?

No, changing the size or alignment of `_storage` is an ABI breaking change. The layout of data members in `view::impl` is flexible within the limits of the predetermined size and alignment of `_storage`, but the size and alignment of `_storage` itself cannot be changed. If `v1::element::view::_storage` needs to be grown to accomodate more `impl` data members, it would have to be implemented as a new `v2::element::view` class. If this were a traditional PIMPL (_pointer_ to `impl`), the size and alignment of `impl` could be changed without breaking ABI. However, `_storage` is an array, not a pointer.

src/bsoncxx/test/v1/types/value.cpp
eramongodb: These annotations are meant to indicate code which are _not_ defined in terms of `BSONCXX_V1_TYPES_XMACRO` which may need to be manually updated when `BSONCXX_V1_TYPES_XMACRO` is updated, e.g. when a new enumerator is added.

src/bsoncxx/test/v1/types/value.cpp
eramongodb: Good catch. This was copy-paste error.

src/bsoncxx/test/v1/types/value.cpp
eramongodb: Both for API stability and to validate current behavior. We do not have documented API compatibility guidelines concerning overload resolution, so I am being conservative with the API stability requirements. `b_<type>` and `view` are currently ambiguous given a copy-initializer `{}`: making this unambiguous would mean either `b_<type>` or `view` (but not both) has an `explicit` default constructor, which would be inconsistent with other `b_<type>` objects and also impact overload resolution with `value` or other user-defined types.

src/bsoncxx/lib/bsoncxx/v1/element/view.hh
eramongodb: Good call. Moved the overloads into `impl` and renamed to `impl::with()`:

```cpp
// Before
return internal::impl(this)->is_valid();

// After
return impl::with(this)->is_valid();
```

src/bsoncxx/test/v1/array/value.cpp
eramongodb: Done.

kevinAlbs (approved):

src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
```diff
@@ -30,6 +42,29 @@ static_assert(
     is_nothrow_moveable<view::const_iterator>::value,
     "bsoncxx::v1::array::view::const_iterator must be nothrow moveable");
 
+view::const_iterator view::find(std::uint32_t i) const {
+    if (!_view) {
+        return this->cend();
+    }
+
+    bsoncxx::itoa key{i};
```
vector-of-bool, line 50: I had missed the addition of this utility type. Very handy.

src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
vector-of-bool: `bson_iter_find` functions may return `false` if they fail to find the element, but also may return `false` if they encounter a parse error (and set `iter.err_off`). Should this also `throw` `invalid_data` in that case?

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
vector-of-bool: Minor: Can this be a `std::array<u8, 5>`? (clang-tidy nags about arrays)

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
```diff
@@ -30,6 +43,167 @@ static_assert(
...
+        return this->cend();
+    }
+
+    bson_iter_t iter;
+
+    if (!bson_iter_init_from_data(&iter, _data, this->size())) {
+        throw v1::exception{code::invalid_data};
+    }
+
+    if (!bson_iter_next(&iter)) {
```
vector-of-bool: As with `iter_find`, this may fail and return `false` with a parse error. (May be useful to have a utility that advances an iterator and throws if it detects an error?)

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
```diff
@@ -12,19 +12,337 @@
...
+    static_assert(_padding_size < sizeof(view::_storage), "sizeof(impl) must not exceed reserved storage size");
+
+    std::uint8_t const* _raw = {};
+    std::uint32_t _length = {};
+    std::uint32_t _offset = {};
+    std::uint32_t _keylen = {};
+
+    BSONCXX_PRIVATE_WARNINGS_PUSH();
+    BSONCXX_PRIVATE_WARNINGS_DISABLE(GNU("-Wunused"));
+    unsigned char _padding[_padding_size]; // Reserved.
```
vector-of-bool: Can be a `std::array`? This should also be zero-initialized with `{}` to prevent uninitialized reads if the object is copied.

src/bsoncxx/lib/bsoncxx/v1/types/id.cpp
```diff
@@ -13,3 +13,47 @@
...
+#pragma push_macro("X")
+#undef X
+#define X(_name, _value)            \
+    case binary_subtype::k_##_name: \
+        return #_name;
+
+    switch (rhs) {
+        BSONCXX_V1_BINARY_SUBTYPES_XMACRO(X)
+
+        default:
```
vector-of-bool, line 51: All subtype octet values `≥0x80` are considered "user" types, not just `0x80`

src/bsoncxx/lib/bsoncxx/v1/types/value.cpp
```diff
@@ -12,19 +12,470 @@
...
+
+    // `_padding_size == 0` given `sizeof(void*) == 8`.
+    static_assert(_padding_size < sizeof(value::_storage), "sizeof(impl) must not exceed reserved storage size");
+
+    bson_value_t _value = {}; // BSON_TYPE_EOD
+
+    ~impl();
+
+    impl(impl&& other) noexcept {
+        _value = other._value; // Ownership transfer.
```
vector-of-bool:
```suggestion
    impl(impl&& other) noexcept : _value(other._value) {
```

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
```diff
@@ -16,13 +16,172 @@
...
+    struct WSAGuard {
+#if defined(_WIN32)
+        ~WSAGuard() {
+            (void)WSACleanup();
+        }
+
+        WSAGuard(WSAGuard&&) = delete;
+        WSAGuard& operator=(WSAGuard&) = delete;
+        WSAGuard(WSAGuard const&) = delete;
+        WSAGuard& operator=(WSAGuard const&) = delete;
```
vector-of-bool, line 50: Recommend moving these deleted definition outside of the conditional block, just to keep the same regularity on non-Windows platforms.

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
```diff
@@ -16,13 +16,172 @@
...
+            if (WSAStartup((MAKEWORD(2, 2)), &wsaData) != 0) {
+                throw v1::exception{
+                    WSAGetLastError(), std::system_category(), "WSAStartup() failed in bsoncxx::v1::oid::oid()"};
+            }
+        }
+#endif
+    };
+
+    bson_oid_t oid;
+    ((void)WSAGuard{}, bson_oid_init(&oid, nullptr));
```
vector-of-bool: Should this immediately destroy the `WSAGuard`, as opposed to making it a (magic) `static`? It would mean that we potentially keep WSA initialized eternally, but it may otherwise be expensive to create a lot of OID objects.

src/bsoncxx/test/v1/array/value.cpp
```diff
@@ -0,0 +1,588 @@
...
+TEST_CASE("exceptions", "[bsoncxx][v1][array][value]") {
+    using code = bsoncxx::v1::document::view::errc;
+
+    std::uint8_t const data[] = {5, 0, 0, 0, 0}; // {}
+
+    REQUIRE(data[0] == sizeof(data));
+
+    auto const make_ptr = [&data] {
+        // Reserve an extra byte to support the `sizeof(data) + 1` test cases.
+        auto res = std::unique_ptr<std::uint8_t[]>(new std::uint8_t[sizeof(data) + 1u]);
```
vector-of-bool, line 48: Could use `make_unique_for_overwrite`?

src/bsoncxx/test/v1/array/view.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+#include <bsoncxx/test/v1/array/view.hh>
+
+//
+
+#include <bsoncxx/test/v1/exception.hh>
+#include <bsoncxx/test/v1/types/value.hh>
+#include <bsoncxx/test/v1/types/view.hh>
+
+#include <climits>
```
vector-of-bool: I believe this header is unused

src/bsoncxx/test/v1/element/view.cpp
```diff
@@ -0,0 +1,308 @@
...
+#include <bsoncxx/test/v1/element/view.hh>
+
+//
+
+#include <bsoncxx/test/v1/document/view.hh>
+#include <bsoncxx/test/v1/exception.hh>
+#include <bsoncxx/test/v1/types/value.hh>
+#include <bsoncxx/test/v1/types/view.hh>
+
+#include <climits>
```
vector-of-bool: Unused header?

src/bsoncxx/test/v1/types/id.cpp
```diff
@@ -0,0 +1,110 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/test/v1/types/id.hh>
+
+//
+
+#include <bsoncxx/test/v1/stdx/string_view.hh>
+
+#include <algorithm>
```
vector-of-bool: Unused header?

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -0,0 +1,587 @@
...
+#include <bsoncxx/v1/types/view.hh>
+
+#include <bsoncxx/test/v1/detail/bit.hh>
+#include <bsoncxx/test/v1/document/view.hh>
+#include <bsoncxx/test/v1/exception.hh>
+#include <bsoncxx/test/v1/types/id.hh>
+#include <bsoncxx/test/v1/types/value.hh>
+#include <bsoncxx/test/v1/types/view.hh>
+
+#include <climits>
```
vector-of-bool: Unused header?

src/bsoncxx/test/v1/types/view.cpp
```diff
@@ -0,0 +1,520 @@
...
+#include <bsoncxx/test/v1/document/view.hh>
+#include <bsoncxx/test/v1/exception.hh>
+#include <bsoncxx/test/v1/types/id.hh>
+#include <bsoncxx/test/v1/types/value.hh>
+#include <bsoncxx/test/v1/types/view.hh>
+
+#include <climits>
+#include <string>
+#include <system_error>
+#include <utility>
```
vector-of-bool: `<climits>` and `<utility>` unused?

src/bsoncxx/lib/bsoncxx/v1/exception.cpp
```diff
@@ -16,6 +16,71 @@
...
+
+            switch (static_cast<code>(e)) {
+                case code::zero:
+                    return "zero";
+                case code::invalid_argument:
+                    return "invalid argument";
+                case code::runtime_error:
+                    return "runtime error";
+                default:
+                    return "unknown: " + std::to_string(e);
```
vector-of-bool, line 71: Minor: The format I've most often seen for "unknown" messages is `"<category-name>:<numeric-value>"`

src/bsoncxx/test/v1/oid.cpp
```diff
@@ -0,0 +1,192 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <bsoncxx/test/v1/oid.hh>
+
+//
+
+#include <bsoncxx/test/v1/exception.hh>
+
+#include <climits>
```
vector-of-bool: Unused header?

vector-of-bool (commented):
Mostly looks good, but leaving open for a behavioral question regarding iteration exception-throwing

src/bsoncxx/lib/bsoncxx/v1/element/view.cpp
eramongodb: Done.

I'd deliberately left it uninitialized since it's bytes are never meant to be accessed, so if they _are_ incorrectly accessed, it can be caught by UBSAN. Copying an object with indeterminate value in this manner shouldn't be UB, but it looks like the relevant wording is only applicable to [C++14 or newer](https://cplusplus.github.io/CWG/issues/1787.html) ([[dcl.init]](https://timsong-cpp.github.io/cppwp/n4140/dcl.init#12)), anyways, so I've zero-initialized the elements as suggested.

src/bsoncxx/lib/bsoncxx/v1/array/view.cpp
eramongodb: Good catch. Yes, it should throw `invalid_data`.

src/bsoncxx/lib/bsoncxx/v1/document/view.cpp
eramongodb: > May be useful to have a utility that advances an iterator and throws if it detects an error?

I think I'll leave it as-is for now to keep the `throw` expressions immediately visible in the relevant function's definitions.

src/bsoncxx/lib/bsoncxx/v1/types/id.cpp
eramongodb: Good catch. I'd completely overlooked this. The [v_noabi implementation](https://github.com/mongodb/mongo-cxx-driver/blob/939c33683944e0a8b3b78d1f4db13cb148392e59/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/types.cpp#L45) returns `"?"` for all values without a named enumerator. Updated the implementation so that `to_string(subtype)` now returns `"user"` for all values in the range `[0x80, 0xFF]`. Also updated related documentation to highlight that `k_user` applies to a _range_ of values, not just `0x80`.

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
eramongodb: Since `WSAGuard` is a no-op on non-Windows targets, I had opted to Rule of Zero the non-Windows definition, as otherwise some awkward special-casing is required.

When the copy+move decls are moved out of the conditional preprocessor block:

```
warning: aggregate initialization of type 'WSAGuard' with user-declared constructors is incompatible with C++20 [-Wc++20-compat]
   63 |     ((void)WSAGuard{}, bson_oid_init(&oid, nullptr));
      |                    ^~
```

-> when changed to `WSAGuard()`:

```
 error: no matching constructor for initialization of 'WSAGuard'
   63 |     ((void)WSAGuard(), bson_oid_init(&oid, nullptr));
      |            ^
```

-> when the default ctor is explicitly defaulted:

```
warning: member function 'WSAGuard' is not needed and will not be emitted [-Wunneeded-member-function]
   55 |         WSAGuard() = default;
      |         ^~~~~~~~
```

However, after applying the "magic static" suggestion, the entire class definition can be moved into the conditional preprocessor block, so ultimately no special-casing for non-Windows targets is necessary: `WSAGuard` simply doesn't exist for non-Windows.

src/bsoncxx/lib/bsoncxx/v1/oid.cpp
eramongodb: I (baselessly) assumed that the call to `WSAStartup()` is relatively cheap, but that might be incorrect. I had reservations with introducing more static variables to the C++ Driver, but since this particular case concerns runtime DLL-loading which is unconditionally required by mongoc anyways (in most cases which users will be using `oid` explicitly or implicitly), it should _probably_ be fine...

Updated the documentation for `v1::oid` to document this static local variable behavior.

src/bsoncxx/test/v1/types/value.cpp
eramongodb: Fixed: this should be `<cstdint>` for `UINT32_MAX`.

src/bsoncxx/lib/bsoncxx/v1/exception.cpp
eramongodb: Updated to use the suggested format, e.g. `bsoncxx::v1::document::view:-1`.

PR #1434, opened 2025-07-30
Enable CSFLE for sanitizer tasks

eramongodb:
Ensure CSFLE code paths are also covered by our sanitizer task coverage. Dropped shared library tasks due to redundancy (can switch to dropping static instead if preferable; I don't think there's a significant difference). Additionally added some log messages to help discern Catch test output by test executable.


eramongodb:
It looks like these changes also exposed [CDRIVER-5998](https://jira.mongodb.org/browse/CDRIVER-5998) on Windows distros. Created CXX-3321 to track this issue (for pending "bump to mongoc 2.1.0" ticket).

eramongodb:
Pinning mongoc to pre-2.1.0 release also brings in CDRIVER-5956. Bumped minimum server versions for tasks from 4.0 to 4.2 accordingly (CXX-3259).

.evergreen/config_generator/components/funcs/install_c_driver.py
```diff
@@ -14,7 +14,7 @@
 # Only MONGOC_DOWNLOAD_VERSION needs to be updated when pinning to an unreleased commit.
```
kevinAlbs: Noting: I expect setting `MONGOC_DOWNLOAD_VERSION` is not needed since most Evergreen tasks use `InstallCDriver` to install the C driver dependency, rather than rely on the C driver downloaded from CMake (which does not build with libmongocrypt).

.evergreen/config_generator/components/integration.py
```diff
@@ -24,8 +24,7 @@
 LINUX_MATRIX = [
     # Linux x86_64 (full).
     # RHEL 8 x86_64: 4.0+.
-    ('rhel80', None, ['Debug'], ['shared', 'static'], [11, 17], [None], ['plain', 'csfle'], [       '4.2', '4.4', '5.0', '6.0', '7.0', '8.0', 'latest'], ['single', 'replica', 'sharded']),
-    ('rhel80', None, ['Debug'], ['shared', 'static'], [11, 17], [None], ['plain',        ], ['4.0',                                                   ], ['single', 'replica', 'sharded']),  # CSFLE: 4.2+.
+    ('rhel80', None, ['Debug'], ['shared', 'static'], [11, 17], [None], ['plain', 'csfle'], ['4.2', '4.4', '5.0', '6.0', '7.0', '8.0', 'latest'], ['single', 'replica', 'sharded']),
```
kevinAlbs, line 28: Suggest also adding to CHANGELOG.md (though could also be done during next release):

```
### Removed
- Support for MongoDB Server 4.0 (due to the minimum required C Driver version).
```

kevinAlbs (approved):

PR #1436, opened 2025-08-04
CXX-3323 drop support for MongoDB Server 4.0 or older

eramongodb:
Resolves CXX-3323. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1429 and https://github.com/mongodb/mongo-cxx-driver/pull/1434.

https://github.com/mongodb/mongo-cxx-driver/pull/1434 already added the corresponding [CHANGELOG entry](https://github.com/mongodb/mongo-cxx-driver/blob/90ff27c0ec29fd756c890ba9ac50d8cc1096e4d5/CHANGELOG.md?plain=1#L19) and removed related EVG tasks.

Scans the codebase for all instances of MongoDB Server version checks against a value that corresponds to 4.0 or older and removes them. Most of these are in test code, but there are a few bits of library code as well, such as in `mongocxx::v_noabi::collection::_drop()` and an obsolete doc comment for `mongocxx::v_noabi::result::bulk_write::modified_count()`.

### DIFF
--- a/src/mongocxx/test/v_noabi/client_helpers.cpp
+++ b/src/mongocxx/test/v_noabi/client_helpers.cpp
@@ -231,7 +231,7 @@ std::int32_t compare_versions(std::string version1, std::string version2) {
     return 0;
 }
 
-bool newer_than(std::string version) {
+bool server_version_is_at_least(std::string version) {
     auto server_version = get_server_version();
     return (compare_versions(server_version, version) >= 0);
 }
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):
LGTM with some suggested additional removals:

- Calls `test_util::newer_than` (which may be better named to `test_util::equal_or_newer_than`) for "4.2" and older.
- ~~[This check](https://github.com/eramongodb/mongo-cxx-driver/blob/c7ea5a5c62fbf5c2f7e9cb4fa88347cfdc62f444/src/mongocxx/test/v_noabi/index_view.cpp#L185-L186) for `commitQuorum` support.~~

eramongodb:
> which may be better named to `test_util::equal_or_newer_than`

Opted for `test_util::server_version_is_at_least` to match its description.

PR #1437, opened 2025-08-05
CXX-2812 enable abi-compatibility-check for stable ABI

eramongodb:
Following https://github.com/mongodb/mongo-cxx-driver/pull/1412, which defines and exports the first v1 ABI symbol `bsoncxx::v1::exception::~exception()`, the `abi-compliance-check` can now support comparing `v1` symbols without errors.

Drive-by improvements include:

- Numerically comparing the [exit code](https://lvc.github.io/abi-compliance-checker/#Error) to distinguish ABI incompatibility (`== 1`) from tool errors (`> 1`).
- Exclude `mongocxx/v_noabi/config/*` headers for consistency with bsoncxx and to avoid triggering the new [v1 macro guard inclusion checks](https://github.com/mongodb/mongo-cxx-driver/blob/8cd8bdc65d5ee9e0a26e92d907c18f114a81afa4/src/mongocxx/include/mongocxx/v1/detail/postlude.hpp#L42) which are [inherited by mongocxx](https://github.com/mongodb/mongo-cxx-driver/blob/8cd8bdc65d5ee9e0a26e92d907c18f114a81afa4/src/mongocxx/include/mongocxx/v_noabi/mongocxx/config/postlude.hpp#L15).
- Exclude the `detail` namespace for both bsoncxx and mongocxx regardless of nesting.

### DIFF
--- a/.evergreen/scripts/abi-compliance-check-test.sh
+++ b/.evergreen/scripts/abi-compliance-check-test.sh
@@ -57,14 +57,15 @@ DOC
   {
     cat <<DOC
 <skip_including>
-  bsoncxx/v_noabi/bsoncxx/enums/
-  bsoncxx/v_noabi/bsoncxx/config/
+  /v_noabi/bsoncxx/enums/
+  /v_noabi/bsoncxx/config/
+  /v_noabi/mongocxx/config/
+  /v1/detail/prelude.hpp
+  /v1/detail/postlude.hpp
 </skip_including>
 
 <skip_namespaces>
-  bsoncxx::detail
-  bsoncxx::v_noabi::detail
-  bsoncxx::v_noabi::stdx::detail
+  detail
 </skip_namespaces>
 
 DOC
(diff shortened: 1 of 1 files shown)


.evergreen/scripts/abi-compliance-check-test.sh
```diff
@@ -117,18 +118,29 @@ fi
...
 echo "Generating stable ABI report..."
 pushd cxx-abi
-if ! abi-compliance-checker -lib mongo-cxx-driver -old old.xml -new new.xml 2>&1; then
-  : # CXX-2812: enable code below once stable ABI symbols exist in the base commit libraries.
-  # declare status
-  # status='{"status":"failed", "type":"test", "should_continue":true, "desc":"abi-compliance-checker emitted one or more errors"}'
-  # curl -sS -d "${status:?}" -H "Content-Type: application/json" -X POST localhost:2285/task_status || true
+declare ret
+abi-compliance-checker -lib mongo-cxx-driver -old old.xml -new new.xml 2>&1 && ret="$?" || ret="$?"
+if [[ "${ret:?}" -gt 0 ]]; then
```
kevinAlbs: Swap if blocks. I expect the `-gt 1` block will never be entered (since the `-gt 0` will be true too).

kevinAlbs (approved):
LGTM with swapping the if conditions.

PR #1438, opened 2025-08-06
 CXX-3324 remove all use of instance::current() 

eramongodb:
Resolves CXX-3224 by removing all use of `mongocxx::v_noabi::instance::current()` except in `src/mongocxx/test/v_noabi/instance.cpp` (to preserve _some_ test coverage of this API until its removal).

The `current()` function was added to `mongocxx::v_noabi::instance` a _long_ time ago by https://github.com/mongodb/mongo-cxx-driver/pull/441 (Jan 2016) to address ASAN and Valgrind memory leak errors. https://github.com/mongodb/mongo-cxx-driver/pull/513 (Sep 2016) later added the clarification that `current()` is "intended primarily for test authors". At this time, Catch was a header-only library, [specifically](https://github.com/mongodb/mongo-cxx-driver/blob/db943d54894f4a1bfa2f9f840f5150179e59b487/src/third_party/catch/include/catch.hpp) the "CATCH v1.0 build 53 (master branch)" [generated on August 2014](https://github.com/catchorg/Catch2/commit/544bf33e73e10ad074a410f2cb25c310351ebfac). This predates [the first stable release](https://github.com/catchorg/Catch2/releases/tag/v1.2.0) by nearly a year. The header was updated to v1.9.4 [on June 2017](https://github.com/mongodb/mongo-cxx-driver/commit/f3b449deff01e647c991743c68c8e9be1cf42d93), to v2.2.1 on [April 2018](https://github.com/mongodb/mongo-cxx-driver/commit/222e67ddee07feaf5f54da389245c87d07551621), then to v3.7.0 on [September 2024](https://github.com/mongodb/mongo-cxx-driver/commit/f183933f4dd8578dd3a1e13a719d342d8a3c9410). That is to say, the codebase has changed quite a lot since `current()` was added.

https://github.com/mongodb/mongo-cxx-driver/pull/1202 added a `catch.cpp` file providing a user-defined `main()` function as part of `__cdecl` checks on Windows (to avoid the Catch2-supplied `main()` from incorrectly inheriting the non-cdecl default calling convention). This `main()` function is the perfect place to create the `mongocxx::v_noabi::instance` object for subsequent use by the test suite. The only exceptions are the `test_instance` and `test_logging` executables which need to create their own instance object to test `mongocxx::v_noabi::instance` itself. All other cases of calling `current()` function is made obsolete, and therefore, removed by this PR.

It is premature to _remove_ `current()` entirely, given it's been presence in both the publically-documented API and ABI for a very long time. However, a deprecation note has been added to CHANGELOG declaring its restricted use and documentation updated to indicate the API is now "for internal use only". AFAICT there is no reason to continue supporting the `current()` function as part of the public API aside from its "initialize if not already" behavior (which is really only meant to be used for testing only), as there is no reason to obtain the "current" instance object for any subsequent purpose (the `instance` class has no meaningful API beyond its constructors + SMFs).

> [!NOTE]
> CXX-1029 proposes adding API to the `instance` class to support inspecting and modifying library logging behavior separately from the mongocxx (and mongoc) initialization/cleanup. We can consider re-introducing a `current()` function alongside such logging-related API, likely as an extension to the v1 API rather than to the v_noabi API.

After this PR, the only remaining use of the `current()` function is in `instance.cpp` to preserve test coverage of the function until it is ultimately removed in an upcoming API major version release. (The tests were also fixed to avoid assuming test-case-declaration execution order.) The `MONGOCXX_TEST_DISABLE_MAIN_INSTANCE` preprocessor macro is defined when building the `test_instance` and `test_logging` test executables to disable the creation of an instance object in `main()`; all other mongocxx test executables create an instance object in `main()` prior to invoking the Catch2 test runner.


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
   - Apple Clang 13.1 with Xcode 13.4.1 (from Apple Clang 5.1 with Xcode 5.1).
   - MSVC 19.0.24210 with Visual Studio 2015 Update 3 (from MSVC 19.0.23506 with Visual Studio 2015 Update 1).
 
+### Deprecated
+
+- `mongocxx::v_noabi::instance::current()` is "for internal use only". The `instance` constructor(s) should be used instead.
```
kevinAlbs, line 30: I see usage of `instance::current()` in [GitHub code search](https://github.com/search?type=code&q=mongocxx%3A%3Ainstance%3A%3Acurrent+NOT+is%3Afork+-repo%3Amongodb%2Fmongo-cxx-driver). I suspect it might be useful to conditionally initialize the C++ driver on first use (for example in an [extension](https://github.com/symbol/symbol/blob/71bd5f703004e9e50f3fe85e371919832ce7f5b7/client/catapult/extensions/mongo/MongoExtension.cpp#L89)).

Is there much harm in keeping `instance::current()`? If there are multiple users, I would rather leave it un-deprecated.

kevinAlbs (commented):
Changes look good. I vote for leaving `instance::current` un-deprecated.

CHANGELOG.md
eramongodb: > Is there much harm in keeping `instance::current()`?

The motivation is to discourage usage patterns which require global state (e.g. see [CDRIVER-1547](https://jira.mongodb.org/browse/CDRIVER-1547); I don't think we have a C++ Driver equivalent ticket yet). The behavior of `instance::current()` is fundamentally global (the function static local variable is an implementation detail). By deprecating and removing this function, users will be forced to explicitly decide their `instance` object allocation strategy; they _have_ to create an `instance` object somewhere. This will ultimately provide us with better opportunities to remove global state (by moving state into the `instance` object) while preserving global mongoc initialization/cleanup behavior _internally_ rather than via a dedicated API (that is, we reduce the scope of the [current global atomic variable](https://github.com/mongodb/mongo-cxx-driver/blob/c39093aa699431dd954cce4b8e617ebcb4e41d37/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/instance.cpp#L80) to handling one-time mongoc init/cleanup only, per [CXX-1029](https://jira.mongodb.org/browse/CXX-1029), until mongoc also hopefully/eventually removes its global state requirements).

The code search results appear to support the case that _most_ users can easily substitute `mongocxx::instance::current();` with `static mongocxx::instance current;`, as all the projects appear to have [only _one_ substantial call](https://github.com/search?type=code&q=mongocxx%3A%3Ainstance%3A%3Acurrent+NOT+is%3Afork+-repo%3Amongodb%2Fmongo-cxx-driver+NOT+path%3Atest+NOT+path%3Aexample+lang%3Acpp) to `current()` within their the codebase (that is not test or example code).

kevinAlbs: > most users can easily substitute `mongocxx::instance::current();` with `static mongocxx::instance current;`

Good point. Suggest expanding the changelog entry to help users migrate. Example:

> If an application only calls `mongocxx::instance::current();` once, it can likely be replaced with `static mongocxx::instance current;`:
> ```
> // Before:
> void handler () {
>     mongocxx::instance::current();
>     // ... code using mongocxx ...
> }
> // After:
> void handler () {
>     static mongocxx::instance current;
>     // ... code using mongocxx ...
> }
> ```

kevinAlbs (approved):

PR #1447, opened 2025-08-20
CXX-3234 refactor bsoncxx::v_noabi to use bsoncxx::v1

eramongodb:
## Summary

Resolves [CXX-3234](https://jira.mongodb.org/browse/CXX-3234). Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1430 and https://github.com/mongodb/mongo-cxx-driver/pull/1445.

This is 4 out of an estimated 7 major PRs which in total are expected to resolve CXX-2745 and CXX-3320

- [x] Top-Level Dependencies (CXX-3239)
- [x] bsoncxx v1 Interfaces (CXX-3232)
- [x] bsoncxx v1 Implementations (CXX-3233) 
- [ ] bsoncxx v_noabi Refactor (CXX-3234) (<-- you are here)
- [ ] mongocxx v1 Interfaces (CXX-3236)
- [ ] mongocxx v1 Implementations (CXX-3237)
- [ ] mongocxx v_noabi Refactor (CXX-3238)

This PR refactors the bsoncxx v_noabi interfaces and implementations to:

- reduce code duplication between v1 and v_noabi,
- document-as-code the discrepancies between v1 and v_noabi behavior,
- indirectly test v1 API using the existing v_noabi test suite, and
- support v1 <-> v_noabi conversion functions to allow incremental migration from v_noabi to v1.

> [!IMPORTANT]
> There should be **_no API breaking changes_** in this refactor!

Unlike prior bsoncxx v1 PRs thus far, this PR is the first case where the changes to v\_noabi and v1 are _not_ an example of how similar changes would be implemented for v1 and v2, since v1 is a _stable_ ABI namespace. This refactor takes advantage of v\_noabi's lack of ABI stability guarantees to reduce code duplication and re-implement v\_noabi in terms of v1 as much as possible. However, when v2 API is introduced, v1 will not be able to be refactored to the same extent, although ABI+API compatible refactors would of course still be permitted.

> [!NOTE]
> Extending support for `BSONCXX_API_OVERRIDE_DEFAULT_ABI` is _not_ in scope for this PR. All changes to the v\_noabi test suite in this PR are _additions only_ (excluding CXX-2120-related error message assertions), meaning there should be _no API breaking changes_. Extending support for `BSONCXX_API_OVERRIDE_DEFAULT_ABI` will necessarily need to be accompanied by corresponding changes to test code to support the API breaking change caused by the changes to root namespace redeclarations.

Per local testing, this PR reduces the ABI footprint for `bsoncxx::v_noabi` from 326 symbols to 169 (-48.2%). Excluding the symbols corresponding to components refactored by this PR (which have v1 counterparts), remaining symbols include:

* `bsoncxx::v_noabi::builder::*`: pending CXX-3275
* `bsoncxx::v_noabi::*_json()` + related: out-of-scope for initial v1 implementation.
* `bsoncxx::v_noabi::*view_or_value`: to be removed in favor of CXX-1546, CXX-1827, etc.
* `bsoncxx::v_noabi::validate` + related: out-of-scope for initial v1 implementation.
* `bsoncxx::v_noabi::vector`: out-of-scope for initial v1 implementation.

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v1/oid.hpp
+++ b/src/bsoncxx/include/bsoncxx/v1/oid.hpp
@@ -193,9 +193,9 @@ class oid {
     class internal;
 
    private:
-    struct uninit_type {};
+    struct for_overwrite_tag {};
 
-    oid(uninit_type) : _bytes{} {} // Required by bsoncxx::v_noabi::oid::oid().
+    /* explicit(false) */ oid(for_overwrite_tag) : _bytes{} {}
 };
 
 } // namespace v1
--- a/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp
+++ b/src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp
@@ -38,7 +38,7 @@ static_assert(is_implicitly_convertible<v1::oid const&, oid>::value, "v1 -> v_no
 
 constexpr std::size_t oid::k_oid_length;
 
-oid::oid() : _oid{v1::oid::internal::make_oid_without_init()} {
+oid::oid() : _oid{v1::oid::internal::make_oid_for_overwrite()} {
     try {
         _oid = v1::oid{};
     } catch (...) {
--- a/src/bsoncxx/lib/bsoncxx/v1/oid.hh
+++ b/src/bsoncxx/lib/bsoncxx/v1/oid.hh
@@ -30,8 +30,8 @@ class oid::internal {
     }
 
     // Required by bsoncxx::v_noabi::oid::oid().
-    static oid make_oid_without_init() {
-        return {oid::uninit_type{}};
+    static oid make_oid_for_overwrite() {
+        return {oid::for_overwrite_tag{}};
     }
 };
 


src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/exception/error_code.cpp
eramongodb, line 1: Defining error condition mappings for `v_noabi::error_code` <-> `v1::*::errc` was considered but rejected due to too much effort and complexity for an API (v_noabi) which we intend to deprecate and remove (in favor of v1). It will also be more reasonable to consider defining mappings between component-specific `v1::*::errc` error codes and _also_ component-specific `v2::*::errc` in the future, than to define mappings to a library-wide `v_noabi::error_code`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/exception/exception.hpp
eramongodb, line 1: Unfortunately, there is no way to general-case-extendably arrange for the v_noabi exception class to IS-A the v1 exception class (or for v1 exceptions to inherit v2 exceptions, etc.). As demonstrated by changes in this PR, any v_noabi exception boundaries must manually catch and rethrow potential v1 exceptions as a v_noabi exception instead to honor their documented "exception thrown by \<abi\>"  contract. Users writing against both v_noabi and v1 API will need to catch `std::system_error` instead of either `v_noabi::exception` or `v1::exception` to handle both cases. Perhaps we can/should consider removing these exception classes in favor of throwing generic `std::system_error` with error codes instead...

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
```diff
@@ -36,43 +34,19 @@ namespace elements {
 
 /// @brief A 32-bit float value in packed little-endian format
 class float32 {
+    template <bsoncxx::detail::endian = bsoncxx::detail::endian::native>
+    void construct(float value);
+
+    template <bsoncxx::detail::endian = bsoncxx::detail::endian::native>
+    float convert() const;
```
eramongodb, line 41: The new [v1/detail/bit.hpp](https://github.com/mongodb/mongo-cxx-driver/blob/19bf883c43ccbe5d1f2c443eab3c6011dbb86299/src/bsoncxx/include/bsoncxx/v1/detail/bit.hpp) header is used to provide endianness checks. Instead of preprocessor macros, the compile-time conditional branch is implemented as template specializations (hopefully eventually replaced with C++17 `if constexpr`).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128-fwd.hpp
```diff
@@ -36,3 +38,6 @@ using ::bsoncxx::v_noabi::decimal128;
 /// @file
 /// Declares @ref bsoncxx::v_noabi::decimal128.
 ///
+/// @par Includes
+/// - @ref bsoncxx/v1/decimal128-fwd.hpp
```
eramongodb, line 42: Related to CXX-3119: all v_noabi headers publically document the inclusion of v1 equivalent headers. This is not only due to the refactor, but also to set precedent for v_noabi headers to continue to provide root namespace redeclarations for users who do not require ABI stability.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -14,11 +14,16 @@
 
 #pragma once
 
+#include <bsoncxx/decimal128-fwd.hpp>
+
+//
+
+#include <bsoncxx/v1/decimal128.hpp>
+#include <bsoncxx/v1/detail/type_traits.hpp>
```
eramongodb, line 22: To keep the scope of potential breaking changes minimal, this PR refrains from changing (removing) include directives even when undocumented. Following releases which publically document transitive includes, the next API major release will hopefully take the opportunity to perform a more thorough header hygiene cleanup.

That being said, component headers _are_ reordered so they are always included first before others.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -38,12 +43,20 @@ namespace v_noabi {
 /// - [BSON Decimal128 (MongoDB Specifications)](https://specifications.readthedocs.io/en/latest/bson-decimal128/decimal128/)
 ///
 class decimal128 {
+   private:
+    v1::decimal128 _d128;
```
eramongodb, line 47: Taking advantage of the lack of ABI stability guarantees, classes in v_noabi are redefined in terms of v1 as data members (composition over inheritance). This level of refactor will not be viable for stable ABI namespaces going forward (e.g. v1 cannot be refactored in terms of v2 to this extent).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -38,12 +43,20 @@ namespace v_noabi {
...
    public:
     ///
     /// Constructs a BSON Decimal128 value representing zero.
     ///
     decimal128() = default;
 
+    ///
+    /// Construct with the @ref bsoncxx::v1 equivalent.
+    ///
+    /* explicit(false) */ decimal128(v1::decimal128 const& d128) noexcept : _d128{d128} {}
```
eramongodb, line 58: All v\_noabi classes (except for `v_noabi::document::value`) define an _implicit_ converting constructor from v1 to v_noabi. This is to facilitate incremental migration from v\_noabi to v1. See static assertions in the `.cpp` file.

> [!NOTE]
> `v1::decimal128` is currently the only bsoncxx class which is ready for `BSONCXX_API_OVERRIDE_DEFAULT_ABI` (that is, no v_noabi -> v1 API breaking changes). The intent and behavior of these implicit conversion functions can be verified by manually changing the root namespace redeclaration in `src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128-fwd.hpp` from `v_noabi::decimal128` to `v1::decimal128`: all existing code should still be well-formed despite `v1::decimal128` being used with API expecting `v_noabi::decimal128`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -63,45 +76,71 @@ class decimal128 {
...
     ///
-    explicit BSONCXX_ABI_EXPORT_CDECL() decimal128(stdx::string_view str);
+    explicit BSONCXX_ABI_EXPORT_CDECL() decimal128(v1::stdx::string_view str);
+
+    ///
+    /// Convert to the @ref bsoncxx::v1 equivalent.
+    ///
+    explicit operator v1::decimal128() const noexcept {
+        return _d128;
+    }
```
eramongodb, line 86: All v\_noabi classes define an _explicit_ user-defined conversion function from v_noabi to v1. This is to facilitate incremental migration from v\_noabi to v1. See static assertions in the `.cpp` file.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -63,45 +76,71 @@ class decimal128 {
...
 
     ///
     /// Converts this decimal128 value to a string representation.
     ///
     /// @return A string representation of a IEEE 754-2008 decimal number.
     ///
-    BSONCXX_ABI_EXPORT_CDECL(std::string) to_string() const;
+    std::string to_string() const {
+        return _d128.to_string();
```
eramongodb, line 94: All v\_noabi behavior is implemented in terms of the v1 API as much as possible (that is, in terms of the v1 data member).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -63,45 +76,71 @@ class decimal128 {
...
 
     ///
-    /// Accessor for high 64 bits.
+    /// @relates bsoncxx::v_noabi::decimal128
+    ///
+    /// Equality comparison operator.
     ///
-    uint64_t high() const {
-        return _high;
+    friend bool operator==(decimal128 const& lhs, decimal128 const& rhs) {
```
eramongodb, line 116: All API which do not need to be exported are made inline to reduce the v\_noabi ABI footprint.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128.hpp
```diff
@@ -63,45 +76,71 @@ class decimal128 {
...
+inline v_noabi::decimal128 from_v1(v1::decimal128 const& v) {
+    return {v};
+}
+
+///
+/// Convert to the @ref bsoncxx::v1 equivalent of `v`.
+///
+inline v1::decimal128 to_v1(v_noabi::decimal128 const& v) {
+    return v1::decimal128{v};
+}
```
eramongodb, line 142: All v_noabi classes are also given explicit `from_v1()` and `to_v1()` conversion functions declared in the v_noabi namespace (except for `v_noabi::*::element`). This is the from/to API described by [this comment](https://github.com/mongodb/mongo-cxx-driver/pull/1182#discussion_r1704329261). These functions account for converting between non-class types (e.g. `bsoncxx::v_noabi::type`) and class types which cannot provide a converting ctor (e.g. `bsoncxx::v_noabi::document::value`). When newer ABI namespaces are added to the library (e.g. v2), similar conversion functions can be defined for equivalent class types (e.g. `from_v2()`/`to_v2()`). These conversion functions can be removed once the v_noabi API is changed to be defined completely in terms of the v1 API.

The incremental library ABI evolution path looks as follows:

```cpp
// v_noabi (default) and v1 (new).
namespace bsoncxx {
  namespace v1 { class decimal128 { ... }; }
  namespace v_noabi {
    class decimal128 { ... };
    v_noabi::decimal128 from_v1(v1::decimal128 v);
    v1::decimal128 to_v1(v_noabi::decimal128 v);
  }
  using v_noabi::decimal128;
}
```
```cpp
// v_noabi (deprecated) and v1 (default)
namespace bsoncxx {
  namespace v1 { class decimal128 { ... }; }
  namespace v_noabi {
    class decimal128 { ... }; // Deprecated.
    v_noabi::decimal128 from_v1(v1::decimal128 v);
    v1::decimal128 to_v1(v_noabi::decimal128 v);
  }
  using v1::decimal128; // <-- API +1: v1 is now the default.
}
```
```cpp
// v_noabi (removed), v1 (only)
namespace bsoncxx {
  namespace v1 { class decimal128 { ... }; }
  using v1::decimal128;
}
```

Going forward, for stable ABI namespaces, the final "removed" step would also require bumping the ABI version number. However, this is not required for the removal of v_noabi API (it is not part of the stable ABI).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/oid.hpp
```diff
@@ -63,14 +76,23 @@ class oid {
     /// @throws bsoncxx::v_noabi::exception if the string isn't an OID-sized hex
     /// string.
     ///
-    explicit BSONCXX_ABI_EXPORT_CDECL() oid(stdx::string_view const& str);
+    explicit BSONCXX_ABI_EXPORT_CDECL() oid(v1::stdx::string_view const& str);
```
eramongodb, line 79: This PR makes an effort to use the v1 API whenever able (stable <- unstable) even for stdx polyfills.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/element.hpp
```diff
@@ -44,271 +50,122 @@ namespace document {
...
-    /// @throws bsoncxx::v_noabi::exception if this element is not a b_maxkey.
+    /// @throws bsoncxx::v_noabi::exception if this element is not the requested type.
     ///
-    /// @return the element's value.
+    /// @{
+    BSONCXX_V1_TYPES_XMACRO(X)
+    /// @}
     ///
-    BSONCXX_ABI_EXPORT_CDECL(types::b_maxkey) get_maxkey() const;
+#pragma pop_macro("X")
```
eramongodb, line 152: The new `BSONCXX_V1_TYPES_XMACRO` pattern is used to reduce verbosity that was formerly required to preserve Doxygen documentation, which was incompatible with the `#include`-based X macro pattern.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/element.hpp
```diff
@@ -324,7 +181,7 @@ class element {
     ///
     /// @return The matching element, if found, or an invalid element.
     ///
-    BSONCXX_ABI_EXPORT_CDECL(element) operator[](stdx::string_view key) const;
+    BSONCXX_ABI_EXPORT_CDECL(element) operator[](v1::stdx::string_view key) const;
```
eramongodb, line 184: This function (+ those involving `v_noabi::types::bson_value::*` and `v_noabi::array::element`) remain exported functions to avoid including non-forward component headers (which causes a cyclic dependency).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/element.hpp
```diff
@@ -374,16 +208,22 @@ class element {
 /// @{
 
 /// @relatesalso bsoncxx::v_noabi::document::element
-BSONCXX_ABI_EXPORT_CDECL(bool) operator==(element const& elem, types::bson_value::view const& v);
+BSONCXX_ABI_EXPORT_CDECL(bool) operator==(element const& lhs, v_noabi::types::bson_value::view const& rhs);
```
eramongodb, line 211: For backward compatibility, operators declared outside the class are preserved as-is, even if they should really be hidden friends.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/element.hpp
```diff
@@ -392,11 +232,27 @@ BSONCXX_ABI_EXPORT_CDECL(bool) operator!=(types::bson_value::view const& v, elem
 } // namespace v_noabi
 } // namespace bsoncxx
 
+namespace bsoncxx {
+namespace v_noabi {
+
+// Ambiguous whether `v1::element::view` should be converted to `v1::array::element` or `v1::document::element.`
+// v_noabi::document::element from_v1(v1::element::view const& v);
```
eramongodb: We cannot define `from_v1(v1::element::view const&)` due to the result type being ambiguous between `v_noabi::document::element` and `v_noabi::array::element`. We _could_, say, default to `v_noabi::document::element`, but this PR proposes leaving the decision to the user (via explicit converting ctors) instead of making the choice for them.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
```diff
@@ -36,50 +41,41 @@ namespace array {
 /// interrogated by calling type() and a specific value can be extracted through
 /// get_X() accessors.
 ///
-class element : private document::element {
+class element : private v_noabi::document::element {
```
eramongodb, line 44: For backward compatibility, the existing structure where "`v_noabi::array::*` is defined in terms of `v_noabi::document::*`" is preserved as-is.

As far as I can tell, CXX-1830 can be resolved per its current description: there is no difference between the API for `v_noabi::array::element` and `v_noabi::document::element` (all public member functions are using-declared) and v1 removes the distinction completely with `v1::element::view`. Instead, CXX-1830 is repurposed to also track `v_noabi::array::*` API consistency as a whole with `v_noabi::document::*`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
```diff
@@ -90,16 +86,24 @@ class element : private document::element {
 /// @{
 
 /// @relatesalso bsoncxx::v_noabi::array::element
-BSONCXX_ABI_EXPORT_CDECL(bool) operator==(element const& elem, types::bson_value::view const& v);
+inline bool operator==(element const& lhs, v_noabi::types::bson_value::view const& rhs) {
+    return static_cast<v_noabi::document::element const&>(lhs) == rhs;
```
eramongodb, line 90: Take advantage of (private) inheritance to defer implementation to `v_noabi::document::element`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -33,6 +38,10 @@ namespace document {
 /// A read-only, non-owning view of a BSON document.
 ///
 class view {
+   private:
+    v1::document::view _view;
+    std::size_t _length;
```
eramongodb, line 43: Following PR feedback in https://github.com/mongodb/mongo-cxx-driver/pull/1412, the v1 API removed this `_length` field in favor of using the "embedded" length field within BSON data. However, For backward compatibility, v_noabi must preserve this separate length field in spite of v1 behavior. Because the potential set of valid values for `_length` is _greater_ than that for v1, the v_noabi class(es) must take care not to invoke v1 API in contexts where v1's length checks may trigger unwanted (new!) exceptions. These special cases have been labeled with "Do NOT use ..." comments to clearly indicate intentional avoidance of v1 API behavior.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -52,7 +66,22 @@ class view {
...
+    /// - If `this->data()` is not null, the size of the storage region pointed to by `data` must be greater than or
+    ///   equal to 5.
+    /// - The "total number of bytes comprising the document" as indicated by the BSON bytes pointed-to by
+    ///   `this->data()` must be less than or equal to the size of the storage region pointed to by `data`.
+    ///
+    /// @note `this->size()` is ignored.
+    ///
+    explicit operator v1::document::view() const {
+        return _view;
+    }
```
eramongodb, line 96: Although the operation is simple, the contract is not. To avoid runtime overhead, the v_noabi -> v1 conversion inherits the same preconditions as v1's constructors rather than checking for the validity of the embedded length (e.g. as is done for v1's [data+length ctors](https://github.com/mongodb/mongo-cxx-driver/blob/19bf883c43ccbe5d1f2c443eab3c6011dbb86299/src/bsoncxx/include/bsoncxx/v1/document/view.hpp#L97-L98)).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -62,17 +91,17 @@ class view {
     ///
     /// @returns A const_iterator to the past-the-end element of the document.
     ///
-    BSONCXX_ABI_EXPORT_CDECL(const_iterator) cend() const;
+    const_iterator cend() const;
```
eramongodb, line 106: Despite inline definitions, these functions must be declared out-of-line after `const_iterator` is a complete type.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -118,30 +149,40 @@ class view {
     ///
     /// @return The length of the document, in bytes.
     ///
-    BSONCXX_ABI_EXPORT_CDECL(std::size_t) length() const;
+    std::size_t size() const {
+        return _length; // Do NOT use _view.size().
+    }
```
eramongodb, line 165: A small drive-by API extension to support both `size()` and `length()`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -151,6 +192,9 @@ class view {
 /// view elements.
 ///
 class view::const_iterator {
+   private:
+    v_noabi::document::element _element;
```
eramongodb, line 207: For backward compatibility, the iterator classes are _not_ (re)defined in terms of v1 equivalents and remain separate types (both for document and for array).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
```diff
@@ -161,37 +205,89 @@ class view::const_iterator {
     using iterator_category = std::forward_iterator_tag;
     using difference_type = std::ptrdiff_t;
 
-    BSONCXX_ABI_EXPORT_CDECL() const_iterator();
-    explicit BSONCXX_ABI_EXPORT_CDECL() const_iterator(element const& element);
+    const_iterator() = default;
+
+    explicit const_iterator(v_noabi::document::element const& element) : _element(element) {}
+
+    reference operator*() {
```
eramongodb, line 223: For backward compatibility, pre-existing API deficiencies (i.e. lack of `const`) are left as-is.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -37,9 +41,38 @@ namespace document {
...
+    ///
+    /// The type of the unique pointer used to manage the underlying BSON binary data.
+    ///
+    using unique_ptr_type = std::unique_ptr<std::uint8_t[], deleter_type>;
+
+    /// @copydoc v_noabi::document::view::const_iterator
+    using const_iterator = v_noabi::document::view::const_iterator;
+
+    /// @copydoc v_noabi::document::view::iterator
+    using iterator = const_iterator;
```
eramongodb, line 63: Drive-by API extension for better consistency with both `v1::document::value` and `v_noabi::document::view`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -37,9 +41,38 @@ namespace document {
...
+
+    ~value() = default;
+
+    value(value&&) = default;
+    value& operator=(value&&) = default;
```
eramongodb, line 68: For backward compatibility, the post-move `_length` is left as-is (not zeroed).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -74,70 +108,103 @@ class value {
...
+    ///
+    /// Assignment used for serialization of user objects. This uses argument-dependent lookup
+    /// to find the function declaration
+    /// `void to_bson(T& t, bsoncxx::v_noabi::document::value doc)`.
+    ///
+    /// @note `T` is not constrained!
+    ///
+    /// @param t
+    ///   A user-defined object to serialize into a BSON object.
+    ///
```
eramongodb, line 147: For backward compatibility, the lack of constraints on the template parameter here and for `operator=(T const& t)` below are preserved as-is. Instead, their documentation has been improved to clarify the situation.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -74,70 +108,103 @@ class value {
...
+    /// @par Preconditions:
+    /// - If `this->data()` is not null, the size of the storage region pointed to by `data` must be greater than or
+    ///   equal to 5.
+    /// - The "total number of bytes comprising the document" as indicated by the BSON bytes pointed-to by
+    ///   `this->data()` must be less than or equal to the size of the storage region pointed to by `data`.
+    ///
+    /// @note `this->size()` is ignored.
+    ///
+    explicit operator v1::document::value() && {
+        _length = 0u;
```
eramongodb, line 181: Unlike the v_noabi move SMFs (which must preserve backward compatibility), the v_noabi -> v1 conversion function zeroes the moved-from `_length` field. This behavior can be removed if consistency with the pre-existing move SMFs is preferrable.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -231,19 +307,25 @@ class value {
...
-    BSONCXX_ABI_EXPORT_CDECL(unique_ptr_type) release();
+    unique_ptr_type release() {
+        auto ptr = _value.release();
+
+        // Invariant: the underlying deleter type MUST be `deleter_type`.
+        auto const deleter_ptr = ptr.get_deleter().target<deleter_type>();
+
+        // Invariant: `ptr` implies `deleter_ptr`, but not the reverse.
+        return {ptr.release(), deleter_ptr ? *deleter_ptr : nullptr};
+    }
```
eramongodb, line 329: Because `deleter_type` is and has always been `void(BSONCXX_ABI_CDECL*)(std::uint8_t*)`, the only situation when `deleter_ptr` is _not_ `deleter_type` is when there is no owned BSON data (`_value == nullptr`).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
```diff
@@ -72,28 +102,116 @@ class value {
...
+
+    ///
+    /// Return true when `this->length() == 5`.
+    ///
+    /// @warning For backward compatibility, this function does NOT check if the underlying BSON bytes represent a valid
+    /// empty document.
+    ///
+    bool empty() const {
+        return _length == 5; // Do NOT use _value.empty().
+    }
```
eramongodb: Drive-by API extension for better consistency with both `v1::array::value` and `v_noabi::document::value` to fulfill the intent of CXX-1830.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/view.hpp
```diff
@@ -53,266 +58,78 @@ namespace bson_value {
...
-    BSONCXX_ABI_EXPORT_CDECL(b_string const&) get_string() const;
-
-    ///
-    /// Returns the underlying BSON document value.
-    ///
-    /// @warning
-    ///   Calling the wrong get_<type> method will cause an exception to be thrown.
-    ///
-    BSONCXX_ABI_EXPORT_CDECL(b_document const&) get_document() const;
+    view() noexcept : _view{} {}
```
eramongodb: For backward compatibility, this `noexcept` default constructor cannot be defaulted due to `v1::types::view::view()` not being `noexcept`. However, the other SMFs can be removed in favor of Rule of Zero, including compatibility with `noexcept`, as `v1::types::view` is semitrivial.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value-fwd.hpp
```diff
@@ -24,6 +26,9 @@ namespace bson_value {
 class value;
 
 } // namespace bson_value
+
+using bson_value::value;
```
eramongodb, line 30: Following https://github.com/mongodb/mongo-cxx-driver/pull/1254, we can reintroduce `v_noabi::types::value` (formerly an alias for `v_noabi::types::bson_value::view`) as a proper alias for `v_noabi::types::bson_value::value` instead. This will help users with incremental v_noabi -> v1 migration, such as via root namespace redeclarations.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
```diff
@@ -46,99 +53,114 @@ namespace bson_value {
...
+    /* explicit(false) */ value(v1::types::value v) : _value{std::move(v)} {}
+
+    ///
+    /// Construct with the @ref bsoncxx::v1 equivalent.
+    ///
+    /// @par Constraints:
+    /// - `from_v1(v)` is found via ADL.
+    ///
+    template <typename T, detail::enable_if_t<has_from_v1<T>::value>* = nullptr>
+    /* explicit(false) */ value(T v) : value{from_v1(v)} {}
```
eramongodb, line 84: `v_noabi::types::bson_value::value` is a bit special: due to all its converting constructors being _implicit_ (unlike in `v1` which made them all _explicit_), this extra converting constructor template is required to support implicit convertion from `v1::types::b_*` for parity with implicit conversion from `v_noabi::types::bson_value::b_*`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
```diff
@@ -148,8 +170,9 @@ class value {
     /// @param sub_type
     ///     an optional binary sub type. Defaults to type::k_binary
     ///
-    BSONCXX_ABI_EXPORT_CDECL()
-    value(std::vector<unsigned char> v, binary_sub_type const sub_type = {});
+    /* explicit(false) */
+    value(std::vector<unsigned char> v, v_noabi::binary_sub_type const sub_type = {})
+        : _value{reinterpret_cast<std::uint8_t const*>(v.data()), v.size(), to_v1(sub_type)} {}
```
eramongodb, line 175: Relative breaking change: v_noabi is `std::vector<unsigned char>`, but v1 is `std::vector<std::uint8_t>`.

src/bsoncxx/lib/bsoncxx/private/convert.hh
eramongodb, line 1: All required conversions to/from `bson_value_t` are now handled by `v1::types::view` and `v1::types::value`.

src/bsoncxx/lib/bsoncxx/private/type_traits.hh
```diff
@@ -75,4 +75,15 @@ struct is_semitrivial : detail::conjunction<
...
+// Equivalent to `is_constructible_v<To, From> && !is_convertible_v<From, To>`.
+template <typename From, typename To>
+struct is_explicitly_convertible : bsoncxx::detail::conjunction<
+                                       std::is_constructible<To, From>,
+                                       bsoncxx::detail::negation<std::is_convertible<From, To>>> {};
+
+// Equivalent to `is_constructible_v<To, From> && is_convertible_v<From, To>`.
+template <typename From, typename To>
+struct is_implicitly_convertible
+    : bsoncxx::detail::conjunction<std::is_constructible<To, From>, std::is_convertible<From, To>> {};
```
eramongodb: To simplify and name the repetitious static assertions for implicit vs. explicit constructibility.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/exception/error_code.cpp
```diff
@@ -96,8 +100,8 @@ class error_category_impl final : public std::error_category {
 } // namespace
 
 std::error_category const& error_category() {
-    static error_category_impl const instance{};
-    return instance;
+    static bsoncxx::immortal<error_category_impl> const instance;
+    return instance.value();
```
eramongodb, line 114: Minor drive-by improvement: this removes exit-time destruction handlers for this instance object.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/decimal128.cpp
```diff
@@ -13,40 +13,28 @@
...
+static_assert(is_explicitly_convertible<decimal128&&, v1::decimal128>::value, "v_noabi -> v1 must be explicit");
+static_assert(is_explicitly_convertible<decimal128 const&, v1::decimal128>::value, "v_noabi -> v1 must be explicit");
+static_assert(is_implicitly_convertible<v1::decimal128&&, decimal128>::value, "v1 -> v_noabi must be implicit");
+static_assert(is_implicitly_convertible<v1::decimal128 const&, decimal128>::value, "v1 -> v_noabi must be implicit");
 
-bool operator!=(decimal128 const& lhs, decimal128 const& rhs) {
-    return !(lhs == rhs);
+decimal128::decimal128(v1::stdx::string_view str) try : _d128{str} {
+} catch (v1::exception const&) {
+    throw v_noabi::exception{v_noabi::error_code::k_invalid_decimal128};
```
eramongodb, line 37: This is an example of how v1 exceptions must be translated into v_noabi equivalents at exception boundaries. To avoid unnecessarily impacting current error messages, this PR proposes merely rethrowing equivalent error codes rather than capturing (often equivalent/redundant) exception messages during the translation process.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp
```diff
@@ -12,93 +12,53 @@
...
+    } catch (...) {
+        // For backward compatibility, ignore any exceptions and initialize anyways.
+        bson_oid_t oid;
+        bson_oid_init(&oid, nullptr);
 
-oid::oid(char const* bytes, std::size_t len) {
-    if (len != this->size()) {
-        throw bsoncxx::v_noabi::exception{error_code::k_invalid_oid};
+        auto& _bytes = v1::oid::internal::bytes(_oid);
+        std::memcpy(_bytes.data(), oid.bytes, sizeof(oid.bytes));
```
eramongodb, line 50: Relative breaking change: v1 may throw due to `WSAGuard`, but v_noabi never threw. The `make_oid_without_init()` function is used to preserve old behavior in the presence of potential exceptions.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/oid.cpp
```diff
@@ -12,93 +12,53 @@
...
-bool operator==(oid const& lhs, oid const& rhs) {
-    return oid_compare(lhs, rhs) == 0;
+oid::oid(stdx::string_view const& str) try : _oid{str} {
+} catch (v1::exception const&) {
+    throw v_noabi::exception{v_noabi::error_code::k_invalid_oid};
 }
 
-bool operator!=(oid const& lhs, oid const& rhs) {
-    return oid_compare(lhs, rhs) != 0;
+oid::oid(char const* bytes, std::size_t len) try : _oid{reinterpret_cast<std::uint8_t const*>(bytes), len} {
```
eramongodb, line 59: Relative breaking change: v_noabi is `char const*`, but v1 is `std::uint8_t const*`.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/element.cpp
```diff
@@ -28,120 +31,83 @@
...
+static_assert(is_implicitly_convertible<v1::element::view const&, element>::value, "v1 -> v_noabi must be implicit");
+static_assert(is_explicitly_convertible<element&&, v1::element::view>::value, "v_noabi -> v1 must be explicit");
+static_assert(is_explicitly_convertible<element const&, v1::element::view>::value, "v_noabi -> v1 must be explicit");
+
+v_noabi::type element::type() const try {
+    return static_cast<v_noabi::type>(_view.type_id());
+} catch (v1::exception const& ex) {
+    throw v_noabi::exception{
+        v_noabi::error_code::k_unset_element,
+        std::string{"cannot return the type of uninitialized element: "} + ex.what()};
```
eramongodb, line 50: As part of CXX-2120 and followup to https://github.com/mongodb/mongo-cxx-driver/pull/1445, this PR changes the exception messages for `v_noabi::document::element` to use the "last known element key" approach for improving diagnostics of accessing invalid elements. This corresponds to the propagation of the v1 exception message during v_noabi exception translation (the `+ ex.what()`). This applies to all `v_noabi::document::element` functions which may throw `v1::element::view::errc::invalid_view`. See the new+updated exception messages in `src/bsoncxx/test/v_noabi/bson_types.cpp`.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp
```diff
@@ -12,161 +12,91 @@
...
+static_assert(is_implicitly_convertible<v1::document::view&&, view>::value, "v1 -> v_noabi must be implicit");
+static_assert(is_implicitly_convertible<v1::document::view const&, view>::value, "v1 -> v_noabi must be implicit");
+
 namespace {
 
-bson_iter_t to_bson_iter_t(element e) {
+bson_iter_t to_bson_iter_t(v_noabi::document::element e) {
     bson_iter_t iter{};
-    bson_iter_init_from_data_at_offset(&iter, e.raw(), e.length(), e.offset(), e.keylen());
+    (void)bson_iter_init_from_data_at_offset(&iter, e.raw(), e.length(), e.offset(), e.keylen());
```
eramongodb, line 42: This PR avoids attempting to extend the error handling improvements in v1 API to the v_noabi implementation beyond minimal requirements, e.g. unchecked return values remain unchecked. However, we have the option to convert UB (i.e. termination) into exceptions (well-defined) as a "bugfix" if preferable.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp
```diff
@@ -12,161 +12,91 @@
...
-    return !(lhs == rhs);
-}
-
 view::const_iterator view::cbegin() const {
     bson_iter_t iter;
 
-    if (!bson_iter_init_from_data(&iter, data(), length())) {
-        return cend();
+    if (!bson_iter_init_from_data(&iter, _view.data(), _length)) {
+        return this->cend();
```
eramongodb, line 69: This PR avoids attempting to extend the error handling improvements in v1 API to the v_noabi implementation beyond minimal requirements, e.g. missing null pointer checks remain unchecked. However, we have the option to convert UB (i.e. termination) into exceptions (well-defined) as a "bugfix" if preferable.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/value.cpp
```diff
@@ -12,17 +12,33 @@
...
+static_assert(is_explicitly_convertible<v1::document::value const&, value>::value, "v1 -> v_noabi: this is a lie!");
 
-value::value(unique_ptr_type ptr, std::size_t length) : _data(std::move(ptr)), _length(length) {}
+// Backward compatibility with unconstrained `operator=(T const&)` permits the following.
+static_assert(std::is_assignable<value, v1::document::value&&>::value, "v1 -> v_noabi: assignment is supported");
+static_assert(std::is_assignable<value, v1::document::value const&>::value, "v1 -> v_noabi: assignment is supported");
+
+// Backward compatibility with lack of default destructor and `value({})` prevent implicit conversions.
+static_assert(!is_implicitly_convertible<v1::document::value&&, value>::value, "v1 -> v_noabi is not supported");
+static_assert(!is_implicitly_convertible<v1::document::value const&, value>::value, "v1 -> v_noabi is not supported");
```
eramongodb, line 41: The pain of the unconstrained template parameters in `v_noabi::document::value` are documented by these series of unfortunate static assertions. 🫠

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/value.cpp
```diff
@@ -32,66 +48,33 @@ void uint8_t_deleter(std::uint8_t* ptr) {
...
-    return std::move(_data);
-}
+    auto const length = v.length();
+    auto const deleter = *deleter_ptr;
 
-void value::reset(document::view view) {
-    _data.reset(new std::uint8_t[static_cast<std::size_t>(view.length())]);
-    _length = view.length();
-    std::copy(view.data(), view.data() + view.length(), _data.get());
+    return {v.release().release(), length, deleter};
```
eramongodb, line 76: Due to `v1::document::value::deleter_type` being a `std::function<void(std::uint8_t*)>` rather than `void(*)(std::uint8_t*)`, some special-casing is required to preserve the invariant that the deleter type is `v_noabi::document::value::deleter_type` (function pointer). Rather than throwing an exception on mismatch, this PR proposes UX-friendliness (at the cost of a hidden performance penalty) by falling back to a copy when the deleter type is incompatible.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/value.cpp
```diff
@@ -12,17 +12,24 @@
...
 namespace array {
 
-value::value(std::uint8_t* data, std::size_t length, deleter_type dtor) : _data(data, dtor), _length(length) {}
+static_assert(is_explicitly_convertible<value&&, v1::array::value>::value, "v_noabi -> v1 must be explicit");
+static_assert(is_explicitly_convertible<value const&, v1::array::value>::value, "v_noabi -> v1 must be explicit");
 
-value::value(unique_ptr_type ptr, std::size_t length) : _data(std::move(ptr)), _length(length) {}
+// Backward compatibility with lack of default destructor and `value({})` prevents the following conversions.
+static_assert(!is_explicitly_convertible<v1::array::value&&, value>::value, "v1 -> v_noabi is not supported");
+static_assert(!is_explicitly_convertible<v1::array::value const&, value>::value, "v1 -> v_noabi is not supported");
```
eramongodb, line 32: Due to lack of support for user-defined conversion functions containing ill-constrained template parameters (... but also, why aren't there any?), `v_noabi::array::value` has fewer issues to deal with than `v_noabi::document::value`.

src/bsoncxx/test/v_noabi/decimal128.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <bsoncxx/stdx/string_view.hpp>
+
+#include <bsoncxx/test/catch.hh>
+
+namespace {
+
+using bsoncxx::v_noabi::decimal128;
+using code = bsoncxx::v_noabi::error_code;
+
+TEST_CASE("exceptions", "[bsoncxx][v_noabi][decimal128]") {
```
eramongodb, line 32: Some minimal test coverage was added to v\_noabi to ensure backward compatibility.

src/bsoncxx/test/v_noabi/decimal128.cpp
```diff
@@ -0,0 +1,68 @@
...
+    CHECK_THROWS_WITH_CODE(decimal128{"invalid"}, code::k_invalid_decimal128);
+}
+
+TEST_CASE("basic", "[bsoncxx][v_noabi][decimal128]") {
+    SECTION("to_string") {
+        CHECK(decimal128{}.to_string() == "0E-6176");
+    }
+}
+
+TEST_CASE("v1", "[bsoncxx][v_noabi][decimal128]") {
```
eramongodb, line 43: All v\_noabi classes have test coverage for their v\_noabi <-> v1 conversion functions.

src/bsoncxx/test/v_noabi/bson_get_values.cpp
```diff
@@ -387,4 +390,243 @@ TEST_CASE("document::values have a superset of document::view's methods") {
...
+
+template <typename Deleter, typename T>
+bsoncxx::v1::stdx::optional<Deleter> get_deleter(T const& v) {
+    if (auto const deleter_ptr = v.get_deleter().template target<Deleter>()) {
+        return *deleter_ptr;
+    }
+    return {};
+}
+
+TEST_CASE("v1", "[bsoncxx][v_noabi][document][value]") {
```
eramongodb, line 458: The refactored `v_noabi::document::value` class received the most thorough test coverage due to complexity of satisfying the deleter type invariant throughout both copy and move operations.

src/bsoncxx/test/v_noabi/bson_get_values.cpp
```diff
@@ -387,4 +390,243 @@ TEST_CASE("document::values have a superset of document::view's methods") {
...
+    CAPTURE(noop_deleter);
+
+    SECTION("from_v1") {
+        auto const size = tmp.size();
+        auto ptr = tmp.release();
+        auto const data = ptr.release();
+        auto const deleter = ptr.get_deleter(); // bson_free_deleter
+
+        SECTION("default") {
+            (void)v_noabi{data, size, deleter}; // Unused.
```
eramongodb, line 485: This is to reacquire ownership the released data above for consistency with other test sections.

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/element.cpp
```diff
@@ -12,35 +12,27 @@
...
-    return !(elem == v);
-}
+namespace bsoncxx {
+namespace v_noabi {
+namespace array {
 
-bool operator!=(types::bson_value::view const& v, element const& elem) {
-    return !(elem == v);
-}
+// MSVC: `std::is_constructible<T, Args...>` does not work with using-declared conversion functions to class type...?
```
eramongodb, line 29: An odd MSVC bug where a using-declared conversion function to a class type does not seem to inherit the `explicit` specifier. See: https://godbolt.org/z/9jTxMan5W

src/bsoncxx/test/v1/types/value.cpp
```diff
@@ -313,6 +314,38 @@ TEST_CASE("b_types", "[bsoncxx][v1][types][value]") {
 #pragma pop_macro("X")
 }
 
+TEST_CASE("get_bson_value", "[bsoncxx][v1][types][value][internal]") {
```
eramongodb, line 317: These test cases were added to document and compare the subtly different requirements concerning empty string behavior by v1's internal implementation vs. what is required by mongocxx's `scoped_bson_value` utility (which is refactored to use `v1::types::value`'s internal `bson_value_t` directly) when interfacing with CSFLE API.

src/mongocxx/lib/mongocxx/private/scoped_bson_value.hh
```diff
@@ -14,61 +14,60 @@
...
 
     // Allow obtaining a pointer to this->value even in rvalue expressions.
     bson_value_t* get() noexcept {
-        return &value;
+        auto& v = bsoncxx::v1::types::value::internal::get_bson_value(value);
+
+        // CSFLE API requires empty strings to be not-null.
+        if (v.value_type == BSON_TYPE_UTF8 && v.value.v_utf8.str == nullptr) {
+            v.value.v_utf8.str = static_cast<char*>(bson_malloc0(1u));
+        }
```
eramongodb, line 38: This requirement is both unexpected and, when not satisfied, a major pain to diagnose. This problem manifests as inscrutible CSFLE errors such as:

```
Corpus
../src/mongocxx/test/v_noabi/client_side_encryption.cpp:1059

../src/mongocxx/test/v_noabi/client_side_encryption.cpp:979: FAILED:
explicitly with message:
  caught an exception for encrypting an allowed field payload=0,algo=rand: BSON
  type invalid for encryption: generic server error
```

with stack trace (libmongocrypt 1.13.0, mongoc 912209d, ignoring mocks) pointing [here](https://github.com/mongodb/libmongocrypt/blob/1.13.0/src/mongocrypt-ctx-encrypt.c#L1739) (with `bson_type == BSON_TYPE_EOD`):

```
_permitted_for_encryption (src/mongocrypt-ctx-encrypt.c:1739)
explicit_encrypt_init (src/mongocrypt-ctx-encrypt.c:1962)
mongocrypt_ctx_explicit_encrypt_init (src/mongocrypt-ctx-encrypt.c)
_mongoc_crypt_explicit_encrypt (src/libmongoc/src/mongoc/mongoc-crypt.c:1772)
mongoc_client_encryption_encrypt (src/libmongoc/src/mongoc/mongoc-client-side-encryption.c:2671)
mongocxx::v_noabi::client_encryption::impl::encrypt (src/mongocxx/lib/mongocxx/private/client_encryption.hh:111)
mongocxx::v_noabi::client_encryption::encrypt (src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client_encryption:43)
_run_corpus_test (src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client_encryption:971)
```

There may be insufficient input validation by mongoc, libmongocrypt, or both, but atm I am not sure which may be at fault. For now, this PR assumes it is mongocxx's responsibility and applies the necessary workaround to ensure `v_utf8.str != nullptr` is always true even for empty strings. This issue is only observable by mongocxx due to accessing `bsoncxx::v1::types::value`'s `bson_value_t` directly.

src/mongocxx/test/private/scoped_bson_value.cpp
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        SECTION("string") {
+            SECTION("empty") {
+                value v{b_string{bsoncxx::stdx::string_view{}}};
+                REQUIRE(v.get()->value_type == BSON_TYPE_UTF8);
+                auto const& v_utf8 = v.get()->value.v_utf8;
+                CHECK(v_utf8.len == 0u);
+                CHECK(static_cast<void const*>(v_utf8.str) != nullptr);
+                CHECK(v_utf8.str[0] == '\0');
```
eramongodb, line 56: These test cases were added to document and compare the subtly different requirements concerning empty string behavior by v1's internal implementation vs. what is required by mongocxx's `scoped_bson_value` utility (which is refactored to use `v1::types::value`'s internal `bson_value_t` directly) when interfacing with CSFLE API.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/view.hpp
kevinAlbs: I suppose the current behavior allows passing a larger buffer length:
```c++
// Document is 12 bytes { 'x': 1}. Buffer is 13 bytes (extra 0xFF).
std::uint8_t const data[] = {12, 0, 0, 0, 16, 'x', '\0', 1, 0, 0, 0, 0, 0xFF};
auto v = bsoncxx::v_noabi::document::view(data, sizeof(data));
// Returns the passed length:
REQUIRE(v.length() == 13);
```

Though I would rather discourage it. Iterating quietly fails:
```c++
REQUIRE(v.begin() == v.end());
```

Consider adding a warning to the `v_noabi` view constructors accepting a length:
```c++
/// @warning For backward compatibility, this function does NOT check if @p length equals the embedded length in the BSON bytes.
```

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/document/view.cpp
kevinAlbs: I slightly prefer checking the return. I expect the unchecked return could result in quiet failures. Example:

```c++
// Document is {'x': 'y'}, except length for 'y' is incorrect.
std::uint8_t const data[] = {
    14, 0, 0, 0,  // Document length.
    2, 'x', '\0', // String and key.
    123, 0, 0, 0, // Bad length for value.
    'y', 0,
    0
};
auto v = bsoncxx::v_noabi::document::view(data, sizeof(data));
REQUIRE(v.begin() == v.end()); // No error indicated!?
```

Though I would rather fix in a separate PR to avoid including behavior changes with the refactor. Reported: CXX-3333.

src/mongocxx/lib/mongocxx/private/scoped_bson_value.hh
kevinAlbs: > which may be at fault

I expect this is related to this [documented quirk in bson_append_utf8](https://mongoc.org/libbson/current/bson_append_utf8.html):

> Due to legacy behavior, passing `NULL` for value appends a null value, *not* a UTF8 value.

kevinAlbs: Confirming: am I correct that this results in no observed behavior change? The previous [make_copy_for_libbson](https://github.com/mongodb/mongo-cxx-driver/blob/7f8ea7bc5e75f1b487c415cb518f4db7535f32f0/src/bsoncxx/lib/bsoncxx/private/convert.hh#L32) appears to also allocate for an empty string.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/element.hpp
```diff
@@ -108,11 +112,27 @@ BSONCXX_ABI_EXPORT_CDECL(bool) operator!=(types::bson_value::view const& v, elem
 } // namespace v_noabi
 } // namespace bsoncxx
 
+namespace bsoncxx {
+namespace v_noabi {
+
+// Ambiguous whether `v1::element::view` should be converted to `v1::array::element` or `v1::document::element.`
```
kevinAlbs:
```suggestion
// Ambiguous whether `v1::element::view` should be converted to `v_noabi::array::element` or `v_noabi::document::element.`
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/view.hpp
```diff
@@ -34,10 +39,58 @@ namespace array {
...
+    ///
+    /// @warning For backward compatibility, `length` is NOT validated. When `length` is inconsistent with the embedded
+    /// length as indicated by the BSON bytes, the BSON bytes may be parsed as "invalid" despite the BSON bytes
+    /// themselves being valid.
+    ///
+    /// @param data
+    ///   A buffer containing a valid BSON array.
+    /// @param length
+    ///   The size of the buffer, in bytes.
+    ///
```
kevinAlbs: Similar to `v1`, consider using `@copydoc` to refer to `v_noabi::document::view` functions.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128-fwd.hpp
kevinAlbs: I may be misunderstanding. Is including the v1 forwarding header needed? Removing the v1 include still compiles.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
```diff
@@ -83,6 +85,34 @@ enum class binary_sub_type : std::uint8_t {
     k_user = 0x80,              ///< User defined.
 };
 
+///
+/// Convert from the @ref bsoncxx::v1 equivalent of `v`.
```
kevinAlbs, line 89:
```suggestion
/// Convert from the @ref bsoncxx::v1 equivalent of `value`.
```

Suggestion also applies to other "Convert from..." and "Convert to..." comments.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
```diff
@@ -35,34 +39,69 @@ namespace array {
...
+    /// @copydoc v_noabi::document::value::deleter_type
     using deleter_type = void(BSONCXX_ABI_CDECL*)(std::uint8_t*);
+
+    /// @copydoc v_noabi::document::value::unique_ptr_type
     using unique_ptr_type = std::unique_ptr<uint8_t[], deleter_type>;
 
+    /// @copydoc v_noabi::document::value::const_iterator
+    using const_iterator = v_noabi::array::view::const_iterator;
+
+    /// @copydoc v_noabi::document::value::iterator
```
kevinAlbs, line 56: Suggest adding docs to `v_noabi::document::view::iterator` to fix warning:
```
[...]/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp:56: warning: @copybrief or @copydoc target 'v_noabi::document::view::iterator' not found
```

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/view.cpp
```diff
@@ -12,39 +12,39 @@
...
 
 #include <bsoncxx/private/bson.hh>
 #include <bsoncxx/private/itoa.hh>
+#include <bsoncxx/private/type_traits.hh>
 
 namespace bsoncxx {
 namespace v_noabi {
 namespace array {
 
+static_assert(is_explicitly_convertible<view&&, v1::array::view>::value, "v_noabi -> v1 must be explicit");
```
kevinAlbs, line 34: I like the added static assert tests, and expect they help to better document and test expected conversions. I expect the `v_noabi <=> v1` conversions will help consumers incrementally upgrade to `v1`.

Am I correct that future API additions can be limited to `v1`? I expect new API additions can still support `v_noabi` equivalents through the conversion operators. I would rather avoid new API additions needing to add both `v_noabi` and `v1` equivalents and conversions.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types.hpp
eramongodb: Switched the suggestion to renaming the parameter to `v` instead to reduce conflation with the various `value` types + consistency with other ABI conversion functions.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
eramongodb: Addressed by [`5e8e749` (#1447)](https://github.com/mongodb/mongo-cxx-driver/pull/1447/commits/5e8e7493a2dfa8807fe1614234ecd15ee41c43e3).

src/bsoncxx/lib/bsoncxx/v_noabi/bsoncxx/array/view.cpp
eramongodb: > Am I correct that future API additions can be limited to v1?

Yes, we should aim to reduce and remove the v_noabi API going forward, such that we minimize the unstable ABI footprint (and corresponding potential for surprise) in upcoming stable ABI releases. Any/all new features (or further API migration) should be implemented in the stable ABI namespace(s).

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/decimal128-fwd.hpp
eramongodb: It is not strictly necessary, _yet_. This is implementing in advance the header structure necessary to support declaring+providing v1 API via v_noabi headers for when root namespace redeclaration will be updated to `v_noabi -> v1` (e.g. `bsoncxx::decimal128` -> `bsoncxx::v1::decimal128`).

src/mongocxx/lib/mongocxx/private/scoped_bson_value.hh
eramongodb: Yes, this should not have any observable behavior changes, since mongoc (and mongocrypt) should not assume null-termination for a `bson_value_t` UTF-8 with explicit length. The bug described above was due to passing a null pointer despite `length == 0` due to the legacy behavior you identified and linked above, which I believe to be the correct root cause. This allocation (formerly unconditional, now only on-demand when passed to CSFLE API) preserves the not-null guarantee to avoid this legacy behavior.

eramongodb:
A closer inspection of abi-compliance-checker's API compatibility report identified some unintended API breaking changes.

---

First, the return type for `v_noabi::types::bson_value::view::get_<type>()` functions were incorrectly changed from `T const&` to `T` (note: `v1`'s equivalent API returns by-value):

> **Change:** Type of return value has been changed from `bsoncxx::v_noabi::types::b_int32 const&` to `bsoncxx::v_noabi::types::b_int32` of different format.
> **Effect:** Recompilation of a client program may be broken.

This can break existing code in the following manner:

```cpp
auto example(v_noabi::types::bson_value::view const& v)
  -> v_noabi::types::b_int32 const&
{
  return v.get_int32(); // Before: reference to this->_b_int32.
                        // After: dangling reference to temporary b_int32. :(
}
```

Therefore, partially reverted the initial changes and refactored the implementation so that `get_<type>()` can once again returns a `T const&` without undefined behavior. Because `v_noabi::b_<type>` classes do not wrap `v1::b_<type>` classes (which would be an API breaking change due to public data members), `v1::types::view` cannot be directly used as the underlying representation for `v_noabi::types::bson_value::view` without reinterpret-casting from the `v1::types::view::_b_<type>` data member(s) to `v_noabi::types::b_<type>`, which is not consistently possible for all `b_<type>` classes (they are not all layout-compatible). Therefore, `v_noabi::types::bson_value::view` is once again given its own `_id` + `_b_<type>` union data members.

---

The `v_noabi::types::bson_value::value` ctor parameters were incorrectly changed from `T` to `T const&`:

> **Change:** Parameters list has been changed.
> **Effect:** Recompilation of a client program may be broken.

This one is comparatively trivial in both scope and fix. These are simply changed back to `T const&`.

---

There are two "private" (and undocumented) hidden friend functions which were overlooked and omitted by the initial changes:

- `bsoncxx::v_noabi::types::bson_value::make_owning_bson(void* internal_value)`
- `bsoncxx::v_noabi::oid_compare(oid const& lhs, oid const& rhs)`

These are clearly _intended_ to be private API, and I was misled by their declaration in `private` access scope to interpret they are also publically inaccessible. However, as hidden friend functions, their actual declaration scope is in the enclosing namespace of the class. Therefore, they are just as eligible for ADL as if it were declared in `public` access scope: the access specifier makes no programmable difference, it is only aesthetic. These functions are therefore vulnerable to Hyrum's Law.

Due to `oid_compare()` being very useful (which motivated its inclusion in v1 API as `v1::oid::compare(v1::oid const& other)`), and given it is already an exported symbol, it is likely enough that users may be using this API despite it being "private" and undocumented. Therefore, this "private" hidden friend function is restored, but it remains undocumented do discourage further usage. Instead, a v1-equivalent `compare()` member will be added in a followup PR to aid in v_noabi -> v1 transitions along with other renamed entities.

On the other hand, `make_owning_bson()` is much less likely to be currently used by users given it requires a `void*` parameter, which resolves to a `bson_value_t const*` (C Driver API). There is no equivalent v1 public API for this operation (which is superceded by `v1::types::value`'s b_type constructors and `value::internal::get_bson_value()`). it is not an exported symbol either (a la v1's `internal` API; the `value` constructor is exported instead). Therefore, this "private" and undocumented hidden friend function remains omitted from this PR's proposed changes.

---

I do not believe the remaining issues in the API compatibility report should be problematic given our circumstances, as they mostly concern user-provided ctors/dtors and implicit defaulted ctors which are now made trivial (reported as "removed"...?) as well as changes to set of implicit template instantiations and specializations (whose template definitions are visible and/or not owned by us).

There is a lot of noise in the v_noabi API compatibility report due to v1 API being included in the unstable ABI report despite the [list of headers to analyze](https://github.com/mongodb/mongo-cxx-driver/blob/ce8f7df995de6de3e3f443f7b8f506cadfdb397e/.evergreen/scripts/abi-compliance-check-test.sh#L93-L96) being limited to v_noabi headers. The XML descriptors may need to be updated to better filter results to v_noabi API only.

src/bsoncxx/include/bsoncxx/v1/oid.hpp
```diff
@@ -189,6 +189,13 @@ class oid {
     friend std::error_code make_error_code(errc v) {
         return {static_cast<int>(v), error_category()};
     }
+
+    class internal;
+
+   private:
+    struct uninit_type {};
+
+    oid(uninit_type) : _bytes{} {} // Required by bsoncxx::v_noabi::oid::oid().
```
vector-of-bool: Minor: rename to `zeroinit_type`? This value-initiailizes the `_bytes` to all zeroes.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
vector-of-bool: Should this be taking the vector parameter by-value, or by `const&`?

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
```diff
@@ -98,6 +72,46 @@ class float32 {
     std::uint8_t bytes[4];
 };
 
+template <>
+inline void float32::construct<bsoncxx::detail::endian::little>(float value) {
+    std::memcpy(bytes, &value, sizeof value);
+}
```
vector-of-bool, line 78: Are template specializations required, or can this just be a regular `if` branch in a single function?

src/bsoncxx/lib/bsoncxx/private/type_traits.hh
```diff
@@ -75,4 +75,15 @@ struct is_semitrivial : detail::conjunction<
                             std::is_trivially_copy_constructible<T>,
                             std::is_trivially_copy_assignable<T>> {};
 
+// Equivalent to `is_constructible_v<To, From> && !is_convertible_v<From, To>`.
+template <typename From, typename To>
+struct is_explicitly_convertible : bsoncxx::detail::conjunction<
+                                       std::is_constructible<To, From>,
+                                       bsoncxx::detail::negation<std::is_convertible<From, To>>> {};
```
vector-of-bool, line 83: Tweak name: implicit-convertbile should imply `is_explicit_convertible`, but this checks that it is *also* **not** implicit-convertible. Maybe `is_only_explicitly_convertible`?

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
eramongodb: It _should_ be taking by `T const&` ([as is done in v1](https://github.com/mongodb/mongo-cxx-driver/blob/031b33bef16b6c9c622eb91525c4f504f5f98cd1/src/bsoncxx/include/bsoncxx/v1/types/value.hpp#L194)), but I've deliberately left it as-is to prioritize API backward compatibility (avoiding changes to the observable API as much as possible) and keeping the `abi-compliance-check` API compatibility report [happy](https://mciuploads.s3.amazonaws.com/mongo-cxx-driver/master/031b33bef16b6c9c622eb91525c4f504f5f98cd1/68c1cc557612af000746e9ba/mongo_cxx_driver_abi_stability_patch_031b33bef16b6c9c622eb91525c4f504f5f98cd1_68c1cc557612af000746e9ba_25_09_10_19_07_04/mongo_cxx_driver_abi_stability_abi_compliance_check_stdlib_cxx17_patch_031b33bef16b6c9c622eb91525c4f504f5f98cd1_68c1cc557612af000746e9ba_25_09_10_19_07_04/0/abi-compliance-check/noabi/compat_report.html) (although it is [nevertheless confused](https://github.com/mongodb/mongo-cxx-driver/pull/1458) for reasons I do not understand).

src/bsoncxx/lib/bsoncxx/private/type_traits.hh
eramongodb: The purpose of these type traits is to more accurately assert the presence _or absence_ of `explicit` for a given constructor or conversion function while also avoiding the `To` vs. `From` ordering inconsistency (which I frequently mixed up when writing the relevant static assertions prior to defining these helpers):

```cpp
struct From {};
struct To { /* explicit(false) */ To(From); };

static_assert(std::is_constructible_v<To, From>, "OK...?"); // Implies nothing about `explicit`.
static_assert(std::is_convertible_v<From, To>, "there is no `explicit`");

static_assert(is_explicitly_convertible_v<From, To>, "there is an `explicit`"); // Fails: a useful check!
static_assert(is_implicitly_convertible_v<From, To>, "there is no `explicit`");
```

I considered naming them `has_(implicit|explicit)_ctor`, but this would not account for conversion functions (indistinguishable from ctors by usage syntax alone). I also considered `has_explicit` + `has_no_explicit`, but then there'd be a `no_` in a type trait identifier... 😑

Therefore I ended up with the current names for the best symmetry: `is_only_explicitly_convertible` would imply an `is_explicitly_convertible` ("not only explicit")... but that would just be equivalent to `std::is_constructible<To, From>` (providing no additional value). I don't think we want/need an `is_only_implicitly_convertible` either (is that even possible?).

That being said, added additional notes to their documentation to clarify that `std::is_constructible` and `std::is_convertible` should be used if the additional implications are not desirable.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/vector/elements.hpp
eramongodb: The specializations are to avoid "conditional expression is constant" compiler warnings without needing to use warning suppressions. This should really be an `if constexpr`, but alas...

src/bsoncxx/include/bsoncxx/v1/oid.hpp
eramongodb: The zero-initialization is meant to be an implementation detail that avoids language-level undefined behavior (in the spirit of `-ftrivial-auto-var-init=zero` and to avoid triggering static analysis warnings for uninitialized data members), but it's still meant to be library-level undefined behavior to access `_bytes` after initializing with `uninit_type`, since the elements are expected to be unconditionally overwritten by `v_noabi::oid::oid()` anyways. I expect/hope the compiler should be able to see this unconditional write-without-read and optimize accordingly (ignoring exception paths):
```cpp
v_noabi::oid::oid()
  // Zeroed bytes are never read after this initialization.
  : _oid{v1::oid::internal::make_oid_without_init()}
{
  // Unconditionally overwritten.
  _oid = v1::oid{};
}
```

eramongodb: On second thought, to avoid the "uninit" confusion, adopted the `*_for_overwrite` suffix from `std::make_unique_for_overwrite`, which hopefully makes the intention much clearer.

PR #1453, opened 2025-08-21
Move ASAN workaround from mongocxx::instance to CI with bypass_dlclose

eramongodb:
Quoting https://github.com/mongodb/mongo-c-driver/pull/1177:

> To silence unhelpful `<unknown module>` leaks, the workaround previously applied [specifically to mongcryptd](https://github.com/mongodb/mongo-c-driver/blob/21275453cd46253b9d252fdfc319c1880b3dca3a/.evergreen/run-tests.sh#L80) was generalized into a `bypass-dlclose.sh` script which wraps a provided command with `LD_PRELOAD` set appropriately. This is applied only to tests run on Evergreen; changes to the CMake configuration were deliberately avoided to prevent introducing downstream effects.

A very long time ago in https://github.com/mongodb/mongo-cxx-driver/pull/527 (Sep 2016), mongocxx implemented a workaround for this issue, but within the implementation of `instance` itself, such that the call to `mongoc_cleanup()` is _compile-time disabled_ if it detects ASAN is enabled. Note this predates the mongocryptd `dlclose` workaround mentioned above (Nov 2019). Looking at related issues (see: [google/sanitizers#89](/google/sanitizers/issues/89)), it doesn't seem like the situation has improved since.

I think we can/should move this migration into CI scripts instead for consistency with libmongocrypt and mongoc. Included a changelog entry in case this change happens to affect anyone downstream building the C++ Driver with ASAN enabled and expect this old behavior for some reason (unlikely, but just in case).

### DIFF
--- a/.evergreen/scripts/test.sh
+++ b/.evergreen/scripts/test.sh
@@ -101,6 +101,7 @@ fi
 export DRIVERS_TOOLS
 popd # "${working_dir:?}/../drivers-evergreen-tools"
 
+. .evergreen/scripts/bypass-dlclose.sh
 . .evergreen/scripts/install-build-tools.sh
 install_build_tools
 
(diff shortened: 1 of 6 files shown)


.evergreen/scripts/test.sh
```diff
@@ -275,21 +276,23 @@ else
...
+    echo "Running ${1:?}..."
+    "${1:?}" "${test_args[@]:?}" || return
+    echo "Running ${1:?}... done."
   }
 
+  declare ld_preload="${LD_PRELOAD:-}"
   if [[ "${TEST_WITH_ASAN:-}" == "ON" || "${TEST_WITH_UBSAN:-}" == "ON" ]]; then
     export ASAN_OPTIONS="detect_leaks=1"
     export UBSAN_OPTIONS="print_stacktrace=1"
+    ld_preload="$(bypass_dlclose):${ld_preload:-}"
```
kevinAlbs, line 288: Is this missing an export of `LD_PRELOAD`?
```sh
export LD_PRELOAD
LD_PRELOAD="${ld_preload:-}"
```

eramongodb: Whoops, I'd meant to _temporarily_ remove the `LD_PRELOAD="${ld_preload:-}"` in `run_test` to verify if the ASAN issue is still present (couldn't observe it in CI). Fixed.

PR #1454, opened 2025-08-25
Format Python and Bash scripts using ruff and shfmt

eramongodb:
Extends automated formatting to Python and Bash scripts using the existing `uv` project file.

> [!IMPORTANT]
> This PR only _selects_ and _provides_ the tools required for automated formatting of Python and Bash scripts. However, it does not _check_ or _enforce_ that these files are correctly formatted. The `lint` task is untouched except to use the new `etc/format.py` script. The `etc/format.py` itself script remains specific to `clang-format` only.

The new `etc/ruff-format-all.sh` and `etc/shfmt-format-all.sh` scripts are included for convenience and reference ("how to use these tools"). They do not need to be used directly (e.g. directly running `uv run ruff format` is just fine). I expect/hope these will be integrated into `etc/format.py` (similar to how it replaces the `etc/clang-format-all.sh` script) in the near future.

Similar changes are planned for the C Driver shortly following this PR.

---

For C and C++ source files, [`tools/format.py`](https://github.com/mongodb/mongo-c-driver/blob/f000f77e8d38a39c3ad80128636fc5712b9100ca/tools/format.py) from https://github.com/mongodb/mongo-c-driver/pull/2050 is imported and reused here as `etc/format.py`, with minor tweaks to support the C++ Driver directory structure. This replaces the `etc/clang-format-all.sh` script. Relevant documentation and references have been updated accordingly.

> [!NOTE]
> Adding support for `uv run format`, similar to [`uv run mc-evg-generate`](https://github.com/mongodb/mongo-c-driver/blob/f000f77e8d38a39c3ad80128636fc5712b9100ca/pyproject.toml#L42), was considered but postponed for now.

---

For Python files, this PR proposes using [Ruff](https://docs.astral.sh/ruff/formatter/). For consistency with our updated ClangFormat preferences, the column width is set to 120. For consistency with recent Python code, `quote-style = "single"` and `indent-style = "space"` are also set. Note multiline strings unconditionally use double-quotes despite `quote-style = "single"` for conformance with [PEP 8](https://peps.python.org/pep-0008/) and [PEP 257](https://peps.python.org/pep-0257/).

The expected commands to format all Python files is documented and supported by `etc/ruff-format-all.sh`. Note the `ruff check --fix` command required to sort and group `import` statements (see: [astral-sh/ruff#8232](/astral-sh/ruff/issues/8232)), first by "standard library", then "third party", then "first party" (see: [Ruff FAQ](https://docs.astral.sh/ruff/faq/#how-does-ruff-determine-which-of-my-imports-are-first-party-third-party-etc)). I don't believe this ordering is customizeable (see: [astral-sh/ruff#7615](/astral-sh/ruff/issues/7615)).

Unfortunately I don't think there's anything we can do about the "two-space lead for trailing comments" formatting either (e.g. `print("example")  # comment`). There is a massive semi-related unresolved issue [astral-sh/ruff#7684](/astral-sh/ruff/issues/7684) concerning formatting of comments. The default behavior is left as-is for now.

Formatting of `pyproject.toml` itself (as a TOML file) is also deferred, as there does not seem to be an immediately-obvious community-supported tool to integrate and use for this purpose at this time (see: [toml-lang/toml#532](/toml-lang/toml/issues/532)). Wait and hope for [astral-sh/ruff#10738](/astral-sh/ruff/issues/10738) to be implemented soon...?

> [!NOTE]
> Ruff also supports [linting](https://docs.astral.sh/ruff/linter/) Python scripts via `ruff check`. This is _not_ integrated into any scripts at this time. However, as a drive-by improvement, all currently-outstanding warnings emitted by `ruff check` are fixed by this PR using both `--fix` and manual edits. Adding `pyright` was considered but postponed to keep this PR focused on _formatting_ only.

---

For Bash scripts, this PR proposes using [shfmt](https://github.com/mvdan/sh). Unlike Ruff, there is no convenient auto-detection of relevant project files, so `etc/shfmt-format-all.sh` uses a simple `find` command to find and filter `.sh` scripts under `.evergreen`, `etc`, and `examples`. For consistency with recent Bash code, indentation is set to two spaces with `-i 2`. Although `shfmt` also supports `--simplify` , but this flag is omitted for now. This flag can be included if preferable.

> [!NOTE]
> Linting Bash scripts can be done using `shellcheck` via the `shellcheck-py` package, but this too is postponed to keep this PR focused on _formatting_ only. However, this can be done manually using `uv run --with shellcheck-py shellcheck -x <files>`.

### DIFF
--- a/etc/format.py
+++ b/etc/format.py
@@ -247,26 +247,5 @@ def _format_files(files: Iterable[Path], *, mode: RunMode, clang_format: str) ->
     return res.returncode == 0
 
 
-def _get_files_matching(pat: str) -> Sequence[Path]:
-    """
-    Obtain files according to a globbing pattern. Checks that at least one file
-    matches.
-    """
-
-    try:
-        if os.path.isabs(pat):
-            # Given an absolute path, glob relative to the root directory
-            root = Path(pat).root
-            ret = tuple(Path(root).glob(str(Path(pat).relative_to(root))))
-        else:
-            # None-relative path, glob relative to CWD
-            ret = tuple(Path.cwd().glob(pat))
-    except Exception as e:
-        raise RuntimeError(f'Failed to collect files for globbing pattern: “{pat}” (See above)') from e
-    if not ret:
-        raise RuntimeError(f'Globbing pattern “{pat}” did not match any files')
-    return ret
-
-
 if __name__ == '__main__':
     sys.exit(main(sys.argv[1:]))


eramongodb:
Moved `ruff` and `shfmt-py` into an optional "format-scripts" dependency group to avoid impacting the `lint` EVG task with the following build requirement and error on the `rhel80` distro:

```
x Failed to build `shfmt-py==3.12.0.2`
├─▶ The build backend returned an error
╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit
    status: 1)
    [stdout]
    running bdist_wheel
    running build
    running fetch_binaries
    https://github.com/mvdan/sh/releases/download/v3.12.0/shfmt_v3.12.0_linux_amd64
    [stderr]
    error: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate
    verify failed: unable to get local issuer certificate (_ssl.c:1035)>
    hint: This usually indicates a problem with the package or the build
    environment.
help: `shfmt-py` (v3.12.0.2) was included because `mongo-cxx-driver:dev`
      (v0.1.0) depends on `shfmt-py`
```

The `*-format-all.sh` scripts have been updated accordingly to explicitly request the optional "format-scripts" dependency group.

etc/format.py
```diff
@@ -0,0 +1,272 @@
...
+    cmd = [clang_format, *mode_args, *map(str, files)]
+    try:
+        res = subprocess.run(cmd, check=False, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+    except Exception as e:
+        raise RuntimeError(f'Failed to spawn [{clang_format}] process for formatting files (See above)') from e
+    sys.stderr.buffer.write(res.stdout)
+    return res.returncode == 0
+
+
+def _get_files_matching(pat: str) -> Sequence[Path]:
```
vector-of-bool: This function is not used (it's not used in the C repo either, I just haven't gotten around to deleting it yet)

vector-of-bool (approved):
I've been meaning to add Ruff format to the C repo for a while, so this is good precedent. LGTM

etc/format.py
eramongodb: Posted https://github.com/mongodb/mongo-c-driver/pull/2102.

PR #1460, opened 2025-09-15
[v4.1] CXX-3302 Update bsoncxx validation examples with offsets (#1416)

kevinAlbs:
Cherry-pick https://github.com/mongodb/mongo-cxx-driver/commit/bf7554cff60b871d1b382e5b7e146707a18c8ab7 to bump the auto-downloaded C driver from 2.0.0 to 2.0.2. Intended to include CDRIVER-5967 to fix building with CMake 4. Reported in CXX-3348.

Tested with this patch build: https://spruce.mongodb.com/version/68c86bf568ff0d0007297883

Though CXX-3302 will be available in C++ 4.2.0, this PR proposes including in a sooner 4.1.3 release since a date for 4.2.0 is not-yet planned.

### DIFF
--- a/etc/augmented.sbom.json
+++ b/etc/augmented.sbom.json
@@ -1,16 +1,16 @@
 {
   "components": [
     {
-      "bom-ref": "pkg:github/mongodb/mongo-c-driver@v2.0.0",
+      "bom-ref": "pkg:github/mongodb/mongo-c-driver@v2.0.2",
       "copyright": "Copyright 2009-present MongoDB, Inc.",
       "externalReferences": [
         {
           "type": "distribution",
-          "url": "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/v2.0.0.tar.gz"
+          "url": "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/v2.0.2.tar.gz"
         },
         {
           "type": "website",
-          "url": "https://github.com/mongodb/mongo-c-driver/tree/v2.0.0"
+          "url": "https://github.com/mongodb/mongo-c-driver/tree/v2.0.2"
         }
       ],
       "group": "mongodb",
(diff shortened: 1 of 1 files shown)


etc/augmented.sbom.json
eramongodb, line 1: I think the augmented SBOM should be regenerated rather than cherry-picked (note: there is no update to the timestamp in the current changes).

CMakeLists.txt
```diff
@@ -45,9 +45,9 @@ else()
 endif()
 
 # Also update etc/purls.txt.
-set(BSON_REQUIRED_VERSION 2.0.0)
-set(MONGOC_REQUIRED_VERSION 2.0.0)
-set(MONGOC_DOWNLOAD_VERSION 2.0.0)
+set(BSON_REQUIRED_VERSION 2.0.2)
+set(MONGOC_REQUIRED_VERSION 2.0.2)
+set(MONGOC_DOWNLOAD_VERSION 2.0.2)
```
eramongodb:
```suggestion
set(BSON_REQUIRED_VERSION 2.0.0)
set(MONGOC_REQUIRED_VERSION 2.0.0)
set(MONGOC_DOWNLOAD_VERSION 2.0.2)
```
Missing corresponding changelog entries + I think only the download version should be updated to avoid bumping the minimum requirements in a patch build.

eramongodb (changes requested):

CMakeLists.txt
kevinAlbs: > avoid bumping the minimum requirements

Good point. Updated as suggested to only bump the auto-download version, but keep the minimum required to 2.0.0.

etc/augmented.sbom.json
kevinAlbs: Done. Reverted changes to the SBOM (`etc/cyclonedx.sbom.js`) and augmented SBOM (`etc/augmented.sbom.js`). Regenerated the SBOM from purls.txt. Regenerated the augmented SBOM with the Evergreen `sbom` task: https://spruce.mongodb.com/version/68c97e0b5a369c0007df476c

PR #1463, opened 2025-09-17
CXX-2745 add aliases and deprecations for v1 renames

eramongodb:
Related to CXX-2745. A minor PR as followup to https://github.com/mongodb/mongo-cxx-driver/pull/1447 to improve the v_noabi -> v1 migration path for renamed entities by providing (and using) equivalent aliases in the current v_noabi API.

Every rename is accompanied by:

- the new name in v_noabi whose behavior is identical to the renamed v_noabi entity, and
- a deprecation note in the old name pointing to the new name to be used instead.

This permits users to easily switch to the renamed v_noabi equivalent without needing to deal with v_noabi vs. v1 relative breaking changes. Users who have switched to the renamed v_noabi equivalent will then be able to manually (via ABI namespace qualification) or automatically (via root namespace redeclaration updates) switch to the v1 equivalent. For example, the migration path for `v_noabi::types::bson_value::view` -> `v1::types::view` may look like:

```cpp
// Current (v_noabi):
bsoncxx::types::bson_value::view v; // v_noabi
v.get_owning_value();

// Renamed (v_noabi):
bsoncxx::types::view v; // v_noabi
v.type_value();

// Root Namespace Redeclaration Update (v_noabi -> v1):
bsoncxx::types::view v; // v1
v.type_value();
```

Without the renames introduced by this PR, the middle step is missing: users would have to jump straight from the old v_noabi name to the new v1 equivalent.

The renamed (deprecated) names may be removed in an upcoming API major release by moving the entity definitions into where the aliases/equivalents are currently declared or by dropping them entirely in favor of v1. Either approach will work depending on the timeline of events (does not need to be decided at this time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
   - `bsoncxx::types::view`: equivalent to `bsoncxx::types::bson_value::view`.
   - `bsoncxx::types::value`: equivalent to `bsoncxx::types::bson_value::value`.
   - `type_view()` in `bsoncxx::document::element` and `bsoncxx::array::element`: equivalent to `get_value()`.
-  - `type_value()` in `bsoncxx::document::element` and `bsoncxx::document::element`: equivalent to `get_owning_value()`.
+  - `type_value()` in `bsoncxx::document::element` and `bsoncxx::array::element`: equivalent to `get_owning_value()`.
 
 ### Fixed
 
@@ -62,7 +62,7 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
   - `bsoncxx::types::bson_value::view` -> `bsoncxx::types::view`
   - `bsoncxx::types::bson_value::value` -> `bsoncxx::types::value`
   - `get_value()` -> `type_view()` in `bsoncxx::document::element` and `bsoncxx::array::element`.
-  - `get_owning_value()` -> `type_value()` in `bsoncxx::document::element` and `bsoncxx::document::element`.
+  - `get_owning_value()` -> `type_value()` in `bsoncxx::document::element` and `bsoncxx::array::element`.
 
 ### Removed
 
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
@@ -311,7 +311,7 @@ class value {
     ///
     /// Equivalent to @ref type() const.
     ///
-    /// To support incremental migration to @ref bsoncxx::v1::types::view.
+    /// To support incremental migration to @ref bsoncxx::v1::types::value.
     ///
     v_noabi::type type_id() const {
         return this->view().type_id();


connorsmacd:
> Relating to [CXX-2118](https://jira.mongodb.org/browse/CXX-2118) (`view` vs. `value` API symmetry), overlooked an old recurring issue with declaring a `.foo()` member function given an existing `foo` namespace-scope declaration with GCC compilers (this is not an issue with Clang or MSVC compilers):
> 
> ```
> In file included from bsoncxx/types/bson_value/value.cpp:15:
> bsoncxx/types/bson_value/value.hpp:307:19: error: declaration of ‘bsoncxx::v_noabi::type bsoncxx::v_noabi::types::bson_value::value::type() const’ changes meaning of ‘type’ [-fpermissive]
>   307 |     v_noabi::type type() const {
>       |                   ^~~~
> In file included from bsoncxx/types/bson_value/view.hpp:31,
>                  from bsoncxx/types/bson_value/value.hpp:34,
>                  from bsoncxx/types/bson_value/value.cpp:15:
> bsoncxx/types.hpp:47:12: note: ‘type’ declared here as ‘enum class bsoncxx::v_noabi::type’
>    47 | enum class type : std::uint8_t {
>       |            ^~~~
> ```
> 
> It remains a mystery to me why `v_noabi::types::bson_value::view` does not seem to trigger this same error despite declaring `type() const` in the same manner as `v_noabi::types::bson_value::value` would.

Does changing the declaration to:

```c++
enum v_noabi::type type() const {}
```

make the error go away? I couldn't repro your exact issue, but this works when the error is expected, e.g.: https://godbolt.org/z/xbhj39cfW

connorsmacd:
> Solved: it's apparently being caused by [this stray unqualified `type`](https://github.com/mongodb/mongo-cxx-driver/blob/6e20fbc3f138f2b842989e670a8d51f5026fe179/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp#L257) in a constructor parameter list (?!): [godbolt.org/z/YnWb9nzKE](https://godbolt.org/z/YnWb9nzKE)
> 
> Removing the ctor declaration or qualifying the declaration with `v_noabi` resolves the GCC compilation error. I'm quite suspicious now that this may be a GCC compiler bug. 😵

Interestingly, I can repro similar errors on MSVC and Clang with some modifications:

- Define the constructor out-of-line: https://godbolt.org/z/aKEWqWzc1
- Declare the member function _before_ the constructor: https://godbolt.org/z/T5Evx4E3E

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
...
 
+### Added
+
+* To support incremental migration, the following entities are defined as equivalent to their renamed counterparts.
+  - `bsoncxx::types::id`: equivalent to `bsoncxx::type`.
+  - `bsoncxx::types::binary_subtype`: equivalent to `bsoncxx::binary_sub_type`.
+  - `bsoncxx::types::view`: equivalent to `bsoncxx::types::bson_value::view`.
+  - `bsoncxx::types::value`: equivalent to `bsoncxx::types::bson_value::value`.
+  - `type_view()` in `bsoncxx::document::element` and `bsoncxx::array::element`: equivalent to `get_value()`.
+  - `type_value()` in `bsoncxx::document::element` and `bsoncxx::document::element`: equivalent to `get_owning_value()`.
```
```suggestion
  - `type_value()` in `bsoncxx::document::element` and `bsoncxx::array::element`: equivalent to `get_owning_value()`.
```

CHANGELOG.md
```diff
@@ -46,6 +56,13 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
       return instance;
     }
     ```
+- These following entities will be renamed (and removed) in an upcoming major release. To support incremental migration, both old and new names are still provided.
+  - `bsoncxx::type` -> `bsoncxx::types::id`.
+  - `bsoncxx::binary_sub_type` -> `bsoncxx::types::binary_subtype`.
+  - `bsoncxx::types::bson_value::view` -> `bsoncxx::types::view`
+  - `bsoncxx::types::bson_value::value` -> `bsoncxx::types::value`
+  - `get_value()` -> `type_view()` in `bsoncxx::document::element` and `bsoncxx::array::element`.
+  - `get_owning_value()` -> `type_value()` in `bsoncxx::document::element` and `bsoncxx::document::element`.
```
```suggestion
  - `get_owning_value()` -> `type_value()` in `bsoncxx::document::element` and `bsoncxx::array::element`.
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/oid.hpp
```diff
@@ -122,6 +122,13 @@ class oid {
         return reinterpret_cast<char const*>(_oid.bytes());
     }
 
+    ///
+    /// Equivalent to [`bson_oid_compare`](https://mongoc.org/libbson/current/bson_oid_compare.html).
+    ///
+    int compare(oid const& other) const {
+        return _oid.compare(other._oid);
+    }
+
```
kevinAlbs, line 131: Consider removing `compare`? I expect existing applications are not calling the private `oid_compare` method (which this method [replaces](https://github.com/mongodb/mongo-cxx-driver/pull/1412/files#r2114400540) in `v1`). Though if this is added as a convenience to existing users, I suppose there is little harm in adding.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
```diff
@@ -297,6 +299,24 @@ class value {
...
+    /// @deprecated Use @ref type_id() const instead (renamed).
+    ///
+    v_noabi::type type() const {
+        return this->view().type();
+    }
+
+    ///
+    /// Equivalent to @ref type() const.
+    ///
+    /// To support incremental migration to @ref bsoncxx::v1::types::view.
```
kevinAlbs:
```suggestion
    /// To support incremental migration to @ref bsoncxx::v1::types::value.
```

Are the added `type`, `type_id` and `get_<type>` methods intended for feature parity with `bsoncxx::v1::types::value`? I expect this may not strictly be needed to ease migration since existing applications would not have calls to these methods.

CHANGELOG.md
```diff
@@ -46,6 +56,13 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
       return instance;
     }
     ```
+- These following entities will be renamed (and removed) in an upcoming major release. To support incremental migration, both old and new names are still provided.
+  - `bsoncxx::type` -> `bsoncxx::types::id`.
```
kevinAlbs, line 60: Consider adding `BSONCXX_DEPRECATED` on the deprecated symbols. I expect that might help inform users with compiler warnings. Though I expect it will certainly trigger (possibly many) deprecation warnings for users.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/oid.hpp
eramongodb: This is added both as convenience and as a v_noabi -> v1 migration path, since `v1::oid` provides `.compare()`.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/types/bson_value/value.hpp
eramongodb: Not for feature parity, but for _name_ parity, so that when root namespace redeclarations are updated from `v_noabi` to `v1`, the required changes can be minimized in advance by downstream users without forced-breaking-changes:

```cpp
bsoncxx::types::view v; // Syntax is the same for v_noabi and v1.
v.type_view();          // Syntax is the same for v_noabi and v1.
```

When `bsoncxx::types::view` is updated from `v_noabi::types::view` to `v1::types::view`, code which have already been incrementally updated (as part of a API _minor_ version release) to use the new names will not be broken by the redeclaration release (as part of a API _major_ version release) to the same extent as if the v_noabi new-name equivalents were not available.

CHANGELOG.md
eramongodb: This cannot be done at the moment without significant noise for both users and ourselves, since the new names are defined as _type aliases_ for the old names. The deprecation warning will end up carrying over to usage of the type aliases:

```cpp
struct __attribute__((__deprecated__)) S {};

_Pragma("GCC diagnostic push")
_Pragma("GCC diagnostic ignored \"-Wdeprecated\"")
using T = S;
_Pragma("GCC diagnostic pop")

T t; // warning: 'T' is deprecated
```

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/oid.hpp
kevinAlbs: Sorry, I expect I am misunderstanding. Is this intended to help with migrating back from v1 -> v_noabi? Since v_noabi does not a `.compare()` equivalent, what would an application be migrating from? 

```c++
bsoncxx::oid oid1, oid2;
REQUIRE(oid1 != oid2); // Works for both v_noabi and v1
// Prior to PR, v_noabi has no `compare` or accessible `oid_compare`.

```

eramongodb: Perhaps "migration" was the wrong word to use for this one: this is a convenience "backport". It's about allowing users to use a trivially-provideable-by-us feature with the current v_noabi API (the `v_noabi::oid` class) to minimize the amount of changes necessary when going from `v_noabi::oid` to `v1::oid`, since the implementation of this `.compare()` is completely equivalent to the `v1` behavior. Current users can adopt use of `.compare()` now (the member function) without early-committing to `v_noabi::oid` -> `v1::oid` (the entire class). Note that `oid_compare()` _is_ accessible (even if not intended/desirable by us).

kevinAlbs: Ah OK. Thank you for explaining.

> Note that `oid_compare()` _is_ accessible (even if not intended/desirable by us).

Doh right. I now remember it is accessible via ADL as a hidden friend:
```c++
bsoncxx::v_noabi::oid oid1, oid2;
oid_compare(oid1, oid2); // Works!
```

PR #1469, opened 2025-09-26
CXX-3320 migrate instance and logger to mongocxx::v1

eramongodb:
Related to CXX-3320. Due to the unique nature of `mongocxx::v_noabi::instance` as a global singleton-like object, this class is migrated in full from `v_noabi` to `v1` as a single, standalone PR. Related types `v_noabi::logger` and `v_noabi::exception` are also minimally migrated as part of this PR to support the `instance` class API. (The `v1::exception` class 

---

First, concerning `v_noabi::logger`: due to having no observable difference in behavior in `v1`, `v_noabi::logger` is merely a redeclaration of `v1::logger` in the `v_noabi` namespace! 🎉 This also avoids complications with trying to support compatibility between both a `v_noabi::logger` and `v1::logger` class given their polymorphic nature.

---

The `v_noabi::instance` class is not so fortunate. Its migration is complicated primary by the realization that there are two unresolvable deficiencies with the `v_noabi::instance` API:

* it does not support using the mongoc default log handler, and
* the custom log handler (if any) is registered _after_ `mongoc_init()` has already been invoked.

As mentioned by mongoc [docs](https://mongoc.org/libmongoc/current/unstructured_log.html):

> Note that in the example above `mongoc_log_set_handler()` is called before `mongoc_init().` Otherwise, some log traces could not be processed by the log handler.

That is, the initialization behavior is currently defined as follows (pseudocode):

```cpp
void v_noabi::init(mongoc_log_func_t handler) {
    mongoc_init(); // May emit log messages!

    if (handler) {
        mongoc_log_set_handler(handler); // Already initialized!
    } else {
        mongoc_log_set_handler(nullptr); // Disable log messages!
    }

    // ... handshake ...
}
```

The `v1::instance` object proposes fixing these issues as follows:

* Support a `v1::default_handler` tag type by which the user can request using the mongoc default log handler.
* Register or disable the custom log handler _before_ calling `mongoc_init()` so that all log messages are handled properly.

That is, the initialization behavior is changed to the following:

```cpp
void v1::init(mongoc_log_func_t handler) {
    v1::init::impl(handler, true); // Set a custom log handler.
}

void v1::init(default_handler tag) {
    v1::init::impl(handler, false); // Use default log handler.
}

void v1::init::impl(mongoc_log_func_t handler, bool set_custom_handler) {
    // Support using the default log handler.
    if (set_custom_handler) {
        if (handler) {
            mongoc_log_set_handler(handler); // Already initialized!
        } else {
            mongoc_log_set_handler(nullptr); // Disable log messages!
        }
    }

    mongoc_init(); // All log messages are captured.

    // ... handshake ...
}
```

For backward compatibility, `v_noabi::instance`'s custom log handler constructor preserves the "register handler after `mongoc_init()`" behavior. If we do not consider this behavior to be worth preserving (how important are log messages which may be emitted by `mongoc_init()`?), we can simplify the implementation of `v_noabi::instance`'s initialization.

The `v_noabi::instance` still supports the (deprecated) `::instance()` helper function for backward compatibility. The implementation of `::instance()` remains specific to `v_noabi`: the implementation `v1::instance` is entirely unaware of `::instance()` support. In place of the `current_instance` "sentinel" value, `v1::instance` instead utilizes a simple `std::atomic_int` counter to track and report exceptions given multiple instance objects.

---

Much of the rest of this PR involves refactors to the test suite for the `instance` and `logger` classes. The `test_logging` executable is removed in favor of grouping all `instance` object testing to `test_instance`. All other test executables can (in theory) be grouped into a single test executable without concern for "multiple instance object" exceptions.

The fork-based pattern to test `instance` objects used by the examples API runner (https://github.com/mongodb/mongo-cxx-driver/issues/1216) is repurposed here for compatibility with the Catch2 test suite as `mongocxx::test::subprocess(op)`. This helper function accepts a single `op: void()` invocable that is executed within a forked subprocess. The result of the subprocess, _including termination_, is translated into well-behaved Catch test assertions. The Catch test suite reporter is reused by the subprocess to support capturing values and evaluating assertions. However, using `SECTION()` is not supported due to non-trivial control flow, either within `op` or in the parent `TEST_CASE()`.

This enabled writing test cases asserting proper log message handling behavior, including _default_ log message handler behavior, by capturing log output emitted by the subprocess (using POSIX pipe API in the `capture_stderr` class). Note this is only supported on non-Windows platforms. Unfortunately, the use of `std::exit()` and `std::terminate()` in the subprocess confuses `valgrind`, which (understandably) reports memory leaks due to the abnormal process termination method. Therefore, leak detection in subprocesses is disabled for the `test_instance` executable using `--trace-children=no`.

### DIFF
--- a/src/mongocxx/include/mongocxx/v1/exception.hpp
+++ b/src/mongocxx/include/mongocxx/v1/exception.hpp
@@ -20,6 +20,13 @@
 
 #include <mongocxx/v1/detail/prelude.hpp>
 
+#include <bsoncxx/v1/detail/macros.hpp>
+
+#include <mongocxx/v1/config/export.hpp>
+
+#include <system_error>
+#include <type_traits>
+
 namespace mongocxx {
 namespace v1 {
 
@@ -28,14 +35,60 @@ namespace v1 {
 ///
 /// @attention This feature is experimental! It is not ready for use!
 ///
-enum class source_errc {};
+enum class source_errc {
+    zero,       ///< Zero.
+    mongocxx,   ///< From the mongocxx library.
+    mongoc,     ///< From the mongoc library.
+    mongocrypt, ///< From the mongocrypt library.
+    server,     ///< From the MongoDB server.
+};
+
+///
+/// The error category for @ref mongocxx::v1::source_errc.
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
+MONGOCXX_ABI_EXPORT_CDECL(std::error_category const&) source_error_category();
+
+///
+/// Support implicit conversion to `std::error_condition`.
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
+inline std::error_condition make_error_condition(source_errc code) {
+    return {static_cast<int>(code), v1::source_error_category()};
+}
 
 ///
 /// Enumeration identifying the type (cause) of a @ref mongocxx::v1 error.
 ///
 /// @attention This feature is experimental! It is not ready for use!
 ///
-enum class type_errc {};
+enum class type_errc {
+    zero,             ///< Zero.
+    invalid_argument, ///< An invalid argument passed to the throwing function.
+    runtime_error,    ///< An erroneous condition was detected at runtime.
+};
+
+///
+/// The error category for @ref mongocxx::v1::type_errc.
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
+MONGOCXX_ABI_EXPORT_CDECL(std::error_category const&) type_error_category();
+
+///
+/// Support implicit conversion to `std::error_condition`.
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
+inline std::error_condition make_error_condition(type_errc code) {
+    return {static_cast<int>(code), v1::type_error_category()};
+}
+
+BSONCXX_PRIVATE_WARNINGS_PUSH();
+BSONCXX_PRIVATE_WARNINGS_DISABLE(MSVC(4251));
+BSONCXX_PRIVATE_WARNINGS_DISABLE(MSVC(4275));
 
 ///
 /// Base class for all exceptions thrown by @ref mongocxx::v1.
@@ -45,10 +98,32 @@ enum class type_errc {};
 ///
 /// @attention This feature is experimental! It is not ready for use!
 ///
-class exception {};
+class exception : public std::system_error {
+   public:
+    ~exception() override;
+
+    exception(exception&&) noexcept = default;
+    exception& operator=(exception&&) noexcept = default;
+    exception(exception const&) = default;
+    exception& operator=(exception const&) = default;
+
+    using std::system_error::system_error;
+};
+
+BSONCXX_PRIVATE_WARNINGS_POP();
 
 } // namespace v1
 } // namespace mongocxx
+
+namespace std {
+
+template <>
+struct is_error_condition_enum<mongocxx::v1::source_errc> : true_type {};
+
+template <>
+struct is_error_condition_enum<mongocxx::v1::type_errc> : true_type {};
+
+} // namespace std
 
 #include <mongocxx/v1/detail/postlude.hpp>
 
--- a/src/mongocxx/include/mongocxx/v1/instance.hpp
+++ b/src/mongocxx/include/mongocxx/v1/instance.hpp
@@ -20,6 +20,14 @@
 
 #include <mongocxx/v1/detail/prelude.hpp>
 
+#include <mongocxx/v1/logger-fwd.hpp>
+
+#include <mongocxx/v1/config/export.hpp>
+
+#include <memory>
+#include <system_error>
+#include <type_traits>
+
 namespace mongocxx {
 namespace v1 {
 
@@ -50,14 +58,134 @@ namespace v1 {
 /// @par Special exemptions
 /// Only the following API are permitted to be used outside the lifetime of an instance object:
 /// - @ref mongocxx::v1::logger
+/// - @ref mongocxx::v1::default_logger
 ///
 /// @see
 /// - [Initialization and Cleanup (mongoc)](https://mongoc.org/libmongoc/current/init-cleanup.html)
 ///
-class instance {};
+class instance {
+   private:
+    class impl;
+    std::unique_ptr<impl> _impl;
+
+   public:
+    ///
+    /// Cleanup the mongocxx (and mongoc) library.
+    ///
+    /// Calls [`mongoc_init()`](https://mongoc.org/libmongoc/current/mongoc_cleanup.html).
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL() ~instance();
+
+    ///
+    /// This class is not moveable.
+    ///
+    instance(instance&&) = delete;
+
+    ///
+    /// This class is not moveable.
+    ///
+    instance& operator=(instance&&) = delete;
+
+    ///
+    /// This class is not copyable.
+    ///
+    instance(instance const&) = delete;
+
+    ///
+    /// This class is not copyable.
+    ///
+    instance& operator=(instance const&) = delete;
+
+    ///
+    /// Initialize the mongoc library with unstructured log messages disabled.
+    ///
+    /// Calls [`mongoc_init()`](https://mongoc.org/libmongoc/current/mongoc_init.html) after disabling unstructured
+    /// log messages by calling `mongoc_log_set_handler(nullptr, nullptr)`.
+    ///
+    /// @important To use mongoc's default log message handler, construct this object with
+    /// @ref instance(v1::default_logger tag) instead.
+    ///
+    /// @throws mongocxx::v1::exception with @ref mongocxx::v1::instance::errc::multiple_instances if an `instance`
+    /// object has already been created.
+    ///
+    /// @see
+    /// - [Custom Log Handlers (mongoc)](https://mongoc.org/libmongoc/current/unstructured_log.html#custom-log-handlers)
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL() instance();
+
+    ///
+    /// Initialize the mongoc library with the custom unstructured log message handler.
+    ///
+    /// Calls [`mongoc_init`](https://mongoc.org/libmongoc/current/mongoc_init.html) after registering the custom
+    /// unstructured log handler by calling `mongoc_log_set_handler()`.
+    ///
+    /// @param handler Disable unstructured logging when null.
+    ///
+    /// @throws mongocxx::v1::exception with @ref mongocxx::v1::instance::errc::multiple_instances if an `instance`
+    /// object has already been created.
+    ///
+    /// @see
+    /// - [Custom Log Handlers (mongoc)](https://mongoc.org/libmongoc/current/unstructured_log.html#custom-log-handlers)
+    ///
+    explicit MONGOCXX_ABI_EXPORT_CDECL() instance(std::unique_ptr<v1::logger> handler);
+
+    ///
+    /// Initialize the mongoc library with its default unstructured log handler.
+    ///
+    /// Calls [`mongoc_init`](https://mongoc.org/libmongoc/current/mongoc_init.html) without registering any custom
+    /// unstructured log handler.
+    ///
+    /// @param tag Unused: only for overload resolution.
+    ///
+    /// @throws mongocxx::v1::exception with @ref mongocxx::v1::instance::errc::multiple_instances if an `instance`
+    /// object has already been created.
+    ///
+    explicit MONGOCXX_ABI_EXPORT_CDECL() instance(v1::default_logger tag);
+
+    ///
+    /// Equivalent to @ref instance(std::unique_ptr<v1::logger> handler) when `handler == nullptr`.
+    ///
+    /// @see
+    /// - [Custom Log Handlers (mongoc)](https://mongoc.org/libmongoc/current/unstructured_log.html#custom-log-handlers)
+    ///
+    explicit MONGOCXX_ABI_EXPORT_CDECL() instance(std::nullptr_t);
+
+    ///
+    /// Errors codes which may be returned by @ref mongocxx::v1::instance.
+    ///
+    /// @attention This feature is experimental! It is not ready for use!
+    ///
+    enum class errc {
+        zero,               ///< Zero.
+        multiple_instances, ///< Cannot construct multiple instance objects in a given process.
+    };
+
+    ///
+    /// The error category for @ref mongocxx::v1::instance::errc.
+    ///
+    /// @attention This feature is experimental! It is not ready for use!
+    ///
+    static MONGOCXX_ABI_EXPORT_CDECL(std::error_category const&) error_category();
+
+    ///
+    /// Support implicit conversion to `std::error_code`.
+    ///
+    /// @attention This feature is experimental! It is not ready for use!
+    ///
+    friend std::error_code make_error_code(errc v) {
+        return {static_cast<int>(v), error_category()};
+    }
+};
 
 } // namespace v1
 } // namespace mongocxx
+
+namespace std {
+
+template <>
+struct is_error_code_enum<mongocxx::v1::instance::errc> : true_type {};
+
+} // namespace std
 
 #include <mongocxx/v1/detail/postlude.hpp>
 
--- a/src/mongocxx/include/mongocxx/v1/logger.hpp
+++ b/src/mongocxx/include/mongocxx/v1/logger.hpp
@@ -20,6 +20,11 @@
 
 #include <mongocxx/v1/detail/prelude.hpp>
 
+#include <bsoncxx/v1/detail/macros.hpp>
+#include <bsoncxx/v1/stdx/string_view.hpp>
+
+#include <mongocxx/v1/config/export.hpp>
+
 namespace mongocxx {
 namespace v1 {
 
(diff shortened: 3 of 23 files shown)


src/mongocxx/include/mongocxx/v1/instance.hpp
```diff
@@ -50,15 +58,135 @@ namespace v1 {
...
+    ///
+    instance(instance const&) = delete;
+
+    ///
+    /// This class is not copyable.
+    ///
+    instance& operator=(instance const&) = delete;
+
+    ///
+    /// Initialize the mongoc library with unstructured log messages disabled.
```
eramongodb, line 100: Updated documentation to clarify that this all applies to "unstructured log handling" to leave open the design space for "structured log handling" support by the C++ Driver.

Documentation also increased references to specific mongoc API to clarify behavior + defer much of the behavioral documentation to mongoc.

src/mongocxx/include/mongocxx/v1/instance.hpp
```diff
@@ -50,15 +58,135 @@ namespace v1 {
...
+
+    ///
+    /// This class is not copyable.
+    ///
+    instance(instance const&) = delete;
+
+    ///
+    /// This class is not copyable.
+    ///
+    instance& operator=(instance const&) = delete;
```
eramongodb, line 97: Relative breaking change: `instance` is made immovable as well as uncopyable. Allowing `instance` to be moved-from does not seem to be well-motivated.

src/mongocxx/lib/mongocxx/v1/instance.cpp
```diff
@@ -13,3 +13,193 @@
...
+// 0: no instance object has been created.
+// 1: there is a single instance object.
+// 2: the single instance object has been destroyed.
+std::atomic_int instance_state{0};
+
+void custom_log_handler(
+    mongoc_log_level_t log_level,
+    char const* domain,
+    char const* message,
+    void* user_data) noexcept {
```
eramongodb, line 58: Small relative change: added `noexcept` for consistency with other C API callback functions.

src/mongocxx/lib/mongocxx/v1/instance.cpp
```diff
@@ -13,3 +13,193 @@
...
+#define STDCXX BSONCXX_PRIVATE_STRINGIFY(__cplusplus)
+#endif
+
+            // Handshake data can only be appended once (or it will return false): this is that "once".
+            // Despite the name, mongoc_handshake_data_append() *prepends* the platform string.
+            // Use " / " to delimit handshake date for mongocxx and mongoc.
+            (void)libmongoc::handshake_data_append(
+                "mongocxx",
+                MONGOCXX_VERSION_STRING,
+                "CXX=" MONGOCXX_COMPILER_ID " " MONGOCXX_COMPILER_VERSION " stdcxx=" STDCXX " / ");
```
eramongodb, line 132: Minor optimization: avoid allocations, given all fields are literals, by concatenating everything into a single string literal.

src/mongocxx/test/subprocess.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto& is_signal = is_signal_ptr ? *is_signal_ptr : is_signal_local;
+
+    pid_t const pid = ::fork();
+
+    // Child: do nothing more than call `fn`.
+    if (pid == 0) {
+        // Use `std::exit()` and `std::terminate()` to prevent continued execution of the Catch2 test suite.
+        try {
+            fn();
+            std::exit(EXIT_SUCCESS);
```
vector-of-bool: Have you checked that these uses of `exit` are correct in the child process? Usually, with a `fork()` without `exec()`, you want to use `_Exit` instead of `exit()` to prevent cleanup handlers from executing in the weird intermediate state of the child process. (Using `_Exit` may or may not also avoid Valgrind weirdness.)

eramongodb: > Usually, with a `fork()` without `exec()`, you want to use `_Exit` instead of `exit()`

Good call. Converted `std::exit()` into `std::_Exit()` within the subprocess.

> Using `_Exit` may or may not also avoid Valgrind weirdness.

Alas, this doesn't entirely address Valgrind weirdness, as the abnormal termination of a subprocess still produces noisy diagnostic messages due to "still reachable" allocations in its exit summary (one for each terminated subprocess). There doesn't seem to be a straightforward way to suppress the diagnostic messages of the subprocesses only while still performing at-exit leak checks. Given the limited use of this pattern, I think I would prefer defer memchecking of subprocesses to ASAN/UBSAN task coverage instead and keep the `--trace-children=no` Valgrind flag for `test_instance` (fwiw it doesn't apply to any other test executables).

src/mongocxx/test/subprocess.hh
```diff
@@ -0,0 +1,60 @@
...
+// Execute `fn` in a forked subprocess.
+//
+// Should be invoked directly from the body of a TEST_CASE(), not within a SECTION(). Keep it simple!
+//
+// @param fn The function to be invoked within a subprocess.
+// @param is_signal_ptr When not null, `*is_signal_ptr` is set to `true` if the return value is a signal rather than an
+// exit code.
+//
+// @returns The exit code of the subprocess (`*is_signal` is `false`) or the signal used to kill the subprocess
+// (`*is_signal` is `true`) .
```
kevinAlbs:
```suggestion
// @returns The exit code of the subprocess (`*is_signal_ptr` is `false`) or the signal used to kill the subprocess
// (`*is_signal_ptr` is `true`) .
```

src/mongocxx/test/subprocess.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+TEST_CASE("skip", "[mongocxx][test][subprocess]") {
+    auto is_signal = false;
+    CHECK_SUBPROCESS([] {
+        // Try to silence noisy Catch2 output.
+        (void)::close(1); // stdout
+        (void)::close(2); // stderr
+
+        SKIP("subprocess");
+    });
```
kevinAlbs:
```suggestion
    CHECK_SUBPROCESS([] {
        // Try to silence noisy Catch2 output.
        (void)::close(1); // stdout
        (void)::close(2); // stderr

        SKIP("subprocess");
    }, &is_signal);
```

src/mongocxx/test/v1/logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+        bsoncxx::v1::stdx::string_view message) noexcept override {
+        (void)level;
+        (void)domain;
+        (void)message;
+    }
+};
+
+} // namespace
+
+// mongocxx::v1::logger must not unnecessarily mpose special requirements on derived classes.
```
kevinAlbs:
```suggestion
// mongocxx::v1::logger must not unnecessarily impose special requirements on derived classes.
```

src/mongocxx/include/mongocxx/v1/instance.hpp
```diff
@@ -50,15 +58,127 @@ namespace v1 {
...
+class instance {
+   private:
+    class impl;
+    std::unique_ptr<impl> _impl;
+
+   public:
+    ///
+    /// Cleanup the mongocxx (and mongoc) library.
+    ///
+    /// Calls [`mongoc_init()`](https://mongoc.org/libmongoc/current/mongoc_cleanup.html).
```
kevinAlbs:
```suggestion
    /// Calls [`mongoc_cleanup()`](https://mongoc.org/libmongoc/current/mongoc_cleanup.html).
```

src/mongocxx/lib/mongocxx/v1/instance.cpp
```diff
@@ -13,3 +13,191 @@
...
+#include <mongocxx/private/config/config.hh>
+#include <mongocxx/private/mongoc.hh>
+
+namespace mongocxx {
+namespace v1 {
+
+using code = v1::instance::errc;
+
+static_assert(!std::is_move_constructible<instance>::value, "bsoncxx::v1::instance must be non-moveable");
+static_assert(!std::is_copy_constructible<instance>::value, "bsoncxx::v1::instance must be non-copyable");
```
kevinAlbs:
```suggestion
static_assert(!std::is_move_constructible<instance>::value, "mongocxx::v1::instance must be non-moveable");
static_assert(!std::is_copy_constructible<instance>::value, "mongocxx::v1::instance must be non-copyable");
```

kevinAlbs (approved):
LGTM with minor comments. I like the new overload to simplify enabling default logging.

> how important are log messages which may be emitted by `mongoc_init()`?

I see two possible logs:
- [1](https://github.com/kevinAlbs/mongo-c-driver/blob/1b0e008d3aab07431827433c783f92680cb0cffe/src/libmongoc/src/mongoc/mongoc-openssl.c#L93) Failure to initialize OpenSSL.
- [2](https://github.com/kevinAlbs/mongo-c-driver/blob/1b0e008d3aab07431827433c783f92680cb0cffe/src/libmongoc/src/mongoc/mongoc-handshake.c#L328) Failure to get OS version for handshake.

I expect neither is likely. (2) might be improved by making handshake non-global (CDRIVER-4142) (errors could be returned on the first client operation).

> For backward compatibility, `v_noabi::instance`'s custom log handler constructor preserves the "register handler after `mongoc_init()`" behavior.

If it helps simplify: I expect having `v_noabi` register the log handler before `mongoc_init` (to match `v1`) is unlikely to harm consumers (some added unlikely-to-occur logs)

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/instance.cpp
```diff
@@ -46,41 +46,13 @@ instance* sentinel() {
...
-
-    impl(std::unique_ptr<logger> handler) try : _instance{v1::default_logger{}}, _handler{std::move(handler)} {
-        // For backward compatibility, allow v1::instance to initialize mongoc with the default log handler, then set
-        // the custom logger AFTER initialization has already completed.
-        if (_handler) {
-            libmongoc::log_set_handler(&custom_log_handler, _handler.get());
-
-            // Inform the user that a custom log handler has been registered.
-            // Cannot use mocked `libmongoc::log()` due to varargs.
-            mongoc_log(MONGOC_LOG_LEVEL_INFO, "mongocxx", "libmongoc logging callback enabled");
```
eramongodb, line 90: Do we want to keep this informational log message?

kevinAlbs: I am slightly in favor of keeping for `v_noabi`. The API examples [expect it](https://github.com/kevinAlbs/mongo-cxx-driver/blob/76bee7853e7b5d2514476a011c9a1ab0e6673cf7/examples/api/mongocxx/examples/logger/basic_usage.cpp#L41), and it might avoid a surprise to consumers. I expect it is not much burden for us to keep it in `v_noabi`. But I do not think it is needed for `v1`.

src/mongocxx/include/mongocxx/v1/exception.hpp
```diff
@@ -28,14 +35,60 @@ namespace v1 {
 ///
 /// @attention This feature is experimental! It is not ready for use!
 ///
-enum class source_errc {};
+enum class source_errc {
+    zero,       ///< Zero.
```
vector-of-bool, line 39: Should this be `zero`, or something like `okay`? Alternatively: set `mongocxx = 1`, give the `enum` a base type like `int`, and then have `source_errc{}` as the stand-in for `zero`, which is the non-error syntax for `std::errc`.

(I don't feel strongly about any of the above options, just something to consider.)

src/mongocxx/lib/mongocxx/v1/exception.cpp
```diff
@@ -13,3 +13,76 @@
...
+            }
+        }
+    };
+
+    static bsoncxx::immortal<type> const instance;
+
+    return instance.value();
+}
+
+exception::~exception() = default;
```
vector-of-bool, line 86: Is there a benefit to providing the defaulted destructor out-of-line? This inhibits inlining of the `exception` destructor without LTO enabled.

src/mongocxx/include/mongocxx/v1/exception.hpp
```diff
@@ -45,11 +98,33 @@ enum class type_errc {};
...
 
+namespace std {
+
+template <>
+struct is_error_condition_enum<mongocxx::v1::source_errc> : true_type {};
+
+template <>
+struct is_error_condition_enum<mongocxx::v1::type_errc> : true_type {};
+
+} // namespace std
```
vector-of-bool: Minor tweak option: You can omit the outer `namespace std` and do:

```c++
template <> 
struct std::is_error_condition_enum<...>
```

with the `std::` qualifier on the `struct` name. This isn't strictly _better_, but may reduce the amount of visual noise.

src/mongocxx/include/mongocxx/v1/logger.hpp
```diff
@@ -37,7 +56,60 @@ enum class log_level {};
...
+    /// @param domain The domain of the message.
+    /// @param message The contents of the log message.
+    ///
+    /// @see
+    /// - [Custom Log handlers (mongoc)](https://mongoc.org/libmongoc/current/unstructured_log.html#custom-log-handlers)
+    ///
+    virtual void operator()(
+        log_level level,
+        bsoncxx::v1::stdx::string_view domain,
+        bsoncxx::v1::stdx::string_view message) noexcept = 0;
```
vector-of-bool, line 107: May be beyond the scope of this PR, but we may want to defer the creation of `message` based on whether `level` is enabled in the logger, otherwise the user pays for the string-formatting of every message for every level, even if they e.g. don't have `trace` enabled. (`mongoc` already does this, but only for the `trace` level.)

Also recommend doing the "private virtual" idiom so that the base class can have logic (i.e. conditional formatting) in `operator()`, and send the message on to a private virtual like `do_write_log(level, domain, msg)`.

src/mongocxx/lib/mongocxx/v1/exception.cpp
eramongodb: This is the [key function](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#vague-vtable) for `v1::exception` in lieu of any other virtual functions:

> [...] if the class has a key function (see below), the tables are emitted in the object for the translation unit containing the definition of the key function. This is unique if the key function is not inline. Otherwise, the tables are emitted in every object that references any of them.

and:

> The key function is the first non-pure virtual function that is not inline at the point of class definition. [...]  Note that if the key function is not inline in the class definition, but its later definition is inline, it will be emitted in every object containing the definition.

The absence of a key function triggers the `-Wweak-vtables` Clang warning.

src/mongocxx/include/mongocxx/v1/exception.hpp
eramongodb: > Should this be `zero`, or something like `okay`? Alternatively: set `mongocxx = 1 [...]

My reason for including the `zero` enumerator was twofold:

* Avoid the need for explicit `= 1` in the otherwise `=`-less list of enumerators (an inconsistency).
* Force explicit handling of the `zero` case in switch cases via `-Wswitch` / `-Wswitch-enum`.

I opted for `zero` instead of `okay` to avoid implying anything beyond a description of the value itself, since "not zero" does not necessarily imply "not okay" or "not valid" given certain error code schemes (i.e. "HTTP status code 200").

> give the enum a base type like `int`

The default underlying type for a scoped enumeration is `int`, so there is no need for explicit `: int` given `enum class E`.

src/mongocxx/include/mongocxx/v1/exception.hpp
eramongodb: I initially attempted this pattern as well, but encountered compilation errors with GCC versions [older than 7.1](https://godbolt.org/z/cbvMxfrsx) (possibly related: [CWG 275](https://cplusplus.github.io/CWG/issues/275.html)). However, we now require GCC 8 or newer as of https://github.com/mongodb/mongo-cxx-driver/pull/1415, so I think the suggestion is [now viable](https://godbolt.org/z/zbea37asa): https://github.com/mongodb/mongo-cxx-driver/blob/cbb70dac6a4f56e5e98bece8a63a1a1ff37e7619/CMakeLists.txt#L28-L33

src/mongocxx/include/mongocxx/v1/logger.hpp
eramongodb: > we may want to defer the creation of `message` based on whether `level` is enabled in the logger

I believe it would be more appropriate to implement this in mongoc (as being done for the trace level) rather than in mongocxx. The `message` parameter is merely a view of whatever `mongoc_log()` has already parsed and formatted.

> recommend doing the "private virtual" idiom

This would will force incompatibility with the current `v_noabi` API, making the `using v1::logger;` alias within `mongocxx::v_noabi` unviable. This may be better suited as a followup extension to `v1` API (i.e. part of [CXX-2586](https://jira.mongodb.org/browse/CXX-2586)).

src/mongocxx/lib/mongocxx/v1/exception.cpp
eramongodb: On second thought, it may be preferable to use a _private_ virtual `.key_function()` for the sole purpose of acting as the key function instead of the destructor, which has the additional benefit of avoiding Ro5 boilerplate:

```cpp
class exception : public std::system_error {
   public:
    using std::system_error::system_error;

   private:
    MONGOCXX_ABI_NO_EXPORT virtual void key_function() const;
};
```

Will propose this pattern for review in a followup PR.

src/mongocxx/include/mongocxx/v1/logger.hpp
vector-of-bool: > I believe it would be more appropriate to implement this in mongoc (as being done for the trace level) rather than in mongocxx

May need to do both, as (IIUC) there is no mongocxx API that sends down a `va_list` to `mongoc`, and there is no `mongoc` API that accepts a `va_list` for logging.

Alternatively: I'm misremembering things and this is a non-issue 🙃

PR #1470, opened 2025-09-26
Replace scoped_bson_t with scoped_bson and scoped_bson_view

eramongodb:
Related to CXX-3320. This PR introduces a new helper utility _for internal use only_ to manage `bson_t` objects and the creation of BSON documents for internal implementation: `mongocxx::scoped_bson` and `mongocxx::scoped_bson_view` (a la `std::string` and `std::string_view`). These two classes replace `mongocxx::libbson::scoped_bson_t` (note: not to be confused with `scoped_bson_value` for `bson_value_t`). The v_noabi implementation is refactored to use the new `scoped_bson(_view)` utilities to verify its correctness + behavior in advance of its widespread use in `mongocxx::v1` implementations.

---

The `scoped_bson_t` class is _very_ old, going back all the way to [the first commit](https://github.com/mongodb/mongo-cxx-driver/blob/fe429b58ffe6e887841940991886affc336781d4/src/bson/libbson.hpp) (Jan 2015). Unfortunately it has been confusing and difficult to use due to preserving too much `bson_t` idiosyncracies in its API, in particular its handling of the"uninitialized `bson_t`" state (e.g. `.init_from_static()`). Its initially simple and primitive API has been expanded over the years to support integration with bsoncxx and mongocxx API, often in tension with `bson_t` behaviors (e.g. owning vs. non-owning and uninitialized vs. empty vs. moved-from). This PR proposes an alternative API which abandons the "uninitialized `bson_t`" state entirely and always preserves bsoncxx-idiomatic well-defined state.

First, both `scoped_bson` and `scoped_bson_view` completely avoid depending on `bson_t` for ownership or managment of underlying BSON data. No operations depend on the `bson_t` aside from exposing `bson_t const*` for mongoc API integration. Instead, the `bsoncxx::v1::document::value` or `bsoncxx::v1::document::view` data member always own and manage associated BSON data. The `bson_t` data member is always and only ever initialized with `bson_init_static()` via `this->sync_bson()` _after_ the bsoncxx value/view object has been updated first. Any applicable "would-be-uninitialized", "null", or "invalid" states are represented as an "invalid" document (`.data() == nullptr`) per v1 API's new well-defined "invalid" state for `value` and `view`.

> [!IMPORTANT]
> There is no "uninitialized `bson_t`" state exposed by the `scoped_bson(_view)` API.

Second, the owning `scoped_bson` class does its best to avoid unnecessary allocations. This includes using `bson_destroy_with_steal()` to acquire ownership of BSON data from `bson_t*` arguments, (re)using `bsoncxx::v1::document::view`'s special empty state whenever an empty document is detected to avoid allocating empty `bson_t` objects, and releasing ownership of underlying BSON data as `bsoncxx::v1::document::value` with rvalue overloads.

Third, `scoped_bson(_view)` provide `.out_ptr()` and `.inout_ptr()` to interface with mongoc API that requires `bson_t*` and `bson_t const**` arguments. This pattern is inspired by [P1132R8](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1132r8.html), although it has been greatly simplified to suit our limited purposes. This ensures the `scoped_bson(_view)` object always remains in a valid, well-defined state regardless of the result of the mongoc API call.

Last, `scoped_bson` implements an `operator+=` overload to minimally yet efficiently support building complex BSON documents in v1 implementations without depending on the v\_noabi builder API. The "concatenation" operation is ultimately the only builder operation needed to build internal BSON documents: all other BSON builder behavior can be handled by the `BCON_*` family of macros, whose resulting `bson_t*` is (typically) immediately used to initialize a `scoped_bson` object:

```cpp
scoped_bson doc{BCON_NEW("x", BCON_INT32(1))};

if (cond) {
  doc += scoped_bson{BCON_NEW("y", BCON_INT32(2))};
}

doc.view(); // {"x": 1} or {"x": 1, "y": 2}
```

This will be the primary method by which the v1 implementation constructs BSON documents for internal use until a new public BSON builder API is implemented with [CXX-3275](https://jira.mongodb.org/browse/CXX-3275). A relaxed extended JSON constructor is also implemented (e.g. `scoped_bson{R"("x": 1)"}`) , but it is primarily for test convenience only. Implementation code should prefer to avoid string parsing performance overhead and should prefer to be explicit about the types of BSON element being constructed.

> [!NOTE]
> Update: this problem is documented by [CDRIVER-6113](https://jira.mongodb.org/browse/CDRIVER-6113).
>
> There was an attempt to default the `scoped_bson_view` copy SMFs and the `scoped_bson` move SMFs. This _should_ be possible in theory given `bson_t` is only ever a view to the accompanying `value` or `view` object's underlying BSON data. However, this led very difficult-to-diagnose undefined behavior. Resulting spurious segmentation faults include stack traces pointing to `std::function<void (unsigned char*)>::function(std::function<void (unsigned char*)>&&)` with `this=0.0` or the `__memcmp_avx2_movbe` intrinsic operation (???). These segfaults were unobservable when sanitizers are enabled (?!). Neither `_DFORTIFY_SOURCE=3` nor `GLIBCXX_ASSERTIONS` were able to identify any problems. My takeaway is that `bson_t` is _not_ trivially-copyable even when only ever zero-initialized or initialized with `bson_init_static()`.


src/mongocxx/lib/mongocxx/private/scoped_bson.hh
```diff
@@ -0,0 +1,364 @@
...
+    };
+
+    out_ptr_type out_ptr() & {
+        return {*this};
+    }
+
+   private:
+    void sync_bson() {
+        if (_view && !bson_init_static(&_bson, _view.data(), _view.length())) {
+            MONGOCXX_PRIVATE_UNREACHABLE;
```
kevinAlbs, line 144: Should this throw instead of abort? Or possibly ignore a failed `bson_init_static` and so later access can throw?

The `document::view` constructor validates less than `bson_init_static`. 

```c++
TEST_CASE("bad data", "[mongocxx][private][scoped_bson_view]") {
    std::uint8_t const data[] = {5, 0, 0, 0, 1};                // Last byte non-0. Validated by `bson_init_static`.
    bsoncxx::v1::document::view const view{data, sizeof(data)}; // Does not throw.
    auto sb = scoped_bson_view{view};                           // Aborts!
}
```

(Though aborting still seems like an improvement over the [previous unchecked `bson_init_static`](https://github.com/kevinAlbs/mongo-cxx-driver/blob/76bee7853e7b5d2514476a011c9a1ab0e6673cf7/src/mongocxx/lib/mongocxx/private/bson.cpp#L27)).

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <cstdint>
+#include <stdexcept>
+
+#include <bsoncxx/private/bson.hh>
+
+namespace mongocxx {
+
+scoped_bson& scoped_bson::operator+=(scoped_bson_view other) {
+    if (!_value) {
+        throw std::logic_error{"mongocxx::scoped_bson::inout_ptr: this->data() == nullptr"};
```
kevinAlbs:
```suggestion
        throw std::logic_error{"mongocxx::scoped_bson::operator+=: this->data() == nullptr"};
```

Similar suggestion applies to other error messages.

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
```diff
@@ -0,0 +1,90 @@
...
+        auto ptr = _value.release();
+
+        std::uint8_t* data = ptr.get();
+
+        // Conditional ownership transfer.
+        auto bson = bson_new_from_buffer(&data, &length, nullptr, nullptr);
+
+        // Ownership transfer failed.
+        if (!bson) {
+            bson_destroy(bson);
```
kevinAlbs: Suggest removing unneeded `bson_destroy` of NULL.

```suggestion
```

src/mongocxx/lib/mongocxx/private/scoped_bson.hh
eramongodb: The choice to terminate is deliberate, as `this->sync_bson()` is invoked by both noexcept move constructors as well as noexcept out_ptr/inout_ptr destructors. Given this is internal API, I expect (hope) we do not ever permit a situation where `bson_init_static` would actually fail. Instead, `operator+=` will be responsible for the construction of complex BSON documents (including with user-provided values) by the mongocxx library. If `bson_concat` ever fails (which we do not really ever expect to happen, otherwise we'd have to verbosely document most of the mongocxx API as potentially throwing `bsoncxx::v1` exceptions), the undocumented `std::logic_error` exception will be thrown so that users at least have the ability to recover. In short, document validity are an unstated general precondition for mongocxx API accepting documents.

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
eramongodb: TIL Catch2 supports [exception matchers](https://catch2-temp.readthedocs.io/en/latest/matchers.html#matching-exceptions) to assert both the exception type _and_ the exception message string. 👏 Fixed and updated tests accordingly.

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
```diff
@@ -0,0 +1,89 @@
...
+        auto bson = bson_new_from_buffer(&data, &length, nullptr, nullptr);
+
+        // Ownership transfer failed.
+        if (!bson) {
+            _value = bsoncxx::v1::document::value{std::move(ptr)}; // Strong exception safety.
+            throw std::logic_error{"mongocxx::scoped_bson::operator+=: bson_new_from_buffer failed"};
+        }
+
+        // Unset BSON_FLAG_NO_FREE: data is now owning.
+        bson->flags = 0;
```
connorsmacd, line 72:
```suggestion
        bson->flags = BSON_FLAG_NONE;
```

Or if the intention is to _only_ unset `BSON_FLAG_NO_FREE`:

```suggestion
        bson->flags &= ~BSON_FLAG_NO_FREE;
```

I don't know if `flags = 0` is more idiomatic in this repo, but I prefer the clarity of explicitly using enumerators.

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
```diff
@@ -0,0 +1,89 @@
...
+        auto const deleter_ptr = _value.get_deleter().target<decltype(&bson_free)>();
+        return deleter_ptr && *deleter_ptr == &bson_free;
+    }();
+
+    // Copy is required for strong exception safety or allocator compatibility.
+    if (is_self_assignment || !is_bson_freeable) {
+        bson_t bson = BSON_INITIALIZER;
+
+        if (bson_concat(&bson, this->bson()) && bson_concat(&bson, other.bson())) {
+            *this = scoped_bson{&bson}; // Ownership transfer.
```
connorsmacd:
```suggestion
            return *this = scoped_bson{&bson}; // Ownership transfer.
```

Personally, I rather see an early return instead of relying on fall-through.

src/mongocxx/lib/mongocxx/private/scoped_bson.hh
```diff
@@ -0,0 +1,364 @@
...
+
+        /* explicit(false) */ operator bson_t const*() const& = delete;
+    };
+
+    out_ptr_type out_ptr() & {
+        return {*this};
+    }
+
+   private:
+    void sync_bson() {
```
connorsmacd, line 142:
```suggestion
    void sync_bson() noexcept {
```

It doesn't look like anything in this function throws (even though `view::data` and `view::length` are not actually marked `noexcept`).

Provided you agree, I can see a number of member functions in this file that could be marked noexcept (e.g., the default constructors) that you may want to look at. I figure you don't need the suggestion spam.

eramongodb: We've adopted a policy going forward of only using the `noexcept` specifier for functions which have a _known_ use case for conditional branching on the nothrow guarantee. The coding guidelines have not yet been updated with this guidance (my fault, sorry). See https://github.com/mongodb/mongo-cxx-driver/pull/1402#discussion_r2096510603 for details.

> The [only situations listed](https://github.com/10gen/mongo/blob/0ec56d807762cd91eff48f5d691fbade66366deb/docs/exception_architecture.md#cases-where-noexcept-is-encouraged) where `noexcept` is explicitly encouraged as valid use cases are the aforementioned "move operations" and "swap operations" (consistent with MIN), "destructors and 'destructor-safe' functions" (which we assume to be a given), and "hash functions" (also consistent with MIN, but not applicable to our codebase).

src/mongocxx/lib/mongocxx/private/scoped_bson.cpp
eramongodb: `BSON_FLAG_*` are [private macros](https://github.com/mongodb/mongo-c-driver/blob/e5c32351f59eaff50b4d6019225b10ddfa414d93/src/libbson/src/bson/bson-private.h#L32-L40) that are not provided by the public bson library headers.

```
error: use of undeclared identifier 'BSON_FLAG_NONE'
      |         bson->flags = BSON_FLAG_NONE;
      |                       ^
```

connorsmacd (approved):
LGTM!

PR #1473, opened 2025-10-07
CXX-3363 bump required C driver to 2.1.2

kevinAlbs:
Bump required C driver to 2.1.2. Motivated by fix to CDRIVER-6112.

Some test scripts detect of a C driver install. These were updated from checking `bson-config.h` to `bson.h` to account for header renames made in CDRIVER-4153.

Coming soon: I plan to backport only a bump to the auto-downloaded version (`MONGOC_DOWNLOAD_VERSION`) not the required version (`MONGOC_REQUIRED_VERSION`) to avoid a minor version dependency bump in a patch release.

### DIFF
--- a/etc/augmented.sbom.json
+++ b/etc/augmented.sbom.json
@@ -1,16 +1,16 @@
 {
   "components": [
     {
-      "bom-ref": "pkg:github/mongodb/mongo-c-driver@v2.1.2",
+      "bom-ref": "pkg:github/mongodb/mongo-c-driver@2.1.2",
       "copyright": "Copyright 2009-present MongoDB, Inc.",
       "externalReferences": [
         {
           "type": "distribution",
-          "url": "https://github.com/mongodb/mongo-c-driver/archive/v2.1.2.tar.gz"
+          "url": "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/2.1.2.tar.gz"
         },
         {
           "type": "website",
-          "url": "https://github.com/mongodb/mongo-c-driver/tree/v2.1.2"
+          "url": "https://github.com/mongodb/mongo-c-driver/tree/2.1.2"
         }
       ],
       "group": "mongodb",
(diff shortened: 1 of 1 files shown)


etc/cyclonedx.sbom.json
```diff
@@ -1,16 +1,16 @@
...
   "components": [
     {
-      "bom-ref": "pkg:github/mongodb/mongo-c-driver@v2.0.2",
+      "bom-ref": "pkg:github/mongodb/mongo-c-driver@v2.1.2",
       "copyright": "Copyright 2009-present MongoDB, Inc.",
       "externalReferences": [
         {
           "type": "distribution",
-          "url": "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/v2.0.2.tar.gz"
+          "url": "https://github.com/mongodb/mongo-c-driver/archive/v2.1.2.tar.gz"
```
eramongodb:
```suggestion
          "url": "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/v2.1.2.tar.gz"
```
Lost the `refs/tags/` here and in the Augmented SBOM, which seems to be necessary (current link is 404). 🤔 Replicated this unexpected change locally. Asked DevProd about this issue. Suggest manually re-adding the `refs/tags/` to the URL here and in the Augmented SBOM for now.

eramongodb: Solved: the purl should not have that `v`. 🤦 
```diff
- pkg:github/mongodb/mongo-c-driver@v2.1.2
+ pkg:github/mongodb/mongo-c-driver@2.1.2
```
Looks like this mistake goes all the way back to https://github.com/mongodb/mongo-cxx-driver/issues/1134 (first SBOM generated). C Driver release tags do not have a `v` prefix.

kevinAlbs: Great spot. And thank you for investigating. Updated and filed CXX-3364 to consider updating previously released SBOMs.

eramongodb (approved):

PR #1607, opened 2026-04-02
CXX-3456: Change CMake `*_SOVERSION` cache variables from INTERNAL to STRING

eramongodb:
Attempts to build experimental Debian packages using `-DBSONCXX_SOVERSION=1` unexpectedly failed due to the following error during the install step:

```
CMake Error at build/src/bsoncxx/cmake/cmake_install.cmake:57 (file):
  file INSTALL cannot find
  "build/src/bsoncxx/libbsoncxx.so.4.3.0-pre":
  No such file or directory.
```

This turned out to be caused by a sequence of unfortunate events:

- `BSONCXX_SOVERSION` is an internal cache variable, which implies `FORCE`.
	- `FORCE` means the cache variable is unconditionally overwritten by the call to `set()`.
	- This is not a problem for typical CMake workflows where build and install steps do not trigger reconfiguration unless the CMake configuration files are modified.
	- Even when the reconfiguration takes place, CMake ensures the subsequent build and install steps are consistent with the updated configuration.
- The package rules build both the `all` and `doxygen-current` targets in sequential order.
- `doxygen-current` uses the `build` directory as its binary directory.
- The `generate-latest-apidocs.sh` script invoked by `doxygen-current` (re)configures `build`.
	- This overwrites `-DBSONCXX_SOVERSION=1` and "resets" its value to "_noabi".
- The packaging routines then performs the equivalent of `cmake --install build`, which does NOT rerun the build step, but instead directly jumps to the install step.
    - The reconfigured cache's `BSONCXX_SOVERSION` has the value `_noabi`, which is the value used by the install step.

After considering the options available to remedy this issue (e.g. using a different binary directory by `generate-latest-apidocs.sh`, using a different binary directory by the Debian packaging routines, or using a temporary binary directory via `doxygen-latest` instead of `doxygen-current`), this PR proposes simply changing the `*_SOVERSION` variables from `INTERNAL` to `STRING` to avoid the reconfiguration -> reset problem. This is not a value which should be "reset" regardless of the circumstances (even when intended for internal use only and even when settings its value via `-DBSONCXX_SOVERSION` is not normally-expected behavior).


rcsanchez97 (approved):
Awesome! Thanks for fixing this. FYI, there is no need to rush out a 4.2.1 release for this. I can simply drop it in as a patch against 4.2.0 and then remove it whenever the next regularly scheduled release is made.

kevinAlbs (approved):
Nice investigation. LGTM

PR #1608, opened 2026-04-09
CXX-3434 Generate top-level `fwd.hpp` headers using CMake

eramongodb:
Resolves CXX-3434 (adding missing `fwd.hpp` for `v1`) by refactoring the CMake configuration such that the top-level `fwd.hpp` in each ABI namespace directory is now generated rather than hand-written.

Given the contents of top-level `fwd.hpp` headers is not dependent on the CMake configuration, I believe it is acceptable to leave its location as-is (notably, not under a `config` subdirectory like most other generated public header files).

This refactor also caught some include directives which were missing in top-level `v_noabi` forward headers:

```
bsoncxx/v_noabi/bsoncxx/
  + id-fwd.hpp
  + value-fwd.hpp
  + view-fwd.hpp
mongocxx/v_noabi/mongocxx/
  + options/text-fwd.hpp
```


vector-of-bool (approved):
LGTM!

PR #1609, opened 2026-04-16
CXX-3466: deprecate support for VS 2015

kevinAlbs:
CXX-3215 proposes dropping support of VS 2015. Prior to doing so, this PR proposes adding a CHANGELOG entry to document support for VS 2015 is deprecated to give users a heads up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
         message(FATAL_ERROR "Visual Studio 13 2015 Update 3 or newer is required")
     endif()
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.10.25017")
-        message(DEPRECATION "Visual Studio 2015 is deprecated. Visual Studio 2017 version 15.0 or newer is required")
+        message(DEPRECATION "Visual Studio 2017 version 15.0 or newer will be required")
         # TODO(CXX-3215) replace DEPRECATION with:
         # message(FATAL_ERROR "Visual Studio 2017 version 15.0 or newer is required")
     endif()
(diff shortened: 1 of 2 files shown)


eramongodb (changes requested):
Also update: https://github.com/mongodb/mongo-cxx-driver/blob/f134a90a644ade049964444140eb1a838aef92fd/CMakeLists.txt#L38-L39

I think the diff should look something like this (tweaking the message format a bit for better consistency with referenced webpages):

```diff
-    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.0.24210")
-        message(FATAL_ERROR "Visual Studio 13 2015 Update 3 or newer is required")
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.10.25017")
+        message(FATAL_ERROR "Visual Studio 2017 version 15.0 or newer is required")
```

kevinAlbs:
> Also update

Good idea. But I added as a DEPRECATION warning rather than a FATAL_ERROR. I intend to only deprecate (not remove) ahead of the 4.3.0 release.

Pending further data (see [slack](https://mongodb.slack.com/archives/GMY98JS86/p1776351020842299)) I would rather err towards deprecating first to give users a heads up.

CMakeLists.txt
```diff
@@ -38,6 +38,11 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.0.24210")
         message(FATAL_ERROR "Visual Studio 13 2015 Update 3 or newer is required")
     endif()
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.10.25017")
+        message(DEPRECATION "Visual Studio 2015 is deprecated. Visual Studio 2017 version 15.0 or newer is required")
```
eramongodb:
```suggestion
        message(DEPRECATION "Visual Studio 2017 version 15.0 or newer is required")
```
Redundant mention of deprecation in a deprecation message, e.g.:
```
CMake Deprecation Warning at CMakeLists.txt:42 (message):
  Visual Studio 2015 is deprecated.  Visual Studio 2017 version 15.0 or newer
  is required
```
Can tweak "is" -> "will be" if preferable.

eramongodb (approved):
Wording tweak; otherwise, LGTM.

CMakeLists.txt
kevinAlbs: Updated (using "will be" to signal VS 2017 is not-yet strictly required).

PR #1612, opened 2026-04-21
CXX-3439 Add `maxAdaptiveRetries` and `enableOverloadRetargeting` URI fields

eramongodb:
Resolves CXX-3439 (CXX-3328) as followup to CDRIVER-6274 (CDRIVER-6073) and CXX-3468.

Because most changes required by the relevant JIRA issues pertain to internal (C driver) behavior or tests which the C++ driver does not implement (due to redundancy or due to blocked-on status), the only meaningful changes required for the C++ driver is extending support for querying URI options (read-only API) to include the new "maxAdaptiveRetries" and "enableOverloadRetargeting" fields.

Extending test coverage to account for these two options reveal inconsistent handling of invalid values relative to other int32 and bool fields following https://github.com/mongodb/mongo-c-driver/pull/2269#discussion_r3081069742, where `mongoc_uri_get_option_as_int32()` is deliberately unused to avoid `0`-special-casing, but also demoted the behavior of invalid values from "error" to "warn and ignore". This manifests as the unexpected absence of exceptions throw by the URI constructor(s) relative to the `_test_int32_options()` and `_test_bool_options()` test helper functions (discrepancies marked with "**NOT** an error" comments for clarity).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,16 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 
 - Bump the minimum required C Driver version to [2.3.0](https://github.com/mongodb/mongo-c-driver/releases/tag/2.3.0).
 
+### Added
+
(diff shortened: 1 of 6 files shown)


kevinAlbs (approved):

connorsmacd (approved):
LGTM

PR #1614, opened 2026-04-22
Post-release changes for 4.3.0

eramongodb:
Resolves CXX-3422.


kevinAlbs (approved):

eramongodb:
Overlooked updating `SITEMAP_URL` in `Doxyfile` prior to tag and release. The `sitemap.xml` file had to be [manually fixed](https://github.com/mongodb/mongo-cxx-driver/commit/6b053733b2ff3733b6bfc744081ed8bbf40e63ec) so that it uses the new API version number in its references to 4.3.0 API documentation pages (the `doxygen-deploy` target uses the `r4.3.0` tag as-is, which contains the wrong `SITEMAP_URL` value).

PR #1620, opened 2026-04-27
CXX-3449 add read_concern to index options classes

EdanStasiuk:
**Related Ticket:** [CXX-3449](https://jira.mongodb.org/browse/CXX-3449)

## Summary
This PR addresses [CXX-3449](https://jira.mongodb.org/browse/CXX-3449), a subtask of [CXX-1748](https://jira.mongodb.org/browse/CXX-1748), which tracks `read_concern` support for individual operations. It introduces support for setting `read_concern` on the following:
- `v1::indexes`
- `v_noabi::options::index_view`

Previously, unlike `write_concern`, `read_concern` could not be set per operation. This change aligns the API with the existing behaviour for write concerns.

## Tests
Added unit tests for `read_concern` in:
- `v1/indexes.cpp`
- `v_noabi/options/index_view.cpp`

### DIFF
--- a/src/mongocxx/include/mongocxx/v1/indexes.hpp
+++ b/src/mongocxx/include/mongocxx/v1/indexes.hpp
@@ -25,6 +25,7 @@
 
 #include <mongocxx/v1/client_session-fwd.hpp>
 #include <mongocxx/v1/cursor-fwd.hpp>
+#include <mongocxx/v1/read_concern-fwd.hpp>
 #include <mongocxx/v1/write_concern-fwd.hpp>
 
 #include <bsoncxx/v1/document/value.hpp>
@@ -430,6 +431,7 @@ class indexes {
     /// - `comment`
     /// - `commit_quorum` ("commitQuorum")
     /// - `max_time` ("maxTimeMS")
+    /// - `read_concern` ("readConcern")
     /// - `write_concern` ("writeConcern")
     ///
     class create_one_options {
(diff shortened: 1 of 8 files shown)


eramongodb:
Upon further inspection of the [Index Management](https://github.com/mongodb/specifications/blob/b519824da64005cdf99ca680fc49c4e278af0ef3/source/index-management/index-management.md) specification and cross-referencing with the [Read and Write Concern](https://github.com/mongodb/specifications/blob/b519824da64005cdf99ca680fc49c4e278af0ef3/source/read-write-concern/read-write-concern.md) specification, the "createIndex" and "dropIndex" should not support setting a "readConcern" field.

Per Index Management spec:

> The `createIndexes` and `dropIndexes` commands take a write concern that indicates how the write is acknowledged. \[...\] As such, the most natural place to indicate write concern is on the client, the database, or the collection itself and not the operations within it. \[...\] It is permitted to allow a write concern option, but since `writeConcern` is a top-level command option, it MUST NOT be specified as part of an `IndexModel` passed into the helper.

Per Read and Write Concern spec:

> The read concern option is available for the following operations:
> - (not listed)

> Command helper methods for commands that write, other than those discussed above, MAY accept a write concern or write concern options in their parameter list. \[...\] These commands that write are:
> - `createIndexes`
> - `dropIndexes`

The [`createIndexes` (database command)](https://www.mongodb.com/docs/manual/reference/command/createIndexes/#transactions) documentation states users may set "readConcern" on the _transaction_ executing "createIndexes", but not on the "createIndex" command itself:

> To use `createIndexes` in a transaction, the transaction must use read concern `"local"`. If you specify a read concern level other than `"local"`, the transaction fails.

Therefore, unlike other command options classes being addressed by CXX-1748, index create/drop command options are correct to support setting _only_ "writeConcern only" without an accompanying "readConcern" field.

Closed CXX-3449 with status "Works as Designed"; also closing this PR accordingly. Apologies for not investigating these requirements in advance.

PR #1623, opened 2026-04-28
Update release instructions and related scripts

eramongodb:
Followup to the 4.3.0 release. Fixes and improves release instructions and relevant scripts.

 ---

Fixes and improvements to `make_release.py` include:

- Adding support for a `--allow-unreleased-changelog-entry` to support pre-release dry-runs against the unmodified CHANGELOG.
- Fix a PyGithub deprecation warning which suggests using `Auth.Token(token)` instead of the old `login_or_token` positional argument.
- Fix the `check_pre_release()` regex to match against the _entire_ tag name (add the end-of-line `$` anchor), not just against a potential-prefix.
- Fix an out-of-place `line` argument when constructing the error message thrown by the duplicate version check.

Fixes and improvements to `test_make_release.py` include:

- Refactored assertions into fine-grained test cases.
- Refactored changelog contents into constant variables to separate test logic from setup and inputs.

Fixes and improvements to release instructions include:

- Move the "Driver Development Status" table update step for `README.md` from "Post-Release Steps" into "Release Steps" so the updated table is included in the release tag.
- Replace the dangerous `git clean -dfx` command and hardcoded paths with temporary directories and "clean repository" instructions.
- Fixed `install-build-tools.sh` so it is not sensitive to the current working directory (exposed by the tmpdir/clean repo changes above).
- Updated reference to CXX-3042 (duplicate) and replaced it with CXX-3062 (successor) in Snyk `--exclude=extras` instructions.
- Added `rsync` to list of required commands prior to Hugo page deployment.
- Updated sitemap update instructions to account for `<priority>` tags (only applicable to the current and latest API doc versions, not older versions).
- Tweaked the GitHub Webhook instructions to account for stricter repository permissions (Admins only).
- Added and updated various commands and examples to improve the flow of instructions.

### DIFF
--- a/etc/make_release.py
+++ b/etc/make_release.py
@@ -168,6 +168,8 @@ def release(
     auth_jira = JIRA(jira_options, token_auth=jira_token)
 
     github_token = read_github_creds(github_token_file)
+    if not github_token:
+        sys.exit(1)
     auth_gh = Github(auth=Auth.Token(github_token))
 
     if not is_valid_remote(remote):
--- a/etc/make_release.py
+++ b/etc/make_release.py
@@ -206,7 +206,7 @@ def release(
         click.echo('Tag {} does not point to HEAD...exiting!'.format(release_tag), err=True)
         sys.exit(1)
 
-    is_pre_release = check_pre_release(release_tag)
+    is_pre_release = is_pre_release_tag(release_tag)
 
     if not quiet:
         found_msg = 'Found release tag {}'.format(release_tag)
@@ -413,13 +413,14 @@ def release_tag_points_to_head(release_tag):
     return tag_commit == repo.head.commit
 
 
-def check_pre_release(tag_name):
+def is_pre_release_tag(tag_name):
     """
-    Check the given tag to determine if it is a release tag, that is, whether it
-    is of the form rX.Y.Z.  Tags that do not match (e.g., because they are
-    suffixed with someting like -beta# or -rc#) are considered pre-release tags.
-    Note that this assumes that the tag name has been validated to ensure that
-    it starts with something like rX.Y.Z and nothing else.
+    Return true when the given (valid) release tag is a pre-release tag.
+
+    A pre-release tag must contain "extra" labels delimited by a '-':
+    - Normal:      r3.9.0
+    - Pre-release: r3.9.0-rc0, r3.9.0-extra, etc.
+    - Invalid:     r3.9.0-, r3.9.0.extra, r3.9.0extra, etc.
     """
 
     match = RELEASE_TAG_RE.fullmatch(tag_name)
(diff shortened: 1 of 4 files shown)


etc/make_release.py
```diff
@@ -414,7 +420,7 @@ def check_pre_release(tag_name):
     it starts with something like rX.Y.Z and nothing else.
     """
 
-    release_re = re.compile('^r[0-9]+\\.[0-9]+\\.[0-9]+')
+    release_re = re.compile('^r[0-9]+\\.[0-9]+\\.[0-9]+$')
```
kevinAlbs: IIUC this function is intended to only return true for pre-release tags. Suggest simplifying to:

```python
def check_pre_release(tag_name):
    """
    Check if the given tag is a pre-release tag
    """
    return RELEASE_TAG_RE.match(tag_name).group('verpre') is not None
```

etc/releasing.md
```diff
@@ -521,12 +539,13 @@ podman pull 901841024863.dkr.ecr.us-east-1.amazonaws.com/release-infrastructure/
...
-
-git add etc/cyclonedx.sbom.json
-git commit -m "update SBOM serial number"
 ```
 
-Update `etc/augmented.sbom.json` by running a patch build which executes the `sbom` task as described above in [SBOM Lite](#sbom-lite).
+Update `etc/augmented.sbom.json` by running a patch build which executes the `sbom` task as described above in [SBOM Lite](#sbom-lite):
+
+```
+evergreen patch -y -p mongo-cxx-driver -t sbom -v sbom -u
```
kevinAlbs: Suggest adding finalize flag (`-f`) for consistency with earlier patch command.

```suggestion
evergreen patch -y -p mongo-cxx-driver -t sbom -v sbom -f -u
```

kevinAlbs (commented):
The updates are much appreciated. LGTM with two minor suggestions.

etc/make_release.py
eramongodb: Also updated to use `.fullmatch()` instead of `.match()` (`^pattern$` vs. `^pattern`).

etc/make_release.py
```diff
@@ -414,9 +422,8 @@ def check_pre_release(tag_name):
     it starts with something like rX.Y.Z and nothing else.
     """
```
kevinAlbs, line 424: As a drive-by, suggest revising the comment of `check_pre_release`. This first sentence seems misleading:

```python
"""
Check if the given tag is a pre-release tag
"""
```

kevinAlbs (approved):

etc/make_release.py
eramongodb: Also renamed the function to `is_pre_release_tag()` to better describe its purpose.

PR #1624, opened 2026-04-29
Support JIRA epic status in release issues

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1623. Epics may be listed in release versions alongside "normal" JIRA issues when they are given a fix version. Unlike normal issues, the "status" field for an epic is set to "Development Complete", not "Closed".


kevinAlbs (approved):
Good spot. LGTM

PR #1625, opened 2026-04-29
Update etc/test_make_release.py with pytest and extended test coverage

eramongodb:
Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1623 and https://github.com/mongodb/mongo-cxx-driver/pull/1624.

Extends test coverage of `make_release.py` to include the bugfixes made in prior PRs (improved regex patterns, JIRA epic status, etc.). Additionally takes this opportunity to migrate from `unittest` to `pytest`, enabling use of more concise assertion syntax and rich features such as `capsys`. The tests may be executed both by running `pytest etc/test_make_release.py` or `etc/test_make_release.py` (via `uv run --group test_make_release`).

Other features such as `tmp_path` and `monkeypatch` may eventually be used to extend test coverage to functions such as `read_github_creds()`, `ensure_c_driver()`, and even `is_valid_remote()` (by using `tmp_path` and `monkeypatch`), but these are excluded from this PR for now to keep its scope manageable. Also considered adding `pytest-xdist` for parallel test execution, but the current number of tests and their duration does not justify its use yet (setting up workers takes longer than simply executing the tests sequentially); this can easily be added later.

### DIFF
--- a/etc/test_make_release.py
+++ b/etc/test_make_release.py
@@ -1,12 +1,13 @@
 import sys
 import textwrap
+import types
 
 import click
 import pytest
 
 import make_release
 
-# Changelog entries for the 3.9.0 release.
+# Changelog entries for the 1.2.3 release.
 CHANGELOG_ENTRIES = textwrap.dedent("""\
 - Add CMake option `MONGOCXX_OVERRIDE_DEFAULT_INSTALL_PREFIX` (default is `TRUE`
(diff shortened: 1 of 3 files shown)


kevinAlbs (approved):
LGTM, I like the convenience of `capsys` to test output.

And TIL `parametrize` is an alternative spelling of `parameterize`.

PR #1629, opened 2026-05-04
CXX-3483 Avoid implementation patterns that require RTTI

eramongodb:
Resolves CXX-3483 as a quality of implementation (QoI) improvement. (This does not constitute official support for `-fno-rtti` build configurations.)

This PR contains three distinct set of changes, ordered by the size of associated diffs:

- Convert `v1::server_exception` into `v_noabi` exceptions using an independent helper function from `v1::exception`.
- Revert `v_noabi` BSON document/array representations to their former standalone `unique_ptr_type + size_t`.
- Revert WiredTiger Storage Options detection to its former type-id-based check.

The `v_noabi::throw_exception<T>(ex)` internal helper function (where `T` derives `v_noabi::operation_exception`) used a `dynamic_cast` to detect and handle `v1::server_error` exceptions. This dynamic cast is a convenience to allow `catch (v1::exception const& ex)` blocks to handle all v1 exceptions thrown by the mongocxx library. However, this convenience also (arguably) obscures the actual set of v1 exceptions which may be thrown by internal v_noabi code reusing v1 API. By separating the `v1::server_error` case into its own catch block, these potential exceptions are made more explicit. Although this requires adding a substantial amount of repetitive catch blocks, I believe the resulting explicitness is a worthwhile improvement.

This investigation also helped identify a missing API documentation entry in `v1::gridfs::downloader`, where `read()` may potentially throw a server error by cursor iteration.

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
+++ b/src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
@@ -42,10 +42,6 @@ namespace array {
 /// should be used sparingly; array::view should be used instead wherever possible.
 ///
 class value {
-   private:
-    v1::array::value _value;
-    std::size_t _length;
-
    public:
     /// @copydoc v_noabi::document::value::deleter_type
     using deleter_type = void(BSONCXX_ABI_CDECL*)(std::uint8_t*);
@@ -53,6 +49,11 @@ class value {
     /// @copydoc v_noabi::document::value::unique_ptr_type
     using unique_ptr_type = std::unique_ptr<uint8_t[], deleter_type>;
 
+   private:
+    unique_ptr_type _data;
+    std::size_t _length{0};
+
+   public:
     /// @copydoc v_noabi::document::value::const_iterator
     using const_iterator = v_noabi::array::view::const_iterator;
 
@@ -89,7 +90,7 @@ class value {
     ///   A user provided deleter.
     ///
     value(std::uint8_t* data, std::size_t length, deleter_type deleter)
-        : _value{data, std::move(deleter)}, _length{length} {}
+        : _data{data, std::move(deleter)}, _length{length} {}
 
     ///
     /// Constructs a value from a std::unique_ptr to a buffer. The ownership
@@ -104,7 +105,7 @@ class value {
     /// @param length
     ///   The length of the document.
     ///
-    value(unique_ptr_type ptr, std::size_t length) : _value{std::move(ptr)}, _length{length} {}
+    value(unique_ptr_type ptr, std::size_t length) : _data{std::move(ptr)}, _length{length} {}
 
     ///
     /// Constructs a value from a view of an array. The data referenced
@@ -128,7 +129,7 @@ class value {
     /// @note `this->size()` is ignored.
     ///
     explicit operator v1::array::value() const& {
-        return _value;
+        return {_data.get(), _length, _data.get_deleter()};
     }
 
     ///
@@ -143,8 +144,9 @@ class value {
     /// @note `this->size()` is ignored.
     ///
     explicit operator v1::array::value() && {
+        auto const length = _length;
         _length = 0u;
-        return std::move(_value);
+        return {_data.release(), length, std::move(_data.get_deleter())};
     }
 
     ///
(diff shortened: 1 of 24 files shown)


src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client_session.cpp
```diff
@@ -58,15 +59,21 @@ void client_session::start_transaction(
     bsoncxx::v_noabi::stdx::optional<options::transaction> const& transaction_opts) try {
     transaction_opts ? _session.start_transaction(v_noabi::options::transaction::internal::as_v1(*transaction_opts))
                      : _session.start_transaction();
+} catch (v1::server_error const&) {
+    MONGOCXX_PRIVATE_UNREACHABLE; // Only client-side errors.
```
kevinAlbs: `mongoc_client_session_start_transaction` may contact the server with [this call](https://github.com/kevinAlbs/mongo-c-driver/blob/18154d15e891a36100b68f9fcd0b5a80cc2e91d7/src/libmongoc/src/mongoc/mongoc-client-session.c#L1073) to `mongoc_cluster_stream_for_writes`. I expect it would be possible for libmongoc to return an error with the `MONGOC_ERROR_CATEGORY_SERVER`.

That said, this may only be a short-term need. The call to `mongoc_cluster_stream_for_writes` seems unnecessary in the C driver. Filed CDRIVER-6318 to remove.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
```diff
@@ -128,7 +129,7 @@ class value {
     /// @note `this->size()` is ignored.
     ///
     explicit operator v1::array::value() const& {
-        return _value;
+        return {_data.get(), _length, _data.get_deleter()};
```
kevinAlbs: I expect the `_data.get()` results in both this object, and the resulting `v1::array::value` having a unique pointer to the same underlying data. This test appears to fail:

```c++
TEST_CASE("v_noabi to v1", "[bsoncxx][v_noabi][array][value]") {
    std::uint8_t const data[] = {5, 0, 0, 0, 0}; // {}
    auto const make_ptr = [&data] {
        auto res = std::unique_ptr<std::uint8_t[]>(new std::uint8_t[sizeof(data)]);
        std::memcpy(res.get(), data, sizeof(data));
        return res;
    };

    auto const deleter = [](std::uint8_t* p) { delete[] p; };
    std::cout << "constructing array_0" << std::endl;
    auto const array_0 = bsoncxx::v_noabi::array::value{make_ptr().release(), 5u, deleter};
    {
        std::cout << "constructing array_1" << std::endl;
        auto const array_1 = static_cast<bsoncxx::v1::array::value>(array_0);
    }
}
```

I expect this instead needs to copy the underlying data.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/document/value.hpp
```diff
@@ -168,7 +169,7 @@ class value {
     /// @note `this->size()` is ignored.
     ///
     explicit operator v1::document::value() const& {
-        return _value;
+        return {_data.get(), _length, _data.get_deleter()};
```
kevinAlbs: Similar to https://github.com/mongodb/mongo-cxx-driver/pull/1629/changes/60632fb3aae6dff6d680820421d97db00bcd5686#r3190771082, I expect this needs to copy.

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client_session.cpp
eramongodb: Nice catch. For correctness, replaced the unreachable macro with a rethrow.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/array/value.hpp
eramongodb: Good catch. Looks like existing v_noabi <-> v1 conversion test coverage missed this due to using only the `from_v1()`/`to_v1()` conversion functions, whose behavior does not exercise this copy UCDF.

Will followup with a PR to extend test coverage for UCDFs + equivalent coverage of `v_noabi::array::value`.

kevinAlbs (approved):

PR #1639, opened 2026-05-08
CXX-2720 add OIDC machine callback

kevinAlbs:
# Summary

Add support for an OIDC machine callback.

## Details

The Java driver defines an interface class [MongoCredential.OidcCallback](https://mongodb.github.io/mongo-java-driver/5.6/apidocs/driver-core/com/mongodb/MongoCredential.OidcCallback.html). However, the spec [is flexible](https://github.com/mongodb/specifications/blob/7039e69945d463a14b1b727d16db063e21f48f53/source/auth/auth.md#oidc-callback) on the callback API:
> The signature and naming of the callback API is up to the driver's discretion.

I opted to use `std::function` rather than a class. This seemed simpler and consistent with the use of `std::function` to define callbacks in `mongocxx::v1::apm`.

---

The name `v1::client::options::oidc_callback` is used for the callback setter/getter. This is the only method in `v1::client::options` without an `_opts` suffix. I thought `oidc_callback` was fitting since the argument is only the callback function.

---

To ensure a C++ exception does not propagate to C calls, a wrapper `oidc_callback_trampoline` function is used to catch a C++ exception and return `nullptr` to report an error to libmongoc.

---

libmongoc's `mongoc_oidc_callback_params_get_timeout` returns a value comparable with `bson_get_monotonic_time`. This is represented as a `std::chrono::steady_clock::time_point`.

## Testing

A subset of [OIDC prose tests](https://github.com/mongodb/specifications/blob/c250bf7a34b09519d163c1edace8e2ab179d363f/source/auth/tests/mongodb-oidc.md#machine-authentication-flow-prose-tests) are implemented:
- Tests requiring modifying the "Client Cache" are not implemented, since these are not exposed by libmongoc.
- Prose test 2.1 is partly skipped until libmongoc is upgraded to include CDRIVER-6310. Without that fix, the username is not passed to the callback.
- Tests for Azure, GCP, and Kubernetes are not planned since they require no code changes.

The OIDC test server requires auth: either with `MONGODB-OIDC` or with a test [user+pass](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4b02eb0136a767b3c6bbdff32b3584c0ec6f5b6f/.evergreen/auth_oidc/setup_oidc.js#L7-L10). Test set-up uses environment variables set by drivers-evergreen-tools `OIDC_ADMIN_USER` / `OIDC_ADMIN_PWD`. I considered a more generic `MONGOCXX_TEST_URI` to support a custom URI to all tests, but decided against to reduce the scope of changes.

The `oidc_callback` option is also added to `v_noabi` to easily permit running the unified spec test. CXX-3490 tracks migrating the unified test runner to `v1`. The `v_noabi` implementation uses `v1` so did not add much complexity. There is only one unified spec test for OIDC (mongodb-oidc-no-retry.json) copied from the latest change: https://github.com/mongodb/specifications/commit/4d1166877f85d6dccbb6260c5e372a624261fda0.

Evergreen: https://spruce.corp.mongodb.com/version/69fe2b80fb654700079fe11d. The ABI task failures appear unrelated to these changes.


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 
 - Added missing subscript operators for class types declared in `bsoncxx::vector::iterators`.
 
+### Added
+- Support for `MONGODB-OIDC` authMechanism.
+
```
eramongodb: Move up into 4.4.0 after "Deprecated" (currently in the 4.3.0 section).

src/mongocxx/include/mongocxx/v1/oidc_callback.hpp
```diff
@@ -0,0 +1,49 @@
...
+} // namespace mongocxx
+
+#include <mongocxx/v1/detail/postlude.hpp>
+
+///
+/// @file
+/// Provides @ref mongocxx::v1::oidc_callback.
+///
+/// @ref mongocxx::v1::oidc_callback is a type alias for a std::function that receives
+/// @ref mongocxx::v1::oidc_callback_params and returns @ref mongocxx::v1::oidc_credential.
```
eramongodb:
```suggestion
```
Redundant details given Doxygen doc page layout:

Image

Note the alias definition is visible and every referenced mongocxx type in its definition is linked-to.

src/mongocxx/include/mongocxx/v1/oidc_callback.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace v1 {
+
+///
+/// An OIDC credential callback function.
+///
+/// The function receives callback parameters and returns an OIDC credential.
+/// If the function throws an exception, the exception is caught, an error is logged, and NULL is returned to libmongoc
+/// to signal failure.
+///
+using oidc_callback = std::function<oidc_credential MONGOCXX_ABI_CDECL(oidc_callback_params const&)>;
```
eramongodb, line 36:
```suggestion
using oidc_credential_cb = std::function<oidc_credential MONGOCXX_ABI_CDECL(oidc_callback_params const&)>;
```
Bikeshed: for better consistency with other (non-APM) mongocxx callback function typenames?

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1377,4 +1427,8 @@ TEST_CASE("client side encryption unified format spec automated tests", "[unifie
     run_unified_format_tests_in_env_dir("CLIENT_SIDE_ENCRYPTION_UNIFIED_TESTS_PATH");
 }
 
+TEST_CASE("auth spec automated tests", "[unified_format_specs],[oidc]") {
```
eramongodb:
```suggestion
TEST_CASE("auth spec automated tests", "[unified_format_specs]") {
```
Suggest avoiding OIDC-specific tag for general "auth" tests. Note Catch2 supports filtering by test case name and sections, e.g.:

```bash
test_unified_format_specs "auth*" -c "mongodb-oidc-no-retry.json"
```

See also: https://github.com/mongodb/mongo-cxx-driver/pull/1639#discussion_r3221391634

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -340,15 +361,31 @@ std::string uri_options_to_string(document::view object) {
     //  'readPreferenceTags' keys in the object.
     REQUIRE_FALSE(object["readPreferenceTags"]);
 
-    auto const json = to_json(object["uriOptions"].get_document());
+    auto uri_options_doc = bsoncxx::builder::basic::document{};
+    for (auto const& el : object["uriOptions"].get_document().value) {
+        auto const key = el.key();
+        auto const value = el.get_value();
+
+        if (key == "authMechanismProperties" && value.type() == bsoncxx::type::k_document &&
```
eramongodb:
```suggestion
        if (key == MONGOC_URI_AUTHMECHANISMPROPERTIES && value.type() == bsoncxx::type::k_document &&
```
Use mongoc URI macro constants? Also noting that `==` is not case-insensitive.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -340,15 +361,31 @@ std::string uri_options_to_string(document::view object) {
...
+            value.get_document().value == make_document(kvp("$$placeholder", 1))) {
+            REQUIRE(object["uriOptions"]["authMechanism"].type() == bsoncxx::type::k_string);
+            REQUIRE(object["uriOptions"]["authMechanism"].get_string().value == "MONGODB-OIDC");
```
eramongodb:
```suggestion
            auto const auth_mechanism = object["uriOptions"]["authMechanism"];
            REQUIRE(auth_mechanism);
            REQUIRE(auth_mechanism.type() == bsoncxx::type::k_string);
            REQUIRE(auth_mechanism.get_string().value == "MONGODB-OIDC");
```
Avoid repeated lookup.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -676,6 +713,16 @@ client create_client(document::view object) {
         auto const server_api_opts = create_server_api(object);
         client_opts.server_api_opts(server_api_opts);
     }
+    if (object["uriOptions"]["authMechanism"] &&
+        object["uriOptions"]["authMechanism"].type() == bsoncxx::type::k_string &&
+        object["uriOptions"]["authMechanism"].get_string().value == "MONGODB-OIDC") {
```
eramongodb:
```suggestion
    {
        auto const auth_mechanism = object["uriOptions"]["authMechanism"];
        if (auth_mechanism && auth_mechanism.type() == bsoncxx::type::k_string &&
            auth_mechanism.get_string().value == "MONGODB-OIDC") {
```
Avoid repeated lookups.

src/mongocxx/test/v1/oidc_credential.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+TEST_CASE("getters", "[mongocxx][v1][oidc_credential]") {
+    oidc_credential without_expiration{"foo"};
+    REQUIRE(without_expiration.access_token() == "foo");
+    REQUIRE(!without_expiration.expires_in());
+
+    oidc_credential with_expiration{"bar", std::chrono::milliseconds{123}};
+    REQUIRE(with_expiration.access_token() == "bar");
+    REQUIRE(with_expiration.expires_in());
+    REQUIRE(with_expiration.expires_in()->count() == 123);
```
eramongodb:
```suggestion
    SECTION("without expiration") {
        oidc_credential creds{"foo"};
        CHECK(creds.access_token() == "foo");
        CHECK(!creds.expires_in());
    }

    SECTION("with expiration") {
        oidc_credential creds{"bar", std::chrono::milliseconds{123}};
        CHECK(creds.access_token() == "bar");
        CHECK(creds.expires_in());
        CHECK(creds.expires_in()->count() == 123);
    }
```
Improve independance of test assertions.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+void admin_command(std::string cmd) {
+    oidc_callback cb = [&](oidc_callback_params const&) { return oidc_credential(read_token_from_file()); };
+
+    auto opts = v1::client::options();
+    opts.oidc_callback(cb);
+
+    auto client =
+        mongocxx::v1::client(v1::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC"), opts);
+    client.database("admin").run_command(scoped_bson(cmd).view());
+}
```
eramongodb:
```suggestion
namespace {

// Read token from /tmp/tokens/test_machine
std::string read_token_from_file() {
    std::ifstream token_file("/tmp/tokens/test_machine");
    REQUIRE(token_file.is_open());
    return std::string((std::istreambuf_iterator<char>(token_file)), std::istreambuf_iterator<char>());
}

void admin_command(std::string cmd) {
    oidc_callback cb = [&](oidc_callback_params const&) { return oidc_credential(read_token_from_file()); };

    auto opts = v1::client::options();
    opts.oidc_callback(cb);

    auto client =
        mongocxx::v1::client(v1::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC"), opts);
    client.database("admin").run_command(scoped_bson(cmd).view());
}

} // namespace
```

Internal helper functions (wrap in unnamed namespace).

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+TEST_CASE("OIDC prose tests", "[oidc]") {
+    if (nullptr == std::getenv("MONGOCXX_TEST_OIDC_AUTH_URI")) {
+        SKIP("Set MONGOCXX_TEST_OIDC_AUTH_URI to run OIDC tests");
+    }
```
eramongodb: There's a check for `MONGOCXX_TEST_OIDC_AUTH_URI`, but it does not seem to be subsequently used by any of the tests below (hardcoded `localhost:27017`). Is this deliberate?

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+   private:
+    bool _is_pooled;
+    bsoncxx::v1::stdx::optional<v1::pool> _pool;
+    bsoncxx::v1::stdx::optional<v1::pool::entry> _pool_entry;
+    bsoncxx::v1::stdx::optional<v1::client> _client;
+};
+
+// Read token from /tmp/tokens/test_machine
+std::string read_token_from_file() {
+    std::ifstream token_file("/tmp/tokens/test_machine");
```
eramongodb:
```suggestion
    std::ifstream token_file(/* OIDC_TOKEN_DIR */);
```
Should this use `OIDC_TOKEN_DIR` instead? Per [test spec](https://github.com/mongodb/specifications/blob/c250bf7a34b09519d163c1edace8e2ab179d363f/source/auth/tests/mongodb-oidc.md#machine-authentication-flow-prose-tests):

> The token file ENVIRONMENT:test is located in OIDC_TOKEN_DIR set by [drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/auth_oidc/README.md) scripts.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -676,6 +713,16 @@ client create_client(document::view object) {
...
+        client_opts.oidc_callback([](mongocxx::v1::oidc_callback_params const&) {
+            std::ifstream token_file("/tmp/tokens/test_machine");
```
eramongodb:
```suggestion
            std::ifstream token_file(/* OIDC_TOKEN_DIR */);
```
Should this use `OIDC_TOKEN_DIR` instead? Per [test spec](https://github.com/mongodb/specifications/blob/c250bf7a34b09519d163c1edace8e2ab179d363f/source/auth/tests/mongodb-oidc.md#machine-authentication-flow-prose-tests):

> The token file ENVIRONMENT:test is located in OIDC_TOKEN_DIR set by [drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/auth_oidc/README.md) scripts.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+            // Expect timeout is roughly 60 seconds:
+            {
+                REQUIRE(params.timeout());
+                auto timeout = params.timeout().value();
+                auto const expected_duration = std::chrono::seconds(60);
+                auto const epsilon = std::chrono::seconds(10);
+                auto low = std::chrono::steady_clock::now() + expected_duration - epsilon;
+                auto high = std::chrono::steady_clock::now() + expected_duration + epsilon;
+                REQUIRE(timeout >= low);
+                REQUIRE(timeout <= high);
```
eramongodb:
```suggestion
                CHECK(timeout >= low);
                CHECK(timeout <= high);
```
General Catch2 test assertion advice: prefer using `CHECK()` for independent assertions to reveal as much info as possible on failure, reserving `REQUIRE()` for assertions where "it is invalid for test to proceed past this check" (enforcing preconditions and invariants).

Advice applies to other instances of `REQUIRE()` in new Catch2 tests.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+            }
+
+            REQUIRE(params.version() == 1);
+
+            if (with_username) {
+                REQUIRE(params.username().has_value());
+                REQUIRE(params.username().value() == "user");
```
eramongodb:
```suggestion
                CHECK(params.username() == "user");
```
Optional `==` supports value-comparison against null values (always false). Recommend using this form so maximal info is revealed by Catch2 expression decomposition.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+        auto opts = v1::client::options();
+        opts.oidc_callback(cb);
+
+        // Spec: "Create an OIDC configured client"
+        bool is_pooled = GENERATE(true, false);
+
+        mongocxx::v1::uri uri =
+            (with_username) ? v1::uri("mongodb://user@localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC")
+                            : v1::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC");
```
eramongodb:
```suggestion
        v1::uri uri = (with_username)
                          ? v1::uri("mongodb://user@localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC")
                          : v1::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC");
```
Qualification consistency.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+        oidc_callback cb = [&](oidc_callback_params const&) {
+            callback_call_count++;
+            return oidc_credential(read_token_from_file());
+        };
+
+        auto opts = v1::client::options();
+        opts.oidc_callback(cb);
+
+        // Spec: "Create an OIDC configured client"
+        bool is_pooled = GENERATE(true, false);
```
eramongodb:
```suggestion
        bool const is_pooled = GENERATE(true, false);
        CAPTURE(is_pooled);
```
For generated test values which are not explicitly used in an assertion (e.g. `CHECK(... == is_pooled);`), capture its value so test assertion failure output includes the state of the generated value.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/client.hpp
```diff
@@ -173,13 +173,38 @@ class client {
...
+    /// Sets the OIDC callback.
+    ///
+    /// @param oidc_callback
+    ///   The OIDC callback.
+    ///
+    /// @return
+    ///   A reference to the object on which this member function is being called. This facilitates
+    ///   method chaining.
+    ///
+    client& oidc_callback(v1::oidc_callback oidc_callback) {
```
eramongodb, line 187: Missing include of `<mongocxx/v1/oidc_callback.hpp>`.

src/mongocxx/include/mongocxx/v1/oidc_credential.hpp
```diff
@@ -0,0 +1,107 @@
...
+    MONGOCXX_ABI_EXPORT_CDECL(oidc_credential&) operator=(oidc_credential const& other);
+
+    ///
+    /// Construct a credential with an access token.
+    ///
+    explicit MONGOCXX_ABI_EXPORT_CDECL() oidc_credential(std::string access_token);
+
+    ///
+    /// Construct a credential with an access token and expiry duration.
+    /// The expiry duration will be evaluated relative to the system clock immediately after the callback returns.
```
eramongodb, line 81:
```suggestion
    /// Construct a credential with an access token and expiry duration.
    ///
    /// The expiry duration will be evaluated relative to the system clock immediately after the callback returns.
```
Split paragraphs so "brief" and "details" are distinct.

src/mongocxx/include/mongocxx/v1/oidc_credential-fwd.hpp
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/v1/detail/prelude.hpp>
+
+namespace mongocxx {
+namespace v1 {
+
+class oidc_credential;
```
eramongodb, line 22: Also add new `v_noabi` headers redeclaring these new v1 API in the root namespace (see post-https://github.com/mongodb/mongo-cxx-driver/pull/1640 `v1::text_options` components as reference):

```cpp
// mongocxx/oidc_credentials-fwd.hpp

#include <mongocxx/v1/oidc_credentials-fwd.hpp> // IWYU pragma: export

namespace mongocxx {

using v1::oidc_credentials;

} // namespace mongocxx
```


```cpp
// mongocxx/oidc_credentials.hpp

#include <mongocxx/oidc_credentials-fwd.hpp> // IWYU pragma: export

//

#include <mongocxx/v1/oidc_credentials.hpp> // IWYU pragma: export
```

src/mongocxx/test/v_noabi/oidc_callback.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+   private:
+    bool _is_pooled;
+    bsoncxx::v_noabi::stdx::optional<v_noabi::pool> _pool;
+    bsoncxx::v_noabi::stdx::optional<v_noabi::pool::entry> _pool_entry;
+    bsoncxx::v_noabi::stdx::optional<v_noabi::client> _client;
+};
+
+TEST_CASE("OIDC (v_noabi)", "[oidc]") {
```
eramongodb:
```suggestion
TEST_CASE("v_noabi", "[mongocxx][v_noabi][oidc_callback]") {
```
Recommend using Catch2 tags for namespace/class scoping.

src/mongocxx/test/v_noabi/oidc_callback.cpp
```diff
@@ -0,0 +1,117 @@
...
+        auto opts = v_noabi::options::client();
+        opts.oidc_callback(cb);
+
+        // Create an OIDC configured client:
+        bool is_pooled = GENERATE(true, false);
+        OIDCTestFixture tf(
+            v_noabi::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC"), opts, is_pooled);
+
+        /// Expect auth to succeed:
+        tf.client().database("test").run_command(bsoncxx::v_noabi::from_v1(scoped_bson{R"({"ping": 1})"}.view()));
```
eramongodb:
```suggestion
                CHECK_NOTHROW(
            tf.client().database("test").run_command(bsoncxx::v_noabi::from_v1(scoped_bson{R"({"ping": 1})"}.view())));
```
Use `CHECK_NOTHROW()` for explicit "expected to _potentially_ fail, but should not" expressions, otherwise the source of the exception is vaguely reported as "within this (sub)section" only.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+#include <mongocxx/v1/client.hpp>
+#include <mongocxx/v1/collection.hpp>
+#include <mongocxx/v1/database.hpp>
+#include <mongocxx/v1/insert_one_result.hpp>
+#include <mongocxx/v1/oidc_callback_params.hpp>
+#include <mongocxx/v1/oidc_credential.hpp>
+#include <mongocxx/v1/pool.hpp>
+#include <mongocxx/v1/uri.hpp>
+
+#include <mongocxx/test/private/scoped_bson.hh>
```
eramongodb, line 30:
```suggestion
#include <mongocxx/v1/uri.hpp>

#include <bsoncxx/test/v1/stdx/optional.hh>

#include <mongocxx/test/private/scoped_bson.hh>
```
For Catch2 stringification of `optional<T>`.

src/mongocxx/test/v1/oidc_credential.cpp
```diff
@@ -0,0 +1,72 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <mongocxx/v1/oidc_credential.hpp>
+
+//
+
+#include <chrono>
```
eramongodb, line 21:
```suggestion
//

#include <bsoncxx/test/v1/stdx/optional.hh>

#include <chrono>
```
For Catch2 stringification of `optional<T>`.

src/mongocxx/test/v_noabi/oidc_callback.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (nullptr == std::getenv("MONGOCXX_TEST_OIDC_AUTH_URI")) {
+        SKIP("Set MONGOCXX_TEST_OIDC_AUTH_URI to run OIDC tests");
+    }
+
+    SECTION("Works") {
+        auto callback_call_count = 0u;
+        v1::oidc_callback cb = [&](v1::oidc_callback_params const&) {
+            callback_call_count++;
+            std::ifstream token_file("/tmp/tokens/test_machine");
+            REQUIRE(token_file.is_open());
```
eramongodb: Dropping a reminder that `REQUIRE()` throws an exception: assertion failure here (and in all other calls to `REQUIRE()` within an OIDC callback function) will trigger the "exception is caught, an error is logged, and NULL is returned to libmongoc" code path.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+        // Spec: "Perform a `insert` operation that succeeds"
+        tf.client().database("test").collection("test").insert_one(scoped_bson{}.view());
+
+        // Spec: "Set a fail point for `insert` commands"
+        admin_command(R"({
+            "configureFailPoint": "failCommand",
+            "mode": { "times": 1 },
+            "data": {
+                "failCommands": [ "find" ],
```
eramongodb:
```suggestion
                "failCommands": [ "insert" ],
```
Should this be "insert" instead of "find"?

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,524 @@
...
+                "errorCode": 391
+            }
+        })");
+
+        // Spec: "Perform a `insert` operation that fails."
+        CHECK_THROWS_WITH(
+            tf.client().database("test").collection("test").find_one(scoped_bson{}.view()),
```
eramongodb, line 496:
```suggestion
            tf.client().database("test").collection("test").insert_one(scoped_bson{}.view()),
```
Should this be `.insert_one()` instead of `.find_one()`?

src/mongocxx/lib/mongocxx/v1/oidc_callback_params.hh
```diff
@@ -0,0 +1,32 @@
...
+//
+
+#include <mongocxx/private/mongoc.hh>
+
+namespace mongocxx {
+namespace v1 {
+
+class oidc_callback_params::internal {
+   public:
+    static oidc_callback_params make(mongoc_oidc_callback_params_t* ptr);
```
eramongodb:
```suggestion
    static oidc_callback_params make(mongoc_oidc_callback_params_t const* ptr);
```
Missing inner const for consistency with ctor and data member.

.evergreen/config_generator/components/oidc.py
```diff
@@ -0,0 +1,92 @@
...
+
+class TestOIDC(Function):
+    name = 'test-oidc'
+    commands = bash_exec(
+        command_type=EvgCommandType.TEST,
+        working_dir='mongo-cxx-driver',
+        script="""\
+            export LD_LIBRARY_PATH=$(pwd)/../mongoc/lib
+            export MONGOCXX_TEST_OIDC_AUTH_URI="mongodb://bob:pwd123@localhost:27017"
+            ./build/src/mongocxx/test/test_driver '[oidc]'
```
eramongodb: This only tests prose tests. Should there also be a call to `test_unified_format_specs 'auth spec automated tests'`?

.evergreen/config_generator/components/oidc.py
```diff
@@ -0,0 +1,92 @@
...
+        TestOIDC.call(),
+    ]
+
+    return [
+        EvgTask(
+            name="oidc-auth-test-task",
+            run_on=[find_small_distro("ubuntu2404").name],
+            commands=commands,
+        )
+    ]
```
eramongodb:
```suggestion
    return [
        EvgTask(
            name='oidc-auth-test-task',
            run_on=[find_large_distro('ubuntu2404').name],
            commands=[
                expansions_update(updates=[KeyValueParam(key='build_type', value='Debug')]),
                InstallCDriver.call(vars={'SKIP_INSTALL_LIBMONGOCRYPT': 1}),
                Compile.call(vars={'ENABLE_TESTS': 'ON'}),
                TestOIDC.call(),
            ],
        )
    ]
```
Task definition is simple enough that this should probably be all inlined + use "large" distro due to the compile step.

eramongodb (changes requested):

CHANGELOG.md
kevinAlbs: Good spot. Fixed.

src/mongocxx/include/mongocxx/v1/oidc_callback.hpp
kevinAlbs: Removed as suggested.

src/mongocxx/include/mongocxx/v1/oidc_callback.hpp
kevinAlbs: Kept as-is for now, but I am open to renaming.

The only other non-APM use of `std::function` appears to be `with_transaction_cb`. The corresponding setter drops the "_cb":
```c++
MONGOCXX_ABI_EXPORT_CDECL(void) with_transaction(with_transaction_cb cb, v_noabi::options::transaction opts = {});
```

Following that pattern would name the setter `oidc_credential`. That seems possibly more confusing (accepts a callback rather than a `oidc_credential` object).

src/mongocxx/test/spec/unified_tests/runner.cpp
kevinAlbs: Updated as suggested.

src/mongocxx/test/spec/unified_tests/runner.cpp
kevinAlbs: Updated and added a `to_lowercase` helper since `MONGOC_URI_AUTHMECHANISMPROPERTIES` is all lowercase.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: The OIDC test server requires auth: either with `MONGODB-OIDC` or with a [user+pass](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4b02eb0136a767b3c6bbdff32b3584c0ec6f5b6f/.evergreen/auth_oidc/setup_oidc.js#L7-L10). `MONGOCXX_TEST_OIDC_AUTH_URI` included the user+pass. Though this file does not connect with the URI, the presence of the environment variable was used to decide the test server supported OIDC.

With the latest changes, this is updated to check for the presence of the `OIDC_TOKEN_FILE`, set by drivers-evergreen-tools.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Updated to use value set by `drivers-evergreen-tools`, and similarly applied to test credentials.

.evergreen/config_generator/components/oidc.py
kevinAlbs: Great catch. Updated, and added a `set -o errexit` to so either test runner failing results in task failure.

src/mongocxx/test/spec/unified_tests/runner.cpp
kevinAlbs: Good idea. Updated to use `OIDC_TOKEN_FILE` environment variable set by drivers-evergreen-tools scripts. That might help avoid a code change if the test setup later changes.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Good idea. Updated many `REQUIRE` assertions to `CHECK` in this PR.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Updated here and elsewhere in this file and in `test/v_noabi/oidc_callback.cpp`.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Good idea. Updated here and elsewhere.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/client.hpp
kevinAlbs: Good catch. Updated.

src/mongocxx/include/mongocxx/v1/oidc_credential-fwd.hpp
kevinAlbs: Great catch. Updated to add root namespace redeclaration headers. Updated `test/v_noabi/oidc_callback.cpp` to use the root namespace redeclarations rather than `v1`.

src/mongocxx/test/v_noabi/oidc_callback.cpp
kevinAlbs: Updated as suggested, and updated the OIDC Evergreen task to add the `[oidc_callback]` tag.

src/mongocxx/test/v_noabi/oidc_callback.cpp
kevinAlbs: Thank you for the suggestion. I think that significantly helps identify the cause of failure. Forcing an exception with a bad command name (`ping2`) shows the before / after:

Before:
```
/Users/kevin.albertson/code/tasks/mongo-cxx-driver-CXX-2590/src/mongocxx/test/v_noabi/oidc_callback.cpp:87: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with messages:
  is_pooled := true
  no such command: 'ping2': generic server error
```

After (with `CHECK_NOTHROW`):
```
/Users/kevin.albertson/code/tasks/mongo-cxx-driver-CXX-2590/src/mongocxx/test/v_noabi/oidc_callback.cpp:94: FAILED:
  CHECK_NOTHROW( tf.client().database("test2").run_command( bsoncxx::v_noabi::from_v1(scoped_bson{R"({"ping2": 1})"}.view())) )
due to unexpected exception with messages:
  is_pooled := false
  no such command: 'ping2': generic server error
```

`CHECK_NOTHROW` was added to other calls.

Note: I left the call in the thread of prose test 1.2 as-is. The exception message appeared garbled. I expect this is related to needing to opt-in to Catch2 [thread-safe assertions](https://github.com/catchorg/Catch2/blob/devel/docs/configuration.md#thread-safety-in-assertions-and-messages). Added a comment.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Oop. Prose test 4.2 is for "find". This operation was correct, but a comment and operation above referencing "insert" were wrong. Updated comment.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Oop. Prose test 4.2 is for "find". This operation was correct, but a comment and operation above referencing "insert" were wrong. Updated comment.

src/mongocxx/test/v_noabi/oidc_callback.cpp
kevinAlbs: Testing with this assert flipped includes the failing line:
```
[...]/src/mongocxx/test/v1/oidc_callback.cpp:82: FAILED:
  REQUIRE( !token_file.is_open() )
with expansion:
  false
with message:
  /tmp/tokens/test_machine

libc++abi: terminating due to uncaught exception of type mongocxx::v1::exception: MONGODB-OIDC callback failed: mongoc_error_code_t:11
[...]/src/mongocxx/test/v1/oidc_callback.cpp:82: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  /tmp/tokens/test_machine
  SIGABRT - Abort (abnormal termination) signal
```

Kept as-is since this the cause of error still seems easy-to-identify.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 
 ## 4.4.0 [Unreleased]
 
+### Added
+- Support for `MONGODB-OIDC` authMechanism.
```
eramongodb: Suggest merging with the existing "Added" section below, after "Deprecated".

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -349,6 +344,11 @@ std::string json_to_uri_opts(std::string const& input) {
     return std::accumulate(std::begin(output) + 1, std::end(output), output[0], join);
 }
 
+std::string to_lowercase(std::string in) {
+    std::transform(std::begin(in), std::end(in), std::begin(in), ::tolower);
+    return in;
+}
```
eramongodb: Suggest reusing the existing `tolowercase()` helper provided by `<src/mongocxx/test/v_noabi/client_helpers.hh>`.

.evergreen/config_generator/components/oidc.py
```diff
@@ -0,0 +1,92 @@
...
+        working_dir='mongo-cxx-driver',
+        script="""\
+            set -o errexit
+            export LD_LIBRARY_PATH="$(pwd)/../mongoc/lib"
+
+            # Sets environment variables: OIDC_TOKEN_FILE, OIDC_ADMIN_USER, and OIDC_ADMIN_PWD.
+            source "${DRIVERS_TOOLS}/.evergreen/auth_oidc/secrets-export.sh"
+            ./build/src/mongocxx/test/test_driver "[oidc],[oidc_callback]"
+            export AUTH_TESTS_PATH="$(pwd)/data/auth/unified"
+            ./build/src/mongocxx/test/test_unified_format_specs "auth*" -c "mongodb-oidc-no-retry.json"
```
eramongodb:
```suggestion
            ./build/src/mongocxx/test/test_unified_format_specs "auth*" -p "c:mongodb-oidc-no-retry.json"
```
Following https://github.com/mongodb/mongo-cxx-driver/pull/1642, suggest using new [path-based section filter flags](https://github.com/catchorg/Catch2/blob/v3.15.0/docs/filtering-execution-path.md).

.evergreen/config_generator/components/oidc.py
eramongodb, line 1: This file should be formatted with `uv run --frozen etc/ruff-format-all.sh`.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,541 @@
...
+    if (nullptr == std::getenv("OIDC_TOKEN_FILE")) {
+        SKIP("Set OIDC_TOKEN_FILE to run OIDC tests");
+    }
+
+    SECTION("1.1 Callback is called during authentication") {
+        auto callback_call_count = 0u;
+        oidc_callback cb = [&](oidc_callback_params const&) {
+            callback_call_count++;
+            return oidc_credential(read_token_from_file());
+        };
```
eramongodb, line 177:
```suggestion
        auto const token = read_token_from_file();
        auto callback_call_count = 0u;
        oidc_callback cb = [&](oidc_callback_params const&) {
            callback_call_count++;
            return oidc_credential(token);
        };
```
In an effort to avoid test failure exceptions from being absorbed by the callback exception handler, recommend relocating initialization of input values out of the lambda expression.

Before relocation (e.g. given a non-existent `OIDC_TOKEN_FILE` + with Catch2's improved nested assertion diagnostics):

```
OIDC prose tests
  1.1 Callback is called during authentication
src/mongocxx/test/v1/oidc_callback.cpp:101

src/mongocxx/test/v1/oidc_callback.cpp:81: FAILED:
  REQUIRE( token_file.is_open() )
with expansion:
  false
with message:
  is_pooled := false

src/mongocxx/test/v1/oidc_callback.cpp:118: FAILED:
  CHECK_NOTHROW( tf.client().database("test").collection("test").find_one(scoped_bson{}.view()) )
due to unexpected exception with messages:
  is_pooled := false
  MONGODB-OIDC callback failed: mongoc_error_code_t:11

===============================================================================
test cases: 1 | 1 failed
assertions: 4 | 2 passed | 2 failed
```

After relocation:

```
OIDC prose tests
  1.1 Callback is called during authentication
src/mongocxx/test/v1/oidc_callback.cpp:101

src/mongocxx/test/v1/oidc_callback.cpp:81: FAILED:
  REQUIRE( token_file.is_open() )
with expansion:
  false

===============================================================================
test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed
```

Suggestion applies to other instances of OIDC callbacks used in test code (e.g. this should also help avoid [termination on test failure](https://github.com/mongodb/mongo-cxx-driver/pull/1639#discussion_r3227629513)).

Note: given Catch2 requires C++14 (thus all test components are also built with C++14), init-captures are also an option:

```cpp
oidc_callback cb = [&, token = read_token_from_file()](oidc_callback_params const&) {
    callback_call_count++;
    return oidc_credential(token);
};
```

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,541 @@
...
+}
+
+void admin_command(std::string cmd) {
+    oidc_callback cb = [&](oidc_callback_params const&) { return oidc_credential(read_token_from_file()); };
+
+    auto opts = v1::client::options();
+    opts.oidc_callback(cb);
+
+    auto client = v1::client(v1::uri("mongodb://localhost:27017/?retryReads=false&authMechanism=MONGODB-OIDC"), opts);
+    client.database("admin").run_command(scoped_bson(cmd).view());
```
eramongodb:
```suggestion
    v1::client{}.database("admin").run_command(scoped_bson(cmd).view());
```
Is it correct for admin commands (primarily used to set failpoints) to use OIDC authentication? I believe the relevant OIDC creds are for `OIDC_ADMIN_USER`, not for `admin`.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,541 @@
...
+            // Expect timeout is roughly 60 seconds:
+            {
+                REQUIRE(params.timeout());
+                auto timeout = params.timeout().value();
+                auto const expected_duration = std::chrono::seconds(60);
+                auto const epsilon = std::chrono::seconds(10);
+                auto low = std::chrono::steady_clock::now() + expected_duration - epsilon;
+                auto high = std::chrono::steady_clock::now() + expected_duration + epsilon;
+                CHECK(timeout >= low);
+                CHECK(timeout <= high);
```
eramongodb:
```suggestion
                auto const timeout = params.timeout();
                auto const now = std::chrono::steady_clock::now();

                REQUIRE(params.timeout());

                auto const expected_duration = std::chrono::seconds(60);
                auto const epsilon = std::chrono::seconds(10);

                auto const low = now + expected_duration - epsilon;
                auto const high = now + expected_duration + epsilon;

                CHECK(timeout.value() >= low);
                CHECK(timeout.value() <= high);
```
Reduce repeated calls to `now()` (explicitly and internally) + reduce potential discrepancy between the time points used to assert expected durations.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,541 @@
...
+            thread.join();
+        }
+
+        // Spec: "Assert that the callback was called 1 time"
+        CHECK(callback_call_count == 1u);
+    }
+
+    SECTION("2.1 Valid Callback Inputs") {
+        auto callback_call_count = 0u;
+        bool with_username = GENERATE(/* true, */ false); // TODO CDRIVER-6310: test 'true' once upgraded to 2.3.1.
```
eramongodb, line 243:
```suggestion
        bool with_username = GENERATE(/* true, */ false); // TODO CDRIVER-6310: test 'true' once upgraded to 2.3.1.
        CAPTURE(with_username);
```
Capture generated value. Alternatively, use `CHECKED_IF(with_username)` below.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/oidc_credential.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <mongocxx/oidc_credential-fwd.hpp> // IWYU pragma: export
+
+//
+
+#include <mongocxx/v1/oidc_credential.hpp> // IWYU pragma: export
+
+///
+/// @file
+/// Declares @ref mongocxx::oidc_credential.
```
eramongodb:
```suggestion
/// Provides @ref mongocxx::oidc_credential.
```
Provides the class definition (not just declaration).

src/mongocxx/include/mongocxx/v_noabi/mongocxx/oidc_callback_params.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <mongocxx/oidc_callback_params-fwd.hpp> // IWYU pragma: export
+
+//
+
+#include <mongocxx/v1/oidc_callback_params.hpp> // IWYU pragma: export
+
+///
+/// @file
+/// Declares @ref mongocxx::oidc_callback_params.
```
eramongodb:
```suggestion
/// Provides @ref mongocxx::oidc_callback_params.
```
Provides the class definition (not just declaration).

src/mongocxx/test/v1/oidc_credential.cpp
```diff
@@ -0,0 +1,77 @@
...
+        target = copy;
+
+        CHECK(copy.access_token() == "source_token");
+        CHECK(target.access_token() == "source_token");
+    }
+}
+
+TEST_CASE("getters", "[mongocxx][v1][oidc_credential]") {
+    SECTION("without expiration") {
+        oidc_credential creds{"foo"};
```
eramongodb:
```suggestion
        oidc_credential const creds{"foo"};
```
Ensure getters are const-qualified.

src/mongocxx/test/v1/oidc_credential.cpp
```diff
@@ -0,0 +1,77 @@
...
+        CHECK(creds.access_token() == "foo");
+        CHECK(!creds.expires_in());
+    }
+    SECTION("with expiration") {
+        oidc_credential creds{"bar", std::chrono::milliseconds{123}};
```
eramongodb:
```suggestion
        oidc_credential const creds{"bar", std::chrono::milliseconds{123}};
```
Ensure getters are const-qualified.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -340,15 +361,39 @@ std::string uri_options_to_string(document::view object) {
...
 
     CAPTURE(json, opts);
     return opts;
 }
 
 std::string get_hostnames(bsoncxx::document::view object) {
-    auto const uri0 = mongocxx::uri("mongodb://localhost:27017");
+    auto uri0 = mongocxx::uri{"mongodb://localhost:27017"};
+    if (auto const* oidc_user = std::getenv("OIDC_ADMIN_USER")) {
+        auto const* oidc_pwd = std::getenv("OIDC_ADMIN_PWD");
```
eramongodb, line 388:
```suggestion
        auto const* oidc_pwd = std::getenv("OIDC_ADMIN_PWD");
        REQUIRE(oidc_pwd);
```
Enforce both env vars are provided together.

src/mongocxx/test/v_noabi/client_helpers.cpp
```diff
@@ -54,26 +54,36 @@ using bsoncxx::builder::basic::kvp;
 using bsoncxx::builder::basic::make_document;
 
 namespace {
+
+mongocxx::uri test_uri() {
+    if (auto const* oidc_user = std::getenv("OIDC_ADMIN_USER")) {
+        auto const* oidc_pwd = std::getenv("OIDC_ADMIN_PWD");
```
eramongodb, line 60:
```suggestion
        auto const* oidc_pwd = std::getenv("OIDC_ADMIN_PWD");
        REQUIRE(oidc_pwd);
```
Enforce both env vars are provided together.

src/mongocxx/test/v_noabi/client_helpers.cpp
```diff
@@ -590,7 +600,7 @@ cseeos_result client_side_encryption_enabled_or_skip_impl() {
 
 #endif // defined(MONGOC_ENABLE_CLIENT_SIDE_ENCRYPTION)
 
-std::string getenv_or_fail(const std::string env_name) {
+std::string getenv_or_fail(std::string const env_name) {
```
eramongodb:
```suggestion
std::string getenv_or_fail(std::string const& env_name) {
```
Should also update the declaration in src/mongocxx/test/v_noabi/client_helpers.hh. (Unclear why these files are not being formatted by etc/format.py. 🤔)

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,541 @@
...
+            pool_entries.push_back(tf.pool().acquire());
+        }
+
+        // Start threads:
+        for (auto i = 0u; i < num_threads; i++) {
+            auto& client = *pool_entries.at(i);
+            threads.emplace_back([&client]() {
+                auto const finds_per_thread = 100u;
+                for (auto i = 0u; i < finds_per_thread; i++) {
+                    // Omit CHECK_NOTHROW. Thread-safe assertions requires Catch2 opt-in.
```
eramongodb, line 226: FYI following https://github.com/mongodb/mongo-cxx-driver/pull/1642 (Catch2 upgraded to 3.15.0), we now have the option to enable thread-safe assertions.

```
--- a/cmake/FetchCatch2.cmake
+++ b/cmake/FetchCatch2.cmake
@@ -41,6 +41,7 @@ function(fetch_catch2)
         # Catch2 config vars.
         set_property(CACHE CATCH_INSTALL_DOCS PROPERTY VALUE OFF)
         set_property(CACHE CATCH_INSTALL_EXTRAS PROPERTY VALUE OFF)
+        set_property(CACHE CATCH_CONFIG_THREAD_SAFE_ASSERTIONS PROPERTY VALUE ON)
 
         message (STATUS "Downloading Catch2... done.")
     endif()
```

eramongodb (changes requested):

CHANGELOG.md
kevinAlbs: Gah, updated.

src/mongocxx/test/v1/oidc_callback.cpp
eramongodb: Note: using `test_bson` as a benchmark (1.2M assertions, measured across 10 samples), without thread-safe assertions (in seconds):

```
Mean:   0.461 (stddev: 0.007)
Median: 0.462 (min: 0.450, max: 0.470)
```

With thread-safe assertions:

```
Mean:   0.501 (stddev: 0.013)          (+8.6%)
Median: 0.491 (min: 0.491, max: 0.536) (+6.3%)
```

Looks like there is indeed a measurable cost to enabling thread-safe assertions, but I think it may be an acceptable tradeoff.

src/mongocxx/test/spec/unified_tests/runner.cpp
kevinAlbs: Oh nice. Updated.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Good idea. Updated as suggested. I opted against the C++14 init-captures (for no good reason other than less familiarity).

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: The OIDC credentials do appear to have permission to set failpoints, but I updated to use the `OIDC_ADMIN_USER` to better separate the intended tests of the `MONGODB-OIDC` auth mechanism.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: I am on the fence, but leaning more towards leaving out thread-safe assertions. [Catch2 notes](https://github.com/catchorg/Catch2/blob/6ee0826dcae55ed1e06b2c5701981221e979e1e6/docs/configuration.md#thread-safety-in-assertions-and-messages):

> Due to the performance cost this imposes even on single-threaded usage, Catch2 defaults to non-thread-safe assertions.

Even if this is only appears as an extra half-second in `test_bson`, IMO the test slowdown is not worth the improved error reporting in one test case. This appears to be the only test that makes assertions in threads.

kevinAlbs (commented):
Review of the [prose tests](https://github.com/mongodb/specifications/blob/c250bf7a34b09519d163c1edace8e2ab179d363f/source/auth/tests/mongodb-oidc.md) clued me to a missing requirement:

> For test cases that create fail points, drivers MUST use a unique `appName` to prevent interaction with other environment processes, and explicitly remove the fail point to prevent interaction between test runs.

A `fail_command_guard` has been added to disable failpoints, and an `appName` is added.


A typo was fixed in the URI for prose test 2.4: `authMechanism=MONGODB-OIDC,authMechanismProperties` => `authMechanism=MONGODB-OIDC&authMechanismProperties`. The expected error occurs on connection (rather than construction) and has been updated.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -676,6 +717,18 @@ client create_client(document::view object) {
...
         client_opts.server_api_opts(server_api_opts);
     }
+    {
+        auto const auth_mechanism = object["uriOptions"]["authMechanism"];
+        if (auth_mechanism && auth_mechanism.type() == bsoncxx::type::k_string &&
+            auth_mechanism.get_string().value == "MONGODB-OIDC") {
+            std::ifstream token_file(test_util::getenv_or_fail("OIDC_TOKEN_FILE"));
+            REQUIRE(token_file.is_open());
+            auto const token =
+                std::string((std::istreambuf_iterator<char>(token_file)), std::istreambuf_iterator<char>());
```
connorsmacd:
```suggestion
                std::string(std::istreambuf_iterator<char>(token_file), std::istreambuf_iterator<char>());
```

Unnecessary parentheses.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+    std::ifstream token_file(oidc_token_file);
+    REQUIRE(token_file.is_open());
+    return std::string((std::istreambuf_iterator<char>(token_file)), std::istreambuf_iterator<char>());
+}
+
+struct fail_command_guard {
+    fail_command_guard(std::string cmd_name, std::int32_t err_code, std::string appname) : _appname(appname) {
+        scoped_bson data(BCON_NEW("failCommands", "[", cmd_name.c_str(), "]"));
+        data += scoped_bson(BCON_NEW("errorCode", BCON_INT32(err_code)));
+        data += scoped_bson(BCON_NEW("appName", appname.c_str()));
```
eramongodb:
```suggestion
    fail_command_guard(std::string const& cmd_name, std::int32_t err_code, std::string appname)
        : _appname(std::move(appname)) {
        scoped_bson data(BCON_NEW(
            "failCommands",
            "[",
            cmd_name.c_str(),
            "]",
            "errorCode",
            BCON_INT32(err_code),
            "appName",
            _appname.c_str()));
```
Allocation reduction.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+        if (uri.appname()) {
+            _appname.emplace(*uri.appname());
+        }
+        if (is_pooled) {
+            _pool.emplace(uri, opts);
+            _pool_entry = _pool->acquire();
+        } else {
+            _client.emplace(uri, opts);
+        }
+    }
```
eramongodb:
```suggestion
    OIDCTestFixture(v1::uri const& uri, v1::client::options opts, bool is_pooled) : _is_pooled(is_pooled) {
        if (is_pooled) {
            _pool.emplace(uri, std::move(opts));
            _pool_entry = _pool->acquire();
        } else {
            _client.emplace(uri, std::move(opts));
        }
    }
```
Allocation reduction + remove unused `_appname`.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+    v1::client& client() {
+        if (_is_pooled) {
+            return *(_pool_entry.value());
+        } else {
+            return *_client;
+        }
```
eramongodb:
```suggestion
        if (_pool_entry) {
            return **_pool_entry;
        } else {
            return *_client;
        }
```
Redundant bool?

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+    }
+    v1::pool& pool() {
+        if (!_is_pooled) {
+            throw std::logic_error("Not a pooled client");
+        }
+        return *_pool;
```
eramongodb:
```suggestion
        if (!_pool) {
            throw std::logic_error("Not a pooled client");
        }
        return *_pool;
```
Redundant bool?

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    ~fail_command_guard() {
+        // Try to disable failpoint. Ignore error.
+        scoped_bson data(BCON_NEW("appName", _appname.c_str()));
+        auto cmd = scoped_bson(R"({
+            "configureFailPoint": "failCommand",
+            "mode": "off"
+        })");
+        cmd += scoped_bson(BCON_NEW("data", BCON_DOCUMENT(data.bson())));
```
eramongodb:
```suggestion
        // Try to disable failpoint. Ignore error.
        auto cmd = scoped_bson(R"({
            "configureFailPoint": "failCommand",
            "mode": "off"
        })");
        cmd += scoped_bson(BCON_NEW("data", "{", "appName", _appname.c_str(), "}"));
```
Allocation reduction.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+        scoped_bson data(BCON_NEW("failCommands", "[", cmd_name.c_str(), "]"));
+        data += scoped_bson(BCON_NEW("errorCode", BCON_INT32(err_code)));
+        data += scoped_bson(BCON_NEW("appName", appname.c_str()));
+
+        auto cmd = scoped_bson(R"({
+            "configureFailPoint": "failCommand",
+            "mode": { "times": 1 }
+        })");
+        cmd += scoped_bson(BCON_NEW("data", BCON_DOCUMENT(data.bson())));
+        admin_command(cmd.view());
```
eramongodb, line 135: If OIDC test server requires auth, this this admin command should also use auth (which is setup below).

src/mongocxx/include/mongocxx/v1/oidc_callback_params.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <cstdint>
+
+namespace mongocxx {
+namespace v1 {
+
+///
+/// Parameters passed to an OIDC callback.
+///
+/// @note This type is not user-constructible. Instances are only provided by the driver as arguments to the registered
+/// credential callback.
```
eramongodb:
```suggestion
/// @note This type is not user-constructible. The lifetime of associate views is limited to the scope of the OIDC
/// callback function to which this argument is passed.
```
Wording tweak suggestion to clarify the expected lifetime of associated views.

src/mongocxx/test/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,585 @@
...
+        }
+        return *_pool;
+    }
+
+   private:
+    bool _is_pooled;
+    bsoncxx::v1::stdx::optional<v1::pool> _pool;
+    bsoncxx::v1::stdx::optional<v1::pool::entry> _pool_entry;
+    bsoncxx::v1::stdx::optional<v1::client> _client;
+    bsoncxx::v1::stdx::optional<std::string> _appname;
```
eramongodb:
```suggestion
```
Remove: appName is now handled by `OIDCTestURI`.

.evergreen/config_generator/components/oidc.py
```diff
@@ -0,0 +1,92 @@
...
+                        'AWS_SESSION_TOKEN',
+                    ],
+                    script='./drivers-evergreen-tools/.evergreen/auth_oidc/setup.sh',
+                ),
+            ],
+            teardown_group=[
+                bash_exec(
+                    command_type=EvgCommandType.SETUP,
+                    script='./drivers-evergreen-tools/.evergreen/auth_oidc/teardown.sh',
+                )
```
eramongodb, line 76:
```suggestion
                bash_exec(
                    script='./drivers-evergreen-tools/.evergreen/auth_oidc/teardown.sh',
                )
```
Setup for pre-test commands, system for post-test commands.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/oidc_callback.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+using v1::oidc_callback;
+
+} // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
+
+///
+/// @file
+/// Declares @ref mongocxx::oidc_callback.
```
eramongodb:
```suggestion
/// Provides @ref mongocxx::oidc_callback.
```

eramongodb (changes requested):

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Updated, and wrapped with `clang-format off` / `clang-format on` to avoid the extra line breaks.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Good spot. This confused me at first since no error was triggered despite using the default URI without auth. On further look, it turns out I mistakenly assumed `configureFailPoint` requires auth, but [server code](https://github.com/10gen/mongo/blob/f1f4a41da7a113d9b2ba402933d6d2ff2bdb52cb/src/mongo/db/commands/fail_point_cmd.cpp#L104) notes:
```
// No auth needed because it only works when enabled via command line.
```

Simplified `fail_command_guard` to remove auth.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Oop. Removed.

src/mongocxx/test/v1/oidc_callback.cpp
kevinAlbs: Good point. Removed.

src/mongocxx/lib/mongocxx/v1/oidc_callback.cpp
```diff
@@ -0,0 +1,63 @@
...
+//
+
+#include <mongocxx/v1/oidc_callback_params.hh>
+#include <mongocxx/v1/oidc_credential.hh>
+
+#include <exception>
+
+#include <mongocxx/private/mongoc.hh>
+
+#include <mongoc/mongoc.h>
```
eramongodb:
```suggestion
#include <mongocxx/private/mongoc.hh>
```
Redundant include.

eramongodb (approved):

connorsmacd (approved):

PR #1640, opened 2026-05-08
CXX-3491 Fix misc. v1::text_options doc comments and includes

eramongodb:
Prompted by questions about new API design for v1 when no v_noabi equivalents already exist, in an attempt to point to `v1::text_options` as reference example, identified several issues with the `v1::text_options`-related headers that were overlooked by https://github.com/mongodb/mongo-cxx-driver/pull/1522:

- Incorrect include of v1 normal header by the v_noabi forward header.
- Incorrect include of v_noabi macro guard headers by the v1 normal header.
- Misc. copy-paste errors in doc comments.
- Update doc comments to refer to the root namespace redeclaration(s) only (users should be using either `mongocxx::text_options` or `v1::text_options`, not `v_noabi::text_options`).

Although `v_noabi::options::text` was deliberatedly added for consistency with existing v_noabi option class declarations, the redeclaration of `v1::text_options` in the v_noabi namespace as `v_noabi::text_options` is incorrect/unnecessary: only the root namespace redeclaration should be provided. Nevertheless, left as-is due to API backward compatibility. New API should not do this going forward.

### DIFF
--- a/src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/text-fwd.hpp
+++ b/src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/text-fwd.hpp
@@ -16,8 +16,6 @@
 
 #include <mongocxx/v1/text_options-fwd.hpp> // IWYU pragma: export
 
-#include <mongocxx/v1/text_options.hpp>
-
 #include <mongocxx/config/prelude.hpp>
 
 namespace mongocxx {
@@ -38,7 +36,7 @@ using v1::text_options;
 
 ///
 /// @file
-/// Declares utilities related to mongocxx logging.
+/// Declares @ref mongocxx::text_options.
 ///
 /// @par Includes
 /// - @ref mongocxx/v1/text_options-fwd.hpp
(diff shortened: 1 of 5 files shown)


src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/text-fwd.hpp
kevinAlbs, line 24: > only the root namespace redeclaration should be provided. Nevertheless, left as-is due to API backward compatibility. New API should not do this going forward.

Suggest adding a comment to note rationale for keeping:
```c++
// v_noabi::text_options is redundant with v_noabi::options::text but kept for API backward compatibility.
```

This might help avoid an accidental future repeat.

kevinAlbs (approved):
Nice spot, and thank you for the fix. LGTM with a suggested comment.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/text-fwd.hpp
eramongodb: Added comments as suggested. Considered documenting as public API, but opted to leave undocumented instead (avoid calling attention to them; similar to undocumented `using` in v3 releases). `mongocxx::text_options` and `mongocxx::v1::text_options` are already publically documented. Also fixed references in `v_noabi::options::encrypt` to use the correct, documented public API (v1) rather than the undocumented redeclarations (v_noabi); because these are all type aliases for the same type, this should be an API backward-compatible change.

PR #1644, opened 2026-05-14
CXX-3492: Support excluding unstable ABI symbols from build

eramongodb:
To (minimally) support packaging policy requirements in advance of CXX-804, this PR introduces new build configuration options which may be used by packaging to completely exclude unstable ABI symbols when building shared libraries. This additionally requires using v_noabi-specific export macros to ensure any downstream code which observes unstable ABI headers (as these changes do not exclude them from installation; that will be handled by the package definition). These options are for internal use only: that is, specifically for stable ABI packaging purposes.


etc/coding_guidelines.md
```diff
@@ -324,6 +324,12 @@ lib/
 
 Analogous `MONGOCXX_`-prefixed variants of the above macros apply to the mongocxx library.
 
+Declarations under `v_noabi/` must use the `_UNSTABLE` variants instead:
```
kevinAlbs, line 325: Suggest moving this section above "Analogous `MONGOCXX_`-prefixed variants". Since there are both `BSONCXX` and `MONGOCXX` versions of these macros.

src/bsoncxx/include/bsoncxx/v_noabi/bsoncxx/config/prelude.hpp
```diff
@@ -14,6 +14,12 @@
 
 #include <bsoncxx/v1/detail/prelude.hpp> // IWYU pragma: export
 
+// bsoncxx/v_noabi/bsoncxx/config/config.hpp (generated by CMake)
+#pragma push_macro("BSONCXX_ABI_EXPORT_CDECL_UNSTABLE")
+#undef BSONCXX_ABI_EXPORT_CDECL_UNSTABLE
+#pragma push_macro("BSONCXX_ABI_EXPORT_UNSTABLE")
+#undef BSONCXX_ABI_EXPORT_UNSTABLE
```
kevinAlbs, line 21: Also add `BSONCXX_EXPORT_UNSTABLE_ABI` (defined by `#cmakedefine01`) to avoid leaking macro.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/config/prelude.hpp
```diff
@@ -14,6 +14,12 @@
 
 #include <mongocxx/v1/detail/prelude.hpp> // IWYU pragma: export
 
+// mongocxx/v_noabi/mongocxx/config/config.hpp (generated by CMake)
+#pragma push_macro("MONGOCXX_ABI_EXPORT_CDECL_UNSTABLE")
+#undef MONGOCXX_ABI_EXPORT_CDECL_UNSTABLE
+#pragma push_macro("MONGOCXX_ABI_EXPORT_UNSTABLE")
+#undef MONGOCXX_ABI_EXPORT_UNSTABLE
```
kevinAlbs, line 21: Also add `MONGOCXX_EXPORT_UNSTABLE_ABI` (defined by `#cmakedefine01`) to avoid leaking macro.

kevinAlbs (approved):

rcsanchez97 (approved):

PR #1646, opened 2026-05-15
Use new abi-dumper approach for abi-compliance-check tasks

eramongodb:
Resolves ongoing `abi-compliance-check-impls-*` EVG task failures. The root cause of failure is still unclear, but the checks appear to be using an absurd amount of memory (over 100GB!?) that eventually triggers the OOM killer. This behavior is not reproducible in an EVG spawn host. It is unclear why this started happening (no change to EVG environment, dependencies, or to library code when failures began). Nevertheless, took this opportunity to migrate the task to use the newer [abi-dumper](https://github.com/lvc/abi-dumper) approach [as recommended by](https://lvc.github.io/abi-compliance-checker/#ABI_Dumper) abi-compliance-checker documentation:

>  This new way is based on the analysis of the debug-info from binary objects. It's more reliable, faster and simple way.

This avoids the memory usage problem while also improving analysis quality. However, the report is now split by library (combined `compat_report.html` -> `bsoncxx_report.html` + `mongocxx_report.html`), and the list of header/source files included in the HTML report is misleading (due to abi-dumper not preserving directory paths -> flat list of header basenames only). Nevertheless, the total set of symbols actually being analyzed and reported appears to be accurate.

Some drive-by improvements include:

- Increasing parallelism of independent commands. (Builds are not in parallel due to each build already maximizing the parallelism of its own compilation commands.)
- Using git worktree instead of stashing to build base commit libraries.
- Skip mongocrypt when building the C driver, as it does not affect the C++ driver's ABI stability.
- Explicitly requesting DWARFv4 debug-info format instead of plain `-g` (abi-dumper has compatibility problems with DWARFv5 due to typeinfo being relocated from `.debug_types` into `.debug_info`)


kevinAlbs (approved):
The fix and speed-ups are much appreciated. LGTM.

Sampling some of the [previous successful tasks](https://spruce.corp.mongodb.com/task/mongo_cxx_driver_abi_stability_abi_compliance_check_stdlib_cxx20_patch_6f83c7f300855b2907125144ce530a49d9f924d4_6a077895b4dadc0008299d25_26_05_15_19_48_40/logs?execution=0) shows a big speedup: 16m24s => 1m45s. Though I expect something was going very wrong before to trigger the large memory usage.

PR #1647, opened 2026-05-20
CXX-2494 Implement improved bulk write API

connorsmacd:
CXX-2494

# Summary

This PR adds an API that essentially mirrors the types found in `mongoc-bulkwrite.h`.

> [!CAUTION]
> There is an unfortunate naming mismatch between the C++ Driver and the C Driver stemming from the two coexisting bulk write APIs. The improved bulk write API is typically referred to as "client bulk write", while the old bulk write API is referred to as "collection bulk write". With this PR, these two APIs are modeled by the following names:
>
> | API | C Driver | C++ Driver |
> | - | - | - |
> | Collection Bulk Write | `mongoc_bulk_operation_t` | `bulk_write` |
> | Client Bulk Write | `mongoc_bulkwrite_t` | `client_bulk_write` |
>
> The lack of distinction between client and collection bulk write in the type names makes it difficult for users to distinguish these APIs at a glance. Moreover, the mapping between the C and C++ types defy intuition as one would expect `bulk_write` to be the C++ version of `mongoc_bulkwrite_t`. We should consider adding typedefs/aliases to both drivers that explicitly differentiate between collection and client bulk write (see CDRIVER-6347 and CXX-3510).

# New Types

With the exception of `mongoc_bulkwrite_t`, most of the types in the C API lack copy functions and getter/setter parity, so the C++ equivalents of these types are implemented using shadow state rather than directly wrapping their corresponding C pointers. The improved bulk write API is comprised of the following types:

* `v1::client_bulk_write`
  - Wraps `mongoc_bulkwrite_t`
* `v1::client_bulk_write::options`
  - Mirrors `mongoc_bulkwriteopts_t`
* `v1::client_bulk_write::result`
  - Mirrors `mongoc_bulkwriteresult_t`
* `v1::client_bulk_write::exception`
  - Mirrors `mongoc_bulkwriteexception_t`
  - Inherits `v1::exception`
* `v1::client_bulk_write::insert_one_options`
  - Mirrors `mongoc_insertoneopts_t`
  - Like `mongoc_insertoneopts_t`, this options type does not currently have any fields, so it is effectively a tag type.
* `v1::client_bulk_write::update_one_options`
  - Mirrors `mongoc_updateoneopts_t`
* `v1::client_bulk_write::update_many_options`
  - Mirrors `mongoc_updatemanyopts_t`
* `v1::client_bulk_write::replace_one_options`
  - Mirrors `mongoc_replaceoneopts_t`
* `v1::client_bulk_write::delete_one_options`
  - Mirrors `mongoc_deleteoneopts_t`
* `v1::client_bulk_write::delete_many_options`
  - Mirrors `mongoc_deletemanyopts_t`

# Spec Tests

The unified test runner now supports client bulk write. However, this PR only adds a subset of the unified tests that are validated in the C Driver. The rationale for excluding certain unified tests is as follows:

* No support for accessing the raw result document (CXX-2543).
  - `crud/client-bulkWrite-errorResponse.json`
* No support for schema version 1.21. Since this schema version was introduced as part of client bulk write, we could argue it is in scope for this PR. However, we also do not support schema version 1.20 (CXX-2780), which is unrelated to client bulk write, so I chose to omit these tests for now.
  - `crud/client-bulkWrite-errors.json`
  - `retryable-writes/client-bulkWrite-clientErrors.json`
  - `retryable-writes/client-bulkWrite-serverErrors.json`
* No support for schema version 1.25.
  - `client-side-encryption/client-bulkWrite-eq.json`
* Some specs are entirely untested in the C++ Driver (Command Logging and Monitoring) or only contain a small subset of the tests we run in the C Driver (Transactions). It is hard for me to figure out why we do or do not run certain tests in the C++ Driver that we run in the C Driver, so I erred on the side of not adding tests in these cases.
  - `command-logging-and-monitoring/monitoring/unacknowledged-client-bulkWrite.json`
  - `transactions/client-bulkWrite.json`


src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <mongocxx/config/prelude.hpp>
+
+namespace mongocxx {
+namespace v_noabi {
+
+using v1::client_bulk_write;
+
+} // namespace v_noabi
+} // namespace mongocxx
+
```
eramongodb:
```suggestion
```
Redeclaration in `v_noabi` is not needed (it is a `v1` entity).

Users will use either `mongocxx::client_bulk_write` (only need API compat) or `mongocxx::v1::client_bulk_write` (need ABI compat).

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write.hpp
```diff
@@ -0,0 +1,47 @@
...
+} // namespace v_noabi
+} // namespace mongocxx
+
+namespace mongocxx {
+
+using v_noabi::client_bulk_write;
+
+} // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
```
eramongodb:
```suggestion
#include <mongocxx/v1/client_bulk_write.hpp> // IWYU pragma: export
```
The forward header handles declarating the class type in the root namespace. This header only needs to provide the class definition via transitive `#include <mongocxx/v1/client_bulk_write.hpp>`.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+///
+/// @file
+/// Provides @ref mongocxx::v_noabi::client_bulk_write.
```
eramongodb:
```suggestion
/// Provides @ref mongocxx::client_bulk_write.
```

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp
```diff
@@ -0,0 +1,45 @@
...
+using v1::client_bulk_write;
+
+} // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
+
+///
+/// @file
+/// Declares @ref mongocxx::v_noabi::client_bulk_write.
```
eramongodb:
```suggestion
/// Declares @ref mongocxx::client_bulk_write.
```

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp
```diff
@@ -0,0 +1,45 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/v1/client_bulk_write-fwd.hpp> // IWYU pragma: export
+
+#include <mongocxx/v1/client_bulk_write.hpp>
+
```
eramongodb:
```suggestion
```
Forward headers must not include normal headers.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <mongocxx/config/prelude.hpp>
```
eramongodb, line 19:
```suggestion
#include <mongocxx/v1/detail/prelude.hpp>
```
v1 headers must never include v_noabi headers (including macro guard headers).

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client.hpp
```diff
@@ -462,12 +462,21 @@ class client {
...
         v_noabi::pipeline const& pipe,
         v_noabi::options::change_stream const& options = {});
 
+    ///
+    /// Return a new client-level bulk write operation for use with this client.
+    ///
+    /// @{
+    MONGOCXX_ABI_EXPORT_CDECL(v_noabi::client_bulk_write) create_bulk_write();
+
+    MONGOCXX_ABI_EXPORT_CDECL(v_noabi::client_bulk_write) create_bulk_write(v_noabi::client_session& session);
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(v1::client_bulk_write) create_bulk_write();

    MONGOCXX_ABI_EXPORT_CDECL(v1::client_bulk_write) create_bulk_write(v_noabi::client_session& session);
```
For consistency with fixed redeclarations. New API should encourage and use v1 namespace entities even in existing v_noabi API, e.g.: https://github.com/mongodb/mongo-cxx-driver/blob/30672d781d6c72fe6737b6afcb9dd1ef9f38aabf/src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/encrypt.hpp#L269-L282

src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -245,6 +246,15 @@ class client {
...
     ///
 
+    ///
+    /// Return a new client-level bulk write operation for use with this client.
+    ///
+    /// @{
+    MONGOCXX_ABI_EXPORT_CDECL(v1::client_bulk_write) create_bulk_write();
+
+    MONGOCXX_ABI_EXPORT_CDECL(v1::client_bulk_write) create_bulk_write(v1::client_session& session);
+    /// @}
```
eramongodb, line 254:
```suggestion
    /// @}
    ///
```
Doxygen requires [_at least two_](https://www.doxygen.nl/manual/docblocks.html) C++ comment lines to denote a doxygen comment block.

Image

I've historically observed inconsistent results when using a single C++ comment line. It appears to work atm for this specific case, but may not reliably continue to do so unless it is accompanied by the empty comment line.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,971 @@
...
+    class impl;
+    std::shared_ptr<impl> _impl;
+
+   public:
+    ///
+    /// A document mapping model index to the corresponding write error.
+    ///
+    /// @returns An empty document if there are no write errors.
+    ///
+    bsoncxx::v1::document::view MONGOCXX_ABI_CDECL write_errors() const;
```
eramongodb, line 448:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::document::view) write_errors() const;
```
Suggest using `MONGOCXX_ABI_EXPORT_CDECL` for consistency with other exported functions, even if somewhat redundant due to the class-level export macro.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,971 @@
...
+
+///
+/// A client bulk write error.
+///
+/// @par Inherits:
+/// - @ref mongocxx::v1::exception
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
+class client_bulk_write::exception : public v1::exception {
```
eramongodb, line 437: Explicitly adding this comment to the review for future reference: a class member exception type is novel, but I think it makes sense given the spec + consistency with class member error categories and error codes in v1 API. I am in favor of using this approach for other component-specific exception types going forward.

I also considered the possibility of inheriting from `v1::server_error` instead of `v1::exception` due to the potential inclusion of server errors. However, I am in favor of a standalone exception class (as currently proposed) to avoid complicating the exception inheritance hierarchy (i.e. trying to map fields of one exception type to another). The spec does not mandate an IS-A relationship, so we should not need to do so either (beyond the `v1::exception` inheritance).

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,971 @@
...
+
+    ///
+    /// Set the "ordered" field.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(options&) ordered(bool ordered);
+
+    ///
+    /// Return the current "ordered" field.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(bool) ordered() const;
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<bool>) ordered() const;
```
Even if the spec requires the _command_ to unconditionally define `ordered: true` in commands when the value is not explicitly specified, the _field_ itself is still optional.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,971 @@
...
+
+    class insert_one_options;
+    class update_one_options;
+    class update_many_options;
+    class replace_one_options;
+    class delete_one_options;
+    class delete_many_options;
+
+    ///
+    /// Destroy this object.
```
eramongodb, line 76:
```suggestion
    /// Destroy this object.
    ///
    /// @warning Invalidates all associated views.
```
For consistency with other classes with view-returning accessors.

eramongodb (changes requested):
Initial review focused on v1 vs. v_noabi API design. Review focused on client bulk write behavior and tests will follow.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
connorsmacd: This is indeed the correct change, but FWIW, I made the field optional to be consistent with `bulk_write::options::ordered`. I see now that the spec dictates that `bulk_write::options::ordered` always has a value, while `client_bulk_write::options::ordered` does not.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp
eramongodb: Sorry, I'd meant to delete this review comment (this is indeed a v_noabi header, not a v1 header).

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
connorsmacd: I made this change on every destructor for types that return views, but for `client_bulk_write` itself, this doesn't seem to apply. The only member function that returns anything that's not `client_bulk_write&` is `execute`, which returns a `result` object that owns all of its BSON. Did you mean to leave this comment on a different destructor? Or am I mistaken?

eramongodb: You're correct, `client_bulk_write` itself does not need the warning, only its member classes.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1388 @@
...
+    if (this != &other) {
+        auto* const tmp = new impl{};
+        delete impl::with(_impl);
+        _impl = tmp;
+    }
+
+    return *this;
+}
+
+client_bulk_write::insert_one_options::insert_one_options() : _impl{new impl{}} {}
```
eramongodb:
```suggestion
class client_bulk_write::insert_one_options::impl {};

client_bulk_write::insert_one_options::~insert_one_options() = default;

client_bulk_write::insert_one_options::insert_one_options(insert_one_options&& other) noexcept = default;

client_bulk_write::insert_one_options& client_bulk_write::insert_one_options::operator=(
    insert_one_options&& other) noexcept = default;

client_bulk_write::insert_one_options::insert_one_options(insert_one_options const& other) = default;

client_bulk_write::insert_one_options& client_bulk_write::insert_one_options::operator=(
    insert_one_options const& other) = default;

client_bulk_write::insert_one_options::insert_one_options() : _impl{nullptr} {}
```
Given `insert_one_options` is currently stateless (no supported option fields), may as well just treat it like a monostate class (out-of-line exported SMFs still leave room for extension in the future if needed).

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1388 @@
...
+#include <system_error>
+#include <utility>
+
+#include <mongocxx/private/scoped_bson.hh>
+#include <mongocxx/private/utility.hh>
+
+namespace mongocxx {
+namespace v1 {
+
+// NOLINTBEGIN(cppcoreguidelines-owning-memory): owning void* for ABI stability.
```
eramongodb: Minor: I'd prefer to narrow the scope of these nolints to the explicit new/delete expressions (in SMFs) instead of the entire component. Feel free to reject/resolve this suggestion.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1388 @@
...
+    }
+
+    static impl* with(void* ptr) {
+        return static_cast<impl*>(ptr);
+    }
+};
+
+client_bulk_write::result::~result() {
+    delete impl::with(this);
+    _impl = nullptr; // scan-build: warning: Use of memory after it is freed [cplusplus.NewDelete]
```
eramongodb, line 275: Remark: I hope we can do away with this workaround someday. 🫠

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1029,56 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+        return;
+
+    auto const actual_result = actual["result"];
+    if (auto const expected_result = expect_error["expectResult"]) {
+        assert::matches(actual_result.get_value(), expected_result.get_value(), get_entity_map());
+    }
+
+    if (auto const is_client_error = expect_error["isClientError"]) {
+        REQUIRE(is_client_error.get_bool());
+    }
```
eramongodb, line 1059: Is this correct given `BulkWriteException` may also contain server errors?

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1029,56 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+    REQUIRE_FALSE(expect_error["errorLabelsContain"]);
+    REQUIRE_FALSE(expect_error["errorLabelsOmit"]);
+
+    REQUIRE_FALSE(/* TODO */ expect_error["errorContains"]);
+    REQUIRE_FALSE(/* TODO */ expect_error["errorCode"]);
+    REQUIRE_FALSE(/* TODO */ expect_error["errorCodeName"]);
+}
+
+void assert_error(
+    mongocxx::v1::client_bulk_write::exception const& exception,
```
eramongodb:
```suggestion
    mongocxx::client_bulk_write::exception const& exception,
```
Suggest defaulting to root namespace redeclarations in v_noabi test code. This should also helps catch API breaking changes in the future when the root namespace redeclarations are updated.

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1029,56 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+    if (auto const is_client_error = expect_error["isClientError"]) {
+        REQUIRE(is_client_error.get_bool());
+    }
+
+    REQUIRE_FALSE(expect_error["errorLabelsContain"]);
+    REQUIRE_FALSE(expect_error["errorLabelsOmit"]);
+
+    REQUIRE_FALSE(/* TODO */ expect_error["errorContains"]);
+    REQUIRE_FALSE(/* TODO */ expect_error["errorCode"]);
+    REQUIRE_FALSE(/* TODO */ expect_error["errorCodeName"]);
```
eramongodb: > No support for accessing the raw result document ([CXX-2543](https://jira.mongodb.org/browse/CXX-2543)).

This should be feasible for v1 exceptions: `v1::exception` provides an internal [`get_reply()`](https://github.com/mongodb/mongo-cxx-driver/blob/30672d781d6c72fe6737b6afcb9dd1ef9f38aabf/src/mongocxx/lib/mongocxx/v1/exception.hh#L64-L65) accessor to store raw reply documents returned by mongoc even for client-side errors [to preserve error labels](https://github.com/mongodb/mongo-cxx-driver/blob/30672d781d6c72fe6737b6afcb9dd1ef9f38aabf/src/mongocxx/lib/mongocxx/v1/exception.cpp#L361-L367). This internal API may also be used to implement error response tests.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1388 @@
...
+    auto const guard = std::unique_ptr<mongoc_bulkwriteexception_t, deleter>{exc};
+
+    auto p = std::unique_ptr<impl>{new impl{}};
+
+    if (auto const* const doc = libmongoc::bulkwriteexception_writeerrors(exc)) {
+        p->_write_errors = scoped_bson_view{doc}.value();
+    }
+
+    if (auto const* const arr = libmongoc::bulkwriteexception_writeconcernerrors(exc)) {
+        p->_write_concern_errors = scoped_bson_view{arr}.value();
```
eramongodb:
```suggestion
        p->_write_concern_errors = scoped_bson_view{arr}.array_view();
```
Suggest eagerly representing `_write_concern_errors` as `array::value` instead of converting doc -> arr in `write_concern_errors() const`.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,983 @@
...
+    /// Copy assignment.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(insert_one_options&) operator=(insert_one_options const& other);
+
+    ///
+    /// Default initialization.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL() insert_one_options();
+
+    class internal;
```
eramongodb:
```suggestion
```
Not needed (yet).

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1459 @@
...
+            return v.bson();
+        })
+        .forever();
+
+    auto const ex = mocks.make();
+
+    CHECK(ex.error_reply() == v.view());
+}
+
+TEST_CASE("ownership", "[mongocxx][v1][client_bulk_write][insert_one_options]") {
```
eramongodb:
```suggestion
}

TEST_CASE("partial_result", "[mongocxx][v1][client_bulk_write][exception]") {
  // ...
}

TEST_CASE("ownership", "[mongocxx][v1][client_bulk_write][insert_one_options]") {
```

Missing test coverage of `partial_result()` behavior.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
connorsmacd: It looks like these changes are causing issues on CI, e.g.:

```
client_bulk_write.hpp(434): error C2487: 'write_errors': member of dll interface class may not be declared with dll interface
```

According to the [docs](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2487?view=msvc-170), you cannot declare members with a DLL interface on a class that is also declared with a DLL interface.

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1459 @@
...
+    auto const source_doc = scoped_bson{R"({"0": {"code": 1, "message": "src", "details": {}}})"};
+    auto const target_doc = scoped_bson{R"({"0": {"code": 2, "message": "tgt", "details": {}}})"};
+
+    client_bulk_write::exception::internal::write_errors(source) = source_doc.value();
+    client_bulk_write::exception::internal::write_errors(target) = target_doc.value();
+
+    REQUIRE(source.write_errors() == source_doc.view());
+    REQUIRE(target.write_errors() == target_doc.view());
+
+    SECTION("copy") {
```
eramongodb, line 531:
```suggestion
    REQUIRE(target.write_errors() == target_doc.view());

    auto const source_data = source.write_errors().data();

    SECTION("move") {
        auto move = std::move(source);

        // source is in an assign-or-move-only state.

        CHECK(move.write_errors().data() == source_data);

        target = std::move(move);

        // move is in an assign-or-move-only state.

        CHECK(target.write_errors().data() == source_data);
    }

    SECTION("copy") {
```

Missing coverage of move SMFs.

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1459 @@
...
+        auto copy = source;
+        target = copy;
+    }
+
+    SECTION("move") {
+        auto moved = std::move(source);
+        target = std::move(moved);
+    }
+}
```
eramongodb: Suggest removing trivial test coverage of a monostate `insert_one_options`.

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1459 @@
...
+        CHECK(copy.write_errors() == source_doc.view());
+
+        target = copy;
+
+        CHECK(copy.write_errors() == source_doc.view());
+        CHECK(target.write_errors() == source_doc.view());
+    }
+}
+
+namespace {
```
eramongodb, line 556:
```suggestion
}

TEST_CASE("default", "[mongocxx][v1][client_bulk_write][exception]") {
  // ...
}

namespace {
```
Suggest adding a test case asserting default state of exception object fields (empty docs, empty arrays, and/or empty optionals).

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
eramongodb: TIL. I support reverting my suggested changes and going back to using `MONGOCXX_ABI_CDECL` for member functions of exported classes. I suppose it's tolerable given exported classes are already a unique case within the codebase.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.hh
```diff
@@ -0,0 +1,90 @@
...
+   public:
+    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bool> const&) bypass_document_validation(
+        options const& self);
+    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bsoncxx::v1::types::value> const&) comment(
+        options const& self);
+    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bsoncxx::v1::document::value> const&) let(
+        options const& self);
+    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bool> const&) ordered(options const& self);
+    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bool> const&) verbose_results(
+        options const& self);
```
eramongodb:
```suggestion
    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bsoncxx::v1::types::value> const&) comment(
        options const& self);
    static MONGOCXX_ABI_EXPORT_CDECL_TESTING(bsoncxx::v1::stdx::optional<bsoncxx::v1::document::value> const&) let(
        options const& self);
```
Data members do not need internal direct accessors when they provide little benefit over using the public API (e.g. cheap-to-copy `int`, `Optional<int>`, etc.). Note these cheap-to-copy data member accessors are not currently in use either (e.g. `bulk_write_options_c` ctor already does the right thing).

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1388 @@
...
+
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <system_error>
+#include <utility>
+
+#include <mongocxx/private/scoped_bson.hh>
+#include <mongocxx/private/utility.hh>
```
eramongodb, line 44:
```suggestion
#include <mongocxx/private/mongoc.hh>
#include <mongocxx/private/scoped_bson.hh>
#include <mongocxx/private/utility.hh>

#include <bson/error.h>
```
IWYU (for `bson_error_t` and mongoc API).

eramongodb (changes requested):

src/mongocxx/test/spec/unified_tests/runner.cpp
connorsmacd: Good catch. It looks like I would need to implement a crude solution similar to what we have [here](https://github.com/mongodb/mongo-cxx-driver/blob/097f026d1354f3aa3bf82feb4ce6002d23c7683e/src/mongocxx/test/spec/unified_tests/runner.cpp#L894).

That said, the two tests that validate client side errors reported via `client_bulk_write::exception` are `crud/client-bulkWrite-errors.json` and `retryable-writes/client-bulkWrite-clientErrors.json`, which have both been omitted from this PR for reasons described in the PR description. If we agree that those tests are out of scope, then I could just assert that the "isClientError" field is not present with a comment indicating that we do not yet support distinguishing between client and server errors for `client_bulk_write::exception`. Otherwise, I will move forward with the pattern matching solution.

eramongodb: > It looks like I would need to implement a crude solution similar to what we have [here](https://github.com/mongodb/mongo-cxx-driver/blob/097f026d1354f3aa3bf82feb4ce6002d23c7683e/src/mongocxx/test/spec/unified_tests/runner.cpp#L894).

I was thinking of something like this instead (deliberately _not_ continuing the string-matching approach used with v_noabi exceptions due to their inability to clearly distinguish client vs. server errors):

```cpp
if (auto const is_client_error = expect_error["isClientError"]) {
    CHECK(e.code() != mongocxx::v1::source_errc::server);
}
```

> I could just assert that the "isClientError" field is not present with a comment indicating that we do not yet support distinguishing between client and server errors

SGTM if the above solution does not work out due to unforeseen complications.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,986 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <mongocxx/v1/detail/prelude.hpp>
```
eramongodb, line 21:
```suggestion
#pragma once

#include <mongocxx/v1/client_bulk_write-fwd.hpp> // IWYU pragma: export

//

#include <mongocxx/v1/detail/prelude.hpp>
```
Missing forward header inclusion.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,986 @@
...
+    /// This class is not copyable.
+    ///
+    client_bulk_write& operator=(client_bulk_write const& other) = delete;
+
+    ///
+    /// Append an "Insert One" operation targeting the given namespace.
+    ///
+    /// @throws mongocxx::v1::exception when a client-side error is encountered.
+    ///
+    /// @{
```
eramongodb:
```suggestion
```
Stray Doxygen group marker.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,986 @@
...
+    MONGOCXX_ABI_EXPORT_CDECL(client_bulk_write&)
+    append(bsoncxx::v1::stdx::string_view ns, bsoncxx::v1::document::value filter, delete_one_options const& opts);
+
+    ///
+    /// Append a "Delete Many" operation targeting the given namespace.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(client_bulk_write&)
+    append(bsoncxx::v1::stdx::string_view ns, bsoncxx::v1::document::value filter, delete_many_options const& opts);
+    /// @}
+    ///
```
eramongodb:
```suggestion
```
Stray Doxygen group marker.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,986 @@
...
+    client_bulk_write& operator=(client_bulk_write const& other) = delete;
+
+    ///
+    /// Append an "Insert One" operation targeting the given namespace.
+    ///
+    /// @throws mongocxx::v1::exception when a client-side error is encountered.
+    ///
+    /// @{
+    MONGOCXX_ABI_EXPORT_CDECL(client_bulk_write&)
+    append(bsoncxx::v1::stdx::string_view ns, bsoncxx::v1::document::value document, insert_one_options const& opts);
```
eramongodb:
```suggestion
    append(bsoncxx::v1::stdx::string_view ns, bsoncxx::v1::document::view document, insert_one_options const& opts);
```
None of these append operations take ownership of the provided document(s). Suggest changing the parameter types to `view` instead.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,986 @@
...
+
+    ///
+    /// Default initialization.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL() options();
+
+    ///
+    /// Set the "bypassDocumentValidation" field.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(options&) bypass_document_validation(bool bypass_document_validation);
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(options&) bypass_document_validation(bool v);
```
To avoid redundancy + verbosity, suggest using simple `v` for the parameter of option field setters (not yet implemented consistently across v1 API, sorry. [Working on it...](https://github.com/mongodb/mongo-cxx-driver/commit/57fbc8545e2167b8818bb7ffdb4f28906220673b)).

eramongodb (changes requested):

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
connorsmacd: This group marker is for `append` overloads. I'm not particularly attached to grouping these, but is there a reason why this overload set doesn't warrant a Doxygen group?

eramongodb: Doxygen does not handle "nested" doc comments within groups well, if at all. The intent of this group appears to be to apply the `\exception` command to all `append()` functions and/or simply-group the set of member functions together for doc readability purposes, but it currently accomplishes neither:

image

Given these overloads are already grouped together by name (per `SORT_MEMBER_DOCS = YES`; note also the `[1/8]` indicating the total number of overloads), I do not think an explicit Doxygen group provides much benefit here.

Despite being boilerplatey, each overload should be given its own explicit `\exception` command in their own doc comment blocks such that the result appears as follows:

image

connorsmacd: Makes sense. Thank you for the explanation.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -336,6 +343,231 @@ void set_sort(Model& model, mongocxx::document::element const& sort) {
     model.sort(sort.get_document().value);
 }
 
+namespace {
+
+bsoncxx::v1::document::value to_v1_value(document::view v) {
+    return bsoncxx::v1::document::value{static_cast<bsoncxx::v1::document::view>(v)};
```
eramongodb:
```suggestion
    return bsoncxx::v1::document::value{bsoncxx::v_noabi::to_v1(v)};
```
Consider using the ABI conversion helpers instead of explicit static casts (here and below).

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -336,6 +343,231 @@ void set_sort(Model& model, mongocxx::document::element const& sort) {
...
+    if (auto const verbose = args["verboseResults"]) {
+        opts.verbose_results(verbose.get_bool().value);
+    }
+
+    if (auto const let = args["let"]) {
+        opts.let(to_v1_value(let.get_document().value));
+    }
+
+    if (auto const comment = args["comment"]) {
+        opts.comment(bsoncxx::v1::types::value{static_cast<bsoncxx::v1::types::view>(comment.get_value())});
```
eramongodb:
```suggestion
        opts.comment(bsoncxx::v1::types::value{comment.type_value()});
```
More direct conversion sequence.

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -336,6 +343,231 @@ void set_sort(Model& model, mongocxx::document::element const& sort) {
...
+    bsoncxx::stdx::string_view ns,
+    bsoncxx::v1::document::view filter,
+    document::element const& update_element,
+    Opts const& opts) {
+    if (update_element.type() == type::k_document) {
+        client_bulk_write.append(
+            ns, filter, static_cast<bsoncxx::v1::document::view>(update_element.get_document().value), opts);
+    } else {
+        client_bulk_write.append(
+            ns, filter, static_cast<mongocxx::v1::pipeline>(build_pipeline(update_element.get_array().value)), opts);
```
eramongodb:
```suggestion
        client_bulk_write.append(
            ns, filter, v_noabi::to_v1(build_pipeline(update_element.get_array().value)), opts);
```
Attempting to validate this suggestion revealed a couple mongocxx classes were missing ABI conversion functions. Posted https://github.com/mongodb/mongo-cxx-driver/pull/1660.

eramongodb (approved):
Some minor suggestions remaining; otherwise, LGTM.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -429,7 +429,9 @@ bsoncxx::v1::stdx::optional<client_bulk_write::result> client_bulk_write::except
 }
 
 client_bulk_write::exception::exception(int code, char const* message, std::unique_ptr<impl> impl)
-    : v1::exception{v1::exception::internal::make(code, std::generic_category(), message)}, _impl{std::move(impl)} {}
+    : v1::exception{v1::exception::internal::make(code, std::generic_category(), message)}, _impl{std::move(impl)} {
+    v1::exception::internal::set_reply(*this, _impl->_error_reply);
```
eramongodb: May be able to avoid redundancy here by removing `_error_reply` in favor of `v1::exception`'s `_reply` data member for server reply messages via `exception::internal::make()` below, e.g.:

```cpp
auto ret = exception{code, message, std::move(p)};

if (auto const* const doc = libmongoc::bulkwriteexception_writeerrors(exc)) {
    v1::exception::internal::set_reply(ret, scoped_bson_view{doc}.value());
}

return ret;
```

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1031,45 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+
+    if (auto const is_client_error = expect_error["isClientError"]) {
+        CHECK(e.code() != mongocxx::v1::source_errc::server);
+    }
+
+    REQUIRE_FALSE(expect_error["errorLabelsContain"]);
+    REQUIRE_FALSE(expect_error["errorLabelsOmit"]);
+
+    if (auto const error_code = expect_error["errorCode"]) {
+        REQUIRE(e.code().value() == error_code.get_int32());
```
eramongodb:
```suggestion
        CHECK(e.code() == std::error_code{error_code.get_int32(), mongocxx::v1::server_error::internal::category()});
```
Assert correct error category as well as the error code value + prefer `CHECK()` instead of `REQUIRE()` when the condition does not guard subsequent test validity (e.g. most of the `REQUIRE` in this function are independent of one another).

src/mongocxx/test/spec/unified_tests/runner.cpp
connorsmacd: I rolled this change into the generic `v1::exception` `assert_error` overload added in ea9782f14c783339f52f34bb4d90c9f6cecf037a.

src/mongocxx/test/spec/unified_tests/operations.cpp
connorsmacd: Good suggestion. I'll make the change to use the ABI conversion function for `pipeline` once your PR is merged.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,1001 @@
...
+
+///
+/// A list of bulk write operations. Unlike @ref mongocxx::v1::bulk_write, this supports writes across more than one
+/// collection.
+///
+/// @see
+/// - [Bulk Write Operations (MongoDB Manual)](https://www.mongodb.com/docs/manual/core/bulk-write-operations/)
+///
+/// @attention This feature is experimental! It is not ready for use!
+///
```
eramongodb:
```suggestion
```
Not needed anymore (here and elsewhere): official v1 support is targeting 4.4.0 release following https://github.com/mongodb/mongo-cxx-driver/pull/1649 and https://github.com/mongodb/mongo-cxx-driver/pull/1659.

eramongodb (approved):
One more suggestion; otherwise, LGTM.

src/mongocxx/test/spec/unified_tests/operations.cpp
eramongodb: Merged: https://github.com/mongodb/mongo-cxx-driver/commit/7f2c2183df0b5e4cabfd3fadd4678b35a4187f60

connorsmacd:
I was seeing test failures as a result of edc66ec1001fd630767154e1448f5bdbd54a5119 due to a failure to properly deduce the error category: https://github.com/mongodb/mongo-cxx-driver/blob/edc66ec1001fd630767154e1448f5bdbd54a5119/src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp#L432

I pushed 8bb9f9de3c23e3ad09a66ef1e75e1273f98e2f96, which exposes the function used for `throw_exception(bson_error_t const& error)` as `exception:internal::make(bson_error_t const& error)`. @eramongodb  re-requesting review because of these changes.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -452,24 +455,22 @@ client_bulk_write::exception client_bulk_write::exception::internal::make(
...
-    auto const code = has_error ? static_cast<int>(error.code) : 0;
+    auto e = exception{v1::exception::internal::make(error), std::move(p)};
 
-    auto const* const message = has_error ? error.message : "";
+    if (auto const* const doc = libmongoc::bulkwriteexception_errorreply(exc)) {
+        v1::exception::internal::set_reply(e, scoped_bson_view{doc}.value());
+    }
 
-    return exception{code, message, std::move(p)};
+    return e;
```
eramongodb:
```suggestion
    bson_error_t error = {};

    if (libmongoc::bulkwriteexception_error(exc, &error)) {
        auto e = exception{v1::exception::internal::make(error), std::move(p)};

        if (auto const* const doc = libmongoc::bulkwriteexception_errorreply(exc)) {
            v1::exception::internal::set_reply(e, scoped_bson_view{doc}.value());
        }

        return e;
    }

    return {v1::exception::internal::make(0, std::generic_category()), std::move(p)};
```
Return value should still guard access to `bson_error_t error;`.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -452,24 +455,22 @@ client_bulk_write::exception client_bulk_write::exception::internal::make(
...
 }
 
 client_bulk_write::exception client_bulk_write::exception::internal::make() {
-    return exception{0, "", std::unique_ptr<impl>{new impl{}}};
+    return exception{v1::exception::internal::make(0, std::generic_category(), ""), std::unique_ptr<impl>{new impl{}}};
```
eramongodb:
```suggestion
    return exception{v1::exception::internal::make(0, std::generic_category()), std::unique_ptr<impl>{new impl{}}};
```
Avoid redundant strings.

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1484 @@
...
+    auto const ex = mocks.make();
+
+    CHECK(ex.write_concern_errors() == v.array_view());
+}
+
+TEST_CASE("error_reply", "[mongocxx][v1][client_bulk_write][exception]") {
+    auto const v = scoped_bson{R"({"ok": 0, "code": 123, "errmsg": "err"})"};
+
+    exception_mocks_type mocks;
+    mocks.error_reply
```
eramongodb, line 675:
```suggestion
    exception_mocks_type mocks;
    mocks.error
        ->interpose([&](mongoc_bulkwriteexception_t const* ptr, bson_error_t* err) -> bool {
            CHECK(ptr == mocks.exc_id);
            err->reserved = 2; // MONGOC_ERROR_CATEGORY
            err->code = 1;
            return true;
        })
        .forever();
    mocks.error_reply
```
To exercise the `bulkwriteexception_error() -> true` code path. Also assert error code below:

```cpp
    CHECK(ex.code() == mongocxx::v1::source_errc::mongoc);
    CHECK(ex.code().value() == 1);
```

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1484 @@
...
+
+        CHECK(copy.write_errors() == source_doc.view());
+        CHECK(target.write_errors() == source_doc.view());
+    }
+}
+
+TEST_CASE("default", "[mongocxx][v1][client_bulk_write][exception]") {
+    auto const ex = client_bulk_write::exception::internal::make();
+
+    CHECK(ex.write_errors().empty());
```
eramongodb, line 550:
```suggestion
    auto const ex = client_bulk_write::exception::internal::make();

    CHECK(ex.code().category() == std::generic_category());
    CHECK(ex.code().value() == 0);

    CHECK(ex.write_errors().empty());
```
To exercise and assert the `bulkwriteexception_error() -> false` code path.

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1484 @@
...
+    mocks.write_concern_errors
+        ->interpose([&](mongoc_bulkwriteexception_t const* ptr) -> bson_t const* {
+            CHECK(ptr == mocks.exc_id);
+            return v.bson();
+        })
+        .forever();
+
+    auto const ex = mocks.make();
+
+    CHECK(ex.write_concern_errors() == v.array_view());
```
eramongodb, line 659:
```suggestion
    CHECK(ex.code().category() == std::generic_category());
    CHECK(ex.code().value() == 0);
    CHECK(ex.write_concern_errors() == v.array_view());
```
Also assert code+category of exceptions in test code (e.g. to highlight the independence of error code from error reply documents and other fields).

eramongodb (changes requested):

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1032,45 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+
+    if (auto const error_code = expect_error["errorCode"]) {
+        CHECK(e.code() == std::error_code{error_code.get_int32(), mongocxx::v1::server_error::internal::category()});
+    }
+
+    if (auto const error_response = expect_error["errorResponse"]) {
+        assert::matches(
+            types::bson_value::value(mongocxx::v1::exception::internal::get_reply(e).value()),
+            error_response.get_value(),
+            get_entity_map());
```
eramongodb:
```suggestion
        auto const& reply = mongocxx::v1::exception::internal::get_reply(e);
        REQUIRE(reply);
        assert::matches(types::bson_value::value(*reply), error_response.get_value(), get_entity_map());
```
Improve diagnostics when reply document is missing (assertion failure instead of unexpected exception within the test section).

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1032,45 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+        return;
+
+    if (auto const expected_result = expect_error["expectResult"]) {
+        REQUIRE(actual.has_value());
+        auto const actual_result = (*actual)["result"];
+        assert::matches(actual_result.get_value(), expected_result.get_value(), get_entity_map());
+    }
+
+    if (auto const is_client_error = expect_error["isClientError"]) {
+        CHECK(e.code() != mongocxx::v1::source_errc::server);
```
eramongodb:
```suggestion
        CHECKED_IF(is_client_error.get_bool().value) {
            CHECK(e.code() != mongocxx::v1::source_errc::server);
        }

        else {
            CHECK(e.code() == mongocxx::v1::source_errc::server);
        }
```
Overlooked in prior suggestion that this field is a bool, per [spec](https://specifications.readthedocs.io/en/latest/unified-test-format/unified-test-format/):

> If true, the test runner MUST assert that the error originates from the client (i.e. it is not derived from a server response). If false, the test runner MUST assert that the error does not originate from the client.

eramongodb (approved):
A couple more test-related suggestions; otherwise, LGTM.

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,983 @@
...
+
+#include <cstdint>
+#include <memory>
+
+namespace mongocxx {
+namespace v1 {
+
+///
+/// A list of bulk write operations. Unlike @ref mongocxx::v1::bulk_write, this supports writes across more than one
+/// collection.
```
kevinAlbs:
```suggestion
/// collection and can group inserts, updates, and deletes in the same network payload.
```

Consider documenting another selling point: `MongoClient.bulkWrite` uses the `bulkWrite` command to group all CRUD operations. `MongoCollection.bulkWrite` uses the `insert`, `update`, and `delete` commands. So a bulk write of mixed operations can reduce round trips.

Consider also documenting the need for server 8.0+. I expect that is a primary reason not to use the new bulk write:

```c++
/// @note
///   @ref mongocxx::v1::client_bulk_write uses the bulkWrite server command. The bulkWrite command is only available on
///   MongoDB servers 8.0 and newer.
```

src/mongocxx/include/mongocxx/v1/client_bulk_write.hpp
```diff
@@ -0,0 +1,983 @@
...
+///
+/// @see
+/// - [Bulk Write Operations (MongoDB Manual)](https://www.mongodb.com/docs/manual/core/bulk-write-operations/)
```
kevinAlbs, line 55: Suggest also linking to the mentioned bulkWrite command documentation:

```suggestion
/// - [Bulk Write Operations (MongoDB Manual)](https://www.mongodb.com/docs/manual/core/bulk-write-operations/)
/// - [bulkWrite (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/command/bulkWrite/)
```

src/mongocxx/test/spec/unified_tests/operations.cpp
```diff
@@ -336,6 +343,230 @@ void set_sort(Model& model, mongocxx::document::element const& sort) {
...
 }
 
+namespace {
+
+bsoncxx::v1::document::value to_v1_value(document::view v) {
+    return bsoncxx::v1::document::value{bsoncxx::v_noabi::to_v1(v)};
+}
+
+bsoncxx::v1::array::value to_v1_value(array::view v) {
+    return bsoncxx::v1::array::value{static_cast<bsoncxx::v1::array::view>(v)};
```
kevinAlbs: Minor: but for consistency with the document `to_v1_value`:

```suggestion
    return bsoncxx::v1::array::value{bsoncxx::v_noabi::to_v1(v)};
```

src/mongocxx/test/spec/unified_tests/runner.cpp
```diff
@@ -1028,6 +1032,49 @@ void assert_error(mongocxx::exception& e, array::element const& ops) {
...
+        CHECKED_IF(is_client_error.get_bool().value) {
+            CHECK(e.code() != mongocxx::v1::source_errc::server);
+        }
+        else {
+            CHECK(e.code() == mongocxx::v1::source_errc::server);
+        }
+    }
+
+    REQUIRE_FALSE(expect_error["errorLabelsContain"]);
+    REQUIRE_FALSE(expect_error["errorLabelsOmit"]);
```
kevinAlbs: To signal the checks are not-yet implemented, as is done with the "errorContains" and "errorCodeName" checks below.

```suggestion
    REQUIRE_FALSE(/ * TODO */ expect_error["errorLabelsContain"]);
    REQUIRE_FALSE(/ * TODO */ expect_error["errorLabelsOmit"]);
```

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1502 @@
...
+    source.comment(source_value);
+    target.comment(target_value);
+
+    REQUIRE(source.comment() == source_value);
+    REQUIRE(target.comment() == target_value);
+
+    SECTION("move") {
+        auto move = std::move(source);
+
+        // source is in an assign-or-move-only state.
```
kevinAlbs: Minor, but related to https://github.com/mongodb/mongo-cxx-driver/pull/1661, suggest replacing all "assign-or-move-only" comments with "assign-or-destroy-only".

src/mongocxx/test/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1502 @@
...
+
+TEST_CASE("error_reply", "[mongocxx][v1][client_bulk_write][exception]") {
+    auto const code = 1;
+    auto const v = scoped_bson{R"({"ok": 0, "code": 123, "errmsg": "err"})"};
+
+    exception_mocks_type mocks;
+    mocks.error
+        ->interpose([&](mongoc_bulkwriteexception_t const* ptr, bson_error_t* err) {
+            CHECK(ptr == mocks.exc_id);
+            err->reserved = 2; // MONGOC_ERROR_CATEGORY
```
kevinAlbs: Minor: but I expect a C driver error returning a server reply would [set the reserved](https://github.com/kevinAlbs/mongo-c-driver/blob/18154d15e891a36100b68f9fcd0b5a80cc2e91d7/src/libmongoc/src/mongoc/mongoc-rpc.c#L135) `bson_error_t` field to `MONGOC_ERROR_CATEGORY_SERVER`

```suggestion
            err->reserved = 3; // MONGOC_ERROR_CATEGORY_SERVER
```

I don't think this has practical impact for the test, but might help make the test better mock the expected behavior. This suggestion requires changing the expectation below to:

```c++
CHECK(ex.code() == mongocxx::v1::source_errc::server);
```

src/mongocxx/test/v1/client_bulk_write.cpp
kevinAlbs, line 1: DRIVERS-716 added [CRUD prose tests](https://github.com/mongodb/specifications/tree/c9b3ee0af76537b37a1a95e8025811f0337e468c/source/crud/tests#5-mongoclientbulkwrite-collects-writeconcernerrors-across-batches) of large batches of documents. Since the C driver abstracts splitting large requests and merging responses, the prose tests might not add meaningful test coverage in the C++ driver. Consider adding a comment in `spec/crud.cpp` to note these prose tests are not implemented. That might help reduce some ambiguity in which tests the C++ driver intentionally does not implement.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1377 @@
...
+        libmongoc::bulkwriteopts_destroy(_opts);
+    }
+
+    bulk_write_options_c(bulk_write_options_c&&) = delete;
+    bulk_write_options_c& operator=(bulk_write_options_c&&) = delete;
+    bulk_write_options_c(bulk_write_options_c const&) = delete;
+    bulk_write_options_c& operator=(bulk_write_options_c const&) = delete;
+
+    explicit bulk_write_options_c(client_bulk_write::options const& opts) : _opts{libmongoc::bulkwriteopts_new()} {
+        if (auto const v = opts.ordered()) {
```
kevinAlbs: Suggest applying same optimization as other `*_c` structs since `mongoc_bulkwrite_execute` [accepts NULL opts](https://github.com/kevinAlbs/mongo-c-driver/blob/18154d15e891a36100b68f9fcd0b5a80cc2e91d7/src/libmongoc/src/mongoc/mongoc-bulkwrite.c#L1545-L1547):

```c++
if (!(opts.ordered() || opts.bypass_document_validation() || opts.comment() || opts.let() || opts.verbose_results() || opts.write_concern())) {
    return;
}

_opts = libmongoc::bulkwriteopts_new();
```

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1377 @@
...
+            libmongoc::bulkwriteopts_set_writeconcern(_opts, write_concern::internal::as_mongoc(*v));
+        }
+    }
+
+    mongoc_bulkwriteopts_t* get() const {
+        return _opts;
+    }
+};
+
+struct bulk_write_return {
```
kevinAlbs: Suggest renaming to signal this wraps the C type `mongoc_bulkwritereturn_t` for consistency with other structs.

```suggestion
struct bulk_write_return_c {
```

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1377 @@
...
+
+    return *this;
+}
+
+client_bulk_write& client_bulk_write::append(
+    bsoncxx::v1::stdx::string_view ns,
+    bsoncxx::v1::document::view filter,
+    v1::pipeline const& update,
+    update_one_options const& opts) {
+    return append(ns, std::move(filter), bsoncxx::v1::document::value{update.view_array()}, opts);
```
kevinAlbs: Remove unnecessary `std::move` since this is a view type.

```suggestion
    return append(ns, filter, bsoncxx::v1::document::value{update.view_array()}, opts);
```

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
```diff
@@ -0,0 +1,1377 @@
...
+
+    return *this;
+}
+
+client_bulk_write& client_bulk_write::append(
+    bsoncxx::v1::stdx::string_view ns,
+    bsoncxx::v1::document::view filter,
+    v1::pipeline const& update,
+    update_many_options const& opts) {
+    return append(ns, std::move(filter), bsoncxx::v1::document::value{update.view_array()}, opts);
```
kevinAlbs: Remove unnecessary `std::move` since this is a view type.

```suggestion
    return append(ns, filter, bsoncxx::v1::document::value{update.view_array()}, opts);
```

kevinAlbs (approved):
Apologies for the delayed review. Great work. LGTM since my comments are not substantial. But feel free to re-request review.

> We should consider adding typedefs/aliases to both drivers that explicitly differentiate between collection and client bulk write

I support this idea as future work. Something like:

```c
// In C driver:
typedef mongoc_bulkwrite_t mongoc_client_bulkwrite_t
typedef mongoc_bulk_operation_t mongoc_collection_bulkwrite_t
```

```cxx
// In C++ driver:
using collection_bulk_write = bulk_write;
```


> It is hard for me to figure out why we do or do not run certain tests in the C++ Driver that we run in the C Driver, so I erred on the side of not adding tests in these cases.

Knowing which tests are "deliberately not run" vs "not-yet implemented" is a long-standing ambiguity and tracked in CXX-2087.

src/mongocxx/lib/mongocxx/v1/client_bulk_write.cpp
connorsmacd: Good catch. This actually revealed a lack of test coverage for the member functions that accept options (`execute` and the `append` overloads).

With 83dc8dd5ec5f729c1f8cf4b7d95aeea627571762, for each options struct with N fields, there are N + 1 test cases:

- 1 test case to ensure `nullptr` is passed when no fields are set.
- N test cases, one per field, to ensure that non-`nullptr` is passed with the expected field set. This ensures that we account for each field in the top-level `if` and remember to set each field when constructing the C type.

src/mongocxx/test/v1/client_bulk_write.cpp
connorsmacd: I think adding a note for this is a good idea, but I'm wondering if `spec/crud.cpp` is the best place for it. AFAIK, all the prose tests implemented in C++ (based on a grep for the word "prose") are found in either `src/mongocxx/test/v_noabi` or `src/mongocxx/test/v1`, not `src/mongocxx/spec`, so I don't expect anyone would look for prose tests in that directory. That said, I don't have any better ideas, so perhaps this fine.

kevinAlbs: > I don't expect anyone would look for prose tests in that directory

Good point. Since no CRUD tests are implemented, this may be better to note the Jira ticket CXX-2494 for anyone referring back to the ticket.

kevinAlbs (approved):

connorsmacd:
> > We should consider adding typedefs/aliases to both drivers that explicitly differentiate between collection and client bulk write
> 
> I support this idea as future work. Something like:
> 
> ```c
> // In C driver:
> typedef mongoc_bulkwrite_t mongoc_client_bulkwrite_t
> typedef mongoc_bulk_operation_t mongoc_collection_bulkwrite_t
> ```
> 
> ```c++
> // In C++ driver:
> using collection_bulk_write = bulk_write;
> ```

Created CDRIVER-6347 and CXX-3510.

PR #1649, opened 2026-05-27
CXX-804 Remove "experimental" commands from v1 API

eramongodb:
Partially resolves CXX-804 by removing all attention paragraphs which document the v1 API as experimental. Features labeled "preview" (notably QE text option enumerators and related class types) retain their attention paragraphs.

As a drive-by fix, removed a misplaced copy-pasted Doxygen comment at the top of the uri.hpp header.

### DIFF
--- a/src/bsoncxx/include/bsoncxx/v1/array/value.hpp
+++ b/src/bsoncxx/include/bsoncxx/v1/array/value.hpp
@@ -36,8 +36,6 @@ namespace array {
 ///
 /// A BSON array.
 ///
-/// @attention This feature is experimental! It is not ready for use!
-///
 class value {
    private:
     v1::document::value _value;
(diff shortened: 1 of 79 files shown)


eramongodb:
On second thought, restored the attention paragraphs for QE Text preview class types (not just enumerators).

kevinAlbs (approved):
Exciting! LGTM.

> restored the attention paragraphs for QE Text preview class types

I support that. That agrees with [the spec notes](https://github.com/mongodb/specifications/blob/44840386103b93e5ebb7d7d595ca605c44eb6e08/source/client-side-encryption/client-side-encryption.md#encryptopts):

> currently unstable API and subject to backwards breaking changes.

And (for better or worse) "text" is soon to be renamed "string" as part of DRIVERS-3531.

PR #1652, opened 2026-05-27
CXX-3502 Fix explicit "upsert: false" behavior for "findOneAnd*" commands

eramongodb:
Resolves CXX-3502. https://github.com/mongodb/mongo-cxx-driver/pull/1568 introduced a regression in the 4.2.0 release for the "findOneAndReplace" and "findOneAndUpdate" commands: the `if (options.upsert().value_or(false))` check was incorrectly refactored to `if (options.upsert())`, where `options.upsert()` returns an `Optional<bool>`. This results in `upsert: true` being set for the command when the field is _set at all_, regardless of its underlying value. No unified spec test or prose test appears to exercise this code path (none explicitly set `upsert: false`; they only explicitly set `upsert: true` or leave it implicitly false).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 
 ### Fixed
 
-- Correctly set `upsert: false` for "findOneAnd*" commands when the option is explicitly set to `false` (regression in 4.2.0).
+- Do not set `upsert: true` in "findOneAnd*" operations when the option is explicitly set to `false` (regression in 4.2.0).
 - Include of `v1/text_options.hpp` (normal header) by `options/text-fwd.hpp` (forward header).
 - Include of v\_noabi macro guard headers by `text_options.hpp` (v1).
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 
 ### Fixed
 
+- Correctly set `upsert: false` for "findOneAnd*" commands when the option is explicitly set to `false` (regression in 4.2.0).
```
```suggestion
- Do not set `upsert` to `true` in `find_one_and_*` operations when the option is explicitly set to `false` (regression in 4.2.0).
```



Suggest tweaking message to match expected behavior.

The C driver only applies `upsert: true` to the `findAndModify` command if the flag [is set](https://github.com/mongodb/mongo-c-driver/blob/66cf09c0f26ad13885daf331e419d1df45857275/src/libmongoc/src/mongoc/mongoc-collection.c#L2315-L2317). I expect this bug caused `upsert: true` to be sent if the the C++ driver option was explicitly set to false.

I also associate "command" with the server command (`findAndModify`) and "operation" with the driver API (`find_one_and_*`).

kevinAlbs (approved):
LGTM with a tweak to CHANGELOG. Thank you for investigating and fixing.

PR #1656, opened 2026-05-28
CXX-3504 Add Mixpanel tracking to latest API doc pages

eramongodb:
Resolves CXX-3504 by manually applying the corresponding changes to API doc pages following https://github.com/mongodb/mongo-cxx-driver/pull/1655 for immediate deployment. Followup to https://github.com/mongodb/mongo-cxx-driver/pull/1651.

API doc pages were manually patched as follows:

- Checkout (detached) the 4.3.0 tag.
- Cherry-pick post-release changes for 4.3.0 (https://github.com/mongodb/mongo-cxx-driver/commit/200bd90340b8540c8d83e8adb250fc542855e792).
- Cherry-pick the changes in this PR (adding Mixpanel tracking).
- Manually set `PROJECT_NUMBER = 4.3.0` in Doxyfile and CMake option `BUILD_VERSION=4.3.0` in the `build` binary directory.
- Generate "current" API doc pages using `etc/generate-latest-apidocs.sh` and explicitly setting .
- Copy generated HTML pages under `build/docs/api/current` into the `api/mongocxx-4.3.0` directory in the `gh-pages` branch.
- Execute `etc/patch-apidocs-index-pages.py` and `etc/patch-apidocs-current-redirects.py`.

Aside from some stray diffs caused by non-deterministic index numbering of deprecation entries, the rest of the diff appears to be clean and as-intended.

Only the latest (4.3.0) API doc pages are patched with the new Mixpanel tracker script for now. We can consider extending tracking to older API doc pages if needed in the future (this may require using a patch script similar to `patch-apidocs-current-redirects`). This may be useful to identify the popularity of specific API doc versions and/or justify the possible removal of old API doc pages.


kevinAlbs (approved):
Thank you, LGTM

PR #1657, opened 2026-05-28
[CXX-3423] Use mongodb-runner instead of orchestration

vector-of-bool:
Refer: CXX-3423

This changeset simply updates the testing scripts to use `mongodb-runner` in drivers-evergreen-tools rather than `mongo-orchestration`.

There was an attempt to debug issues related to execution on certain macOS hosts, but it appears to be an intermittent issue that is unrelated to any changes that are present in this PR. I'm not sure how common those issues actually are, as it took a long time even to run a single patch. It is possible that this merge could create flaky tests on certain platforms, or it may have been bad luck on the patches that I tried to run.

Because of the script's lack of commentary and the reliance on environment variables to implicitly punch through abstraction layers, I'm not sure what else in the script is only relevant to orchestration and can be removed or modified. Does `runner` reuse the same environment variables as `orchestration`? I do not know.

### DIFF
--- a/.evergreen/config_generator/components/funcs/stop_mongod.py
+++ b/.evergreen/config_generator/components/funcs/stop_mongod.py
@@ -12,7 +12,7 @@ class StopMongod(Function):
             script="""\
             set -o errexit
             if test -d drivers-evergreen-tools; then
-                pushd drivers-evergreen-tools
+                cd drivers-evergreen-tools
                 .evergreen/run-mongodb.sh stop
             fi
             """,
--- a/.evergreen/config_generator/components/funcs/stop_mongod.py
+++ b/.evergreen/config_generator/components/funcs/stop_mongod.py
@@ -6,19 +6,18 @@ from config_generator.etc.utils import bash_exec
 
 class StopMongod(Function):
     name = 'stop_mongod'
-    commands = bash_exec(
-        command_type=EvgCommandType.SYSTEM,
-        script="""\
+    commands = [
+        bash_exec(
+            command_type=EvgCommandType.SYSTEM,
+            script="""\
             set -o errexit
-            set -o pipefail
-            if cd drivers-evergreen-tools/.evergreen/orchestration 2>/dev/null; then
-                . ../venv-utils.sh
-                if venvactivate venv 2>/dev/null; then
-                    mongo-orchestration stop
-                fi
+            if test -d drivers-evergreen-tools; then
+                pushd drivers-evergreen-tools
+                .evergreen/run-mongodb.sh stop
             fi
-        """,
-    )
+            """,
+        )
+    ]
 
 
 def functions():
--- a/.evergreen/scripts/start-mongod.sh
+++ b/.evergreen/scripts/start-mongod.sh
@@ -86,5 +86,9 @@ if [[ "${TOPOLOGY:-}" == replica_set ]]; then
   echo "Waiting for replset member 27017 to become primary... done."
 fi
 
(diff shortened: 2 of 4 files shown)


.evergreen/config_generator/components/funcs/stop_mongod.py
```diff
@@ -6,19 +6,18 @@
...
+            script="""\
             set -o errexit
-            set -o pipefail
-            if cd drivers-evergreen-tools/.evergreen/orchestration 2>/dev/null; then
-                . ../venv-utils.sh
-                if venvactivate venv 2>/dev/null; then
-                    mongo-orchestration stop
-                fi
+            if test -d drivers-evergreen-tools; then
+                pushd drivers-evergreen-tools
```
eramongodb:
```suggestion
                cd drivers-evergreen-tools
```
Suggest using `cd` when `popd` is not required.

.evergreen/scripts/start-mongod.sh
eramongodb:
```suggestion
```
These env vars should no longer be necessary: now internally handled by [handle-paths.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4b02eb0136a767b3c6bbdff32b3584c0ec6f5b6f/.evergreen/handle-paths.sh#L62-L63).

eramongodb (changes requested):
Also update this comment: https://github.com/mongodb/mongo-cxx-driver/blob/097f026d1354f3aa3bf82feb4ce6002d23c7683e/.evergreen/config_generator/components/funcs/test.py#L21

Should be:

```python
'CRYPT_SHARED_LIB_PATH',  # Set by run-mongodb.sh in "start_mongod"
```

.evergreen/scripts/start-mongod.sh
```diff
@@ -32,8 +30,7 @@ export ORCHESTRATION_FILE
 
 export PATH="${MONGODB_BINARIES:?}:${PATH:-}"
```
eramongodb: Stray references to removed env vars should also be removed here and below (including the `mongocryptd` copy, which shouldn't be necessary for DET scripts anymore).

eramongodb (changes requested):

.evergreen/scripts/start-mongod.sh
kevinAlbs: Though `mongodb-runner` may not check `MONGODB_BINARIES`, updating the `PATH`  may still be needed since `mongosh` is used below:

```suggestion
export PATH="${DRIVERS_TOOLS:?}/mongodb/bin:${PATH:-}"
```


I expect that is the [download path](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/4b02eb0136a767b3c6bbdff32b3584c0ec6f5b6f/.evergreen/orchestration/drivers_orchestration.py#L185) of `mongodb-runner`.

kevinAlbs (approved):

eramongodb (approved):

vector-of-bool:
This is merge-ready, but waiting on https://github.com/mongodb-js/devtools-shared/pull/770 so that we don't introduce flaky tasks into the mainline.

kevinAlbs:
@vector-of-bool https://github.com/mongodb-js/devtools-shared/pull/770 was merged.

hiSandog:
The mongodb-runner migration is a good place to document topology parity. It would be useful to make sure the new path covers the same standalone, replica set, sharded, auth, and TLS cases that orchestration covered, and that cleanup happens reliably after failed tests. A short checklist in the PR description would help reviewers focus on behavior rather than only the tool swap.

PR #1669, opened 2026-06-10
CXX-3489 add CSFLE prose test 27 cases 8-11

kevinAlbs:
This PR implements test cases 8-11 from https://github.com/mongodb/specifications/commit/2facc54d8a8fc8502a3430e20d9314efa9ec88d0. This includes the change noted in DRIVERS-3470:

> Cases 8 and 9 accidentally specify queryType "prefix" and "suffix". Use "prefixPreview" and "suffixPreview".

The libmongocrypt compile script is updated to use 1.18.1.

As a drive-by fix, majority write concern is added to inserts in all cases of prose test 27 test cases to match spec.

The new test cases will be soon updated to resolve CXX-3467. But I chose to isolate this PR for a smaller review.

`encrypted_client` is renamed to `explicit_encrypted_client` to match the spec and distinguish from the new `auto_encrypted_client` for cases 8-11.


src/mongocxx/test/v_noabi/client_side_encryption.cpp
```diff
@@ -3511,7 +3511,13 @@ TEST_CASE("27. Text Explicit Encryption", "[client_side_encryption]") {
...
     auto tpl = _setup_explicit_encryption(key1_document, &key_vault_client);
     auto client_encryption = std::move(std::get<0>(tpl));
-    auto encrypted_client = std::move(std::get<1>(tpl));
+    auto explicit_encrypted_client = std::move(std::get<1>(tpl));
+    options::insert insert_opts_majority;
+    {
+        write_concern wc_majority;
+        wc_majority.acknowledge_level(write_concern::level::k_majority);
+        insert_opts_majority.write_concern(wc_majority);
+    }
```
eramongodb:
```suggestion
    auto const insert_opts_majority = std::move(v1::insert_one_options{}.write_concern(v1::write_concern{}.majority()));
```
Minor simplification: one-liner using v1 API (no other code changes needed due to implicit v1 -> v_noabi conversion).

src/mongocxx/test/v_noabi/client_side_encryption.cpp
```diff
@@ -3525,28 +3531,48 @@ TEST_CASE("27. Text Explicit Encryption", "[client_side_encryption]") {
...
         auto const encrypted_foobarbaz = client_encryption.encrypt(make_value("foobarbaz"), encrypt_opts);
 
-        coll_substring.insert_one(make_document(kvp("_id", 0), kvp("encryptedText", encrypted_foobarbaz)));
+        coll_substring.insert_one(
+            make_document(kvp("_id", 0), kvp("encryptedText", encrypted_foobarbaz)), insert_opts_majority);
     }
 
     auto const foobarbaz_doc = make_document(kvp("_id", 0), kvp("encryptedText", "foobarbaz"));
 
+    auto const ci_di_text_opts = [&]() { return text_options().case_sensitive(false).diacritic_sensitive(false); };
```
eramongodb:
```suggestion
    auto const ci_di_text_opts = [&]() {
        text_options ret;
        ret.case_sensitive(false).diacritic_sensitive(false);
        return ret;
    };
```
Performance optimization: permit NRVO instead of forced copy of return value due to lvalue (`setter() -> T&`).

src/mongocxx/test/v_noabi/client_side_encryption.cpp
eramongodb:
```suggestion
    auto const default_encrypt_opts = [&]() {
        options::encrypt ret;
        ret.key_id(key1_id).algorithm(options::encrypt::encryption_algorithm::k_textPreview).contention_factor(0);
        return ret;
    };
    auto const default_text_opts = [&]() {
        text_options ret;
        ret.case_sensitive(true).diacritic_sensitive(true);
        return ret;
    };
```
Performance optimization for consistency with NRVO suggestion below.

src/mongocxx/test/v_noabi/client_side_encryption.cpp
```diff
@@ -3671,6 +3697,132 @@ TEST_CASE("27. Text Explicit Encryption", "[client_side_encryption]") {
...
+        }
+    }
+
+    SECTION("Case 9: can find an auto-encrypted diacritic-insensitively indexed document by prefix and suffix") {
+        if (test_util::server_version_is_at_least("9.0")) {
+            SKIP("MongoDB server 9.0 and newer does not support prefixPreview or suffixPreview");
+        }
+
+#define E_ACCENT "\xC3\xA9"
+#define A_UMLAUT "\xC3\xA4"
```
eramongodb:
```suggestion
#define E_ACCENT "\xC3\xA9"
#define A_UMLAUT "\xC3\xA4"

    SECTION("Case 9: can find an auto-encrypted diacritic-insensitively indexed document by prefix and suffix") {
        if (test_util::server_version_is_at_least("9.0")) {
            SKIP("MongoDB server 9.0 and newer does not support prefixPreview or suffixPreview");
        }
```
Macro constants misleadingly scoped in Case 9 when used in other sections below.

data/client_side_encryption/explicit-encryption/encryptedFields.json
```diff
@@ -27,4 +27,4 @@
       "bsonType": "string"
     }
   ]
-}
+}
```
eramongodb: Stray diff?

src/mongocxx/test/v_noabi/client_side_encryption.cpp
```diff
@@ -3525,28 +3531,48 @@ TEST_CASE("27. Text Explicit Encryption", "[client_side_encryption]") {
     auto const suffix_opts = text_options::suffix().str_max_query_length(10).str_min_query_length(2);
     auto const substring_opts =
         text_options::substring().str_max_length(10).str_max_query_length(10).str_min_query_length(2);
-    auto coll_prefix_suffix = encrypted_client["db"]["prefix-suffix"];
+    auto coll_prefix_suffix = explicit_encrypted_client["db"]["prefix-suffix"];
     if (!test_util::server_version_is_at_least("9.0")) {
         // Only test prefixPreview and suffixPreview on server < 9.0. Server 9.0 removes support.
```
eramongodb:
```suggestion

    // Only test prefixPreview and suffixPreview on server < 9.0. Server 9.0 removes support.
    if (!test_util::server_version_is_at_least("9.0")) {
```
Position of comment is a bit misleading in this context. Suggest moving it up a bit.

eramongodb (approved):
Minor suggestions; otherwise, LGTM.

src/mongocxx/test/v_noabi/client_side_encryption.cpp
kevinAlbs: Updated to add a `std::move` call since that seemed to better document the intended optimization. (I likely would not have realized the reason for the named variable was for NRVO).

data/client_side_encryption/explicit-encryption/encryptedFields.json
kevinAlbs: Oop. Fixed.

src/mongocxx/test/v_noabi/client_side_encryption.cpp
```diff
@@ -3671,6 +3697,132 @@ TEST_CASE("27. Text Explicit Encryption", "[client_side_encryption]") {
...
+#define E_ACCENT "\xC3\xA9"
+#define A_UMLAUT "\xC3\xA4"
+
+    SECTION("Case 9: can find an auto-encrypted diacritic-insensitively indexed document by prefix and suffix") {
+        if (test_util::server_version_is_at_least("9.0")) {
+            SKIP("MongoDB server 9.0 and newer does not support prefixPreview or suffixPreview");
+        }
+
+        auto_encrypted_client["db"]["prefix-suffix-ci-di"].insert_one(
+            make_document(kvp("encryptedText", "caf" E_ACCENT "barb" A_UMLAUT "z")), insert_opts_majority);
```
mdb-ad, line 3751:
```suggestion
            make_document(kvp("encryptedText", "cafébarbäz")), insert_opts_majority);
```
I think we can use Unicode strings in C++. Add add_compile_options(/utf-8) to MSVC (introduced in 2015 Update 3, our oldest supported version).

mdb-ad (approved):

src/mongocxx/test/v_noabi/client_side_encryption.cpp
kevinAlbs: TIL about the `/utf-8` flag. I considered this but decided against. AFAICT the library makes no charset assumptions. I could add the UTF-8 build flags to only test binaries, but that may further diverge the test build from the library, and using UTF-8 literals only appears to be a small readability improvement in this test.

PR #1674, opened 2026-06-24
Apply pre-4.4.0 release changes

kevinAlbs:
Follows the "Update CHANGELOG..." release step for the 4.4.0 release.

- Use "v1" in the "ABI Stability" column. This is a new convention, since previous values were "None".
- Update announcement note of "no promises".
    - Remove "API" since I expect API stability is guaranteed (even in `v_noabi`) per [API Versioning](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/api-versioning/#std-label-cpp-api-source-compatibility)

### DIFF
--- a/etc/apidocmenu.md
+++ b/etc/apidocmenu.md
@@ -2,6 +2,7 @@
 
 <h1>Driver Documentation By Version</h1>
 
+[4.4.0](../mongocxx-4.4.0) |
 [4.3.1](../mongocxx-4.3.1) |
 [4.3.0](../mongocxx-4.3.0) |
 [4.2.0](../mongocxx-4.2.0) |
@@ -58,8 +59,8 @@
 | Version     | ABI Stability   | Development Stability       | Development Status |
 | :---------: | :-------------: | :-------------------------: | :----------------: |
 | master      | N/A             | _Do not use in production!_ | Active             |
-| 4.3.1       | None            | Ready for Use               | Bug Fixes Only     |
-| 4.3.0       | None            | Ready for Use               | Not Supported      |
+| 4.4.0       | v1              | Ready for Use               | Bug Fixes Only     |
+| 4.3.1       | None            | Ready for Use               | Not Supported      |
 | ...         | ...             | ...                         | ...                |
 | 4.0.0       | None            | Ready for Use               | Not Supported      |
 | 3.11.1      | None            | Ready for Use               | Not Supported      |
--- a/etc/apidocmenu.md
+++ b/etc/apidocmenu.md
@@ -56,10 +56,10 @@
 > [!IMPORTANT]
 > Relevant bug fixes will be backported from the current major version to the previous major version for a period of one year after the new major version is released.
 
-| Version     | ABI Stability   | Development Stability       | Development Status |
+| Version     | Soversion       | Development Stability       | Development Status |
 | :---------: | :-------------: | :-------------------------: | :----------------: |
 | master      | N/A             | _Do not use in production!_ | Active             |
-| 4.4.0       | v1              | Ready for Use               | Bug Fixes Only     |
+| 4.4.0       | 1               | Ready for Use               | Bug Fixes Only     |
 | 4.3.1       | None            | Ready for Use               | Not Supported      |
 | ...         | ...             | ...                         | ...                |
 | 4.0.0       | None            | Ready for Use               | Not Supported      |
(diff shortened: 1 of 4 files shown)


etc/apidocmenu.md
```diff
@@ -58,8 +59,8 @@
 | Version     | ABI Stability   | Development Stability       | Development Status |
 | :---------: | :-------------: | :-------------------------: | :----------------: |
 | master      | N/A             | _Do not use in production!_ | Active             |
-| 4.3.1       | None            | Ready for Use               | Bug Fixes Only     |
-| 4.3.0       | None            | Ready for Use               | Not Supported      |
+| 4.4.0       | v1              | Ready for Use               | Bug Fixes Only     |
```
eramongodb: Suggest avoiding the `vN` syntax to refer to the ABI version number in this table, to avoid misleading users into thinking `v1` refers to the namespace (e.g. `bsoncxx::v1::*`) rather than the soversion (e.g. `libbsoncxx1.so`). Per the "Important Note" near the bottom of the [ABI Versioning](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/abi-versioning/#root-namespace-redeclarations) page:

> The integer used by an ABI namespace does _NOT_ directly correspond to an ABI version number. The ABI version number is bumped whenever _any_ binary incompatible change occurs, even when it is the removal of a _single_ symbol from the stable ABI. An ABI version number bump from `A` to `B` does NOT imply the deprecation or removal of symbols declared in the ABI namespace `vA` (if one exists).

Intead, suggest updating the table label "ABI Stability" -> "Soversion" to complement the existing "Version" column and leave no room for ambiguity. e.g. a hypothetical version table where API v5.0 makes an ABI breaking change, then API v5.1 makes another ABI breaking change, may look like:

| Version     | Soversion | Development Stability       | Development Status |
| :---------: | :-------: | :-------------------------: | :----------------: |
| master      | N/A       | _Do not use in production!_ | Active             |
| 5.1.0       | 3         | Ready for Use               | Bug Fixes Only     |
| 5.0.0       | 2         | Ready for Use               | Not Supported      |
| 4.8.0       | 1         | Ready for Use               | Bug Fixes Only     |
| 4.7.0       | 1         | Ready for Use               | Not Supported      |
| ...         | ...       | ...                         | ...                |
| 4.4.0       | 1         | Ready for Use               | Not Supported      |
| 4.3.1       | None      | Ready for Use               | Not Supported      |
| ...         | ...       | ...                         | ...                |
| 4.0.0       | None      | Ready for Use               | Not Supported      |

etc/releasing.md
```diff
@@ -828,7 +828,7 @@ Please note that this version of mongocxx requires [MongoDB C Driver A.B.C](http
 
 See the [MongoDB C++ Driver Manual](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/) and the [Driver Installation Instructions](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/installation/) for more details on downloading, installing, and using this driver.
 
-NOTE: The mongocxx X.Y.x series does not promise API or ABI stability across patch releases.
+NOTE: The mongocxx X.Y.x series does not promise ABI stability of entities in the `v_noabi` namespace.
```
eramongodb:
```suggestion
NOTE: ABI stability is NOT supported for entities declared in the `v_noabi` namespace.
```

Wording tweak: stronger language for clarity.

Confirming: removal of the old note suggests we are committing to _not_ having [an escape hatch](https://github.com/mongodb/mongo-c-driver/pull/2312#discussion_r3350088426) for making API/ABI breaking changes in a patch release. This is consistent with the C++ Driver's current [Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/), but I'd overlooked the note in this document.

eramongodb (changes requested):

etc/apidocmenu.md
kevinAlbs: I like the suggestion to switch to "Soversion" for clarity. Will update as suggested.

etc/releasing.md
kevinAlbs: Updated as suggested.

> Confirming: removal of the old note suggests we are committing to _not_ having [an escape hatch](https://github.com/mongodb/mongo-c-driver/pull/2312#discussion_r3350088426)

I did not intend to commit to no escape hatch. I expect this note was a relic of old practices. The C++ driver used to have an (unfortunate) practice of deprecating API in a minor release, and removing in a later minor release. Example: CXX-1578.

eramongodb (approved):

PR #1678, opened 2026-07-09
CXX-3274 add API for sending client metadata on-demand

connorsmacd:
CXX-3274
DRIVERS-2985

# Summary

This PR adds the following member functions to the public API:

- `v1::client::append_metadata` (wraps `mongoc_client_append_metadata`)
  - Since `mongoc_client_append_metadata` only indicates errors by returning `false` (no `bson_error_t` out-param), this throws `mongocxx::v1::exception` with the new error code `mongocxx::v1::client::errc::append_metadata_failure` in the case of error.
- `v1::pool::append_metadata` (wraps `mongoc_client_pool_append_metadata`)
  - Similar to `v1::client::append_metadata`, error handling uses an exception thrown with the new error code `mongocxx::v1::pool::errc::append_metadata_failure`.

> [!NOTE]
> Because `client::errc::append_metadata_failure` and `pool::errc::append_metadata_failure` are very similar, I questioned whether or not they should be different error codes. On the one hand, `pool` already has its own error codes, so throwing an exception from a separate set of error codes may be unexpected. That said, `pool::pool(v1::uri, options)` can throw either `client::errc::tls_not_enabled` or `client::errc::tls_not_supported`, so it would not be unprecedented.

Since these new member functions required minimal integration with libmongoc, I chose not to add the relevant spec tests.


src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -274,9 +298,10 @@ class client {
     /// Errors codes which may be returned by @ref mongocxx::v1::client.
     ///
     enum class errc {
-        zero,              ///< Zero.
-        tls_not_enabled,   ///< TLS is not enabled by URI options.
-        tls_not_supported, ///< TLS is not supported by the mongoc library.
+        zero,                    ///< Zero.
+        tls_not_enabled,         ///< TLS is not enabled by URI options.
+        tls_not_supported,       ///< TLS is not supported by the mongoc library.
+        append_metadata_failure, ///< Failed to append the given metadata.
```
eramongodb:
```suggestion
        append_metadata_failure, ///< Failed to append client metadata.
```
Suggest consistently using "client metadata" in docs for clarity.

src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -254,6 +254,30 @@ class client {
     /// @}
     ///
 
+    ///
+    /// Append the given metadata to the handshake command sent as part of the initial connection handshake.
```
eramongodb:
```suggestion
    /// Append the given client metadata to the handshake command sent as part of the initial connection handshake.
```
Suggest consistently using "client metadata" in docs for clarity.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client.hpp
```diff
@@ -478,6 +478,31 @@ class client {
...
+    /// - The resulting handshake document would exceed the size limit.
+    ///
+    /// @see
+    /// - [hello (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/command/hello/)
+    ///
+    void append_metadata(
+        bsoncxx::v_noabi::stdx::string_view name,
+        bsoncxx::v_noabi::stdx::string_view version,
+        bsoncxx::v_noabi::stdx::string_view platform) {
+        _client.append_metadata(name, version, platform);
```
eramongodb: Move into `.cpp` file and wrap in try-catch block to convert from v1 exception to `v_noabi::operation_exception` for consistency with the rest of the `v_noabi::client` API, e.g. see `list_database_names()`: https://github.com/mongodb/mongo-cxx-driver/blob/0589bdf281a1917a9334c5035388a53a91aed4ab/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp#L300-L302

Also applies to `v_noabi::pool` equivalent.

src/mongocxx/lib/mongocxx/v1/client.cpp
```diff
@@ -400,6 +413,8 @@ std::error_category const& client::error_category() {
                     return "TLS is not enabled by the URI option";
                 case code::tls_not_supported:
                     return "TLS is not supported by the mongoc library";
+                case code::append_metadata_failure:
+                    return "could not append the given metadata";
```
eramongodb:
```suggestion
                    return "could not append client metadata";
```
For consistency with suggested doc wording.

src/mongocxx/lib/mongocxx/v1/client.cpp
```diff
@@ -430,6 +446,7 @@ std::error_category const& client::error_category() {
                 switch (static_cast<code>(v)) {
                     case code::tls_not_enabled:
                     case code::tls_not_supported:
+                    case code::append_metadata_failure:
```
eramongodb: Suggest categorizing this as a runtime error rather than an invalid argument, as failure conditions depend on the current state of the client metadata fields (i.e. size limits).
```cpp
case code::append_metadata_failure:
  return type == condition::runtime_error;
```

src/mongocxx/test/v1/client.cpp
eramongodb, line 1: Also update the "error code" test case at top of this file to cover the new error codes in the "source" and "type" sections.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client.hpp
eramongodb, line 1: IWYU: include `<bsoncxx/v1/stdx/string_view.hpp>` + update `v_noabi::stdx::string_view` -> `v1::stdx::string_view` (for consistency with pool)?

src/mongocxx/include/mongocxx/v_noabi/mongocxx/pool.hpp
eramongodb, line 1: IWYU: include `<bsoncxx/v1/stdx/string_view.hpp>`.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client.hpp
```diff
@@ -478,6 +478,31 @@ class client {
...
+    /// - `this` is from @ref "mongocxx::v_noabi::pool": use @ref mongocxx::v_noabi::pool::append_metadata instead.
+    /// - The resulting handshake document would exceed the size limit.
+    ///
+    /// @see
+    /// - [hello (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/command/hello/)
+    ///
+    void append_metadata(
+        bsoncxx::v_noabi::stdx::string_view name,
+        bsoncxx::v_noabi::stdx::string_view version,
+        bsoncxx::v_noabi::stdx::string_view platform) {
```
eramongodb:
```suggestion
        bsoncxx::v1::stdx::string_view name,
        bsoncxx::v1::stdx::string_view version,
        bsoncxx::v1::stdx::string_view platform) {
```
For consistency with pool API?

eramongodb (changes requested):

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp
```diff
@@ -326,6 +326,15 @@ v1::client_bulk_write client::create_bulk_write(v_noabi::client_session& session
     return bulk_write;
 }
 
+void client::append_metadata(
+    bsoncxx::v1::stdx::string_view name,
+    bsoncxx::v1::stdx::string_view version,
+    bsoncxx::v1::stdx::string_view platform) try {
+    _client.append_metadata(name, version, platform);
+} catch (v1::exception const& ex) {
```
eramongodb, line 336:
```suggestion
} catch (v1::server_error const&) {
    MONGOCXX_PRIVATE_UNREACHABLE; // Only client-side errors.
} catch (v1::exception const& ex) {
```
Explicitly document and enforce absence of server errors in this context (see https://github.com/mongodb/mongo-cxx-driver/pull/1629 for context).

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/pool.cpp
```diff
@@ -130,6 +130,15 @@ bsoncxx::v_noabi::stdx::optional<pool::entry> pool::try_acquire() {
     return ret;
 }
 
+void pool::append_metadata(
+    bsoncxx::v1::stdx::string_view name,
+    bsoncxx::v1::stdx::string_view version,
+    bsoncxx::v1::stdx::string_view platform) try {
+    _pool.append_metadata(name, version, platform);
+} catch (v1::exception const& ex) {
```
eramongodb, line 140:
```suggestion
} catch (v1::server_error const&) {
    MONGOCXX_PRIVATE_UNREACHABLE; // Only client-side errors.
} catch (v1::exception const& ex) {
```

src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client.cpp
```diff
@@ -326,6 +326,15 @@ v1::client_bulk_write client::create_bulk_write(v_noabi::client_session& session
     return bulk_write;
 }
 
+void client::append_metadata(
+    bsoncxx::v1::stdx::string_view name,
+    bsoncxx::v1::stdx::string_view version,
+    bsoncxx::v1::stdx::string_view platform) try {
+    _client.append_metadata(name, version, platform);
```
eramongodb:
```suggestion
    check_moved_from(_client).append_metadata(name, version, platform);
```
For `v_noabi::client` consistency.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client.hpp
```diff
@@ -478,6 +480,28 @@ class client {
...
+    /// Append client metadata to the handshake command sent as part of the initial connection handshake.
+    ///
+    /// @param name
+    ///   The name of the wrapping driver. Must not be empty.
+    /// @param version
+    ///   The optional version of the wrapping driver.
+    /// @param platform
+    ///   The optional information about the current platform, for example configure options or compile flags.
+    ///
+    /// @throws mongocxx::v_noabi::operation_exception when of the following occurs:
```
eramongodb:
```suggestion
    /// @throws mongocxx::v_noabi::operation_exception when one of the following occurs:
```

src/mongocxx/test/v1/client.cpp
```diff
@@ -2167,5 +2169,40 @@ TEST_CASE("reset", "[mongocxx][v1][client]") {
...
+
+    bool called = false;
+
+    auto append_metadata = libmongoc::client_append_metadata.create_instance();
+    append_metadata
+        ->interpose([&](mongoc_client_t* ptr, char const* name, char const* version, char const* platform) -> bool {
+            CHECK(ptr == mocks.client_id);
+            CHECK(bsoncxx::v1::stdx::string_view{name} == "name");
+            CHECK(bsoncxx::v1::stdx::string_view{version} == "version");
+            CHECK(bsoncxx::v1::stdx::string_view{platform} == "platform");
```
eramongodb:
```suggestion
            CHECK_THAT(name, Catch::Matchers::Equals("name"));
            CHECK_THAT(version, Catch::Matchers::Equals("version"));
            CHECK_THAT(platform, Catch::Matchers::Equals("platform"));
```
Suggest using Catch2 string matchers instead (for consistency with `v1::pool` tests).

eramongodb (changes requested):

src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -254,6 +254,30 @@ class client {
...
+    /// - The resulting handshake document would exceed the size limit.
+    ///
+    /// @see
+    /// - [hello (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/command/hello/)
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(void)
+    append_metadata(
+        bsoncxx::v1::stdx::string_view name,
+        bsoncxx::v1::stdx::string_view version,
+        bsoncxx::v1::stdx::string_view platform);
```
kevinAlbs: Consider using `bsoncxx::v1::stdx::optional` for `version` and `platform`.

The C driver seemingly handles a NULL or empty string for `version` and `platform` [equivalently](https://github.com/mongodb/mongo-c-driver/blob/6b7eeb961270222ade9a22674c119646d1be2c03/src/libmongoc/src/mongoc/mongoc-topology-scanner.c#L700-L701), so I expect users can pass an empty string to skip the value. But a `bsoncxx::v1::stdx::optional` might better match user expectation.

Similar suggestion applies to other `append_metadata` helpers.

src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -254,6 +254,30 @@ class client {
...
+    ///   The name of the wrapping driver. Must not be empty.
+    /// @param version
+    ///   The optional version of the wrapping driver.
+    /// @param platform
+    ///   The optional information about the current platform, for example configure options or compile flags.
+    ///
+    /// @throws mongocxx::v1::exception with @ref mongocxx::v1::client::errc::append_metadata_failure when one of the
+    /// following occurs:
+    /// - `this` is from @ref "mongocxx::v1::pool": use @ref mongocxx::v1::pool::append_metadata instead.
+    /// - The resulting handshake document would exceed the size limit.
```
kevinAlbs, line 270:
```suggestion
    /// - The resulting handshake document would exceed the size limit.
    /// - A string argument contains the metadata delimiter " / "
```

For consistency with note in [C driver API](https://mongoc.org/libmongoc/2.3.3/mongoc_client_append_metadata.html). Appending a string with " / " results in an exception.

Similar suggestion applies to other `append_metadata` helpers.

eramongodb (approved):

src/mongocxx/include/mongocxx/v1/client.hpp
connorsmacd: Good point. Similarly, I'm inclined to also use default args for `version` and `platform` to simplify the case where they may be omitted, i.e.:

```C++
MONGOCXX_ABI_EXPORT_CDECL(void)
append_metadata(
    bsoncxx::v1::stdx::string_view name,
    bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> version = {},
    bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> platform = {});
```

kevinAlbs: The coding guidelines includes:

> Avoid using default arguments for parameters of exported ABI functions.

Though admittedly I do not remember the rationale. @eramongodb do you recall the motivation? Was avoiding defaults more of an API consistency concern.

eramongodb: There were three distinct reasons for this guideline:

- default parameters obscures the actual signature of the exported function.
- default parameters force parameter types to be complete, preventing use of forward declarations.
- default parameters permit 0-to-N parameters to be defaulted even when all-or-none may be more correct.

The v_noabi API heavily used `= {}`, which prevented use of forward headers to improve compilation performance and (in certain circumstances, usually when multiple parameters are defaulted) led problems with overload resolution getting in the way of adding new overloads to an existing set (e.g. `foo(a, b, c = {})` -> `foo(a, b, c = {}, d = {})`). Certain API (e.g. `v1::collection`) are so overload-heavy that the convenience of supporting default parameters arguably outweighs the drawbacks listed above. That being said, given how the metadata append API works, I'm not strongly opposed to using `= {}` in this circumstance (the 0-to-N is deliberate).

src/mongocxx/include/mongocxx/v1/client.hpp
```diff
@@ -254,6 +254,31 @@ class client {
...
+    /// - A string argument contains the metadata delimiter " / ".
+    ///
+    /// @see
+    /// - [hello (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/command/hello/)
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(void)
+    append_metadata(
+        bsoncxx::v1::stdx::string_view name,
+        bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> const& version = {},
+        bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> const& platform = {});
```
eramongodb:
```suggestion
        bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> version = {},
        bsoncxx::v1::stdx::optional<bsoncxx::v1::stdx::string_view> platform = {});
```
Trivially-copyable types should be passed by-value ([CppCoreGuidelines F.15](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information)).

kevinAlbs (approved):

PR #1679, opened 2026-07-14
CXX-3529 pin to CMake 4.3.4

kevinAlbs:
To temporarily work around CDRIVER-6367 until the C driver dependency is upgraded. Intended to avoid widespread test failures on Evergreen.

The long-term fix is to update the auto-downloaded C driver to 2.3.0. But this seems to need a more substantial change to account for deprecated QE experimental APIs and tests (CXX-3488).

### DIFF
--- a/.evergreen/scripts/install-build-tools.sh
+++ b/.evergreen/scripts/install-build-tools.sh
@@ -23,7 +23,8 @@ install_build_tools() {
   # PyPI `cmake` requires a sufficiently recent Python version.
   uv python install --no-bin -q || uv python install -q || return
 
-  uv tool install -q cmake || return
+  # TODO: replace "cmake==4.3.4" with "cmake" once the C driver dependency is updated to 2.3.3 to include CDRIVER-6367.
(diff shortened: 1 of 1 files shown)


.evergreen/scripts/install-build-tools.sh
```diff
@@ -23,7 +23,8 @@ install_build_tools() {
   # PyPI `cmake` requires a sufficiently recent Python version.
   uv python install --no-bin -q || uv python install -q || return
 
-  uv tool install -q cmake || return
+  # TODO: replace "cmake==4.3.4" with "cmake" once the C driver dependency is updated to 2.3.3 to include CDRIVER-6367.
+  uv tool install -q "cmake==4.3.4" || return
```
eramongodb:
```suggestion
  uv tool install -q "cmake<4.4" || return
```
Suggest tweaking dependency spec to better describe our intent ("pre-4.4", not "exactly 4.3.4").

eramongodb (approved):

PR #1680, opened 2026-07-15
CXX-3350 Expose atClusterTime parameter in snapshot sessions

mdb-ad:
- Expose snapshot time field from C driver
- Add unified test ops
- Add snapshot unified test
- Bump C driver to 2.4.0


.evergreen/config_generator/components/funcs/install_c_driver.py
```diff
@@ -12,7 +12,7 @@
 # - the default value of --c-driver-build-ref in etc/make_release.py
 # If pinning to an unreleased commit, create a "Blocked" JIRA ticket with
 # a "depends on" link to the appropriate C Driver version release ticket.
-MONGOC_VERSION_MINIMUM = '2.3.3'
+MONGOC_VERSION_MINIMUM = '2.4.0'
```
kevinAlbs, line 15: Many Evergreen tasks fail with:

> Required library (libmongocrypt) not found.

C driver 2.4.0 requires libmongocrypt 1.20.0, so I expect [install-c-driver.sh](https://github.com/mongodb/mongo-cxx-driver/blob/master/.evergreen/scripts/install-c-driver.sh#L85) needs to be updated to bump the libmongocrypt install. I also expect the QE text tests will fail because libmongocrypt 1.19.0 dropped `TextPreview`. I suggest skipping those tests and adding a comment to unskip as part of CXX-3467. I marked CXX-3467 with fixVersion 4.6.0, since I expect bumping to C driver 2.4.0 will make the QE text API unusable until addressing CXX-3467.

kevinAlbs (changes requested):

src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_session.hpp
```diff
@@ -105,6 +105,24 @@ class client_session {
...
 
+    ///
+    /// Get the session's snapshot time ("atClusterTime"), as a BSON timestamp. The snapshot time is
+    /// either the value configured via mongocxx::v_noabi::options::client_session::snapshot_time or
+    /// the value established by the server on the first read operation of a snapshot session.
+    ///
+    /// @return
+    ///   The snapshot time, or an empty optional if the snapshot time has not yet been established.
+    ///
+    /// @throws mongocxx::v_noabi::operation_exception if this session is not a snapshot session.
```
kevinAlbs, line 116: The wrapped `_session.snapshot_time()` throws `mongocxx::v1::exception`. I expect this needs a try-catch to translate the exception to `v1`. See other `try-catch` [example](https://github.com/mongodb/mongo-cxx-driver/blob/59fe517bf407e7121fbac398d609d525fa7df991/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/client_session.cpp#L74-L78) for reference.

Related: https://github.com/mongodb/mongo-cxx-driver/pull/1629 notes the reason for two catch blocks:

> By separating the `v1::server_error` case into its own catch block, these potential exceptions are made more explicit.

kevinAlbs (commented):
Looks good with one fix to the thrown exception in `v_noabi`.

kevinAlbs (approved):

PR #1683, opened 2026-07-17
[CXX-2586] Expose structure logging controls from the C driver in the C++ API

vector-of-bool:
Refer: CXX-2586

Note that this change builds upon #1682 

# Summary

This changeset does not implement structured logging itself, but exposes the existing structured logging APIs from the C driver libraries to the C APIs. This mostly concerns setting structured logging parameters and handlers. This also defines C++ types that wrap the C library types for structure logging entries.

Despite the large size, it is mostly rote handling of the ABI boundary conditions.

> [!NOTE]
> For reviewers: This is my first significant PR that significantly does a lot of work at the ABI boundary since the ABI refactoring. This is largely AI assisted so it closely imitates the existing ABI boundary code. As such, I am not fully confident in the way it handles the ABI boundary, so pay special attention to that aspect.


src/mongocxx/include/mongocxx/v_noabi/mongocxx/structured_log-fwd.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace mongocxx {
+namespace v_noabi {
+
+using v1::structured_log_component;
+using v1::structured_log_entry;
+using v1::structured_log_level;
+
+} // namespace v_noabi
+} // namespace mongocxx
+
```
eramongodb:
```suggestion
```
Only root namespace redeclarations are necessary for v_noabi headers that do not define their own `v_noabi` implementation. They also do not need to strictly follow v_noabi namespace conventions. See the `client_bulk_write` component [for the most up-to-date reference](https://github.com/mongodb/mongo-cxx-driver/blob/59fe517bf407e7121fbac398d609d525fa7df991/src/mongocxx/include/mongocxx/v_noabi/mongocxx/client_bulk_write-fwd.hpp#L23).

src/mongocxx/include/mongocxx/v_noabi/mongocxx/structured_logging-fwd.hpp
```diff
@@ -0,0 +1,47 @@
...
+} // namespace v_noabi
+} // namespace mongocxx
+
+namespace mongocxx {
+namespace options {
+
+using v_noabi::options::structured_logging;
+
+} // namespace options
+} // namespace mongocxx
```
eramongodb, line 25:
```suggestion
namespace mongocxx {

using v1::structured_logging;

} // namespace mongocxx
```
Move and rename this component to `mongocxx/v_noabi/mongocxx/structured_logging` (no `options/`) to mirror the v1 component structure. `v_noabi` patterns should not be followed by v1-only API; they only provide root namespace redeclarations for the v1 entities so that `mongocxx::structured_logging` -> `mongocxx::v1::structured_logging`.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/structured_log.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+using v1::structured_log_component_from_string;
+using v1::structured_log_level_from_string;
+
+using v1::structured_log_handler;
+
+} // namespace v_noabi
+} // namespace mongocxx
+
+namespace mongocxx {
```
eramongodb:
```suggestion
namespace mongocxx {
```
The root namespace redeclarations are sufficient. No `v_noabi` namespace redeclarations are needed (it is not `v_noabi` API).

CHANGELOG.md
```diff
@@ -22,6 +22,22 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 - `mongocxx::logger_guard` (v1): a scope guard to temporarily replace the current
   unstructured log message handler.
 
+### Added
+
+- Structured logging support wrapping libmongoc's per-client/per-pool structured
```
eramongodb:
```suggestion
- Structured logging support (v1) wrapping libmongoc's per-client/per-pool structured
```
Suggest mentioning the ABI stability of these new APIs.

CHANGELOG.md
```diff
@@ -22,6 +22,22 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
...
+    enums, with `to_string()` and `structured_log_level_from_string()` /
+    `structured_log_component_from_string()` helpers.
+  - `mongocxx::structured_log_entry`, a view over a structured log message with
+    `level()`, `component()`, `message()`, and `message_as_bson()` accessors.
+  - `mongocxx::structured_log_handler`, a `std::function` handler type for
+    structured log messages.
+  - `mongocxx::options::structured_logging`, configuring a handler, per-component
+    and all-component maximum levels, maximum document length, and environment
+    overrides, applied per-client and per-pool via
+    `mongocxx::options::client::structured_logging_opts()`.
```
eramongodb:
```suggestion
  - `mongocxx::structured_logging`, configuring a handler, per-component
    and all-component maximum levels, maximum document length, and environment
    overrides, applied per-client and per-pool via
    `structured_logging_opts()`.
```
Avoid being over-specific about API shared by both v1 vs. v_noabi given different declaration namespaces (`v_noabi::options::client` vs. `v1::client::options`).

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/structured_logging.hpp
```diff
@@ -0,0 +1,51 @@
...
+namespace mongocxx {
+namespace options {
+
+using v_noabi::options::structured_logging;
+
+} // namespace options
+} // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
+
```
eramongodb:
```suggestion
#include <mongocxx/config/prelude.hpp>
```
When forward header already declared everything that needs to be redeclared in root namespace, no work besides including the v1 normal header is necessary.

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/structured_logging.hpp
```diff
@@ -0,0 +1,51 @@
...
+///
+/// @file
+/// Provides @ref mongocxx::v_noabi::options::structured_logging.
```
eramongodb:
```suggestion
/// Provides @ref mongocxx::structured_logging.
```

src/mongocxx/include/mongocxx/v_noabi/mongocxx/options/structured_logging-fwd.hpp
```diff
@@ -0,0 +1,47 @@
...
+using v_noabi::options::structured_logging;
+
+} // namespace options
+} // namespace mongocxx
+
+#include <mongocxx/config/postlude.hpp>
+
+///
+/// @file
+/// Declares @ref mongocxx::v_noabi::options::structured_logging.
```
eramongodb:
```suggestion
/// Declares @ref mongocxx::structured_logging.
```

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    MONGOCXX_ABI_EXPORT_CDECL() structured_logging();
+
+    ///
+    /// Set the handler invoked for each structured log message.
+    ///
+    /// A copy of `handler` is stored. When empty (null), structured logging is disabled.
+    ///
+    /// @important When configured on a pool, `handler` MUST be thread-safe.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) handler(v1::structured_log_handler handler);
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) handler(v1::structured_log_handler v);
```
Use `v` for parameter names for consistency with v1 API's avoidance of redundant setter identifiers (e.g. `thing(T thing)` -> `thing(T v)`).

Suggestion also applies to other setters in this class.

src/mongocxx/include/mongocxx/v1/structured_log.hpp
```diff
@@ -0,0 +1,181 @@
...
+///
+/// The severity level of a structured log message.
+///
+/// Levels follow the syslog severities. A lower value is more severe. Filtering selects a
+/// per-component "maximum level": a message is emitted when its level is at or below the
+/// configured maximum for its component.
+///
+/// @see
+/// - @ref mongocxx::v1::structured_log_entry::level
+/// - @ref mongocxx::v1::options::structured_logging
```
eramongodb: Fix Doxygen references to v1 entities:
```
mongocxx/v1/structured_log.hpp:169: warning: unable to resolve reference to 'mongocxx::v1::options::structured_logging' for \ref command
mongocxx/v1/structured_log.hpp:65: warning: unable to resolve reference to 'mongocxx::v1::options::structured_logging' for \ref command
mongocxx/v1/structured_log.hpp:43: warning: unable to resolve reference to 'mongocxx::v1::options::structured_logging' for \ref command
```

eramongodb (changes requested):
> I am not fully confident in the way it handles the ABI boundary, so pay special attention to that aspect.

This initial review was focused specifically on aspects of the v1 API extension. See https://github.com/mongodb/mongo-cxx-driver/pull/1647 for similar feedback and references.

etc/coding_guidelines.md
```diff
@@ -88,7 +88,7 @@ src/<library>/
 ```
 
 > [!NOTE]
->
+>
```
eramongodb:
```suggestion
>
```
Stray whitespace?

etc/coding_guidelines.md
```diff
@@ -260,11 +266,77 @@ Headers under `v_noabi/` declare both unstable (`v_noabi`) AND stable ABI
 interfaces.
 
 > [!IMPORTANT]
->
+>
```
eramongodb:
```suggestion
>
```
Stray whitespace (here and elsewhere below)?

etc/coding_guidelines.md
```diff
@@ -260,11 +266,77 @@ Headers under `v_noabi/` declare both unstable (`v_noabi`) AND stable ABI
...
+- The `v_noabi/` re-export headers MUST mirror the directory layout and name of
+  the `v1` component. Do NOT introduce `v_noabi`-only structure that the `v1`
+  component does not have (e.g. an `options/` grouping subdirectory): a
+  `mongocxx::v1::foo` entity is re-exported by
+  `v_noabi/<library>/foo{-fwd}.hpp`, not
+  `v_noabi/<library>/some_group/foo{-fwd}.hpp`.
+- The root namespace redeclarations follow the usual split (see
+  [Component Design](#component-design)): class-type and enumeration
+  redeclarations belong in the forward header; function and variable
+  redeclarations belong in the normal header.
```
eramongodb: Minor remark: imperative (agentic) language seems inconsistent with the more neutral tone used by the existing document.

src/mongocxx/include/mongocxx/v1/structured_log.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+    ///
+    /// Return the standardized message identifier (e.g. `"Command started"`).
+    ///
+    /// @warning The returned view is valid only for the duration of the handler invocation.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::string_view) message() const;
+
+    ///
+    /// Return an owned copy of the message as a BSON document.
```
eramongodb:
```suggestion
    /// Return the message as a BSON document.
```
"Owned" is redundant given the type declaration: `document::value` is always owning.

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    ///
+    /// Return the current handler.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(v1::structured_log_handler) handler() const;
+
+    ///
+    /// Set the maximum (least severe) level emitted for a single component.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&)
+    max_level_for_component(v1::structured_log_component component, v1::structured_log_level level);
```
eramongodb:
```suggestion
    max_level_for_component(v1::structured_log_component component, v1::structured_log_level v);
```

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+    ///
+    /// Return the maximum level configured for a single component, if any was set on this object.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<v1::structured_log_level>)
+    max_level_for_component(v1::structured_log_component component) const;
```
eramongodb:
```suggestion
    max_level_for_component(v1::structured_log_component v) const;
```

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<v1::structured_log_level>)
+    max_level_for_all_components() const;
+
+    ///
+    /// Read the per-component maximum levels from the environment (the `MONGODB_LOG_*` variables)
+    /// when this configuration is applied.
+    ///
+    /// Applied after any programmatic level settings, so environment values take precedence.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) max_levels_from_env(bool toggle);
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) max_levels_from_env(bool v);
```

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<std::size_t>) max_document_length() const;
+
+    ///
+    /// Read the maximum document length from the environment
+    /// (`MONGODB_LOG_MAX_DOCUMENT_LENGTH`) when this configuration is applied.
+    ///
+    /// Applied after any programmatic document-length setting, so the environment value takes
+    /// precedence.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) max_document_length_from_env(bool toggle);
```
eramongodb:
```suggestion
    MONGOCXX_ABI_EXPORT_CDECL(structured_logging&) max_document_length_from_env(bool v);
```

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,160 @@
...
+namespace v1 {
+
+namespace {
+
+// An opaque tag used as the identity of a fabricated mongoc_structured_log_entry_t.
+struct identity_type {};
+
+} // namespace
+
+TEST_CASE("structured_log_level to_string", "[mongocxx][test][v1][structured_log]") {
```
eramongodb:
```suggestion
TEST_CASE("structured_log_level to_string", "[mongocxx][v1][structured_log]") {
```
Unnecessary `[test]` tag? (Here and elsewehere.)

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,160 @@
...
+                 structured_log_component::k_server_selection,
+                 structured_log_component::k_connection,
+             }) {
+            auto const parsed = structured_log_component_from_string(to_string(component));
+            REQUIRE(parsed);
+            CHECK(*parsed == component);
+        }
+    }
+
+    SECTION("returns nullopt for an unknown name") {
```
eramongodb, line 110:
```suggestion
    }

    SECTION("is case-insensitive") { ... }

    SECTION("returns nullopt for an unknown name") {
```
For symmetry with the `structured_log_level_from_string` test case.

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,160 @@
...
+    CHECK(to_string(structured_log_level::k_debug) == "debug");
+    CHECK(to_string(structured_log_level::k_trace) == "trace");
+}
+
+TEST_CASE("structured_log_component to_string", "[mongocxx][test][v1][structured_log]") {
+    // Names must match the standardized logging specification.
+    CHECK(to_string(structured_log_component::k_command) == "command");
+    CHECK(to_string(structured_log_component::k_topology) == "topology");
+    CHECK(to_string(structured_log_component::k_server_selection) == "serverSelection");
+    CHECK(to_string(structured_log_component::k_connection) == "connection");
```
eramongodb, line 56:
```suggestion
    CHECK(to_string(structured_log_component::k_connection) == "connection");
    CHECK(to_string(static_cast<structured_log_component>(-1)) == "unknown");
```
Coverage for unknown values.

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,160 @@
...
+TEST_CASE("structured_log_level to_string", "[mongocxx][test][v1][structured_log]") {
+    CHECK(to_string(structured_log_level::k_emergency) == "emergency");
+    CHECK(to_string(structured_log_level::k_alert) == "alert");
+    CHECK(to_string(structured_log_level::k_critical) == "critical");
+    CHECK(to_string(structured_log_level::k_error) == "error");
+    CHECK(to_string(structured_log_level::k_warning) == "warning");
+    CHECK(to_string(structured_log_level::k_notice) == "notice");
+    CHECK(to_string(structured_log_level::k_info) == "info");
+    CHECK(to_string(structured_log_level::k_debug) == "debug");
+    CHECK(to_string(structured_log_level::k_trace) == "trace");
```
eramongodb, line 48:
```suggestion
    CHECK(to_string(structured_log_level::k_trace) == "trace");
    CHECK(to_string(static_cast<structured_log_level>(-1)) == "unknown");
```
Coverage for unknown values.

CHANGELOG.md
```diff
@@ -22,6 +22,22 @@ Changes prior to 3.9.0 are documented as [release notes on GitHub](https://githu
 - `mongocxx::logger_guard` (v1): a scope guard to temporarily replace the current
   unstructured log message handler.
 
+### Added
```
eramongodb, line 12: Fix/Merge this new "Added" section + the one above into a single "Added" section below.

```md
## 4.6.0 [Unreleased]

### Removed

- Support for Visual Studio 2015 ...

### Added

- Support appending client handshake metadata ...
- `mongcxx::log_hander` (v1): ...
- Structured logging support (v1) ...
```

src/mongocxx/lib/mongocxx/private/mongoc.hh
```diff
@@ -473,6 +475,24 @@ MONGOCXX_PRIVATE_MOCKED_EXPORT(session_opts_new);
 MONGOCXX_PRIVATE_MOCKED_EXPORT(session_opts_set_causal_consistency);
 MONGOCXX_PRIVATE_MOCKED_EXPORT(session_opts_set_default_transaction_opts);
 MONGOCXX_PRIVATE_MOCKED_EXPORT(session_opts_set_snapshot);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_entry_get_component);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_entry_get_level);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_entry_get_message_string);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_entry_message_as_bson);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_get_component_name);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_get_level_name);
```
eramongodb:
```suggestion
```
No need to re-export an unused symbol.

src/mongocxx/lib/mongocxx/private/mongoc.hh
```diff
@@ -473,6 +475,24 @@ MONGOCXX_PRIVATE_MOCKED_EXPORT(session_opts_new);
...
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_get_named_component);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_get_named_level);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_opts_destroy);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_opts_get_max_document_length);
+MONGOCXX_PRIVATE_MOCKED_EXPORT(structured_log_opts_get_max_level_for_component);
```
eramongodb:
```suggestion
```
No need to re-export unused symbols.

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+//
+
+#include <bsoncxx/v1/stdx/optional.hpp>
+
+#include <mongocxx/v1/structured_log.hh>
+
+#include <array>
+#include <cstddef>
+#include <iostream>
```
eramongodb, line 26:
```suggestion
#include <cstddef>
#include <exception>
#include <iostream>
```
IWYU for explicit `std::terminate()`.

src/mongocxx/lib/mongocxx/v1/structured_log.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+bsoncxx::v1::stdx::string_view to_string(structured_log_component component) {
+    if (auto const name =
+            libmongoc::structured_log_get_component_name(static_cast<mongoc_structured_log_component_t>(component))) {
+        return name;
+    }
+
+    return "unknown";
```
eramongodb: `to_string()` for `structured_log_level` handles unknown values via the `default` switch case. However, this `to_string()` unconditionally narrow-casts from `structured_log_component` (typed as `int`) into `mongoc_structured_log_component_t` (untyped), which is UB for values unrepresentable by mongoc's enum (note: caught by UBSAN given suggested addition of "unknown" test coverage). Suggest avoiding UB by widening this function's contract to gracefully handle unknown values like with `structured_log_level`.

eramongodb (changes requested):

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+v1::structured_log_handler structured_logging::handler() const {
+    return impl::with(*this)._handler;
+}
+
+structured_logging& structured_logging::max_level_for_component(
+    v1::structured_log_component component,
+    v1::structured_log_level level) {
+    impl::with(*this)._component_levels[static_cast<std::size_t>(component)] = level;
```
kevinAlbs, line 96: Consider using `at` so an out-of-range index throws an exception rather than UB.

This may be over-cautious, but it might avoid UB if we later add to `v1::structured_log_component` and forget to resize `_component_levels`.

Similar suggestion applies to the getter.

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    /// Return the "all components" maximum level, if one was set on this object.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<v1::structured_log_level>)
+    max_level_for_all_components() const;
+
+    ///
+    /// Read the per-component maximum levels from the environment (the `MONGODB_LOG_*` variables)
+    /// when this configuration is applied.
+    ///
+    /// Applied after any programmatic level settings, so environment values take precedence.
```
kevinAlbs, line 133:
```suggestion
    /// Applied after any programmatic level settings, so environment values take precedence.
    /// By default, programmatic level settings are applied after environment settings.
```

Suggest revising this comment to note that the environment variables are read by default (even without this setting). But this setting just changes the ordering. On initial read, I thought this meant that environment variables were only read if this this setting was enabled.

src/mongocxx/include/mongocxx/v1/structured_logging.hpp
```diff
@@ -0,0 +1,167 @@
...
+    /// Return the maximum document length, if one was set on this object.
+    ///
+    MONGOCXX_ABI_EXPORT_CDECL(bsoncxx::v1::stdx::optional<std::size_t>) max_document_length() const;
+
+    ///
+    /// Read the maximum document length from the environment
+    /// (`MONGODB_LOG_MAX_DOCUMENT_LENGTH`) when this configuration is applied.
+    ///
+    /// Applied after any programmatic document-length setting, so the environment value takes
+    /// precedence.
```
kevinAlbs, line 152:
```suggestion
    /// Applied after any programmatic document-length setting, so the environment value takes
    /// precedence.
    /// By default, programmatic level settings are applied after environment settings.
```

src/mongocxx/test/v1/client.cpp
```diff
@@ -1085,6 +1088,47 @@ TEST_CASE("apm_opts", "[mongocxx][v1][client]") {
...
+
+    auto opts_destroy = libmongoc::structured_log_opts_destroy.create_instance();
+    opts_destroy->interpose([&](mongoc_structured_log_opts_t* ptr) { CHECK(ptr == opts_id); }).forever();
+
+    auto set_handler = libmongoc::structured_log_opts_set_handler.create_instance();
+    set_handler->interpose(
+        [&](mongoc_structured_log_opts_t* ptr, mongoc_structured_log_func_t func, void* ctx) -> void {
+            CHECK(ptr == opts_id);
+            CHECK(func != nullptr);
+            CHECK(ctx != nullptr);
```
kevinAlbs, line 1108: Minor: suggest adding a bool signal that this function was called, similar to `installed`:

```c++
bool installed_handler = false;
// ...
installed_handler = true;
// ...
CHECK (installed_handler);
```

That may prevent an accidental omission of calling `mongoc_structured_log_opts_set_handler`.

etc/coding_guidelines.md
vector-of-bool: My auto-formatter keeps adding these. At this point it is producing more questions in PRs than it is worth, so I'm just going to disable it.

src/mongocxx/lib/mongocxx/private/mongoc.hh
vector-of-bool: ~~`structured_log_opts_new` is needed and kept.~~ ignore, I got confusion by the suggestion diff.

etc/coding_guidelines.md
vector-of-bool: I softened the language for the paragraph. Should read a bit easier on the eyes now.

src/mongocxx/test/v1/structured_logging.cpp
```diff
@@ -0,0 +1,207 @@
...
+struct pool_identity {};
+struct entry_identity {};
+
+mongoc_structured_log_opts_t* fake_opts(opts_identity* id) {
+    return reinterpret_cast<mongoc_structured_log_opts_t*>(id);
+}
+
+} // namespace
+
+TEST_CASE("structured_logging builder", "[mongocxx][test][v1][structured_logging]") {
```
eramongodb, line 44:
```suggestion
TEST_CASE("structured_logging builder", "[mongocxx][v1][structured_logging]") {
```

src/mongocxx/test/v1/structured_logging.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    SECTION("is copyable") {
+        cfg.max_document_length(500);
+        auto const copy = cfg;
+        CHECK(copy.max_document_length() == std::size_t{500});
+    }
+}
+
+TEST_CASE("structured_logging apply_to", "[mongocxx][test][v1][structured_logging]") {
```
eramongodb, line 77:
```suggestion
TEST_CASE("structured_logging apply_to", "[mongocxx][v1][structured_logging]") {
```

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,167 @@
...
+    CHECK(to_string(static_cast<structured_log_level>(-1)) == "unknown");
+}
+
+TEST_CASE("structured_log_component to_string", "[mongocxx][v1][structured_log]") {
+    // Names must match the standardized logging specification.
+    CHECK(to_string(structured_log_component::k_command) == "command");
+    CHECK(to_string(structured_log_component::k_topology) == "topology");
+    CHECK(to_string(structured_log_component::k_connection) == "connection");
```
eramongodb, line 56:
```suggestion
    CHECK(to_string(structured_log_component::k_topology) == "topology");
    CHECK(to_string(structured_log_component::k_server_selection) == "serverSelection");
    CHECK(to_string(structured_log_component::k_connection) == "connection");
```
Restore `serverSelection` assertion.

src/mongocxx/test/v1/structured_log.cpp
```diff
@@ -0,0 +1,160 @@
...
+        bson_t* const expected = BCON_NEW("message", "Command started", "commandName", "find");
+
+        auto message_as_bson = libmongoc::structured_log_entry_message_as_bson.create_instance();
+        message_as_bson->interpose([&](mongoc_structured_log_entry_t const* ptr) -> bson_t* {
+            CHECK(static_cast<void const*>(ptr) == &identity);
+            // The caller (mongocxx) takes ownership; hand back a fresh allocation.
+            return bson_copy(expected);
+        });
+
+        auto const doc = entry.message_as_bson();
```
kevinAlbs, line 150: Minor: can use `scoped_bson` to declare `expected`:

```c++
scoped_bson const expected{R"({"message": "Command started", "commandName": "find"})"};
```

And simplify this comparison:
```c++
CHECK(entry.message_as_bson().view() == expected.view());
```

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+namespace mongocxx {
+namespace v1 {
+
+class structured_logging::impl {
+   public:
+    v1::structured_log_handler _handler;
+    bsoncxx::v1::stdx::optional<v1::structured_log_level> _all_components_level;
+    // Indexed by `static_cast<std::size_t>(structured_log_component)`.
+    std::array<bsoncxx::v1::stdx::optional<v1::structured_log_level>, 4> _component_levels;
```
eramongodb, line 40: Suggest using a named constant for this `4` magic number to make clearer that this is a `component -> level` map, e.g.:

```cpp
// Number of enumerators in structured_log_component.
constexpr std::size_t structured_log_component_count = 4u;
```

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return *this;
+}
+
+v1::structured_log_handler structured_logging::handler() const {
+    return impl::with(*this)._handler;
+}
+
+structured_logging& structured_logging::max_level_for_component(
+    v1::structured_log_component component,
+    v1::structured_log_level level) {
```
eramongodb, line 95:
```suggestion
    v1::structured_log_level v) {
```
More parameter consistency here and elsewhere in this file. (See: [readability-inconsistent-declaration-parameter-name](https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html))

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,214 @@
...
+v1::structured_log_handler const& structured_logging::internal::handler(structured_logging const& self) {
+    return impl::with(self)._handler;
+}
+
+mongoc_structured_log_opts_t* structured_logging::internal::make_opts(structured_logging& self) {
+    auto const opts = libmongoc::structured_log_opts_new();
+    auto& i = impl::with(self);
+
+    if (i._all_components_level) {
+        libmongoc::structured_log_opts_set_max_level_for_all_components(
```
eramongodb, line 165: These setters discard the `bool` return value which indicates potential failures:

> Returns `true` on success, or `false` if the supplied parameters were incorrect.

Suggest explicitly discarding with `(void)` casts + optionally add a comment to document that these sets should be no-fail by class invariant.

src/mongocxx/lib/mongocxx/v1/structured_logging.cpp
```diff
@@ -0,0 +1,214 @@
...
+        std::cerr << "fatal error: structured log handler " << source << " exited via an exception" << std::endl;
+        std::terminate();
+    }
+}
+
+// The C callback registered with mongoc; recovers the structured_logging from user_data.
+void handle_structured_log(mongoc_structured_log_entry_t const* entry, void* user_data) noexcept {
+    auto const& self = *static_cast<structured_logging const*>(user_data);
+    auto const view = structured_log_entry::internal::make(entry);
+    exception_guard(__func__, [&] { structured_logging::internal::handler(self)(view); });
```
eramongodb, line 151:
```suggestion
    exception_guard(__func__, [&] {
        if (auto const& handler = structured_logging::internal::handler(self)) {
            handler(view);
        } else {
            // Invariant: `structured_logging::internal::make_opts()` never registers `handle_structured_log()` with a
            // null handler.
            MONGOCXX_PRIVATE_UNREACHABLE;
        }
    });
```
Explicitly document this not-null (`std::function<T>`) invariant forbidding the possibility of a `std::bad_function_call` exception.

eramongodb (changes requested):

src/mongocxx/test/v1/structured_logging.cpp
vector-of-bool: I thought I cought them all, and I thought it was a weird thing that the agent was inserting it, but I just did a quick search for literal `[test]` an it appears to be all over the codebase for months. Should I remove them all or just the new ones?

eramongodb: The `[test]` tag is for test cases that validate the behavior of _test utilities_ rather than library API, e.g. `Catch::StringMaker<T>` specializations. Removing them from this PR alone (which does not add new test utilities) should be sufficient.

vector-of-bool: Good to note. Is this information recorded anywhere that's discoverable for users and agents?

eramongodb: AFAIK no (not easily). It was first explicitly documented in the PR description for https://github.com/mongodb/mongo-cxx-driver/pull/1430:

> * `[namespace][to][component]` for better [test case filtering](https://catch2-temp.readthedocs.io/en/latest/command-line.html#specifying-which-tests-to-run).
>   * `[error]` tag for testing error-code-specific behavior (e.g. mapping codes to conditions).
>   * `[test]` tag for test-suite-specific behavior (e.g. Catch stringification).

but its first use was back in [Dec 2024](https://github.com/mongodb/mongo-cxx-driver/commit/db2bdfedeb757df270acd1f2b5605ff039e32f5a#diff-ce2ae6e090e9bb37eeed8dbac7543e1f89057252db2b2e59e279cbff0a688956R29). The v_noabi test suite is rather inconsistent with its use of tags (including misplacing them in `SECTION()` instead of `TEST_CASE()` at one point), so I tried to establish a more consistent scheme for v1 tests.

Feel free to add this info to the coding guidelines refinements or as a separate PR.

etc/coding_guidelines.md
```diff
@@ -380,11 +452,10 @@ lib/
...
 >
-> - `inline` variables require C++17 and newer.
-> - `constexpr` implies `inline` for variables only in C++17 and newer.
-> - Before C++17, non-`inline` `constexpr` variables which are ODR-used
->   require an out-of-line definition.
-> - Use `BSONCXX_PRIVATE_INLINE_CXX17` for pre-C++17 compatibility.
+> - `inline` variables require C++17 and newer. - `constexpr` implies `inline`
+> for variables only in C++17 and newer. - Before C++17, non-`inline`
+> `constexpr` variables which are ODR-used require an out-of-line definition. -
+> Use `BSONCXX_PRIVATE_INLINE_CXX17` for pre-C++17 compatibility.
```
kevinAlbs, line 458: Revert diff that accidentally puts all bullets on one line.

kevinAlbs (commented):
Sorry for the delayed review. The updates to the coding guidelines are much appreciated.

PR #1692, opened 2026-08-10
Close file/pipe descriptors in capture_stderr dtor

eramongodb:
Minor followup to review of https://github.com/mongodb/mongo-cxx-driver/pull/1682, upon discovering the existing relocated `capture_stderr` helper was potentially leaking file descriptors stored as `_stderr` and `_pipes[0]` (`_pipes[1]` is closed in the ctor). Duped file descriptors must be independently closed even when they refer to the same file description. Both the read _and_ write ends (file descriptors) must be closed to cleanup the entire pipe.


vector-of-bool (approved):
LGTM

PR #1697, opened 2026-08-26
CXX-3563 prevent macOS dotfiles in tarball

kevinAlbs:
# Summary

- Set `COPYFILE_DISABLE=1` to prevent macOS adding dotfiles to the distribution tarball.
- Error earlier if a computed example target `name` is empty.

# Details
CXX-3563 reports "same name already exists" errors running CMake on the release tarball for r4.5.0.

```
CMake Error at examples/CMakeLists.txt:66 (add_executable):
  add_executable cannot create target "mongocxx-mongodb.com-" because another
  target with the same name already exists.  The existing target is an
  executable created in source directory
  "C:/____build/Mongo_Driver/mongo-cxx-driver-r4.5.0/mongo-cxx-driver-r4.5.0/examples".
  See documentation for policy CMP0002 for more details.
```

The cause is the release tarball contains extra dot-files produced by macOS `tar`.  I can reproduce by running `cmake --build cmake-build --target dist` on my macOS laptop and extracting the tarball on a Windows host:

Screenshot 2026-08-26 at 8 44 48 AM

macOS `tar` may include [metadata dotfiles](https://unix.stackexchange.com/questions/9665/create-tar-archive-of-a-directory-except-for-hidden-files). The glob to get examples sources finds these dotfiles, parses the `name` as an empty string, and later fails when adding duplicate targets.

To fix, the `tar` command now uses the environment variable `COPYFILE_DISABLE=1`, which tested locally and checked dotfiles are no longer generated. There appears to be no utility to the existing dotfiles. Claude's analysis says of r4.5.0 dotfiles: "they carry exactly one thing: an empty com.apple.provenance xattr.".

Once fixed, I plan to regenerate the r4.5.0 release tarball.

### DIFF
--- a/cmake/make_dist/MakeDist.cmake
+++ b/cmake/make_dist/MakeDist.cmake
@@ -16,8 +16,7 @@ include (MakeDistFiles)
 
 function (MAKE_DIST PACKAGE_PREFIX MONGOCXX_SOURCE_DIR BUILD_SOURCE_DIR)
 
-   set (CMAKE_COMMAND_TMP "")
-   set (CMAKE_COMMAND_TMP ${CMAKE_COMMAND} -E env)
+   set (CMAKE_ENV_COMMAND ${CMAKE_COMMAND} -E env)
 
    # -- Remove any existing packaging directory.
 
@@ -69,7 +68,7 @@ function (MAKE_DIST PACKAGE_PREFIX MONGOCXX_SOURCE_DIR BUILD_SOURCE_DIR)
 
    # Set `COPYFILE_DISABLE=1` to prevent macOS from adding AppleDouble `._<name>` files.
    execute_process_and_check_result (COMMAND
-      ${CMAKE_COMMAND_TMP} COPYFILE_DISABLE=1 tar cf ${PACKAGE_PREFIX}.tar ${PACKAGE_PREFIX}
+      ${CMAKE_ENV_COMMAND} COPYFILE_DISABLE=1 tar cf ${PACKAGE_PREFIX}.tar ${PACKAGE_PREFIX}
       WORKING_DIRECTORY .
       ERROR_MSG "tar command to create ${PACKAGE_PREFIX}.tar failed."
    )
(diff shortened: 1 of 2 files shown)


cmake/make_dist/MakeDist.cmake
```diff
@@ -67,16 +66,17 @@ function (MAKE_DIST PACKAGE_PREFIX MONGOCXX_SOURCE_DIR BUILD_SOURCE_DIR)
 
    # -- Create the tarball.
 
+   # Set `COPYFILE_DISABLE=1` to prevent macOS from adding AppleDouble `._<name>` files.
    execute_process_and_check_result (COMMAND
-      ${CMAKE_COMMAND} -E tar cf ${PACKAGE_PREFIX}.tar ${PACKAGE_PREFIX}
+      ${CMAKE_ENV_COMMAND} COPYFILE_DISABLE=1 tar cf ${PACKAGE_PREFIX}.tar ${PACKAGE_PREFIX}
```
eramongodb: Recommend adding a comment noting that `cmake -E tar` does _not_ see `COPYFILE_DISABLE`, which is imo an unexpected incompatibility between the two approaches.

Given the use of a system `tar` command, may also want to use [find_program()](https://cmake.org/cmake/help/latest/command/find_program.html) (+ one for gzip below); feel free to defer this suggestion as out-of-scope.

eramongodb (approved):
One suggestion; otherwise, LGTM.

rcsanchez97 (approved):

cmake/make_dist/MakeDist.cmake
kevinAlbs: Added comment noting that CMake's built-in `tar` does not respect `COPYFILE_DISABLE`. 

> may also want to use [find_program()](https://cmake.org/cmake/help/latest/command/find_program.html)

I like that idea to produce an earlier error. Updated. Testing with a temporary bad name:

```
  Could not find TAR_EXECUTABLE using the following names: tar_bad_test
```

PR #1715, opened 2026-09-15
Replace `-keep-registers-and-offsets` with `-quiet`

eramongodb:
A quick workaround to fix the failing (false-positive) [abi-compliance-check-impls-cxx17](https://spruce.corp.mongodb.com/task/mongo_cxx_driver_abi_stability_abi_compliance_check_impls_cxx17_dabc8ff93b616747c64007183f04399d9636468d_26_08_27_20_02_03/logs?execution=1) EVG task. `-keep-registers-and-offsets` was added in https://github.com/mongodb/mongo-cxx-driver/pull/1646 to suppress Clang-specific false-positive diagnostics concerning `-Og -g` flags, but it seems to have also added unnecessary sensitivity to stack layouts, which does not affect ABI compatibility. Note the corresponding [abidiff-impls-cxx17](https://spruce.corp.mongodb.com/task/mongo_cxx_driver_abi_stability_abidiff_impls_cxx17_dabc8ff93b616747c64007183f04399d9636468d_26_08_27_20_02_03?execution=1) EVG task correctly reports no compatibility issues.

To unblock analysis of upcoming releases, this PR replaces the flag with `-quiet` to suppress the noisy warnings until a more targeted workaround is found and implemented (we do not want to keep `-quiet`, as it also suppresses useful diagnostic messages).


kevinAlbs (approved):
Thank you!

